PR #52, opened 2020-01-09
Start command Flags

esopian:
**Purpose:**  To allow for overriding variables at runtime.

**User Story:** 

> As an Optic User
> I Need to change my port at runtime
> So that I can match different test environments

**Updated Command Examples:**
`api start --port 8012`
`api start --target "http://externalservice:4343"`

### DIFF
--- a/api-cli/src/commands/start.ts
+++ b/api-cli/src/commands/start.ts
@@ -113,7 +113,7 @@ export default class Start extends Command {
       target,
       port: proxyPort
     })
-    this.log(fromOptic(`Starting ${colors.bold(config.name)} on Port: ${colors.bold(config.proxy.port.toString())}, with ${colors.bold(config.commands.start)}`))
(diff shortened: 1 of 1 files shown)


api-cli/src/commands/start.ts
```diff
@@ -98,10 +106,12 @@ export default class Start extends Command {
...
-    const target = processSetting(config.proxy.target, inputs)
+    const proxyTarget = flags.target ? flags.target : config.proxy.target
+    const proxyPort = flags.port ? parseInt(flags.port, 10) : config.proxy.port
+    const target = processSetting(proxyTarget, inputs)
     await proxySession.start({
       target,
-      port: config.proxy.port
+      port: proxyPort
     })
     this.log(fromOptic(`Starting ${colors.bold(config.name)} on Port: ${colors.bold(config.proxy.port.toString())}, with ${colors.bold(config.commands.start)}`))
```
acunniffe: Hey @esopian  -- can you update the log statement to use the new variables?

esopian: @acunniffe missed this comment.  You've probably moved passed this but I went ahead and changed anyway

PR #74, opened 2020-03-26
Render basic report details using example data

JaapRood:


workspaces/ui/src/components/dashboards/TestingDashboard.js
```diff
@@ -47,13 +47,36 @@ async function ExampleReportTestingDashboardServiceBuilder(exampleId) {
...
-function TestingDashboardHome() {
-  debugger
-  const routingNavigationContext = useContext(RoutingNavigationContext);
-  const {baseUrl} = routingNavigationContext;
+function TestingDashboardHome(props) {
+  const { match } = props;
+  const baseUrl = match.url;
+
+  // TODO: consider factoring this as a useCaptures hook, for re-use elsewhere
+  // ApolloClient's useQuery might serve as inspiration for this: https://github.com/apollographql/apollo-client/blob/master/src/react/hooks/utils/useBaseQuery.ts
```
devdoshi: Will check it out! Another alternative is to provide it via context, which might allow for greater control.

workspaces/ui/src/components/dashboards/TestingDashboard.js
```diff
@@ -47,13 +47,36 @@ async function ExampleReportTestingDashboardServiceBuilder(exampleId) {
   return new ExampleReportTestingDashboardService(orgId);
 }
 
-function TestingDashboardHome() {
-  debugger
-  const routingNavigationContext = useContext(RoutingNavigationContext);
-  const {baseUrl} = routingNavigationContext;
```
devdoshi: I notice you removed this. It isn't exactly necessary at this root component, but it's best to keep paths as paths, as opposed to replacing a matched subpath with the corresponding url. Also, because we will have multiple top level components reusing this, child components should be able to link transparently throughout the common subtree. For example /orgs/:orgId/reports-by-capture and /example-reports/reports-by-capture are both base urls that we will have, but all of the links internally should be independent of that, i.e. `<Link to={`${baseUrl}/captures/capture-1`}>` which we can make a context-aware link wrapper component to simplify. So, paths should use basePath and links should use baseUrl.

Since it doesn't affect this version, no changes required. But we will most likely bring it back later.

JaapRood: Makes sense! My thought was that if we need the path somewhere deeper I'd use one of ReactRouter's Hooks to get to it, since they already provide the router context, but I guess that doesn't account for the changing base url. Once we need it I'll make sure to wrap logic that in some neat way!

workspaces/ui/src/components/dashboards/TestingDashboard.js
```diff
@@ -1,27 +1,21 @@
-import React, {useContext, useEffect, useState} from 'react';
-import Loading from '../navigation/Loading';
-import {Link, Switch, Route} from 'react-router-dom';
-import {opticEngine} from '@useoptic/domain';
+import React, { useContext, useEffect, useState } from "react";
+import Loading from "../navigation/Loading";
+import { Link, Switch, Route, Redirect } from "react-router-dom";
+import { opticEngine } from "@useoptic/domain";
```
devdoshi: Do you have a strong preference for single vs double quotes? I prefer single quotes for typing and reading

JaapRood: No preference on my side, that's just Prettier doing its thing. I'll setup a project config file for it, so we can match it to the existing style where needed!

JaapRood: Fixed with https://github.com/opticdev/optic/pull/74/commits/9eabc2e582a50631415c16af10939c0a3176e8b7

PR #75, opened 2020-03-27
End-End Diff Flow. 

acunniffe:
View captures, add new URLs, update the spec.


workspaces/cli-server/src/routers/spec-router.ts
```diff
@@ -142,7 +141,10 @@ ${events.map((x: any) => JSON.stringify(x)).join('\n,')}
   router.get('/captures', async (req, res) => {
     const { captures } = req.optic.session;
     res.json({
-      captures: sortBy(captures, i => i.taskConfig.startTime).reverse().map(i => i.taskConfig.captureId)
+      captures:
+        sortBy(captures, i => i.taskConfig.startTime)
+        .reverse()
+        .map(i => ({captureId: i.taskConfig.captureId, lastUpdate: i.taskConfig.startTime, hasDiff: false}))
```
devdoshi: What's your plan for this hasDiff?

acunniffe: Probably only show the last 10 captures and compute it each time on the backend. Open to suggestions and not too married to this being a data point the frontend has

PR #78, opened 2020-03-31
Add NavBar to Testing Dashboard page

JaapRood:


JaapRood:
@acunniffe there's a small conflict between this and #73, because the _menuItems_ for the `Navbar` got moved. Let me know if that causes any trouble in the merge, more than happy to help out with that 🙂.

PR #80, opened 2020-04-01
Feature/testing dashboard

devdoshi:
These changes should allow Jaap to continue iterating on the reporting UI


JaapRood:
Merged through #81, in which I cleaned up a few small details so it can live in on the codebase until I get back to this.

PR #85, opened 2020-04-03
Adopted new routing structure w/ caveats

acunniffe:
Hey all -- as discussed, we've migrated over to @JaapRood's new router structure and stayed true to @devdoshi's requirement that context is injected from entry points above the routers. 

There are now two entry points: 
- development (2/3 of the routes implemented) 
- local-cli (weekend project) 

I prioritized getting the example sessions loading since we use those in development, and planned to spend Saturday porting the local-cli spec service over for those routes. 

At day's start, this seemed kinda daunting, but I'm pretty happy with where it landed and as predicted it's made it easier to see what code can be deleted (hint it's a lot) and made everything more readable. 

@JaapRood  -- one todo for you. We don't want concerns about the data loaders to leak past the entry point, so I think you should factor your report hook into a context and provide it from above. Open to different styles and not using HOCs -- but the main thing that matters is the proper separation of concerns.  

I repurposed your useDebugX hooks to make it easy to load the mock sessions and applied your style to loading in the spec service (see code for `ApiSpecServiceLoader`). You should be able to wrap `ApiRoutes` with your context and be good to go. 

```
export default function Development(props) {

  const {match} = props;
  const {sessionId} = useParams();

  const debugSession = useMockSession({
    sessionId: sessionId,
    path: match.url
  });

  return (
    <DebugSessionContextProvider value={debugSession}>
      <ApiSpecServiceLoader>
        <ApiRoutes />
      </ApiSpecServiceLoader>
    </DebugSessionContextProvider>
  );
}
```

Finally, links are still a bit funky, I didn't go through and debug them -- seemed fine to put off till cleanup week. The main thing I wanted to get done today was to unblock everyone and get our code merged.


workspaces/ui/src/components/diff/v2/CaptureManagerPage.js
```diff
@@ -155,52 +154,66 @@ class _AllCapturesStore extends React.Component {
...
-                <TrafficSessionStore sessionId={props.match.params.captureId} specService={specService}>
-                  <DiffPageNew {...props} />
-                </TrafficSessionStore>
-              );
-            }}/>
-          </Switch>
-      </IgnoreDiffStore>
-    </AllCapturesStore>
+  return (
+    <Page title="Optic Live Contract Testing Dashboard">
```
JaapRood, line 164: You'll probably want to change this page title here :)

workspaces/ui/src/routes.js
```diff
@@ -1,65 +1,44 @@
...
+import {useRouterPaths} from './RouterPaths';
 import SharedLoader from './components/loaders/SharedLoader';
-import { ExampleTestingServiceLoaderComponent } from './components/loaders/TestingDashboardLoader';
+import TestingDashboardLoader from './components/loaders/TestingDashboardLoader';
+import ApiSpecServiceLoader from './components/loaders/ApiLoader';
+import {CaptureManagerPage} from './components/diff/v2/CaptureManagerPage';
+import {DocsPage} from './components/docs/DocsPage';
 
+export function ApiRoutes(props) {
+  const routerPaths = useRouterPaths();
```
JaapRood, line 14: I keep getting confused with stuff being named "Api*", as I instantly go "wait, are we serving an API with this?". Only to remember that we're making an app about documenting APIs. I think we could do with better names for these different parts of the entire system / targets 😅.

workspaces/ui/src/components/docs/DocsPage.js
```diff
@@ -0,0 +1,124 @@
...
+
+}));
+
+export const DocsPage = ({match, specService}) => {
+
+  const routerPaths = useRouterPaths();
+  return (
+    <Page title="Optic Live Contract Testing Dashboard">
+      <Page.Navbar
+        mini={true}
```
JaapRood, line 26: Another title prop that should probably not say "Testing Dashboard" 🙂

workspaces/ui/src/entrypoints/development.js
```diff
@@ -0,0 +1,25 @@
...
+  const debugSession = useMockSession({
+    sessionId: sessionId,
+    path: match.url
+  });
+
+  return (
+    <DebugSessionContextProvider value={debugSession}>
+      <ApiSpecServiceLoader>
+        <ApiRoutes />
+      </ApiSpecServiceLoader>
```
JaapRood, line 21: I'm not sure we'll want to wrap all the `AppRoutes` (`ApiRoutes`?.. 🤭) in the `ApiSpecServiceLoader`, as the service that is fetched and the stores applied don't apply to all routes => `TestingDashboard`. As I argued in the `LoaderFactory` comment, I reckon you'll want to define this per page or logical group of pages, but not blanket for all routes.

workspaces/ui/src/contexts/MockDataContext.js
```diff
@@ -0,0 +1,79 @@
...
+    }
+  }, deps);
+
+  return { available, data, loading, error };
+}
+
+export function useBathUrl() {
+  const debugSession = useContext(MockDataContext);
+  return debugSession.path;
+}
```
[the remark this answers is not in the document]
acunniffe: haha make it so. we need a dyslexia gitbot :)

workspaces/ui/src/entrypoints/index.js
```diff
@@ -0,0 +1,24 @@
...
+          component={Development}
+        />
+        <Route
+          strict
+          path="/apis/:apiId"
+          component={() => {
+            return <div>APIs</div>
+          }}
+        />
+      </Switch>
```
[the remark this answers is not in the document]
acunniffe: Agreed, I think the layers are as follows
- TopLevel (routing between these envs, possibly turning some of for certain builds)
- Data Fetching / Saving
- Actual App Router (shared among all the envs). 

In the model, the only thing that should really change is the midtier

workspaces/ui/src/components/loaders/LoaderFactory.js
```diff
@@ -109,37 +109,40 @@ class LoaderFactory {
...
+                                 component={withSpecServiceContext(ApiOverview)}/>
+                          <Redirect to={routerPaths.apiDashboard(basePath)}/>
+                        </Switch>
+                      {/*</Navigation>*/}
+                    </ApiOverviewContextStore>
+                  </IgnoreDiffStore>
+                </AllCapturesStore>
+              </RfcStoreImpl>
+            </InitialRfcCommandsStore>
+          </SpecServiceStore>
```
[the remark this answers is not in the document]
acunniffe: Yep ^ this is dead code now. 

In general, I agree. Spec service is used almost everywhere so I put it at the top, but traffic/session is only used by the diff page so it just lives there right now

core/optic/shared/src/main/scala/com/useoptic/contexts/requests/projections/AllEndpointsProjection.scala
```diff
@@ -0,0 +1,31 @@
+package com.useoptic.contexts.requests.projections
+
+import com.useoptic.contexts.requests.Commands.PathComponentId
+import com.useoptic.contexts.rfc.RfcState
+
+import scala.scalajs.js.annotation.JSExportAll
+
+object AllEndpointsProjection {
```
devdoshi, line 8: A "Projection" must be built from the event stream. I would only name things built from the event stream *Projection

workspaces/ui/src/components/diff/v2/AddUrlModal.js
```diff
@@ -89,40 +89,45 @@ export const NewUrlModal = withNavigationContext(withRfcContext(({pushRelative,
...
-        <Button onClick={() => setNaming(false)}>
-          Back
-        </Button>
-        <Button onClick={handleCreate} color="secondary" disabled={!matches}
-                endIcon={<NavigateNextIcon/>}>Finish</Button>
-      </DialogActions>
-    </Dialog>
-  );
+  function NamingDialog() {
+    const [purpose, setPurpose] = React.useState('');
```
devdoshi, line 101: @acunniffe I moved this into here, since it was re-rendering the parent on every change before. Let's try to stay conscious of that

workspaces/ui/src/components/diff/v2/CaptureManagerPage.js
```diff
@@ -126,23 +124,24 @@ class _AllCapturesStore extends React.Component {
 
   render() {
 
-    const { location } = this.props;
+    const {location} = this.props;
     const matched = matchPath(location.pathname, {
+      //@todo fix this
```
devdoshi, line 129: What exactly is the requirement/feature here?

workspaces/ui/src/components/diff/v2/CaptureManagerPage.js
```diff
@@ -155,52 +154,66 @@ class _AllCapturesStore extends React.Component {
...
+      <Page.Navbar
+        mini={true}
+        baseUrl={match.url}
+      />
+      <Page.Body>
+        <Switch>
+          <AllCapturesStore>
+            <IgnoreDiffStore>
+              <Switch>
+                <Route exact path={routerPaths.diffPage()}
```
devdoshi, line 174: let's move these to *Root components. *Root components are glue between react-router and our app components

workspaces/ui/src/components/diff/v2/CaptureManagerPage.js
```diff
@@ -155,52 +154,66 @@ class _AllCapturesStore extends React.Component {
...
   const rfcState = rfcService.currentState(rfcId);
 
   function handleChange(event) {
     const captureId = event.target.value;
-    pushRelative(`/diff/${captureId}`);
+    history.push(`diff/${captureId}`);
   }
 
   if (captureContext.captures.length > 0 && !captureId) {
-    pushRelative(`/diff/${captureContext.captures[0].captureId}`);
```
devdoshi, line 197: push relative to baseUrl is what we want, right?

workspaces/ui/src/components/loaders/LoaderFactory.js
```diff
@@ -109,37 +109,40 @@ class LoaderFactory {
...
+                <AllCapturesStore>
+                  <IgnoreDiffStore>
+                    <ApiOverviewContextStore specService={specService}>
+                      {/*<Navigation*/}
+                      {/*  notifications={notificationAreaComponent}*/}
+                      {/*  entryBasePath={entryBasePath}*/}
+                      {/*  shareButtonComponent={shareButtonComponent}>*/}
+                        <Switch>
+                          <Route path={routerPaths.request(basePath)}
+                                 component={withSpecServiceContext(RequestsDetailsPage)}/>
```
devdoshi, line 124: let's move these to *Root as well, so they aren't recreated every time

workspaces/ui/src/contexts/InitialRfcCommandsContext.js
```diff
@@ -2,25 +2,30 @@ import * as React from 'react';
...
-
-        return (
-            <InitialRfcCommandsContext.Provider value={{rfcId, initialCommandsString, initialEventsString}}>
-                {this.props.children}
-            </InitialRfcCommandsContext.Provider>
-        )
-    }
+  render() {
+    const {rfcId, initialCommandsString, initialEventsString} = this.props;
+    global.devlog2.push({
```
devdoshi, line 12: we can remove this. was helpful for me for debugging though, via watching console.table(devlog2)

workspaces/ui/src/routes.js
devdoshi: Agreed. Let's iterate on a glossary in Notion and then get it into the codebase / public docs

workspaces/ui/src/routes.js
```diff
@@ -1,65 +1,44 @@
...
-          <Route exact path={'/'} component={Welcome} />
-          <Redirect to={'/'} />
-        </Switch>
-      </div>
-    );
-
-  }
+  return (
+    <ApiSpecServiceLoader>
+    <Switch>
```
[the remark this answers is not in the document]
devdoshi: Here's one use case for my displayName convention. Once we're in react devtools we can filter to only show our component hierarchy from different lenses. Or we can set up some tests that assert invariants about the heirarchy

PR #86, opened 2020-04-05
Updated UX for Diff flow

acunniffe:
Based on our internal design session and some of the user feedback we got Friday I rebuilt the diff flow and tried to better balance the improved functionality w/ the user experience. 

https://www.loom.com/share/14f51a1a36fb4048bf8ce08667e8d7b6


JaapRood:
Not sure how stable this branch is, but keen for this to get merged, so I can work with similar design language in the testing dashboard, re-use stuff, etc. If it's stable enough, additional PRs could be used to further iterate on it.

PR #88, opened 2020-04-06
Adapt Page layout and implement basic layout for testing dashboard

JaapRood:


JaapRood:
@acunniffe in order to make sure a `Page` can render any type of UI it wants (a viewport heigh one, in case of the testing dashboard), some of it's basics have changed. While I tried to keep the default behaviour intact, this _might_ have an impact on #86, so you might want to give that a look when you merge that in.

PR #91, opened 2020-04-07
Made local CLI compatible with the new frontend routing structure 

acunniffe:


workspaces/ui/src/contexts/BaseUrlContext.js
```diff
@@ -0,0 +1,13 @@
+import React, { useContext, useEffect, useMemo, useState } from 'react';
+
+const BaseUrlContext = React.createContext(null);
+BaseUrlContext.displayName = 'BaseUrlContext';
```
JaapRood, line 4: Seemed like a silly name when I first saw it, but can't think of anything else as descriptive and obvious 😅.

PR #100, opened 2020-04-09
Fixed a serious bug in the side-by-side diff viewer

acunniffe:
Because of the way the Example rendering traversal was written (by me), any Lists of Objects would show the last item in the array for every item. The diff and suggestions were still correct, but w/o the proper UI everything was quite confusing. 

The issue boiled down to the fact that I tried to make the parallel traversal of spec + examples yield a map of ID -> RenderX where IDs were shared across both our spec and example domain. 

You would basically lookup two RenderX's and merge them

Unfortunately, sharing IDs like that allowed details from domain to bleed over into JSON land. In domain you can only have one listItem so the same ID was being used for every list item. You'd still get the right # of items, but each would be the last item in the array. oops

Now IDs from the JSON traversal are BodyTrail.toString and they contain a reference to their specShape. 

If I were doing this from scratch I would stop trying to have RenderX be shared between example and spec. 

For now, it's alright, but long term RenderExampleX and RenderSpecX will diverge more.

### DIFF
--- a/core/optic/shared/src/main/scala/com/useoptic/diff/shapes/JsonLikeTraverser.scala
+++ b/core/optic/shared/src/main/scala/com/useoptic/diff/shapes/JsonLikeTraverser.scala
@@ -76,7 +76,7 @@ class JsonLikeTraverser(spec: RfcState, visitors: JsonLikeVisitors) {
         })
         val jsonTrail = bodyTrail.withChild(JsonObjectKey(key))
 
-        visitors.objectVisitor.visit(key, value, bodyTrail, resolvedTrail, resolved.bindings)
+        visitors.objectVisitor.visit(key, value, jsonTrail, resolvedTrail, resolved.bindings)
 
         traverse(Some(value), jsonTrail, resolvedTrail)
       })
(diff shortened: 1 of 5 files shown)


core/optic/shared/src/main/scala/com/useoptic/diff/shapes/JsonLikeTraverser.scala
```diff
@@ -76,7 +76,7 @@ class JsonLikeTraverser(spec: RfcState, visitors: JsonLikeVisitors) {
         })
         val jsonTrail = bodyTrail.withChild(JsonObjectKey(key))
 
-        visitors.objectVisitor.visit(key, value, bodyTrail, resolvedTrail, resolved.bindings)
+        visitors.objectVisitor.visit(key, value, jsonTrail, resolvedTrail, resolved.bindings)
```
acunniffe, line 79: I think this may be causing some of the misplaced diffs we get in nested objects. each field had the wrong trail :(

acunniffe: @devdoshi -- am I missing something? All other behavior seems to work fine with the change

PR #101, opened 2020-04-10
Render capture information as part of ReportSummary

JaapRood:


JaapRood:
@acunniffe there might be some conflicts with this and #104, so let me know if you need my help in resolving those the right way!

PR #104, opened 2020-04-12
Cleaned up unused code, deps and imports

acunniffe:
Hey all -- I took the time to go through the rest of the UI codebase and trim all the unused parts. 

@JaapRood  --  you were using the flattenOperations method from API Overview -- that's been deprecated a while (my bad for not looking at all your PRs). I removed it, and made your code use the preferred method, the `endpoints` query. 

I also went ahead and made your `createSpec` method produce a context that simulates whatever spec the report was made against. This will allow you to use `EndpointContext` and any components from the documentation page you wish :) 

I tested everything pretty well but wanted to call these changes out so they didn't get lost in the torrent of other changes. tbh I should have made it a separate PR but I was already pretty deep in the refactor and did not want to step back and start a few parts over. 

```
const SimulatedContext = (props) => (
    <InitialRfcCommandsStore
      instance="the one in simulated command context"
      rfcId={rfcId}
      initialEventsString={stringifiedEvents}>
      <ImmutableRfcStore>
        {props.children}
      </ImmutableRfcStore>
    </InitialRfcCommandsStore>
  );
```


JaapRood:
> @JaapRood -- you were using the flattenOperations method from API Overview -- that's been deprecated a while (my bad for not looking at all your PRs). I removed it, and made your code use the preferred method, the endpoints query.

Thanks for the clean-up! We knew those methods were deprecated, but just had to get something on the screen (endpoints query didn't exist yet at that time). Perfect time to swap that all out :)

workspaces/ui/src/components/dashboards/TestingDashboard.js
```diff
@@ -1,14 +1,12 @@
...
-import { stuffFromQueries } from '../../contexts/RfcContext';
-
+import {stuffFromQueries} from '../../contexts/RfcContext';
 // Components and hooks
 // --------------------
-import { Switch, Route, Redirect } from 'react-router-dom';
-import { makeStyles } from '@material-ui/core/styles';
+import {Redirect, Route, Switch} from 'react-router-dom';
+import {makeStyles} from '@material-ui/core/styles';
 import ReportSummary from '../testing/ReportSummary';
```
JaapRood, line 10: There's a bunch of these reformatting changes in here, I imagine from the built-in WebStorm formatter. We've got a Prettier config file setup as part of the repo, which is the ecosystem defect standard nowadays. To prevent reformatting the codebase all in one go and losing lots of Git history, would it be an idea to [set WebStorm up to use  Prettier](https://prettier.io/docs/en/webstorm.html)? That way we can all format consistently, migrate file by file and end this tug of war 😅.

JaapRood:
Given the scale of the PR, I'd say we'd want to make it a priority to get it stable and merged. I'm already running into conflicts with #101 😅.

workspaces/ui/src/components/testing/ReportSummary.js
```diff
@@ -31,67 +28,79 @@ export default function ReportSummary(props) {
...
+                        <PathAndMethod method={endpointDescriptor.method} path={endpointDescriptor.fullPath}/>{' '}
+                        {endpointDescriptor.endpointPurpose}: ({counts.compliant} {'/'}
+                        {counts.interactions} interactions compliant)
+                      </li>
+                    );
+                  }}
+                </EndpointsContext.Consumer>
+              </EndpointsContextStore>
+            ))}
+          </ul>
```
JaapRood, line 65: I appreciate the idea of trying to re-use the endpoints domain logic to help render the endpoints correctly, but I don't think context is the right way to go here. The fact that the `EndpointContextStore` is defined and it's direct child instantly consumes is a big indicator that we're using a hammer to drive down a screw, rather than a screw driver.

Instead, I'll opt to extract the domain / view model logic from EndpointsContext as a pure function, as I've done with the view model of the summary thus far.

Using contexts this way is actually one of the major sources of confusion and barriers for me getting to grip with the UI code. In my mind, using contexts is a great place to hold application state, or to provide an interface to a service. However, domain or view-model logic, which is a lot of what I see in these contexts, deserves it's own spot,  so it can be re-used statelessly. All that _does_ belong in a Context / Store, should be accessed through a Hook. With a bit of luck, that'll prevent the endless wrapping components, super deep element trees, opaquely injected properties, etc.

The TestingDashboard feature is my exploration in how to balance these concerns in a way that  so lets keep that pattern out of it. Let's revert this to the latest what's in `feature/testing-dashboard` and I'll move on the domain logic separately, keeping this PR as an example of how the domain is used 🙂.

JaapRood: I've taken this approach as part of https://github.com/opticdev/optic/pull/104/commits/307c804dac71415e91aec2415575d6a3bb9286fb, housing some of the logic that was in the now removed `APIOverviewContext` in the `ReportSummary`, ready to be factored away by using the endpoints seq in a separate PR.

JaapRood:
Back to stable from my perspective, ready to get merged.

workspaces/ui/src/components/dashboards/TestingDashboard.js
acunniffe: Bingo! Good idea.

workspaces/ui/src/components/testing/ReportSummary.js
acunniffe: This makes sense to me, I think there have been very few occurrences when I have not directly consumed the context. I like the pattern you have here and will do something similar soon :) For now let's merge and I can have something new up here in the course of the week.

JaapRood: Restored the changes in the "new style" through #105.

PR #112, opened 2020-04-17
Rewrote the Side / Side Diff Renderer

acunniffe:
As we got deeper into testing the new diff flow a lot of related bugs started to emerge from how I was doing the parallel rendering between the example Optic observed and the spec. 

As discussed in [#100](https://github.com/opticdev/optic/pull/100), the design focus on reuse between different contexts (example only, shape only, side-by-side, etc) was creating a lot of code comprehension problems, logic branches, and ultimately adding a lot of surface area for bugs and unexpected behavior. 

Over the last day, I took a step back and build the rendering logic from scratch with an eye towards stability, comprehension and making it testable in Scala for faster iteration. 

A lot of LOC and code smell was cut and I think this will be both more reliable and easier to fix should we find issues. 

Note: The Shape Rendering for the docs page has been disabled, but I'll get that working tomorrow in another PR. I just wanted to get this out ASAP for everyone to do testing on.


workspaces/ui/src/services/TestingService.js
```diff
@@ -89,18 +93,30 @@ export async function createExampleTestingService(exampleId = 'todo-report') {
...
+      rfcId={rfcId}
+      initialEventsString={stringifiedEvents}>
+      <ImmutableRfcStore>
+        {props.children}
+      </ImmutableRfcStore>
+    </InitialRfcCommandsStore>
+  );
+
+  return {queries, rfcState, SimulatedContext};
 }
```
JaapRood, line 122: I think you might have been working of an older branch, because all of this and formatting changes to `TestingService` and `TestingDashboardContext`  re-appeared! Can't think how that would be related, so I'll go ahead and revert those changes.

JaapRood: I was pretty confused about why this would happen, but I think the answer is in the git tree.
Here's the commits:

Screenshot 2020-04-17 at 11 44 39

The orange line is the current branch of this PR, as you can see, the changes made that are missing here never got merged to it. @acunniffe I suspect you might have merged the PR locally, rather than through Github, and never pulled down my commits!

From there, I don't think you went back to `feature/testing-dashboard` and pulled the latest version from Github either, but instead kept on merging those changes back in. 

Screenshot 2020-04-17 at 12 30 53

To prevent this in the future, I'd suggest pulling the "main development" branch and branching from that, when starting a new branch. That way, you make sure your new changes are on the latest "stable" development version. Also, merging PRs are best done through the Github interface; it even allows rebasing or squashing as 1 commit, rather than a merge!

JaapRood: Fixed with 10b8781193eddcb02cb97794373c94e86f1865b8. @acunniffe be sure to pull these changes this time 😂.

JaapRood:
When I test this against my F1 API example project, I get the following when trying to access an endpoint with a diff (the one where the example wasn't rendered properly):

Screenshot 2020-04-17 at 14 24 10

This might be expected as this PR might not have been ready for review / merging yet, but I thought I'd share it nonetheless.

workspaces/ui/src/services/TestingService.js
acunniffe: Thanks for walking me through exactly all missteps re: gitflow. I'm still getting used to this process and appreciate all the guidance.

JaapRood:
Replaced by #113!

PR #119, opened 2020-04-20
Add some reasonably detailed Page titles to all pages

JaapRood:


workspaces/ui/src/components/Page.js
```diff
@@ -19,14 +21,23 @@ const useStyles = makeStyles((theme) => ({
...
   const classes = useStyles();
-  const title = props.title || 'Optic API Dashboard';
+  const pageContext = useState([]);
+  usePageTitle(props.title, pageContext);
+
+  const [pageTitles] = pageContext;
+  const titles = pageTitles.filter((title) => !!title);
+  const title = titles[titles.length - 1] || '';
 
   useEffect(() => {
```
acunniffe, line 31: Why not use Helmet? https://www.npmjs.com/package/react-helmet

JaapRood: Because why use Helmet? Maybe if we were writing way more page specific things and have to write heaps of html in head , but `document.title` is pretty functional, allows us full control, way less complexity, one less dependency, etc 😊

PR #123, opened 2020-04-20
Clean up React warnings

JaapRood:
Mostly by rendering lists with keys.

### DIFF
--- a/workspaces/ui/src/components/diff/v2/DiffHelperCard.js
+++ b/workspaces/ui/src/components/diff/v2/DiffHelperCard.js
@@ -75,8 +75,9 @@ export const DiffHelperCard = (props) => {
 
         <FormControl component="fieldset" className={classes.formControl}>
           <RadioGroup>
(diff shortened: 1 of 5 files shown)


JaapRood:
@acunniffe you'll need to review this one, as I'm not familiar enough with the diff domain model to know what good stable keys for the lists would make. For now, I'm taking the index of the collection that's being iterated over with `mapScala`, which gets rid of the warnings, but will break if the underlying changes in certain ways. It's better for the key to be an identifier that describes the element's entity uniquely relative to the others being rendered.

acunniffe:
> @acunniffe you'll need to review this one, as I'm not familiar enough with the diff domain model to know what good stable keys for the lists would make. For now, I'm taking the index of the collection that's being iterated over with `mapScala`, which gets rid of the warnings, but will break if the underlying changes in certain ways. It's better for the key to be an identifier that describes the element's entity uniquely relative to the others being rendered.

Everything in the diff domain as a unique ID so I'll update to those references :)

PR #127, opened 2020-04-22
Optic v8

JaapRood:
We'll want to merge v8 as it's released back to master, and keep that as the stable and current branch, so master always reflects what has been published.


JaapRood:
Locally this merge seemed to work without any issues!

PR #136, opened 2020-04-28
New Shape Diff flow to support some gaps with polymorphism

devdoshi:
@acunniffe I tested against the diff-test-cases examples and it seems to work but the UI doesn't seem to update all the time. It's probably related to my changes to the trails, so we should pair on that like we planned


acunniffe:
@devdoshi -- looks good. [Schedule a pair session](https://calendly.com/optic-onboarding/1hr-chat)?

JaapRood:
@devdoshi I'm afraid this broke something, triggered by the `TestingService` accessing `PathAndMethodDiffManager.diffRegions`. Commit before this everything is peachy, since this merge:

Screenshot 2020-05-04 at 12 22 58

To reproduce, run UI dev server and visit http://localhost:3000/development/aidan-2/testing/captures/capture-1/endpoints/path_MwY9nZ6x25.POST. Having a hard time tracing it any further myself, the error returned by Scala js doesn't have a stack trace :/

PR #140, opened 2020-05-01
coverage report should use real counts for compliant requests

devdoshi:


JaapRood:
I've updated the target to `develop` branch, now that v8 has been merged to `master` and `develop`.

JaapRood:
@devdoshi is this safe to merge in, even if it's not complete? Seems to have an effect on the example reports, so would be nice to work against :)

devdoshi:
Yup - and it should be complete. diffsCount will come later

PR #142, opened 2020-05-02
Diffs presented on the endpoints page should be grouped by their shape trails

acunniffe:
Hey all -- @ross-nordstrom raised an issue on the Slack about duplicate diffs. 

Use case: given an array where many of the items have a missing field, Optic produces a diff for each item

Now all diffs are grouped by shape diffs. 

https://www.loom.com/share/0576f4f3d9df4941895a4ac4fcb9b4aa

@devdoshi -- anything I'm missing here?


devdoshi:
Nice! Seems like the other choice would be to change the InteractionDiffResult equality so they end up in the same group to begin with if the JsonTrails are otherwise equivalent (ignoring the array item index). The way you did it keeps it isolated to the UI so that seems fine with me.

acunniffe:
That was my thinking. Seems like filtering and grouping should always remain separate from core diffing on the UI and Backend

acunniffe:
Included in #146

PR #143, opened 2020-05-02
Distribution aware shape builder (WIP)

acunniffe:
https://www.loom.com/share/0f33dd3e102b41908f469a9cd3028ac7

### DIFF
--- a/core/optic/shared/src/main/scala/com/useoptic/diff/initial/DistributionAwareShapeBuilder.scala
+++ b/core/optic/shared/src/main/scala/com/useoptic/diff/initial/DistributionAwareShapeBuilder.scala
@@ -46,15 +46,15 @@ object DistributionAwareShapeBuilder {
       id
     }
 
-    fromJsons(root, JsonTrail(Seq.empty), false)
+    fromJsons(root, JsonTrail(Seq.empty), false, trailValues.totalSamples)
   }
 
-  private def fromJsons(values: Vector[JsonLike], trail: JsonTrail, inner: Boolean)(implicit trailValues: TrailValueMap, idGenerator: () => String): ShapesToMake = {
-    val isOptional = values.size != trailValues.totalSamples
+  private def fromJsons(values: Vector[JsonLike], trail: JsonTrail, inner: Boolean, totalSamples: Int)(implicit trailValues: TrailValueMap, idGenerator: () => String): ShapesToMake = {
+    val isOptional = values.size != totalSamples
     val kinds = values.groupBy(v => Resolvers.jsonToCoreKind(v))
 
     if (isOptional && !inner) {
-      val optionalShape = OptionalShape(fromJsons(values, trail, true), trail, idGenerator())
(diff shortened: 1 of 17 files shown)


core/optic/shared/src/main/scala/com/useoptic/diff/initial/DistributionAwareShapeBuilder.scala
```diff
@@ -0,0 +1,265 @@
...
+
+    val jsonLikeTraverser = new JsonLikeTraverser(RfcState.empty, visitor)
+
+    bodies.foreach(body => jsonLikeTraverser.traverse(Some(body), JsonTrail(Seq.empty), None))
+
+    aggregator
+  }
+
+
+  def toShapes(implicit trailValues: TrailValueMap, seed: String = s"${Random.alphanumeric take 6 mkString}"): ShapesToMake = {
```
devdoshi, line 100: We should inject the ID generator this time

core/optic/shared/src/main/scala/com/useoptic/diff/interactions/interpreters/MissingValueInterpreter.scala
```diff
@@ -18,7 +18,8 @@ class MissingValueInterpreter(rfcState: RfcState) extends InteractiveDiffInterpr
   private val basicInterpretations = new BasicInterpretations(rfcState)
   private val descriptionInterpreters = new DiffDescriptionInterpreters(rfcState)
 
-  override def interpret(diff: InteractionDiffResult, interaction: HttpInteraction): Seq[InteractiveDiffInterpretation] = {
+  override def interpret(diff: InteractionDiffResult, interactions: Vector[HttpInteraction]): Seq[InteractiveDiffInterpretation] = {
+    val interaction = interactions.head
```
devdoshi, line 22: I think we could have kept this interface intact and added a new one that takes multiple interactions instead

core/optic/shared/src/main/scala/com/useoptic/diff/initial/DistributionAwareShapeBuilder.scala
```diff
@@ -0,0 +1,265 @@
...
+
+  def toMap: Map[JsonTrail, Vector[Option[JsonLike]]] = _internal.toMap
+
+  def valuesForTrail(trail: JsonTrail): Vector[Option[JsonLike]] = _internal.getOrElse(trail, Vector.empty)
+
+  def hasTrail(trail: JsonTrail) = _internal.contains(trail)
+
+}
+
+class ShapeBuilderVisitor(aggregator: TrailValueMap) extends JsonLikeVisitors {
```
devdoshi, line 227: We should use the JsonLikeAndSpecVisitors from now on - let me know if you want a hand migrating

acunniffe: Let’s do it during our session tomorrow. Didn’t realize it was new!

acunniffe:
In #146

PR #145, opened 2020-05-04
Fix bug where shape with nullable trail did not yield a shape description

JaapRood:
Fixes bug introduced by #136, to unblock #138, allowing `EndpointReport` to render with mock data.


JaapRood:
My attempt at fixing the issue, after finding the stack trace hidden in the thrown error. Not sure how to get a shape description for a `NullableItem` which doesn't hold a `shapeId`. Won't merge without your approval, as I know you guys might be working on this in parallel. Please either close the PR if you guys covered separately, or merge if the other work is going to be a while :)

JaapRood:
@devdoshi I can't see any conflicts with #146, but it _is_ still blocking the merging of other work, so I'm going to go ahead and merge this. Worst case, the feature behind a flag remains broken, which it already is anyway 😅.

PR #146, opened 2020-05-04
Mightrated Shape Render & Distribution Aware Shape Diff Over to new Traversers

acunniffe:


core/optic/shared/src/main/scala/com/useoptic/diff/interpreters/Interpreter.scala
```diff
@@ -7,5 +7,6 @@ import scala.scalajs.js.annotation.{JSExportAll, JSExportDescendentClasses}
 
 
 trait InteractiveDiffInterpreter[T] {
-  def interpret(diff: T, interaction: HttpInteraction): Seq[InteractiveDiffInterpretation]
+  def interpret(diff: T, interactions: Vector[HttpInteraction]): Seq[InteractiveDiffInterpretation]
+  def interpret(diff: T, interaction: HttpInteraction): Seq[InteractiveDiffInterpretation] = interpret(diff, Vector(interaction))
```
devdoshi, line 11: So it looks like you did follow the suggestion to keep the previous interface intact. However, you did it by mapping onto the one that takes many. There should be a distinction between interpreters that interpret based on one interaction only, and interpreters that are meant to interpret many interactions. So, all of the existing interpreters should revert to keeping the original interface that only takes 1 interpretation, instead of supporting the interface that takes multiple interpretations and then always choosing the first one.

core/optic/shared/src/main/scala/com/useoptic/ux/DiffPreviewer.scala
```diff
@@ -85,206 +84,447 @@ object DiffPreviewer {
...
-          diffs = diffsByTrail(bodyTrail)
+          json.asJson,
+          diffs = diffsByTrail(jsonTrail)
         ))
       }
     }
   }
 }
 
+//class ExampleRenderVisitor(spec: RfcState, diffs: Set[ShapeDiffResult]) extends JsonLikeVisitors with ExampleRenderVisitorHelper {
```
devdoshi, line 333: we can delete all the dead code now

devdoshi:
Other than my small comments, looks good! Haven't played around with edge cases but am planning to have the automatic testing setup running against a subset of this functionality soon.

acunniffe:
Great -- lmk if you need any help wiring up something that can test the SideBySideDiffViewer Headleess

core/optic/shared/src/main/scala/com/useoptic/types/capture/Capture.scala
```diff
@@ -12,9 +12,9 @@ case class GroupingIdentifiers(agentGroupId: String,
 case class ShapeHashBytes(bytes: Vector[Byte])
 
 @JSExportAll
-case class ArbitraryData(asShapeHashBytes: Option[ShapeHashBytes],
-                         asJsonString: Option[String],
-                         asText: Option[String])
+case class ArbitraryData(asShapeHashBytes: Option[ShapeHashBytes] = None,
```
devdoshi, line 15: shouldn't we do companion object .empty here instead?

acunniffe: It's more for the case where you want to create ArbitraryData and only want to specify one of the properties ie `ArbitraryData(asJsonString = Some(example))`

devdoshi: Same principle - it would be better to have builders. ArbitraryData.fromJson(...)

PR #152, opened 2020-05-06
Never learn the request body for an interaction that yields a non 2xx resposne

acunniffe:
In Optic 6 + 7 requests that yielded a non-2xx response were never used while learning the spec. 

This logic could be in the visitor -- but since we are using that same visitor to do testing and verification it didn't seem right to include this documentation-only concern there. 

I elected to put the logic in the Diff Manager right after we compute the diff for the set of interactions. First Optic filters out interactions for `UnmatchedRequest*` that are not 2xx, then it removes keys from the Diff -> Interaction Map that are now empty.


devdoshi:
Good job maintaining the separation of concerns! It should be 2xx and 3xx, not just 2xx though right?

PR #153, opened 2020-05-07
Fix bug where contributions are not saving

devdoshi:


JaapRood:
@devdoshi there seem to be a lot of reformatting changes. Would you mind [configuring Prettier for WebStorm](https://prettier.io/docs/en/webstorm.html)? Ever since @acunniffe did that it stopped the tug-of-war and muddling of git history :)

devdoshi:
> @devdoshi there seem to be a lot of code-formatting changes. Would you mind [configuring Prettier for WebStorm](https://prettier.io/docs/en/webstorm.html)? Ever since @acunniffe did that it stopped the tug-of-war and muddling of git history :)

My Webstorm was already set up to apply Prettier so these formatting changes are reflecting that -- it seems like Prettier was not run on all of our files. Maybe we should do a one time normalization? Anyway I'll try to commit the formatting changes separately from my actual changes until we do that.

JaapRood:
@devdoshi ahhh, you're totally right! I was reading the diffs the wrong way around, it's actually Prettier applying it's formatting. Sorry, my bad.

> Maybe we should do a one time normalization?

Losing a bunch of Git history is the biggest issue with that, so formatting file by file as we work on it for a while seemed like good way to go. There's alternatives, like [precise-commits](https://www.npmjs.com/package/precise-commits), but last I used that it had some really funky outcomes and didn't broke git's line-by-line staging. Code changing between hitting "commit" and the commit being written turns out to a huge loss in control.

devdoshi:
Yeah doing it as we go is fine with me.

acunniffe:
Nice work -- tested this and it's working nicely.

PR #158, opened 2020-05-07
Analytics should cover error cases, and key parts of the user journey 

acunniffe:


devdoshi:
Can you rearticulate the desired properties? I'm not sure a new process for every analytics event is warranted here. If we control when the process exits then we can wait for things to finish right?

acunniffe:
Yeah there are valuable analytics that are not known until the end of the CLI command. For instance, it's helpful to know how many interactions Optic saw whenever a user runs `api start`. 

The best case for flushing the analytics buffer is ~3 seconds. In my testing today I noticed the process was hanging too long. Since this is something we want them to feel comfortable constantly starting/stopping there shouldn't be a lag. 

The main use case for the `api start` command is using it interactively as you develop. So
1. developers make a code change
2. restarts server w/ Optic. 
    - CTRL-C, up-arrow, enter should be really easy. 
3. tests change
[repeat] 


--- 

There's a normal `track(<event>)` in place for all other cases.

PR #160, opened 2020-05-07
Memoized the Spec Resolvers

acunniffe:
@devdoshi asked me to take a look at memoizing the Spec Resolvers today because they account for a lot the compute in diffing process. 

An example of one of the Spec Resolver is `resolveTrailToCoreShape`. This resolver takes the spec and a shape trail and returns a ShapeEntity, CoreShapeKind, and bindings so the differ knows what to compare the observed value to. 

If you had 100 interactions for an endpoint, this computation would happen >=100 times in most cases, and possibly many more if the Shape Trail matches a repeating list item. 

All the Spec Resolvers are pure functions, and while not particularly expensive, their call #s can grow exponentially as payloads grow/nest.

----

I ended up using scalaz's `immutableHashMapMemo` which lets us memozie by the input tuple for each function. 

scalaz effectively manages building the cache, looking up values, and computing an output if no cache entry is present.

It doesn't manage the lifecycle of the cache so I wrapped it in `MemoizedResolvers`, a class that is passed around the diff traversed and visitors implicitly.  If we only wrapped the current methods, we'd end up building an unbounded cache since they live in a Static Object.

For the diff compute, I have the implicit instantiating outside of the interactions iterator so the cache is used across all (n) interactions before being released from memory. This should make big diffs a lot more efficient. 

The next time you run the diff it'll start w/ a fresh cache, but presumably, that's alright because the diff will only be computed in response to new interactions or events that yield changes to the RfcState. 

I already A, B tested this with the poke-empty example session and things are noticeably faster -- still slow, but definitely an improvement


acunniffe:
@devdoshi Made some updates here -- you can now instantiate either the: 
```
new UncachedSpecResolvers(<spec>)
```

```
new MemoizedResolvers(<spec>)
```

Now we can properly benchmark the impact of the memorization by toggling between implementations. 

I also moved the Json resolvers to their own static Object. Those do not rely on the spec and don't seem like they would benefit much from caching.

core/build.sbt
```diff
@@ -19,6 +19,7 @@ lazy val optic =
       version := "0.1-SNAPSHOT",
       libraryDependencies ++= Seq(
         "io.github.cquiroz" %%% "scala-java-time" % "2.0.0-RC2",
```
devdoshi, line 21: side note, I'm probably going to get rid of this in favor of the pluggable approach we've been using so the js side can provide an implementation of this when we need it. Hope it brings the bundle size down!

core/optic/shared/src/main/scala/com/useoptic/ux/DiffPreviewer.scala
```diff
@@ -28,6 +28,8 @@ object DiffPreviewer {
   ///@todo make return optional
   def previewDiff(jsonLike: Option[JsonLike], spec: RfcState, shapeIdOption: Option[ShapeId], diffs: Set[ShapeDiffResult]): Option[SideBySideRenderHelper] = shapeIdOption map { shapeId =>
 
+    implicit val Resolvers = new MemoizedResolvers(spec)
```
devdoshi, line 31: once I play around with this a little bit I'll probably change this to be a parameter since it should be constructed once when we build the `RfcContext` alongside the `cachedQueryResults`

JaapRood:
We should be sure to create profiles before and after these changes, so we can:

- measure whether it makes any difference in responsiveness of the UI
- make sure this trade-off of time vs space isn't causing issues elsewhere (memory bloat, leaks, GC pauses, etc)

devdoshi:
Nice work, I've applied it to #177

PR #164, opened 2020-05-11
Debug Captures can be generated from a simple command

acunniffe:
Users can now generate a shape hashed debug capture from the developer console. 

I also wrote [Product docs we can share with users](https://docs.useoptic.com/debugging-diffs) and put links to this on the Error Page.

### DIFF
--- a/workspaces/ui/src/App.js
+++ b/workspaces/ui/src/App.js
@@ -80,7 +80,12 @@ function AppError() {
               {' '}
               email it to our team
             </a>
-            . Just open Developer Tools and type `debugOptic()`.{' '}
+            or{' '}
+            <a href="https://github.com/opticdev/optic/issues">
+              {' '}
+              open a GitHub issue
+            </a>{' '}
+            Just open Developer Tools and type `debugOptic()`.{' '}
             <a target="_blank">Read more here</a>
           </li>
           <li>
--- a/workspaces/ui/src/App.js
+++ b/workspaces/ui/src/App.js
@@ -66,7 +66,7 @@ function AppError() {
 
         <h5>
           Please feel free to{' '}
-          <a href="mailto:hi@useoptic.com?subject=Optic App crash report">
+          <a href="mailto:aidan@useoptic.com?subject=Optic App crash report">
             contact us about it.
           </a>
         </h5>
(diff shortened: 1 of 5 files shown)


workspaces/ui/src/utilities/debug-dump.js
```diff
@@ -0,0 +1,33 @@
...
+    console.log('sanitizing data...');
+    sessionCleaned.samples.forEach((i) => {
+      i.request.body.value.asJsonString = null;
+      i.request.body.value.asText = null;
+
+      i.response.body.value.asJsonString = null;
+      i.response.body.value.asText = null;
+    });
+
+    const output = JSON.stringify({
```
devdoshi, line 18: I'd probably `JSON.stringify(..., null, 2)` so it's more friendly to look at by default

workspaces/ui/src/utilities/debug-dump.js
```diff
@@ -0,0 +1,33 @@
...
+      session: sessionCleaned,
+      examples: {},
+    });
+
+    const blob = new Blob([output], { type: 'application/json' });
+    const url = window.URL.createObjectURL(blob);
+
+    const link = document.createElement('a');
+    link.href = url;
+    link.download = `debug-capture-${captureId}.json`;
```
devdoshi, line 29: In addition to the captureId I'd put a timestamp so the system doesn't have to rename them

acunniffe: good idea!

workspaces/ui/src/utilities/debug-dump.js
acunniffe: The shape hash makes it pretty ugly to look at when formatted. each bit gets it's own line. And it takes up a whole lot more disk space

workspaces/ui/src/utilities/debug-dump.js
```diff
@@ -0,0 +1,33 @@
+import * as deepCopy from 'deepcopy';
+
+export function debugDump(specService, captureId) {
+  return async function () {
+    const events = await specService.listEvents();
+    const session = await specService.listCapturedSamples(captureId);
+    const sessionCleaned = deepCopy(session);
```
devdoshi, line 7: I think we would be fine with a `const copy = JSON.parse(JSON.stringify(x))` here instead of adding a dependency

workspaces/ui/src/App.js
```diff
@@ -66,30 +66,29 @@ function AppError() {
...
         </h5>
 
         <p>Meanwhile, you could try the following:</p>
 
         <ul>
+          <li>
+            Create a debug file and{' '}
+            <a href="mailto:aidan@useoptic.com?subject=Optic App crash report">
+              {' '}
+              email it to our team
```
devdoshi, line 81: How do we feel about saying they can open a github issue with it as well?

workspaces/ui/src/App.js
```diff
@@ -66,30 +66,34 @@ function AppError() {
...
-          <li>
-            Try your browser in{' '}
-            <a
-              href="https://en.wikipedia.org/wiki/Privacy_mode"
-              target="_blank"
-            >
-              "privacy" or "incognito" mode
-            </a>
-            .
-          </li>
```
JaapRood, line 92: Removed some legit cache-busting / broken cookie / local storage / service worker advice! Not that we have any of that, at the moment 😅

acunniffe: Yeah since it probably wouldn't help I didn't want to give them a false path to go down, esp if it destroyed their current session (which I wanted them to provide) @JaapRood

PR #167, opened 2020-05-12
Add development route for easy loading of user debug sessions while preventing publishing it

JaapRood:
Now that we've made it easier for users to share their debug captures / sessions with us, we should make sure we can work with them easily, while making sure we don't ever leak it any further.

At the core, this change basically gives us a git-ignored version of `public/example-sessions`, so we won't accidentally commit one of these sessions. Any capture files placed in `public/customer-sessions/` will be available through `localhost:3000/development/customer-sessions/:sessionId`.

Since the debug context is aware of this being a customer file, the hook is there to add additional measures in the future, if we uncover the need for them. For example, a little debug banner warning we're looking at customer data, something that suggests deleting after we've been using it for a while, etc. 

In addition, I can imagine a post-install script that purges old captures from this folder, to make sure we don't keep these captures hanging around locally for any longer than we have to.

For now, though, I think this is a good first step worth adding.

### DIFF
--- a/workspaces/ui/.gitignore
+++ b/workspaces/ui/.gitignore
@@ -25,3 +25,5 @@ yarn-error.log*
 
 # ide
 .idea/
+
+public/customer-sessions
--- a/workspaces/ui/src/contexts/MockDataContext.js
+++ b/workspaces/ui/src/contexts/MockDataContext.js
@@ -6,13 +6,14 @@ MockDataContext.displayName = 'MockDataContext';
 // only expose the Provider component, but hide the Consumer in favour of hooks
 export const { Provider } = MockDataContext;
 
-function createDebugSession({ sessionId }) {
+function createDebugSession({ sessionId, customerSession }) {
+  const collection = customerSession ? 'customer-sessions' : 'example-sessions';
   let fetchedData = null;
   async function getData(refresh = false) {
     if (!fetchedData || refresh) {
       // only fetch data once
-      fetchedData = fetch(`/example-sessions/${sessionId}.json`, {
-        headers: { accept: 'application/json' }
+      fetchedData = fetch(`/${collection}/${sessionId}.json`, {
+        headers: { accept: 'application/json' },
       }).then((response) => {
         if (response.ok) {
           return response.json();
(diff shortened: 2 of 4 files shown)


workspaces/ui/src/entrypoints/index.js
```diff
@@ -1,15 +1,20 @@
...
-import {Route, Switch} from 'react-router-dom';
+import { Route, Switch } from 'react-router-dom';
 import Development from './development';
 import LocalCli from './localcli';
 import WelcomePage from '../components/support/WelcomePage';
 
 export default function TopLevelRoutes() {
   return (
     <Switch>
+      <Route
```
devdoshi, line 10: I think the path that would have fewer conditionals and allow greater composition would be to fork the Development component into a PrivateSession/CustomerSession component.

devdoshi: which we would then mount here as ```<Route strict path="/testing/private-examples/:sessionId" component={PrivateExampleRoot} />```

workspaces/ui/src/entrypoints/development.js
```diff
@@ -1,27 +1,33 @@
...
+} from '../contexts/MockDataContext';
+import { ApiRoutes } from '../routes';
+import { Provider as BaseUrlContext } from '../contexts/BaseUrlContext';
 
 export default function Development(props) {
-
   const match = useRouteMatch();
-  const {sessionId} = useParams();
+  const { sessionId } = useParams();
+  const isCustomerSession = !!matchPath(
```
devdoshi, line 14: if we compose the components differently, as described in my other comments, then we would not need this here

workspaces/ui/src/contexts/MockDataContext.js
```diff
@@ -6,13 +6,14 @@ MockDataContext.displayName = 'MockDataContext';
 // only expose the Provider component, but hide the Consumer in favour of hooks
 export const { Provider } = MockDataContext;
 
-function createDebugSession({ sessionId }) {
+function createDebugSession({ sessionId, customerSession }) {
+  const collection = customerSession ? 'customer-sessions' : 'example-sessions';
```
devdoshi, line 10: if we compose things as I describe in my other comments, this parameter would be applied/injected directly from outside

workspaces/ui/.gitignore
```diff
@@ -25,3 +25,5 @@ yarn-error.log*
 
 # ide
 .idea/
+
+public/customer-sessions
```
devdoshi, line 29: consider renaming to the more generic "private-sessions"

workspaces/ui/src/entrypoints/index.js
JaapRood: I considered this, but decided adding another Entrypoint like `Development` would be more confusing than this 1 additional parameter is worth, especially as there is only a single parameter as of yet (session id). Let the private session entrypoint prove the need for it's existence, that it's truly some novel concept and we can easily change this.

PR #169, opened 2020-05-13
Gracefully handle report resources not being able to be found

JaapRood:


workspaces/ui/src/contexts/TestingDashboardContext.js
```diff
@@ -59,3 +86,66 @@ export function useEndpointPath(captureId, endpointId) {
 }
 
 export { queriesFromEvents } from '../services/TestingService';
+
+// Reducers
+// --------
+
+function initialServiceState() {
```
devdoshi, line 93: have you checked out xstate? https://xstate.js.org/

workspaces/ui/src/services/TestingService.js
```diff
@@ -144,6 +166,24 @@ export async function createExampleTestingService(exampleId = 'todo-report') {
   return new ExampleTestingService(orgId);
 }
 
+export class TestingServiceError extends Error {
+  static type = 'testing-service-error';
```
devdoshi, line 170: Would it be better to use a Symbol here?

workspaces/ui/src/services/TestingService.js
```diff
@@ -144,6 +166,24 @@ export async function createExampleTestingService(exampleId = 'todo-report') {
...
+  // don't rely on `instanceof` in JS, it's a mistake
```
devdoshi, line 171: Interesting, I'll have to double check what our typescript code is generating too. https://github.com/Microsoft/TypeScript-wiki/blob/master/Breaking-Changes.md#extending-built-ins-like-error-array-and-map-may-no-longer-work

workspaces/ui/src/services/TestingService.js
JaapRood: You could if type should be private, but exposing `type` as string on Error seems to be becoming the defect way in the wider ecosystem, so 🤷‍♂️.

workspaces/ui/src/contexts/TestingDashboardContext.js
JaapRood: Never have! Pretty happy with the simplicity of this, but I can see how it might be useful it the amount of states and state transitions increased.

workspaces/ui/src/services/TestingService.js
devdoshi: I think i'd do both, use symbol for checking so you don't have to worry about collisions, type string for printing, and since I presumably have to fix it in our typescript code in several places makes sense to make it a generic pattern.

workspaces/ui/src/contexts/TestingDashboardContext.js
devdoshi: I'd say this is right at the point where I'd want to switch over to state machines, only because we don't have it set up yet. So I wouldn't block this change, but if we had it set up I'd want to use it by default. I think even in the simplest case it's never worse / more complex.

JaapRood: Must admit my personal point of friction lies quite a bit further down the road. As long as I can sequentially read over a function and see what's going on without massive if statements to only allow the right transitions, I've never really had trouble getting them to work correctly and robustly. Either way, for now I've left this as is, choosing to ship less code. Since it's completely unabstracted and pure code, the path for applying an abstraction on top of it when necessary is completely open.

workspaces/ui/src/services/TestingService.js
JaapRood: I'll soon be porting this to Typescript so the mock and production implementation can share an interface, so let's tackle this then.

PR #172, opened 2020-05-14
Define a TestingService interface

JaapRood:


JaapRood:
Trying to create an effective interface for the `TestingService` and with the goal of making sure the mock and production versions behave reasonably, I was confronted with how to handle errors that we expect such a service to have, like a 404. 

Depending on the endpoint, there are status codes we _do_ and _don't_ expect, and want to treat as such. We also want to make sure that the example mocked version can simulate the production ones, so we can in turn make sure the UI supports it properly. An example of this was a 404 on loading a capture with unknown id causing an app crash.

Looking at various approaches of other TypeScript projects, I landed on an `Result<Ok, Err>` approach, so that per endpoint we can determine what errors are expected. Using type guards, the `result.isOk()` and `result.isErr()` used in branch, will allow further correct use as well.

@devdoshi with all that in mind, I'm hoping to get some review on a conceptual level. Do you think this is useful, or just complicating things further than we would want. Nit picks and optimisations aside, see any glaring holes in the way it's implemented?

devdoshi:
Nice! I like where your head is at and the implementation does give a lot more explicit ways to handle failures. Usually I keep it simple, where a promise either fulfills or it throws an exception/rejects, and there is no fine-grained error handling by default, since there is typically nothing the user can do about it aside from wait and/or refresh, and that is often sufficient at this stage. The mocking experience is basically the same; just provide an implementation that throws. 

I'm happy to use it, as it isn't a high-churn piece of code so it doesn't seem like it will be costly to maintain. However, I think the programming model could be more ergonomic for the end-developer, like pattern matching js-style:
```
performRequest()
  .then(matchResponseStatus<HintThatGivesConfidenceAboutTheKeysYouNeedInTheHandlerMap>({
    200(body) {
    },
    // or equivalently 
    [Ok](body) {

    }
    404(body) {
    },
    // or equivalently
    [NotFound](body) {
    }
  })
```
But, I'd have to see what the actual service implementation looks like in your approach with applying the same wrappers to fetch().

JaapRood:
Good to hear you like the direction. The pattern matching looks neat and expressive as far as JS goes, an enhancement we should be able to add on top of this. For now I'll keep it to the level we've got it at now, just so we can move on. Anyone touching this code again and dealing with multiple handling of the response should feel free to add this!

PR #173, opened 2020-05-14
agent-cli groundwork

devdoshi:
This refactoring fixes some bugs and structural issues and sets us up to:
- have an agent-cli that sends data to our saas backend
- be able to have the ui grab diffs from the cli-server instead of loading all the interactions and computing diffs in the ui

This is not 100% ready to merge; agent-cli should be separated from local-cli and needs to have some more backend pieces wired up for the full functionality but it should work for sending interactions to the saas backend and computing reports

To test, please use the `install-local 8.0.7` command/alias and try it out and see if you run into any issues with `api start` `api agent:start` etc.


workspaces/ui/src/utilities/EndpointUtilities.js
```diff
@@ -0,0 +1,92 @@
+"use strict";
```
devdoshi, line 1: didn't mean to commit this

JaapRood: Was gonna say, we've got this as TS, right? 😅

JaapRood:
I rebased this PR onto the latest `develop` branch. It took multiple attempts, but I believe to have resolved all conflicts correctly. 

In a private convo with @devdoshi we discussed how we should get this merged ASAP, as the reorganisation of code happening in it as well in `develop` causes conflicts that'll be harder and harder to manage. The added `agent:start` local-cli command we can leave either undocumented or put behind some flag, as that's not the spot it's supposed to end up at, nor is it ready for use.

I'll do some testing locally to get an idea of whether it's stable enough. Given the amount of changes, perhaps it's a good idea for someone else to do the same.

core/optic/shared/src/main/scala/com/useoptic/ux/DiffManager.scala
```diff
@@ -33,7 +33,6 @@ class DiffManager(initialInteractions: Seq[HttpInteraction], onUpdated: () => Un
 
   def updateInteractions(httpInteractions: Seq[HttpInteraction]): Unit = {
     if (_interactions != httpInteractions) {
-      println("LOOK HERE "+ httpInteractions.filter(_.request.path == "/backend/api/user/123/realtime-rule-paginated"))
```
acunniffe, line 36: this can be removed

workspaces/cli-config/src/helpers/ignore-parser.ts
```diff
@@ -1,17 +1,17 @@
...
+  shouldIgnore(method: string, url: string): boolean;
 }
 
 export interface IIgnoreRunnable {
-  rules: IIgnoreRunnablePredicate[],
+  rules: IIgnoreRunnablePredicate[];
 
-  shouldIgnore(method: string, url: string): boolean
+  shouldIgnore(method: string, url: string): boolean;
 }
```
acunniffe, line 16: Is the Intent here that ignore rules run on the agent-cli before sending?

workspaces/cli-shared/src/captures/avro/file-system/index.ts
```diff
@@ -0,0 +1 @@
+export const captureFileSuffix = '.optic-capture.avro';
```
acunniffe, line 1: should we begin using explicit versions here? `.optic-captureV1`?

acunniffe:
@devdoshi lmk if you need help moving the memo resolvers over to this branch. Resist the urge to reduce boilerplate. I already sank 4hrs into the effort. This is about the best we get w/o custom macro.

workspaces/cli-config/src/helpers/ignore-parser.ts
devdoshi: There is a filter that the proxy uses to discard interactions that it knows should not be logged. All other interactions are recorded, and then this filter is used to limit which interactions are considered for diffs. This lets you adjust the filters without needing to re-do a capture

workspaces/cli-shared/src/captures/avro/file-system/index.ts
devdoshi: The evolution plan is to add a new implementation of the interface that uses its own suffix, and optionally another one that composes implementations to migrate or unify older versions, depending on our requirements

PR #176, opened 2020-05-18
(WIP) build and publish CLI to github package registry under a development channel 

acunniffe:
Goal: easier testing of PRs

PR #177, opened 2020-05-19
groundwork for agent-cli plus memoization and bug fixes

devdoshi:
This adds on to the work in https://github.com/opticdev/optic/pull/173 and any new PRs should build on this. I kept them separate but you should review and test this one. I found one regression with the UI not updating properly on http://localhost:3000/development/diff-test-cases/diffs/example-session so I'll need to work with @acunniffe briefly on that but otherwise seems like an overall improvement. There are still some diff edge cases but I need to look at some of the other failing tests in the meantime


core/optic/js/src/main/scala/com/useoptic/QueriesJsFacade.scala
```diff
@@ -22,12 +21,13 @@ import scala.scalajs.js.annotation.{JSExport, JSExportAll, JSExportTopLevel}
 @JSExport
 @JSExportAll
 object OASProjectionHelper {
-  def fromEventString(eventString: String): js.Any = {
+  //@TODO: pass in apiName
```
devdoshi, line 24: If I understand correctly, we are no longer using this part of the event stream and instead we are using the value from the optic.yml correct? I'm taking this opportunity to get rid of dead code.

core/optic/shared/src/main/scala/com/useoptic/diff/shapes/resolvers/ShapesResolvers.scala
```diff
@@ -0,0 +1,42 @@
...
+object ShapesResolvers {
+  case class ChoiceOutput(parentTrail: ShapeTrail, additionalComponents: Seq[ShapeTrailPathComponent], shapeId: ShapeId, coreShapeKind: CoreShapeKind, bindings: ParameterBindings) {
+    def shapeTrail(): ShapeTrail = parentTrail.withChildren(additionalComponents: _*)
+  }
+
+  case class ResolvedTrail(shapeEntity: ShapeEntity, coreShapeKind: CoreShapeKind, bindings: ParameterBindings)
+  type ParameterBindings = Map[ShapeParameterId, Option[ProviderDescriptor]]
+}
+
+trait ShapesResolvers {
```
devdoshi, line 18: We should have enforced this more strictly earlier, but we should stop accessing RfcState/ShapesState/RequestsState directly, and only access them through some additional query layer, like ShapesResolvers. Most of the places where I did not remove shapesState from the arguments are only using some simple queries like shapesState.shape() and shapesState.field() so it should be easy to move those into ShapesResolvers and remove ShapesState from the arguments. I'm imagining it should be easy to change ShapesResolvers into SpecResolvers which contains shapesResolvers and requestsResolvers, etc. when the time comes

workspaces/ui/src/contexts/RfcContext.js
```diff
@@ -93,7 +86,16 @@ function BaseRfcStore(props) {
         const eventStore = Facade.makeEventStore();
         global.eventStore = eventStore;
         if (initialEventsString) {
+          const markerNameStart = 'rfcContextAddEventsStart';
+          const markerNameStop = 'rfcContextAddEventsStop';
+          //performance.mark(markerNameStart);
```
devdoshi, line 91: I did a little profiling and this is usually taking about 30-50ms compared to about 200ms for DiffManager.updateInteractions. Once I implement timing and add some larger scenarios to our snapshot tests we should get a better sense of any bottlenecks within the core.

workspaces/ui/src/services/testing/TestingService.js
```diff
@@ -0,0 +1,3 @@
+"use strict";
```
devdoshi, line 1: sigh...is there a way to automatically ignore these? I would have put them in a separate package in our workspace so it uses the same typescript pathway as (most of) the rest of the packages

workspaces/ui/src/components/docs/DocsPage.js
```diff
@@ -256,10 +256,10 @@ export const EndpointDocs = (props) => {
   const fingerPrint = `${endpointDescriptor.httpMethod} ${endpointDescriptor.fullPath}`;
   usePageTitle(`${purpose || fingerPrint} - API Documentation`);
 
-  const { rfcService, rfcId } = useContext(RfcContext);
-
+  const { rfcService, rfcId, cachedQueryResults } = useContext(RfcContext);
+  const { shapesResolvers } = cachedQueryResults;
   const currentRfcState = rfcService.currentState(rfcId);
-
+  const diffPreviewer = new DiffPreviewer(shapesResolvers, currentRfcState);
```
devdoshi, line 262: I suspect my ui bug is here

workspaces/ui/src/services/testing/TestingService.js
JaapRood: It shouldn't be building them in-place in the first place. Perhaps something your IDE is doing?

JaapRood:
@devdoshi @acunniffe be sure to pull before continuing your work on this, as #173 got merged and I rebased this onto `develop` (reducing this pr back to 1 commit, the diff between #173 and this).

workspaces/ui/src/contexts/RfcContext.js
```diff
@@ -181,12 +183,9 @@ function LocalRfcStore(props) {
...
   async function handleChange({ eventStore, rfcId }) {
     try {
-      const response = await specService.saveEvents(eventStore, rfcId);
-      if (!response.ok) {
-        const text = await response.text();
-        throw new Error(text);
-      }
+      await specService.saveEvents(eventStore, rfcId);
     } catch (e) {
+      debugger;
```
JaapRood, line 188: Bit of left over debug code.

workspaces/ui/src/contexts/RfcContext.js
JaapRood: Did you end up saving any profiles to share? Would be interesting to see how it compares between machines / deployments.

core/optic/js/src/main/scala/com/useoptic/QueriesJsFacade.scala
acunniffe: Yes sir, API name should come from `optic.yml`, it's not currently used anywhere in the UX so safe to make changes here

core/optic/shared/src/main/scala/com/useoptic/diff/shapes/resolvers/ShapesResolvers.scala
acunniffe: Agreed. That seems reasonable and correct to me, but probably better for a cool-down period, not a step on the critical path.

workspaces/ui/src/components/docs/DocsPage.js
acunniffe: I'll test this out.

devdoshi:
> From a scan of the many changes and some little nitpicks, it all looks good from my perspective. @devdoshi I'll plan some manual testing from the UI tomorrow morning. Are there any particular areas that you want me to look at / are concerned about being affected? Since it's memoization efforts, I'll assume to be on the look out for stale values.

Cool - the two areas of focus stem from the refactor prior to 177 and then the memoization changes in 177. In the refactor, the risks are regressions around all of the cli commands, and in 177 the risks are around stale data like you said.

devdoshi:
The other focus is making sure it publishes and installs properly (which @acunniffe is focusing on in #176

acunniffe:
I went through end-end manual testing based on the reqs on Optic8 Basecamp page. Fixed a few issues, but for the most part, seeing no regressions. 

As for user issues, the rouge diffs based on incorrect trails are no longer happening and the DistributionAwareShapeDiff is working as expected again (after inferring polymorphism there are never any diffs). 

I found a small issue that Dev is looking at now in one of the private sessions since it's user data  I will not share here.

devdoshi:
Thanks @JaapRood @acunniffe - I'll remove printlns in another PR

PR #182, opened 2020-05-21
Integrating with FeaturePeek

acunniffe:
Aidan Cunniffe is beginning to integrate this repo with [FeaturePeek](https://featurepeek.com)!

FeaturePeek creates deployment previews for every frontend pull request automatically. Your team can view your running code earlier in the product development cycle, while leaving feedback, taking screenshots and screen recordings, and filing bugs directly from the running page.

Once your deployment preview is ready, a link will appear in this PR pointing to the FeaturePeek deployment. Then it's safe to merge in your changes, and all subsequent pull requests will be watched by FeaturePeek.

### DIFF
--- a/.github/workflows/featurepeek.yml
+++ b/.github/workflows/featurepeek.yml
@@ -23,6 +23,6 @@ jobs:
     - name: Install dependencies
       run: yarn install
     - name: Build
-      run: yarn run build-ui
+      run:  cd workspaces/ui && yarn run build-local
     - name: Ping FeaturePeek
       run: bash <(curl -s https://peek.run/ci)
--- a/.github/workflows/featurepeek.yml
+++ b/.github/workflows/featurepeek.yml
@@ -1,0 +1,19 @@
+
+name: Build and ping FeaturePeek
+on: push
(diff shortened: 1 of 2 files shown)


devdoshi:
it's worth consolidating the workflows into one, so we don't have 3 doing the same stuff.

devdoshi:
if we run into issues with this way, seems like they also have a way to run a docker image - should be easy to build using the published @useoptic/ui and a static server

acunniffe:
Featurepeak team fixed this issue -- appears that these work as expected now. 

Re: Separate builders -- I don't see a benefit, doesn't seem like GitHub throttles or queues them and I'd rather keep task pure and focused on one concern.

devdoshi:
I noticed that FeaturePeek proxies everything -- it made our logo became janky, which I hadn't noticed when running locally (which makes sense). No big deal but typically when I have an image I set its box size so it doesn't disrupt the layout when it isn't there. And also it should be a link somewhere useful -- I know I have clicked on it a bunch locally expecting it to take me somewhere.

PR #185, opened 2020-05-22
Changes required for backend

devdoshi:


package.json
```diff
@@ -6,7 +6,7 @@
     "copy-domain": "echo copied",
     "build-ui": "cd workspaces/ui && yarn run build-local",
     "copy-ui": "rm -rf ./workspaces/cli-server/resources/ui && mkdir -p ./workspaces/cli-server/resources/ui && cp -R ./workspaces/ui/build/* ./workspaces/cli-server/resources/ui/",
-    "postpull": "yarn install && yarn run build-domain && yarn run build-workspace && yarn run build-ui && yarn run copy-ui && echo 'Run $ source export_development_aliases.sh'",
+    "postpull": "printf 'Run these commands:\n$ source sourceme.sh\n$ optic-install\n'",
```
JaapRood, line 9: Haha, running this and seeing the message made me laugh 😆. All these other yarn scripts still seem perform their own stuff, yet postpull is the only one changed. I wonder if that has anything to do with me mentioning I still use it 🤔 🤫.

All of that aside, it makes sense to have all of that logic in one place, but I think we should reconsider the names of some of these functions. _I_ can barely keep `optic-install`, `install-and-publish` and `install-local` differentiated in my head, not to mention how anyone not familiar with these commands would figure out the difference and decide which is appropriate in which circumstance?

I'd suggest starting off renaming `optic-install` to `build-optic`, matching `watch-optic`. Hopefully that communicates that this is all about making sure the modules are built. I know that doesn't include building the ui workspace, but it's confusing how that's not part of an install anyway (until you work on it), so something to be considered separately.

From there, we could rename `install-and-publish` to `build-and-publish-locally`, indicating we're making a fresh build, to then publish to the local registry. Especially the "local" part is important here I think, as I've double checked more than twice now to make sure I wasn't actually writing to public registry (even though I don't have permission for that atm).

With the other two tagged as "build", that should differentiate `install-local` from those commands as the installing of the built and published product. Perhaps something as `install-from-local-registry` might do an ever better job of that, but might also be unnecessarily verbose.

devdoshi: Happy to rename them however, I'd prefer a common prefix like optic-* for autocomplete like we have on the backend equivalent. Just need to update anything like our Github Actions that might depend on them

devdoshi:
The backend build should be able to pick up changes from this repo without any manual intervention now - let me know if that's not the case @JaapRood

package.json
JaapRood: > Just need to update anything like our Github Actions that might depend on them

Because of that, lets rename these functions as a separate PR. I've got some other incoming  structural changes based on this PR, that are better off merged sooner than later, too.

JaapRood:
Haven't made enough changes yet to really notice @devdoshi, time will have to tell. Lets get this merged, I've already been making changes depending on it that should be merged sooner rather than later as well. Even making a PR for those changes looks wrong now, as targeting develop it includes all the changes of this PR as well.

PR #187, opened 2020-05-24
Fixed analytics

acunniffe:


workspaces/cli-client/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@useoptic/cli-client",
-  "version": "8.0.7",
+  "version": "8.0.8",
```
JaapRood, line 3: Are you planning to instantly publish this afterwards? If not, we shouldn't be bumping versions until we roll a release. While it would seem not to matter, with the more people we work on this, the more likely it becomes that we've got concurrent version changes through pull requests. Add to that manually resolving conflicts, and you soon get stale versions, skipped versions, etc. The closer to atomic these version bumps can be (it's own "bump version" commit right before a publish, the better :)

acunniffe: I agree on the philosophy. sad to say it was an oversight -- unbumped in latest commit.

PR #189, opened 2020-05-25
Change all "main" development aliases to optic-* style.

JaapRood:
A little left over from #185, where we discussed how the development aliases:

- could do with a bit of disambiguation, making it less confusing what command does what.
- would be great if all prefixed with `optic-`, for sweet autocomplete support.


JaapRood:
FYI @devdoshi @acunniffe, don't be surprised you'll get a "Not a command" next time you source aliases and use the old ones.

acunniffe:
This is one of those welcome changes I didn’t know I wanted till it was done.

JaapRood:
Benefit of fresh eyes 😅 👀.

PR #190, opened 2020-05-25
Implement saas-client obtaining auth token from saas endpoint

JaapRood:


workspaces/local-cli/src/index.ts
```diff
@@ -1 +1,3 @@
-export {run} from '@oclif/command'
+import * as dotEnv from 'dotenv';
+dotEnv.config();
```
devdoshi, line 2: I'm not 100% sure if/when this file ever gets run. As far as I can tell, the entry point(s) are in /bin where we call setLogger. So, I imagine local-cli/src/config.ts is always getting the default value. Other than that I'm fine with there being a hardcoded default and env override

workspaces/local-cli/src/config.ts
```diff
@@ -0,0 +1,5 @@
+export default {
+  apiGatewayUrl:
+    process.env.OPTIC_LOCAL_CI__API_GATEWAY ||
```
devdoshi, line 3: LOCAL_CLI*

workspaces/local-cli/tsconfig.json
```diff
@@ -7,9 +7,8 @@
     "rootDir": "src",
     "strict": true,
     "target": "es2017",
-    "skipLibCheck": true
+    "skipLibCheck": true,
+    "esModuleInterop": true
```
devdoshi, line 11: Was there any particular reason this was required?

workspaces/local-cli/package.json
```diff
@@ -45,8 +46,10 @@
     "@types/mocha": "^5",
     "@types/node": "^10",
     "@types/react-dev-utils": "^9.0.2",
+    "@types/url-join": "^4.0.0",
     "@types/uuid": "^7.0.3",
     "chai": "^4",
+    "dotenv": "^8.2.0",
```
devdoshi, line 52: These should be regular dependencies right?

devdoshi:
Nothing blocking merge, just need to make sure they get addressed eventually as they will affect publishing and installation

workspaces/local-cli/package.json
JaapRood: You're right!

workspaces/local-cli/src/index.ts
JaapRood: It seemingly does, at least through `apidev agent:start`. Let me verify!

workspaces/local-cli/tsconfig.json
JaapRood: Yeah, basically for any CommonJS module that you try to `import` which used the `module.exports = {}` syntax instead of `exports.something = {}`, as far as I remember. With the latter `import * as` works, but not with the former. In this case, its as `url-join` that required it to be enabled.

PR #192, opened 2020-05-26
Add command to enable live contract testing

JaapRood:


workspaces/local-cli/src/commands/testing/enable.ts
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    const baseUrl = UrlJoin(Config.apiGatewayUrl, 'api/v1');
+    const saasClient = new SaasClient(baseUrl);
+    developerDebugLogger('fetching auth token');
+
+    const gettingApiToken = saasClient
+      .getApiAuthToken(config.name || 'Unknown API')
+      .catch(() => {
+        cli.action.stop;
```
devdoshi, line 36: seems like we're missing a cli.action.start() before this, and missing the invocation here? Also, are we using both this and ora for indeterminate?

JaapRood: Just a bit of code I hadn't removed yet 😅

PR #193, opened 2020-05-26
backend integration and tweaks

devdoshi:
There were some issues after the sourceme.sh changes so I fixed those and moved to underscores to make the names actually valid though they were tolerated in bash.

I reverted the esModuleInterop change to keep the clis consistent with all the other packages, except ui. I can't tell what should be the default but as far as I can tell it's never necessary. If there's a reason to switch over we should do it for all of the workspaces.

Note that with Jaap's change to the clis' `bin/run` the cli no longer magically runs the latest code and depends on lib having code so you'll have to optic_watch or rebuild it before running it again


JaapRood:
> I reverted the esModuleInterop change to keep the clis consistent with all the other packages, except ui. I can't tell what should be the default but as far as I can tell it's never necessary. If there's a reason to switch over we should do it for all of the workspaces.

Keeping the config consistent is a good idea. That said, I think the changes of UrlJoin to use `require` instead of import, is the code smell of why it's necessary. Mixing CommonJS modules and ESM, two separate module systems, within a single module is not a good idea. 

Here's what it solves:

```js
import * as fs from `fs`
// this works fine
fs.file(/* .. */)

import * as UrlJoin from 'url-join'

// this doesn't work, as UrlJoin is callable, but ES6 modules expects an object
UrlJoin('https://example.com', 'docs')

// This would allowed to be a callable in ES6, but it looks for the `default` export
import UrlJoin from 'url-join`

// inside UrlJoin, and many other common js modules
module.exports = function UrlJoin() { /*...*/ }
``` 

The single tweak `esModuleInterop` makes is to assign a `module.exports` in absence of any other exports, to be the `default` export in ESM terms. Basically any TS codebase working with Node and Node packages should have it enabled!

I'll get a PR out to switch all workspaces to have it enabled.

PR #195, opened 2020-05-27
Enable and use `esModuleInterop` for all workspaces

JaapRood:
In #193 I [try to explain why this is a good idea](https://github.com/opticdev/optic/pull/193#issuecomment-634509969):

> > I reverted the esModuleInterop change to keep the clis consistent with all the other packages, except ui. I can't tell what should be the default but as far as I can tell it's never necessary. If there's a reason to switch over we should do it for all of the workspaces.
> 
> Keeping the config consistent is a good idea. That said, I think the changes of UrlJoin to use `require` instead of import, is the code smell of why it's necessary. Mixing CommonJS modules and ESM, two separate module systems, within a single module is not a good idea.
> 
> Here's what it solves:
> 
> ```js
> import * as fs from 'fs'
> // this works fine
> fs.file(/* .. */)
> 
> // this doesn't, as ESM looks for a default export that isn't there
> import fs from 'fs'
> 
> // this import succeeds
> import * as UrlJoin from 'url-join'
> 
> // but this doesn't work, as UrlJoin is callable, but ES6 modules expects an object
> UrlJoin('https://example.com', 'docs')
> 
> // This would allowed to be a callable in ES6, but it looks for the `default` export
> import UrlJoin from 'url-join`
> 
> // inside UrlJoin, and many other common js modules
> module.exports = function UrlJoin() { /*...*/ }
> ```
> 
> Without `esModuleInterop`, TS is completely agnostic towards CommonJS, which is why you need the `import * as fs` syntax in the first place. With it, it takes the reasonable mapping between the two, making both `import fs from 'fs'` work as expected and allowing callable modules. Basically any TS codebase working with Node and Node packages should have it enabled!
> 
> I'll get a PR out to switch all workspaces to have it enabled.

This is said PR! I did the following:

1. Enabled `esModuleInterop` for all the workspaces.
2. Replaced all `import something = require('something')` with normal imports. Found bugs in our code because of it (see below), as the compiler now understands this is a module we're importing.
3. Replaced `import * as something from 'something'` with `import something from 'something` for CommonJS modules. I carefully for each dependency looked up to see if they're exposing CommonJS or ESM, to make sure no imports that should actually be `import *` were affected (like `uuid`).

The bug I found by replacing the `import = require` type statements, was `snapshot-tests` not being registered as a workspace and not having its dependencies installed properly. It could very well be that this was intentional, since when it was building, I had to stub out a method that didn't adhere to the required type. From what I could see, this code isn't actually run anywhere yet, but at least now it's building, giving us chance to keep it building in the future (code that doesn't run is code with bugs!).


JaapRood:
Tested a bunch locally, all seems good. Sorry for the bits I might have missed, see PR description for how to fix!

devdoshi:
Looks good, thanks for clarifying - I had a feeling it should be enabled but it was disabled by default in ts presumably for backwards compatibility and I never took the time to dig deeper. Glad all the workspaces are in sync now!

PR #198, opened 2020-05-28
readme: update

LouManglass:
Needs a bit of work, as we should add the demo link once it goes live. Until then, I left that out. Opening as draft


acunniffe:
SO far -- like what i see

LouManglass:
Added contribution guidelines, developer setup per drafts. Added demo link to readme, PR is ready for review.

PR #201, opened 2020-06-01
QA Fixes / Prep for End-End test suite

acunniffe:
- Made sure that install local alias adds the ci-cli and agent-cli as well as local-cli
- Fixed a missing dependency in agent-cli
- Distribution aware shape builder was running on every diff interpretation instead of when users turned it on (big performance hit)


core/optic/shared/src/main/scala/com/useoptic/ux/DiffManager.scala
```diff
@@ -288,13 +288,19 @@ abstract class PathAndMethodDiffManager(pathComponentId: PathComponentId, httpMe
 
         val previewRender = (interaction: HttpInteraction) => diffPreviewer.previewBody(interaction.response.body)
 
-        val bodies = interactions.map(_.response.body).flatMap(BodyUtilities.parseBody).toVector
-        val preview = diffPreviewer.shapeOnlyFromShapeBuilder(bodies)
 
         val previewShape = (interaction: HttpInteraction, inferPolymorphism: Boolean) => {
+          println("ABC LOOK HERE")
```
devdoshi, line 293: test harness should fail when we leak printlns

acunniffe: I removed this and added some checks. 

I think it's much wiser to check for the statements themselves in rather than hoping whatever test we run has enough coverage of domain that it catches them all. 

See changes re package-optimized @devdoshi 

The only part that doesn't work is getting the OPTIC_PUBLISH_SCOPE passed down when the publish script is run. I don't see anywhere in that script that invokes yarn run build-domain

devdoshi: Sorry, I should have clarified:
Any println, console.log, or other output that isn't part of the specified output should fail. This should be covered in our snapshot tests. It isn't the printlns specifically; for example, a client with the logger set to console.log would have the same undesired effect. We should be monitoring our command output for regressions anyway

devdoshi: I would remove the enforcement from the scripting side

devdoshi: i would have separate commands for each build. if we're in ci doing a release, use the optimized one

acunniffe: gotcha -- removed those changes. 

I like the idea in principle, but figure it won't catch everything at first, we have yet to craft some bodies we know cover the entire surface area of the diff. But in the meantime what you've proposed will catch a lot. I'll add to end-end

PR #202, opened 2020-06-01
Use a single debug .env file for all CLI packages

JaapRood:
As I started to use the `agent-cli` and `ci-cli` for local debugging, I realised that each requires it's own `.env` file, even though they all host the single flag. This is annoying, as evidenced by this ENV variable making it into the _sourceme.sh_.

This PR allows all CLI packages to use the single env file, which is only read for the dev aliases from _sourceme.sh_ of the CLIs. This should now be _the_ place for anyone to inject their own env config. 

Note that the UI workspace still has it's own .env files: these serve a different purpose (feature flags) and _should_ be kept separate: we don't want the frontend code to have access to all these various env variables.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 .idea
 node_modules
+.env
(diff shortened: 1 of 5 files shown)


devdoshi:
Cool. Glad you kept it separate from the ui env. I put it in the sourceme so the url was accessible in the bash scope for making exploratory http calls in the same shell, but we can work around that by processing the env file in there. The final consideration is that we just need to be sure each CLI config does not depend on the presence of the env file, so every config has a default

PR #207, opened 2020-06-02
Agent CLI should be less-brittle when starting

acunniffe:
I made some changes to the Agent CLI to make it less brittle in production. 

The `run` command would crash on any invalid input to `--config`. Invalid tokens, strings, json of a different shape, etc. I wrapped that destabilization in a try/catch to we can branch the code. 

Branch 1 is followed when your start config and monitoring config is valid and starts the agent as usual 

Branch 2 is followed when the start config is valid and the monitoring config is not. It starts the API and proxy as usual and swaps out the agent persistence manager for the `DoNothingCapture` -- just a proxy with no recording, batching, etc :) 

Optic CI API having downtime, or some accidental clearing of the config environment variable should never affect your ability to deploy your code to prod. 

- I also renamed `--masquerade` to `--listen` to match other proxies/http CLIs. Also, the context around the word `masquerade` is not welcome cognitive load for security teams.


workspaces/agent-cli/src/commands/run.ts
```diff
@@ -1,25 +1,30 @@
...
+import {
+  fromOptic,
+  ICaptureSaver,
+  SaasCaptureSaver,
+} from '@useoptic/cli-shared';
 import { Config } from '../config';
 //@ts-ignore
 import jwtDecode from 'jwt-decode';
+//@ts-ignore
+import niceTry from 'nice-try';
```
JaapRood, line 11: 🚨 swallowing errors, pretty much never the thing we want to do. I totally agree that the UX should be nice and that Optic service being down shouldn't stop services. Swallowing errors, however, is the total enemy to visibility and complete ignores that thrown errors could have put the process in an undefined state. 

I don't see it used anywhere, but it shouldn't even be imported. We have to do the work of actually hardening this, not just hide the issues!

JaapRood:
> Optic CI API having downtime, or some accidental clearing of the config environment variable should never affect your ability to deploy your code to prod.

Totally agree here. Expected errors in operating of a distributed system should never bring the process down. Network partitions, downtime of other systems, all of that is expected.

However, misconfiguration should always be as loud as possible and fail a deployment. There's nothing as frustration when running multiple instances on servers, with log outputs all collated, to have no idea why things aren't working, only to find out there is 1 tiny log line amongst millions indicating a failed parse somewhere. 

When users choose to deploy Optic in a place, that's a direct intent, so if they have misconfigured things, it's our jobs to let them know.

acunniffe:
Good points @JaapRood  -- I'll find a better balance between preventing Optic from being the thing that takes down your production infra and providing better visibility.

JaapRood:
Past-Jaap who spent hours sifting through aggregated logs looking for similar issues thanks you!

workspaces/agent-cli/test/run.test.ts
```diff
@@ -0,0 +1,46 @@
...
+  capture: { id: 'c7acd41a-1797-4bff-86a0-73d3c05c6bba' },
+  agentToken:
+    'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJvcHRpY0NvbnRleHQiOnsiYWdlbnRHcm91cElkIjoiNzRkMTM0ZTUtMDZhMi00NDgyLWIzMTAtZGVlZDFmZWI0MGJkIiwiY2FwdHVyZUlkIjoiYzdhY2Q0MWEtMTc5Ny00YmZmLTg2YTAtNzNkM2MwNWM2YmJhIiwib3JnSWQiOiJkMzM0ZDQzOS0yYmY2LTQ5MjQtYTk5OS1iZmVkNTNjZTFmNjUifSwiaWF0IjoxNTkxMDQ2NDgwLCJhdWQiOiJvcHRpYy1iYWNrZW5kLXYxIiwiaXNzIjoiaHR0cHM6Ly91c2VvcHRpYy5jb20ifQ.IcD8no4evid0enfkXUX3kG6ZcW3APMsCtQgYbbf6DEc',
+});
+
+const invalidStartConfig = JSON.stringify({
+  capture: { id: 'c7acd41a-1797-4bff-86a0-73d3c05c6bba' },
+  agentToken: 'eyJ___badtoken',
+});
+const invalidStartConfig2 = JSON.stringify('{NOT EVEN JSON}');
```
devdoshi, line 13: What happens if you remove the JSON.stringify?

PR #218, opened 2020-06-04
scalajs Counters should be usable from javascript

devdoshi:


devdoshi:
i'll bring these changes in my next pr

PR #220, opened 2020-06-05
Partition diffs by pathId and method

acunniffe:
This PR addresses some low-hanging, high-value fruit on the diff manager performance. 

My research/performance audit from today revealed a stale assumption: 
Every suggestion we approve requires us to re-run the entire diff. **Stale**

**Now that all shapes are distinct and aren't referenced across the endpoint boundaries, we can confidently partition the interactions by endpoint. The diff is the expensive thing so only having to run a subset of the interactions through it is an instant win** 

look at that partition

I don't think this is the final word on performance, but it made my experience using Optic with a large session at least 5x faster -- like almost usable :)


JaapRood:
Great! Knowing how an operation partitions into a valid unit of concurrency is often half the battle :) From there, not only can you avoid work, but also think about horizontal scaling.

PR #221, opened 2020-06-15
Distribution Aware Shape Builder should be memory efficient

acunniffe:
This is one of those small things I had hoped to fix during cooldown

Previous implementations of this required aggregating the possible values for each trail. and then reducing those down to a set of commands. That allocated a lot of memory and was quite slow. 

This new version records affordances for each trail, greatly reducing memory footprint and making the computation much more efficient.

My test cases all passed, and I'm confident this does the exact same thing, better


acunniffe:
closing see #223

PR #223, opened 2020-06-23
Make ids deterministic

acunniffe:
Snapshot testing the diff domain is difficult today because IDs are randomly generated. We don't want to leak domain logic into our assertions and walk the graph so this PR adds convention for passing down an implicit idGenerator that is chosen based on the entry point (deterministic for testing, random for production)

This opens a lot of doors for testing, and since it touches a ton of files it'd be helpful to get this through soon


devdoshi:
looks good! next thing to abstract out is the date-time stuff

PR #225, opened 2020-06-24
API Check failures should be reported

acunniffe:
@LouManglass found an issue where API Check was not reporting failures. It seems like an issue with how the processes are being forked. For now, it seems that removing the process.exits fix the problem, this causes 1-2 seconds of lag in some cases after the test finishes, but that's acceptable. 

We should think about how we want to ensure we get analytics from other important lifecycles ie capture finished. 

I also expanded the props so that a JSON string wasn't being sent into mixpanel. This should help us learn a lot more from how users call check 

Screen Shot 2020-06-24 at 10 27 10 AM
Screen Shot 2020-06-24 at 10 26 35 AM


LouManglass:
API check looking great! FYI, Run is still a JSON string (see screenshot)

If it's quick, would be cool to fix in this PR. If not, it can be a separate thing. I'd be willing to take it on as a learning exercise.

Screen Shot 2020-06-24 at 10 51 39 AM

acunniffe:
Yeah they're different pathways, I can do that next.

PR #226, opened 2020-06-26
added key component to response descriptions

trulyronak:
Fixed issue where changing response descriptions appeared to not make any changes

the issue was that react wasn't updating the components. since Response Descriptions were mapped via `ContentTabs`, it needed a `key` component in order to trigger an update when the props changed

also in this pr
- added `different-responses.json` to test this fix
- updated `yarn.lock`


devdoshi:
Nice work!

devdoshi:
Looks like you're taking a look at the other usages of the component for the same issue - awesome!

trulyronak:
modified it to use a key for all things for ContentTab - i’m not 100% if this is the best key - right now I'm combining the `contentTypeTab`, `responseStatusCode,` and `responseContentType` to create a key for each one, but if there's a uuid field I'm missing that would be much better

acunniffe:
Seems like a fine place to me! nice work sir. Merging now

PR #227, opened 2020-06-26
Fixed issue where scrolling was preserved when navigating

trulyronak:
If you were all the way scrolled down on the documentation page and clicked on a request, you'd nothing because you’re scrolled down (from before) - I made it scroll you back to the top or something after you click on a route (well really whenever you navigate to a page)

Added [Scroll to Top Component](https://reacttraining.com/react-router/web/guides/scroll-restoration/scroll-to-top) to fix this, following [stack overflow](https://stackoverflow.com/questions/36904185/react-router-scroll-to-top-on-every-transition).


trulyronak:
That makes a lot of sense - I'll take a look at implementing it via the `Page` component

trulyronak:
@JaapRood I think I addressed your concerns in the latest change - I placed the `ScrollToPosition` component in the `Page` component. While it defaults to (0,0), you can override the x or y accordingly.

One thing I think that might be useful (I can look into adding this) is maybe making it auto scroll to a specific component. Unsure how I'd go about finding a specific component on the page, but I think it should be doable

trulyronak:
> it might be worth asking ourselves whether this logic needs its own component `ScrollToTop`

Honestly, when looking at the current `Page` component (and seeing how it handles the title logic within the component already), I think it's worth just adding the scroll code in the component, as I can't think of another place where I'd use it. There's an argument to be made about abstracting all logic out into its own component, but I think in this case it just makes most sense to keep the behavior all inside the `Page` component — I went ahead and made that change, but I'm not 100%. on how I did it.

workspaces/ui/src/components/Page.js
```diff
@@ -34,6 +34,13 @@ export default function Page(props) {
     document.title = `${title} - Optic`;
   }, [title]);
 
+  const { pathname } = useLocation();
+  const scrollToTop = true; // change to be determined by page context
+
+  useEffect(() => {
+    window.scrollTo(0, 0);
+  }, [pathname && scrollToTop]);
+
```
JaapRood, line 43: Great job, almost there! 🙂

- if `scrollToTop` comes from the `props` object, it can be used like `<Page scrollToTop={true}></Page>` in other components. There's [an excellent guide by the ReactJS team about props and how to use them](https://reactjs.org/docs/components-and-props.html), if you'd like to look that up.
- the `scrollToTop` must be used inside the `useEffect` to trigger the window scrolling conditionally. The second parameter `[pathname && scrollToTop]` is monitored to see whether the effect must be run. _Any change in that value_ causes the effect to be ran again. For more details, see [this guide to how useEffect works](https://reactjs.org/docs/hooks-overview.html#effect-hook).

JaapRood:
> Honestly, when looking at the current Page component (and seeing how it handles the title logic within the component already), I think it's worth just adding the scroll code in the component, as I can't think of another place where I'd use it. 

I totally agree! 👍 Even _if_ something can be it's own function or component, doesn't mean you should. As soon as something is it's own function that can be called from elsewhere, other code _will_ eventually use it. 9 out of 10 times that code has slightly different requirements, making it hard to make changes to it in the future. Unless some code is particularly error-prone, complicated or you have a really good understanding of what it _should_ and _shouldn't_ do, inlining is the more pragmatic and durable choice.

workspaces/ui/src/components/Page.js
trulyronak: adjusted it! since the default behavior is not to scroll to the top, I also specified for `DocsPage` to pass in `scrollToTop`

PR #229, opened 2020-06-30
Surfaced proxy port to user

trulyronak:
### The issue
It isn't entirely clear when using Optic where the user should be using their API — most api boilerplates and start methods have some sort of default log that outputs where the API is listening, but when using optic, this results in the wrong port being displayed.

By logging it to the user, we can avoid any confusion, and make it more clear which port they should be using to make sure optic monitors their requests.


acunniffe:
@trulyronak see `fromOptic` which does formatting like this
Screen Shot 2020-07-06 at 9 01 55 AM

Also I think the copy can better. "Optic will monitor requests made to ....."

trulyronak:
Gotcha - fixed it accordingly!

PR #231, opened 2020-07-01
cli: updated documentation

LouManglass:
Some minor CLI documentation tweaks. dev builds succeed, ran help against each command/topic as well.


devdoshi:
Looks good - do we want to include docs links too?

LouManglass:
That's something Aidan and I were discussing - what would be the most helpful to different user personas. There's some upcoming docs work we're still sorting out, and part of that will be more directly useful docs to each step of the flow (capturing & documenting diffs, organizing capture sessions, capturing & reporting). That'll be the best place to link, though if you think it's a blocker we could always link to our troubleshooting page for now and update it later.

PR #232, opened 2020-07-01
moved logfile to be inside optic folder and ignored it closes #68

### LINKED ISSUE #68: [Enhancement] Auto .gitignore .optic-daemon.log
Upon installation, a log file is created in the local directory but is not automatically gitignored.

`.optic-daemon.log` could be added to .gitignore automatically.

Alternatively, it could be placed in .optic and gitignored there.

trulyronak:
## Key changes
- `.optic-daemon.log` is no longer hidden, and is now in `.optic/`
- `optic-daemon.log` is in the default created gitignore

### DIFF
--- a/workspaces/cli-config/src/index.ts
+++ b/workspaces/cli-config/src/index.ts
@@ -234,7 +234,6 @@ export async function createFileTree(config: string, basePath: string) {
       path: gitignorePath,
       contents: `
 captures/
-optic-daemon.log
 `,
     },
     {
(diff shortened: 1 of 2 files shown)


workspaces/cli-config/src/index.ts
```diff
@@ -234,6 +234,7 @@ export async function createFileTree(config: string, basePath: string) {
       path: gitignorePath,
       contents: `
 captures/
+optic-daemon.log
```
devdoshi, line 237: I believe to hide any existing file you want ".optic-daemon.log"

trulyronak: Didn't mean to make that change, whoops

PR #233, opened 2020-07-02
surfaced proxy port via env var

trulyronak:
## Reasoning
Several APIs like to output a message like "your app is listening on localhost:3000", where they use the port that was passed in to create the message. While that makes sense, when using optic the port that is passed in isn't the port the user should be using for testing purposes - they should be going to the proxy port. 

## Solution
Two things should be done (I've only addressed one in this PR)
- the user should be told to go to the proxied port (hey your api is serving @ localhost:proxyPort
- the api should be given this information (to do this, i've surfaced it via the `OPTIC_PROXY_PORT` environment variable


acunniffe:
I'm happy with this, can definitely see why it's helpful for the boilerplate projects.

PR #234, opened 2020-07-02
init: updated optic.yml template

LouManglass:
Per Slack discussion

Manual tests: Created new API to validate the optic.yml file, `apidev check start` passes start exists, is a valid task.


workspaces/local-cli/src/commands/init.ts
```diff
@@ -44,12 +44,13 @@ export default class Init extends Command {
...
   # The default task, invoke using \`api run start\`
-  # Learn how to finish setting up Optic at http://docs.useoptic.com/setup
+  # Learn how to finish setting up Optic at https://docs.useoptic.com/setup
   start:
     command: echo "Setup A Valid Command to Start your API!"
     baseUrl: http://localhost:4000
 ignoreRequests:
-- OPTIONS *`.trimLeft();
+# For more information on configuration, visit https://docs.useoptic.com/captures
+- OPTIONS (.*)`.trimLeft();
```
trulyronak, line 53: do you need to call `trimLeft()` ?

JaapRood: Yup! To get rid of the new line after the opening tick.

PR #235, opened 2020-07-03
Make Diff Performance Usable

acunniffe:
Hey guys! Some big improvements to both performance and the user experience of the diff mode are ready to merge. The entire flow has been reworked based on learnings from the last few months. 

_regressions:_ 
- The diff compass is disabled 
- Live mode is semi-gone, you get whatever is in the capture folder when you refresh, or restart diffs (which you'll be doing a lot of). 

_rough edges:_
- Giant bodies are still slow to preview, apply suggestions to, and approve. That's the only domain js work happening on the UX now. 

_also of note:_ 
We also have to talk about the DiffManager, which is now deprecated, it's being used on the Testing Dashboard and should have equivalent functionality available too it in the new DiffResultHelper


workspaces/ui/src/entrypoints/testing-sessions.js
```diff
@@ -17,10 +26,71 @@ export default function TestingSessions(props) {
...
+
+  const diffServiceFactory = async (
+    specService,
+    captureService,
+    _rfcState,
+    additionalCommands,
+    config,
+    captureId
+  ) => {
+    async function computeInitialDiff() {
```
acunniffe, line 41: @devdoshi this is the last thing blocking example sessions. Line 43 always gets events '[]' even after you make changes. Can you look at lifecycle here?

PR #237, opened 2020-07-07
Review Only - exceptions: demo of a potential custom exception handler

LouManglass:
For review and discussion only, do not merge into the code!

I mocked up a sample custom exception last night that logs errors with various verbosity depending on the mode (normal run vs. DEBUG=* run for now, which is probably fine) and passes the full exception and stack trace originally thrown in verbose mode. It  also passes the trace out to analytics for triage and potential follow-up.

Of note: I tried to follow what I read as best practices, but it may not be idiomatic. Also, I had circular dependencies in auth and analytics, so I just duplicated them in the workspace for now...any full solution would probably involve either refactoring these into shared or into their own workspace. I'm not sure if/how they're currently made available to other CLI tools and this may be useful for more than just exceptions.

Happy to chat about this in depth, I wanted an artifact available to show off the idea. This shouldn't be merged into opticdev/optic.

### DIFF
--- a/workspaces/cli-shared/src/analytics-shared.ts
+++ b/workspaces/cli-shared/src/analytics-shared.ts
@@ -1,0 +1,84 @@
+/*
+  @TODO:
+  - Analytics should be abstracted with an interface: track(...)
+  - There should be one implementation that does nothing
+  - It should be easy to switch between implementations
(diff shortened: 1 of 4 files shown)


acunniffe:
This seems fine to me. Generally I like the idea of being able to have domain errors that provide us with information we need to debug.

acunniffe:
closing this since we have sentry now

PR #239, opened 2020-07-07
query parameters should save in capture

acunniffe:
This PR adds query parameters to captures, the first step in making Optic handle query parameters natively. Super basic, keeping it near the proxy for the time being. After the first person requests customization, we can isolate it and make it configurable.


workspaces/cli-shared/src/httptoolkit-capturing-proxy.ts
```diff
@@ -191,6 +195,30 @@ export class HttpToolkitCapturingProxy {
     }
   }
 
+  extractQueryParameters(req: mockttp.CompletedRequest): IArbitraryData {
+    const rawQuery = url.parse(req.url).query;
+    developerDebugLogger('extracting query params', { rawQuery, url: req.url });
+    if (rawQuery) {
+      const jsonLikeValue = this.config.queryParser.parse(rawQuery);
+      return {
+        asJsonString: this.config.flags.includeShapeHash
```
devdoshi, line 204: this should be a different flag - perhaps includeQueryString

workspaces/cli-shared/src/httptoolkit-capturing-proxy.ts
```diff
@@ -191,6 +195,30 @@ export class HttpToolkitCapturingProxy {
...
+          ? null
+          : JSON.stringify(jsonLikeValue),
+        asText: rawQuery,
```
devdoshi, line 207: Should this always be included?

workspaces/cli-shared/babel.config.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
```
devdoshi, line 1: I think this file can be removed

workspaces/cli-shared/package.json
```diff
@@ -25,6 +25,7 @@
     "bottleneck": "^2.19.5",
     "colors": "^1.4.0",
     "fs-extra": "^9.0.0",
+    "mockttp": "^0.19.2",
```
devdoshi, line 28: this should be a devdependency, and I think the other changes can be reverted

devdoshi:
Just a few small issues remaining to clean up, and the rest can be addressed when we add in multiple strategies

devdoshi:
Nice work!

acunniffe:
@devdoshi I think I did the flag right. appreciate any further feedback

workspaces/cli-shared/package.json
```diff
@@ -28,11 +28,11 @@
     "ora": "^4.0.4",
     "tree-kill": "^1.2.2",
     "uuid": "^8.0.0",
-    "mockttp": "^0.19.2",
     "shape-hash": "^1.0.6",
     "whatwg-mimetype": "^2.3.0"
   },
   "devDependencies": {
+    "@babel/preset-typescript": "^7.10.4",
```
devdoshi, line 35: This line should be removed, and mockttp should be added as a devDependency

workspaces/cli-shared/src/query/query-parser-interfaces.ts
```diff
@@ -0,0 +1,14 @@
+export interface IQueryParser {
+  parse(rawQueryString: string): object; // this should always return something. bugs should be swallowed.
```
devdoshi, line 2: LIke we discussed, this should be the only part of the interface

yarn.lock
```diff
@@ -16,6 +16,13 @@
   dependencies:
```
devdoshi, line 16: I would revert this file

workspaces/cli-shared/src/query/build-query-string-parser.ts
```diff
@@ -0,0 +1,6 @@
+import { IQueryParser } from './query-parser-interfaces';
+import { DefaultQueryParser } from './parsers/query-string/DefaultQueryParser';
+
+export function buildQueryStringParser(): IQueryParser {
+  return new DefaultQueryParser();
```
devdoshi, line 5: In the future this is the spot where we inspect the user's optic config, determine the selected strategy, and return the appropriate function?

acunniffe: yes

PR #240, opened 2020-07-09
Create new Release Workflow

trulyronak:
## Workflow

- As PRs are merged into `develop`, release-drafter updates a draft release that logs all changes that have been made
- When we want to release something, we simply need to make a PR from `develop` to `release`
    - **note**: before merging this PR, make sure to bump the version ahead of time
    - upon merge, this wil trigger
        - npm publish
        - brew package publish
        - debian package publish

## Key Changes

- Adds [release drafter](https://github.com/release-drafter/release-drafter)
- Adds release script (triggers on changes to `release`)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,7 +6,7 @@ name: Release
 
 env:
   HOMEBREW_REPO: opticdev/homebrew-optic
-  BUCKET_NAME: optic-deb-packages
+  BUCKET_NAME: optic-packages
   PACKAGE_NAME: api
   AWS_SECRET_ACCESS_KEY: ${{secrets.AWS_SECRET_ACCESS_KEY}}
   AWS_ACCESS_KEY_ID: ${{secrets.AWS_ACCESS_KEY_ID}}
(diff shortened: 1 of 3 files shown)


trulyronak:
Works in tandem with [homebrew-optic](https://github.com/opticdev/homebrew-optic)

.github/workflows/release.yml
```diff
@@ -0,0 +1,68 @@
...
+      # - name: Install APT Utils
+      #   run: sudo apt-get install apt-utils
+      # - run: sudo npm i -g @oclif/dev-cli
+      # - run: sudo oclif-dev pack:deb
+      # - uses: actions/setup-ruby@v1
+      #   with:
+      #     ruby-version: '2.6'
+      # - run: sudo gem install deb-s3
+      # - name: Upload to s3
+      #   run: sudo deb-s3 upload --access-key-id=${{ env.AWS_ACCESS_KEY_ID }} --secret-access-key=${{ env.AWS_SECRET_ACCESS_KEY }} --bucket $BUCKET_NAME --codename ${{ env.PACKAGE_NAME }} --preserve-versions dist/deb/legume_$(npm view ${{ env.PACKAGE_NAME }} version)-1_amd64.deb
```
notnmeyer, line 54: lets use the `-e`/`--encryption` argument as well. we want bucket objects encrypted even when they're public

.github/workflows/release.yml
```diff
@@ -0,0 +1,68 @@
+# Releases the latest version of Optic on NPM/Brew/Debian
+# Assumes the version has been updated
+# Triggers on all changes to the release branch
+
+name: Release
+
+env:
+  HOMEBREW_REPO: opticdev/homebrew-optic
+  BUCKET_NAME: optic-deb-packages
```
notnmeyer, line 9: moving over this question from slack, do we want `optic-deb-packages` or `optic-packages` and then we'd specify a `--prefix` for deb-s3 below that would result in bucket path that is like, `optic-packages/deb/filename`

trulyronak: changed to `optic-packages`!

acunniffe:
This looks good to me! You should define these tags somewhere on Notion so we know how to use them 
```
 labels:
      - 'fix'
      - 'bugfix'
      - 'bug'
```

PR #289, opened 2020-07-22
Fix bug where DiffCompass degraded DiffViewer performance

JaapRood:


devdoshi:
nice!

JaapRood:
Closing as it's become outdated. We're no longer using the diff viewer component implementation this applies to.

PR #290, opened 2020-07-22
feature/demo - adds a demo project to optic

trulyronak:
Before I submit this for real, I wanted to get feedback on my implementation — I'm not an expert on React by any means, and I wanted to get this looked at.

In general, the biggest change is that there are a TON more `track` calls than usual. `DemoSessions` essentially displays snackbar / info toasts on specific events emitted from track.

The biggest point where I think I need a second look at is where I called track — I put track wherever an important user behavior action seemed to be happening, but not everything is consistent, so would definitely appreciate a second look there.


workspaces/ui/src/components/diff/v2/CaptureManagerPage.js
```diff
@@ -428,6 +437,11 @@ function EndpointDiffs(props) {
                       className={classes.row}
                       component={Link}
                       to={to}
+                      onClick={() => {
+                        if (i.count > 0) {
+                          setTimeout(() => track("Viewing Endpoint Diff", i), 500)
```
devdoshi, line 442: Can you explain the goal here?

workspaces/ui/src/components/diff/v2/Finalize.js
```diff
@@ -76,7 +71,18 @@ export default function FinalizeDialog(props) {
       JsonHelper.jsArrayToVector([EndBatchCommit(batchId)])
     );
     await specService.saveEvents(newEventStore, rfcId);
-    setTimeout(() => history.push(`${baseUrl}/diffs/${captureId}`), 500);
+    
+    setTimeout(() => {
```
devdoshi, line 75: I'm curious why this was delayed to begin with

workspaces/ui/src/components/diff/v2/Finalize.js
```diff
@@ -76,7 +71,18 @@ export default function FinalizeDialog(props) {
...
+      history.push(`${baseUrl}/diffs/${captureId}`)
+    }, 500);
+
+    setTimeout(() => {
+      track('Committed Changes to Endpoint', {
+        message: commitMessage,
+        captureId,
+        suggestions: acceptedSuggestions.length,
+      });
+    }, 1000)
```
devdoshi, line 85: What's the goal with the delay here as well?

workspaces/ui/src/routes.js
```diff
@@ -41,7 +41,7 @@ export function ApiRoutes(props) {
             />
           )}
 
-        <Redirect to={routerPaths.docsRoot} />
+        <Redirect to={props.demo ? routerPaths.diffsRoot : routerPaths.docsRoot} />
```
devdoshi, line 44: Conceptually, I'd prefer the generic routes to not have knowledge of things like debugging and demos. The pathways for having different behaviors between each entry point should be injecting things at the entry point and factoring extension points of the UI into the shared tree and providing components for those slots. When does this pathway get hit?

workspaces/ui/src/components/loaders/ApiLoader.js
```diff
@@ -39,7 +39,7 @@ export function ApiSpecServiceLoader(props) {
     }
   }, [service]);
 
-  if (!service) {
+  if (!events) {
```
devdoshi, line 42: If this is a separate fix, we should merge it separately

trulyronak: makes sense — made #293 to address this

workspaces/ui/src/components/diff/v2/CaptureManagerPage.js
trulyronak: the timeout isn't necessary — that was something I had added because it wasn't working before, but I think now I do state management better so that's not an issue.

The if statement is so that the onClick action is only added to URL modals that are diffs (as technically a ListItem here can potentially have no diffs, and thus should not trigger any event upon a click. However, it seems that if there are no diffs, the onclick never gets triggered, so I don't think its needed

workspaces/ui/src/components/diff/v2/Finalize.js
trulyronak: The timeout here exists really just so it comes after the history push finishes from above — otherwise another event occurs and takes priority, making it so the postcommit message doesn't show up

workspaces/ui/src/components/diff/v2/Finalize.js
trulyronak: no idea, it was here before I started (doesn't seem needed though) — when I remove it, core functionality seems unaffected

workspaces/ui/src/components/diff/v2/Finalize.js
trulyronak: haven't found a way to do this without the delay here 😕

workspaces/ui/src/routes.js
trulyronak: This gets hit from `demo-sessions.js` https://github.com/opticdev/optic/pull/290/files#diff-0c55cfef92a59cc08e309f67f2b0bae8R230

agree that there should be a better way — I was looking for a way to simply specify a default page - I think a general prop called "default" could be good to include, considering we may want to specify a default page (or have the ui remember the users last page)

workspaces/ui/src/components/diff/v2/Finalize.js
trulyronak: went ahead and removed the timeout for `history.push` for now

workspaces/ui/src/routes.js
devdoshi: Yeah I think adding a defaultRoute prop set to docs by default would make sense

trulyronak:
Reviewed with @devdoshi and @LouManglass , I think this is ready to go for our initial run!

## Steps to Run the Demo

```bash
# build the ui project
optic_build
cd workspaces/ui
optic_watch
# in new tab
yarn start-local
```
navigate to [localhost:3000/demos/todo](http://localhost:3000/demos/todo)

PR #271, opened 2020-07-16
Run End to End Tests before Release

trulyronak:
This runs optic's end-to-end tests before any release, blocking release if they do not pass.

Before merging, note that will require a new org secret — a repo access token for end-to-end tests, called `END_TO_END_ACCESS_KEY`

PR #291, opened 2020-07-23
 Run End to End Tests before Release #271 

trulyronak:
This runs optic's end-to-end tests before any release, blocking release if they do not pass.

Before merging, note that will require a new org secret — a repo access token for end-to-end tests, called `END_TO_END_ACCESS_KEY`

After adding the `END_TO_END_ACCESS_KEY` token to the org secrets, this should be good to go

### DIFF
--- a/.github/workflows/endToEndTests.yml
+++ b/.github/workflows/endToEndTests.yml
@@ -1,0 +1,40 @@
+name: Test end to end
+
+on:
+  pull_request:
+    branches:
+      - develop
+      - release
+  push:
+    branches:
+      - develop
+      - release
+
+jobs:
+  test-end-to-end:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/setup-node@v1
+      with:
+        node-version: 12
+    - run: echo "::add-path::$(yarn global bin)"
+    - name: Check out Optic Repo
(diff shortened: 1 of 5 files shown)


.github/workflows/endToEndTests.yml
```diff
@@ -0,0 +1,40 @@
...
+
+on:
+  pull_request:
+    branches:
+      - develop
+      - release
+  push:
+    branches:
+      - develop
+      - release
```
trulyronak, line 11: this is one thing I'm not sure about — when should the end-to-end tests run? Right now I have it on any PR to `develop` or `release`, should I also be having it run once they merge into `develop` and `release` as well?

trulyronak:
right now you can see that the tests fail since the `END_TO_END_ACCESS_KEY` doesn't yet exist - this should be set to be some repo_access_token that can read the optic_end_to_end_tests

devdoshi:
I added the token, waiting for it to finish running now

.github/workflows/endToEndTests.yml
devdoshi: Seems fine for now, I'll be working on the e2e suite so I'll revisit once I look at it a bit more

PR #296, opened 2020-07-28
Fixed bug where two scrollbars showed up on diffs page

trulyronak:
2 scrollbars

On firefox/chrome, whenever you went to the diffs page, you would see 2 scrollbars. This was because `CaptureManagerPage` rendered things inside its own container, making it so it had its own overflow — thus adding another scrollbar to be rendered.

I fixed this by conditionally hiding the outer scrollbar when you are on the diffs page. It doesn't appear to affect / hide anything, as all elements for the diffs page are rendered within the `CaptureManagerPage`.

only one scrollbar


acunniffe:
Nice I could use featurepeak to test and it worked!

workspaces/ui/src/components/diff/v2/CaptureManagerPage.js
```diff
@@ -136,6 +136,17 @@ function AllCapturesStore(props) {
 export function CaptureManagerPage(props) {
   const routerPaths = useRouterPaths();
 
+  // set scrolling
```
acunniffe, line 139: @trulyronak any reason this can't just be a CSS style?

trulyronak: Totally could - we'd instead be swapping classes. Wasn't sure where to put the class definitions / what policy was on CSS files vs useStyles, happy to change as needed

workspaces/ui/src/components/diff/v2/DiffReviewPage.js
```diff
@@ -260,7 +257,8 @@ const useStyles = makeStyles((theme) => ({
     position: 'fixed',
     width: 'inherit',
     height: '100vh',
-    overflowY: 'scroll',
+    overflowY: 'visible',
+    overflowX: 'hidden',
```
JaapRood, line 261: Is there any thing specific that made you hide overflow on the x-axis? Unless there's a specific reason for it, I tend to stay away form it, as it prevents a whole bunch of enhancements to be made. The most common enhancement I see, especially in navigation, is rendering tooltips over links, buttons, etc. Without hiding the overflow, these tooltips would be allowed to render outside the navigation, overlapping some of main page content (which for a tooltip is fine). When overflow is hidden, stuff gets a _lot_ more complicated.

workspaces/ui/src/components/diff/v2/DiffReviewPage.js
```diff
@@ -290,7 +287,7 @@ const useStyles = makeStyles((theme) => ({
     paddingLeft: 20,
     paddingRight: 20,
     paddingTop: 20,
-    paddingBottom: 400,
+    paddingBottom: 100,
```
JaapRood, line 290: Great, that gets rid of a whole bunch of arbitrary white space. Have you seen the `theme.spacing(x)` API? I'm trying to get us all to use whole numbers of that for any spacing between items, so our white spaces become very nicely consistent throughout the entire app.

workspaces/ui/src/components/diff/v2/CaptureManagerPage.js
```diff
@@ -656,16 +655,15 @@ const useStyles = makeStyles((theme) => ({
...
-    height: '100vh',
     justifyContent: 'center',
   },
   navRoot: {
     flexGrow: 1,
     position: 'fixed',
     width: 'inherit',
     height: '100vh',
-    overflowY: 'scroll',
+    overflowY: 'visible',
```
JaapRood, line 665: I think this is where the left over problem you've described comes from: while we want the normal page content to be in the main scroll context, the left hand nav can have it's own! Therefor, it should be fine for this one to have a `overflowY: scroll`. When set to `visible` it will just bleed past the container. Probably not a big deal with the small amount of content (overflow is pretty unlikely), but perhaps a fix worth trying regardless.

trulyronak: doesn't fix anything :| no matter which one, it still results in being unable to scroll to the two tabs. I'm actually going to keep with visible, because otherwise it adds a scrollbar at all times there, and it looks kind of ugly

https://www.loom.com/share/32dab57277b04a6580d91fa68126986a

workspaces/ui/src/components/diff/v2/DiffReviewPage.js
trulyronak: I had hidden it because it was adding an unnecessary scrollbar — switched it to visible so that if we need to scroll, it'll be able to scroll (https://www.loom.com/share/c0d79ac07dba4f7280bb63f10a0297f8)

workspaces/ui/src/components/diff/v2/DiffReviewPage.js
trulyronak: worked perfectly — switched to `theme.spacing(1)`, since it really just needed a minimal amount of padding because otherwise the card was getting a little cut off 
Screen Shot 2020-08-05 at 9 46 46 AM

workspaces/ui/src/components/diff/v2/DiffReviewPage.js
JaapRood: Hmm, that's fair enough. It's an indicator of an underlying problem, as that left nav shouldn't overflow in the first place: it should be the set width that it is and all content should shrink to fit or wrap to new lines. I wonder if there's a missing `flex-shrink: 0` would fix things, but chances are it's not just that. Don't bother too much with that though, unless you think it's fun.

workspaces/ui/src/components/diff/v2/DiffReviewPage.js
JaapRood: Excellent!

PR #299, opened 2020-08-04
Implement flattened-rows rendering ShapeViewer behind feature flag

JaapRood:
This pull request adds a new approach to implementing the `DiffViewer`, using a "view model" approach to control what's rendered. For now, it's hidden behind a feature flag and rendered by adding the `/flattened` suffix to a diff review URL.

So far, we're only exploring how the component itself should work and whether it will give the amount of control and expressiveness to better control the responsiveness of the UI. Eventually, for this to bear fruit, the view model should be generated in a more direct way from the domain, without performing the entire diff render all at once. That is outside the scope of this PR: proving this kind of view model can power this type of UI in the first place. We shouldn't also care too much about this the debug route is wired up: a lot of that maze of react components might not even make it to the version that hits production, so caring about the ergonomics of that now is quite likely wasted time.


JaapRood:
@acunniffe I would have just merged this, if it wasn't for the failing checks. To me those failures seem totally unrelated to the changes I've made, but it would be good to get your confirmation on that as well.

acunniffe:
Not seeing anything would ‘fail’ this. Rerunning, and will checkout if it still failed.

JaapRood:
@acunniffe it builds, but end-to-end tests job fails, not able to check them out. The way that works seems to have been last changed recently through #297.

JaapRood:
Double checked everything and was confident this wouldn't break anything. @acunniffe still keen on a retrospective review, just to make sure I've siloed it off behind a feature flag properly.

PR #300, opened 2020-08-06
Render new ShapeViewer by directly traversing HttpInteraction

JaapRood:
Side-stepping a whole bunch of the domain, any browser is perfectly capable of traversing a JSON structure itself. In our case, traversing the `body` of either the `Request` or `Response` of an `HttpInteraction` to render the basic shape of the shape viewer, powering the row-level view model.

This depends on #299 to get merged, first. Through that, this all happens behind a feature flag, so not much to review. All missing functionality `RenderShape` provided, like marking missing fields and folding of collections, will be reimplemented incrementally and is beyond the scope of this PR.


JaapRood:
@acunniffe a sanity check in traversals could never hurt, but feel free to ignore as it's one of those things of it either working, or not.

acunniffe:
@JaapRood this looks right to me. Definitely nice to see some of those domain logic moving out. You made quick work of this.

PR #301, opened 2020-08-06
Implement default folding behaviour of ShapeViewer and ability to unfold

JaapRood:
This PR implements default folding of the interaction when first rendered in the ShapeViewer.

Rather processing the entire interaction body all at once, we're only rendering rows that will actually be visible. To reduce the amount of visible rows for first render, we only have the first element of every array unfolded, with the rest hidden. When clicking on a folded line, the view model can be updated by only rendering the bit of missing shape and inserting it in the right place, rather than having to recalculate the entire thing.

The exact behaviour of what and what not to fold is arbitrary, but what we have now is a good default until we start considering diffs (out side the scope of this PR). In case of the example I'm testing with, instead of processing and rendering `781` rows straight from the beginning, we collapsed `19` trails and only render `96` rows initially.

Since we start with a small amount of initial rows and can then incrementally process additional subsets, that also gives us the partitioning we need to decide what to perform additional transformations on, like annotating each row with the expected type.

Another thing that's worth mentioning is that the view model expressed using the reducer pattern. This makes all the state transformations into stateless functions, which doesn't only help us get the logic right in this first implementation, but is also a first great step to it being extracted to it's own module later (for use of Typescript, easier testing, re-use of logic, etc). Any additional enhancements are easily expressed as actions as well.

Final note: this is still behind a disabled-by-default feature flag + opt-in URL.

### DIFF
--- a/workspaces/ui/package.json
+++ b/workspaces/ui/package.json
@@ -25,6 +25,7 @@
     "date-fns": "^2.11.1",
     "lodash.compose": "^2.4.1",
     "lodash.debounce": "^4.0.8",
+    "lodash.get": "^4.4.2",
     "lodash.groupby": "^4.6.0",
     "lodash.isequal": "^4.5.0",
     "lodash.sortby": "^4.7.0",
--- a/workspaces/ui/package.json
+++ b/workspaces/ui/package.json
@@ -26,6 +26,7 @@
     "lodash.compose": "^2.4.1",
     "lodash.debounce": "^4.0.8",
     "lodash.groupby": "^4.6.0",
+    "lodash.isequal": "^4.5.0",
     "lodash.sortby": "^4.7.0",
     "memoize-weak": "^1.0.2",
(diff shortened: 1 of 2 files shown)


JaapRood:
@acunniffe see the PR description for what the interesting stuff is to review. I'm very happy with the approach, so it would be good to know if you can follow what's going on before we build on this approach further.

acunniffe:
@JaapRood this looks great.

> Rather processing the entire interaction body all at once, we're only rendering rows that will actually be visible. To reduce the amount of visible rows for first render, we only have the first element of every array unfolded, with the rest hidden. When clicking on a folded line, the view model can be updated by only rendering the bit of missing shape and inserting it in the right place, rather than having to recalculate the entire thing.

I definitely expected it to end up with properties like you describe. Thought it would come through via React tree folding -- the view model approach makes way more sense and I like how its implemented.

PR #309, opened 2020-08-10
Add highlighting of rows with recorded diffs in ShapeViewer

JaapRood:
This PR adds (still behind a feature flag) the highlighting of the row that contains a diff to the new `ShapeViewer`. It manages to do so by a the domain's new ability to return the `jsonTrails` of where the current diff was detected, introduced in `@useoptic/domain: 10.0.6`. This allows efficient matching with the json trails tracked for each row rendered.

Taking the diffs into account, the automatic collapsing of items has been altered for the first occurrence of the diff to always be unfolded. Any other collapsed array items that contain diffs are marked as such. Diffs that describe fields being missing cause the missing field to be rendered in the place where it would have been expected.

Applying one of the suggested actions updates the highlighted rows / collapsed items, giving feedback on the action that will be applied. This in contrast to the current "production" viewer, where updated field type information is displayed, the domain logic changes required to make work efficiently currently not worth reproducing this.


JaapRood:
@acunniffe merged right away as this mostly concerns UI and UX of a hidden feature, that's still to be tested and go through QA (which will be a separate PR as we ship with an enabled feature flag / default route). The part worth reviewing might be the view model logic of how we take the diffs and map them to the rows.

PR #311, opened 2020-08-13
Prevent ShapeViewer in diff review to be re-mounted upon choosing a suggestion

JaapRood:
This PR fixes a bug where the new `ShapeViewer` (still behind a feature flag and opt-in route) was re-mounted when a suggested was chosen to resolve the observed diff. Not only did this remounting incur a heavy performance penalty as the view model was generated again from scratch, it also caused the unwanted behaviour of all unfolded items to collapse again.

It was fixed by moving the `ShapeViewer` up the component hierarchy, from the `DiffViewSimulation` to the `DiffReviewExpanded`. The `SimulatedRfcContext` rendered by the former caused the complete re-render, as it previews what the updated spec must look like with the suggested change applied. However, we're not using any of this spec information and most likely will not again for the foreseeable future (until we find a way to query the domain efficiently for it).

Note that in the case where no diff is detected in the body rendered (as might happen if we're rendering bodies for both request and response), we fall back on the current production viewer. That's because the current implementation of `ShapeViewer` requires a `diff`, something we'll be addressing separately, at which time we'll revisit this fallback strategy.

### DIFF
--- a/workspaces/ui/src/components/diff/v2/DiffReviewExpanded.js
+++ b/workspaces/ui/src/components/diff/v2/DiffReviewExpanded.js
@@ -27,6 +27,7 @@ import { useDiffDescription, useInteractionWithPointer } from './DiffHooks';
 import LinearProgress from '@material-ui/core/LinearProgress';
 import { DiffReviewLoading } from './LoadingNextDiff';
 import { DiffViewSimulation } from './DiffViewSimulation';
+import ShapeViewer from './shape_viewers/ShapeViewer';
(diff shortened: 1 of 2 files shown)


JaapRood:
@acunniffe I'm confident I've got the semantics of this right, not needing the simulated rfc state in the way we're rendering the new `ShapeViewer`, to the point where I'll probably merge this on my own accord. In case you see any issue with it, let me know and we can address that assumption separately.

PR #313, opened 2020-08-13
Use ShapeViewer (InteractionBodyViewer) for rendering of all interaction bodies in BodyDiffs and NewRegions

JaapRood:
This PR renames the `ShapeViewer` to `InteractionBodyViewer` and applies it everywhere in the Diffing where interaction examples are rendered.

The rename has been done to better describe the situations where the viewer can currently be used, especially as "Shape" has a more specific meaning (a `Shape` as in described by the stored or simulated spec). In the future, one can imagine a `Shape` instances having it's own way to generate the required view model for the viewer, allowing it to be use as well, but that's clearly beyond the scope of what we're currently targeting. We're better off being explicit in the meantime.

The additional places where the viewer is now used (still behind feature flag and `/flattened` URL suffix) are:
- rendering an example interaction for a new region
- rendering a body of either response or request when that's not the body containing the diff.

To make this work, the `InteractionBodyViewer` now no longer relies on the `diff` prop being provided, only rendering anything diff related if it was. To make sure we're rendering the right body, the `body` itself is required as a prop, replacing the `interaction` prop (from which the body was previously parsed using the `diff`).

### DIFF
--- a/workspaces/ui/src/components/diff/v2/DiffNewRegions.js
+++ b/workspaces/ui/src/components/diff/v2/DiffNewRegions.js
@@ -44,6 +44,7 @@ import { DiffToolTip } from './shape_viewers/styles';
 import Pagination from '@material-ui/lab/Pagination';
 import { RfcContext } from '../../../contexts/RfcContext';
 import DiffHunkViewer from './DiffHunkViewer';
+import InteractionBodyViewer from './shape_viewers/InteractionBodyViewer';
 import { ShapeExpandedStore } from './shape_viewers/ShapeRenderContext';
 import { ShapeBox } from './DiffReviewExpanded';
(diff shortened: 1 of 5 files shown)


JaapRood:
@acunniffe another one to review retrospectively. With this one, you'll time be best spent looking at the way `NewRegions` uses the viewer. I'm pretty confident I've got a good handle on how these things fit together now, but a sanity check always makes sense when replacing existing components!

acunniffe:
` be best spent looking at the way NewRegions uses the viewer. I'm pretty confident I've got a good handle on how these things fit together now, but a sanity check always makes sense when replacing existing components!`

Thanks for the focus -- this looks right to me. Basically yeah -- in this usage there won't be a diff, so you're purely relying on the body viewing capabilities. Your component, as written, doubles a JSON Viewer and is perfectly valid to be used as such. Nice work!

PR #315, opened 2020-08-13
Implement RowCompass for InteractionBodyViewer

JaapRood:
More info to come once we make it out of draft.


JaapRood:
Going to close this one, as we won't be working on this any time soon and it'll just get staler meanwhile.

PR #318, opened 2020-08-14
Enable the new DiffViewer by default in development builds

JaapRood:
We'll want to start testing the new DiffViewer, and what best way to enable the feature flag for anyone who's developing Optic.

In case you want to disable it and get to the current production version, add (or edit) `workspaces/ui/.env.development.local` and make sure it contains the following line:

```
REACT_APP_FLATTENED_SHAPE_VIEWER=false
```


JaapRood:
@acunniffe fyi!

PR #321, opened 2020-08-17
Migration to new analytics

acunniffe:
This is a draft of the new approach to analytics we want to adopt. We're starting to use user events from across all the various parts of Optic (local UI, CLI, SaaS) to power demos, on-boarding flows, support interactions, etc. These are going to be key to the user experience and we don't want an extra space or a small change in the tense of an event name to break one of these experiences. 

- These changes introduce types for the analytics (open to recommendations to make this cleaner. it is quite hard to get TS runtime + compile time validation). 

- Consistent naming for analytics events

- All User Tracking Events are published through Event Emitters -- segment, our backend, the demo + in-app interactions etc read off these streams. 

---

This is in draft form. It will need to be approved before we can release on-boarding step 3 which consumes these events and persists some of them to the backend to power the on-boarding flow. 

@trulyronak we need to talk about the demo's events. 
@LouManglass some of the names here change


workspaces/analytics/src/interfaces/Events.ts
```diff
@@ -0,0 +1,26 @@
...
+  ExitedTaskWithLocalCLI: null,
+
+  //diffs
+  UserChangedCaptureOverviewTab: null,
+  NewBodyDiffRendered: null,
+  InferPolymorphismEnabled: null,
+  UserBeganAddingNewUrl: null,
+  UserFinishedAddingNewUrl: null,
+  DiffWasReset: null,
+  PreviewSuggestion: null,
```
devdoshi, line 23: This one doesn't fit the grammar, and it seems like we're inconsistently applying the User prefix.

workspaces/analytics/src/interfaces/Events.ts
```diff
@@ -0,0 +1,26 @@
+//@ts-ignore
+import keymirror from 'keymirror';
+export const Events = keymirror({
+  //on-boarding
+  ApiCreated: null,
+  ApiCheckCompleted: null,
+
+  //Errors
+  JavascriptErrorDetectedInFrontend: null,
+  RequirementForDiffsToHaveASuggestionFailed: null,
```
devdoshi, line 10: Lou might have a suggestion for assigning these numeric codes or something to make it easier to add more, otherwise seems fine

workspaces/analytics/src/interfaces/RegisterEvent.ts
```diff
@@ -0,0 +1,50 @@
...
+  type: string,
+  schema: any,
+  propsToSentence: (propsInstance: T) => string
+): RegisteredEvent<T> {
+  const event = {
+    withProps(properties: T): TrackingEventBase<T> {
+      return {
+        type: type,
+        context: {
+          clientAgent: '',
```
devdoshi, line 16: Some of these we'll need to figure out where they come from. if all tracking is authenticated we can fill in clientId on the server side

workspaces/analytics/tests/DeclaringEvents.test.ts
```diff
@@ -0,0 +1,7 @@
+const { ApiCreated } = require('../src/events/onboarding');
+
+test('Can creeate an event type', () => {
```
devdoshi, line 3: The opportunity to fix spelling errors is now, otherwise you'll also have to edit the snapshot file

workspaces/cli-server/src/analytics.ts
```diff
@@ -0,0 +1,79 @@
...
+//event bus for tracking events
+const analyticsEvents: AnalyticsEventBus = newAnalyticsEventBus(
+  async (batchId: string) => {
+    const user = await getCredentials();
+    const clientAgent = user ? user.token : consistentAnonymousId;
+
+    const clientContext: ClientContext = {
+      clientAgent: clientAgent,
+      clientId: clientId,
+      clientSessionInstanceId: batchId,
```
devdoshi, line 27: clientSessionInstanceId corresponds to the client lifecycle. For example, in a browser each tab would have a different one, and each cli run would have a different one. batchId is an additional field, like a correlation or causation id, helping to group events by their originating cause

workspaces/analytics/.prettierrc
```diff
@@ -0,0 +1,5 @@
+{
```
devdoshi, line 1: if it is in the workspace, it should inherit the workspace prettierrc correct?

workspaces/analytics/src/consistentAnonymousId.ts
```diff
@@ -0,0 +1,2 @@
+import { v4 as uuidv4 } from 'uuid';
+export const consistentAnonymousId = `anonymous_user_${uuidv4()}`;
```
devdoshi, line 2: might want to document here that it must start with anonymous_, or capture that in some test

workspaces/analytics/src/interfaces/Events.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  //running tasks
+  StartedTaskWithLocalCLI: null,
```
devdoshi, line 13: To be consistent with Api, etc., these should say Cli instead of CLI

workspaces/local-cli/src/shared/analytics.ts
```diff
@@ -37,36 +42,51 @@ export async function getUser(): Promise<IUser | null> {
...
+  const daemonState = await ensureDaemonStarted(
+    lockFilePath,
+    Config.apiBaseUrl
+  );
+
+  const cliServerClient = new Client(
+    `http://localhost:${daemonState.port}/api`
+  );
+
+  cliServerClient.postTrackingEvents([event]);
```
devdoshi, line 57: This is probably fine since it's local but in the future we may wish to queue & batch flush

workspaces/analytics/src/interfaces/RegisterEvent.ts
acunniffe: it stubs empty and each client sets these themselves

workspaces/analytics/.prettierrc
acunniffe: yeah it started separated. cleaning it now

workspaces/local-cli/src/shared/analytics.ts
```diff
@@ -37,34 +42,17 @@ export async function getUser(): Promise<IUser | null> {
...
-      if (user) {
-        analytics.track({ userId: user.sub, event, properties }, resolve);
-      } else {
-        analytics.track({ anonymousId: 'anon', event, properties }, resolve);
-      }
-    });
-  });
-}
+export async function trackUserEvent(event: TrackingEventBase<any>) {
+  const daemonState = await ensureDaemonStarted(
```
acunniffe, line 46: @devdoshi it's here

PR #325, opened 2020-08-27
[GitBot] Add Publish and Compare Commands

trulyronak:
This PR contains features needed for the gitbot. This edits one workspace, `ui`. The UI changes are to enable the apidocs site, that lets users access their documentation online.

Core Changes
- deployment workflow for apidocs site
- updates to the latest version of @useoptic/domain


workspaces/ui/src/contexts/SpecViewerContext.js
```diff
@@ -0,0 +1,57 @@
+import React, { useMemo } from 'react';
+import axios from 'axios';
```
devdoshi, line 2: any reason to use axios instead of fetch?

trulyronak: i was running into issues with `fetch` — but that was before we did a lot of fixing — just retried it and it all worked so I switched over

devdoshi: awesome. Thank you!

trulyronak:
ignore my bot doing strange things :P (its resolved now)

PR #333, opened 2020-09-01
8.3.1

acunniffe:


trulyronak:
notes for tomorrow — this seems to be failing because `cli-client` doesn't have access to the @useoptic/analytics package — #331 addresses this, and after that gets merged into develop we should be good to go for releasing this

acunniffe:
@devdoshi I'm seeing some unexpected logging, not sure if this is from consoles you added during the Rust work. It happens on every interaction. 

Where should I be looking to remove these? The PR doesn't show any obvious logs. 

```
optic:optic-demo-project aidancunniffe$ apidev start
[optic] Review the API Diff at http://localhost:3000/apis/1/diffs
[optic] Optic is observing requests made to http://localhost:3005

> tododemo@0.1.0 server-start /Users/aidancunniffe/Desktop/optic-demo-project
> babel-node server/server.js

q {
  il: null,
  Al: null,
  Hga: null,
  EQ: wm { kb: qx { va: n3a {} }, Eb: no {} },
  SW: false,
  stackdata: [Circular]
}
q {
  il: null,
  Al: null,
  Hga: null,
  EQ: wm { kb: qx { va: n3a {} }, Eb: no {} },
  SW: false,
  stackdata: [Circular]
}
q {
  il: null,
  Al: null,
  Hga: null,
  EQ: wm { kb: qx { va: n3a {} }, Eb: no {} },
  SW: false,
  stackdata: [Circular]
}
```

devdoshi:
@acunniffe  you have conflicting versions of domain installed, which is causing a match error when computing diffs as interactions come in. `yarn bump-core 10.0.8` (replacing 10.0.8 with whatever version you intend) should fix it.

PR #335, opened 2020-09-02
disabled editing for spec viewer

trulyronak:
**Please merge this before #333 **

This is required for the release of **GitBot**.

This prohibits users from editing the public spec viewer site on apidocs.useoptic.com. Technically, they already couldn't (since we're not saving anything they do), but it feels like you can (until you refresh). Rather than confuse users, this PR ensures that no edits can happen.

This PR also does a quick refactor of the buildSpec process, moving common env vars into a `specReqs` file


devdoshi:
I wonder if there's not a quick thing we can do to make that urge to edit sticky. Like have a call to action to visit the repo it came from or something

workspaces/ui/src/components/diff/v2/shape_viewers/ShapeOnlyShapeRows.js
```diff
@@ -252,7 +252,11 @@ function FieldDescription({ fieldId }) {
 
   const description = contributions.getOrUndefined(fieldId, DESCRIPTION);
 
-  const [editing, setEditing] = useState(false);
+  const [editing, _setEditing] = useState(false);
+  const setEditing = (condition) => {
+    if (process.env.REACT_APP_DONT_ALLOW_EDITING === 'true') { return; }
```
devdoshi, line 257: Double negatives are confusing. I would do ...ENABLE_SPEC_EDITING = yes/no. And then the pattern we would follow in the code base is to make a context that encapsulates this concern and separates the state from the env

trulyronak:
Like an "edit on github" button?

devdoshi (changes requested):

workspaces/ui/src/components/diff/v2/shape_viewers/ShapeOnlyShapeRows.js
trulyronak: my main rationale by making it "DONT ALLOW EDITING" was because I didn't want to make it something required across the entire app

devdoshi: You can also achieve that by only checking if x === 'no' which is an opt in way. Either way, it should only appear once near the top when setting the react context value, and after that it would be consumed as a Boolean from react context anywhere below

acunniffe:
> I wonder if there's not a quick thing we can do to make that urge to edit sticky. Like have a call to action to visit the repo it came from or something

I like this idea, maybe if you click a field to edit it, you get a modal with instructions to clone and open it -- not perfect, but might be right for this stage.

PR #338, opened 2020-09-08
Fix analytics + Diff Refresh

acunniffe:


workspaces/ui/src/components/diff/v2/CaptureManagerPage.js
```diff
@@ -363,6 +363,11 @@ function CaptureDiffWrapper(props) {
   const rfcContext = useContext(RfcContext);
   const services = useServices();
 
+  useEffect(() => {
```
devdoshi, line 366: should this be removed?

workspaces/ui/src/contexts/CaptureContext.js
```diff
@@ -57,6 +57,16 @@ class _CaptureContextStore extends React.Component {
     }
   }
 
+  componentDidUpdate(prevProps, prevState, snapshot) {
```
devdoshi, line 60: can you explain the change?

PR #340, opened 2020-09-08
fixed refresh and loading issue

acunniffe:


workspaces/ui/src/contexts/CaptureContext.js
```diff
@@ -51,6 +51,16 @@ class _CaptureContextStore extends React.Component {
     );
   };
 
+  componentDidUpdate(prevProps, prevState, snapshot) {
```
devdoshi, line 54: how often does this run, is it performance intensive, and is there a better way?

devdoshi (approved):

workspaces/ui/src/contexts/CaptureContext.js
acunniffe: I think at some point during the demo implementation that pathway to refresh the diff after finalize was circumvented. I'm able to replicate this issue from Victor on every new endpoint. It only runs once. 

Audited the code to make sure. Start diff only happens on....
- componentDidMount
- new commands applied
- whenever the outer rfc state changes

Screen Shot 2020-09-08 at 2 27 03 PM

PR #344, opened 2020-09-09
Fix bug where text bodies of interactions would render an empty InteractionBodyViewer

JaapRood:
PR as described in the title. It was found while testing what would happen to html, malformed json and other non-json bodies. Just like the `DiffHunkViewer`, it renders this a string, by virtue of how non-json interactions are represented.


JaapRood:
End to end test is failing because of a misconfig, like it has with past pull requests, so merging!

PR #347, opened 2020-09-09
Enable feature flag for new InteractionBodyViewer for all builds

JaapRood:
This enables the new `InteractionBodyViewer` for all builds, including production. It replaces the current `DiffHunkViewer` for wherever bodies of interactions are rendered, with one that renders using an incremental, row-by-row strategy.

It's gone through an extensive QA phase, testing against many example and captured debug sessions, working as expected.


JaapRood:
@acunniffe since you're managing the releases, I'm leaving this for you to merge. Hopefully we can do that soon, so there's days left in the cycle to deal with any unforeseen issues.

PR #350, opened 2020-09-14
Add experimental diff engine for internal review

JaapRood:
This PR is a rebased version of [a branch where we've started to implement a new diff engine in Rust](https://github.com/opticdev/optic/commit/ce9ea2076c53463c19402cef340ab925be24f049). It aims to add the new experimental engine so that it can build and run depending on a feature flag being set.

The new engine is still highly experimental, incomplete and unproven, so should not be used in any production capacity. Rather, the goal is for Optic to evaluate the benefits this alternative implementation might yield, plan future development on it, etc. As such, it should be built and ran by those who can work with the development toolchain, and only when opted-in.

The mechanism for opting in, so far, is providing the `OPTIC_RUST_DIFF_ENGINE` env variable set to `true` to **both building and running**. It will require the rust compiler and cargo to be installed (documentation pending).

Compared to the branch where development happened, this has been rebased on the latest `develop` branch which moved a lot since we first started. Additionally, it's been moved to the `workspaces/diff-engine` package, where also the Node.js glue-code lives for interacting with the compiled Rust binary. So far, it does so by spawning a new process and connecting up it's `stdin`, `stdout` and `stderr`.


JaapRood:
There's a bunch of things to do before things can be merged:

- [ ] Add a README.md to the diff-engine package with details on how it can be evaluated.
- [ ] Add a Github Team and assign it as the code-owner of the diff-engine package (Rust team?).
- [ ] Verify the publishing has been set-up right: it doesn't have to be built as this is not for outside consumption, but other packages that depend on it have (`cli-shared`).
- [ ] Fix bug that causes where UI shows diff process as never having been finished.

JaapRood:
@acunniffe I've pushed a version where I've outlined some basic documentation on how to get this running. Since you seemed keen to give it a go, perhaps you can give it a go, see what I've missed / what is broken / what is lacking? It's detailed in `workspaces/diff-engine/README.md`.

JaapRood:
I pushed a version where ownership has been assigned to the newly created rust-core team, and where publishing should be fixed. 

I had a look at ways we could publish a _working_ version, with the binary included or being installed from having been uploaded as part of a release, but all of it is too involved for now. To keep `cli-shared` from failing installation, for now, we'll just ship the compiled Typescript and not the binary / Rust source. That's fine, because there is intentionally no path way to using a published version of optic with the new engine yet.

JaapRood:
@devdoshi would appreciate a review regarding the optic toolchain and publishing mechanics. As long as building and running is successfully hidden behind the feature flag, we should be able to publish new versions of Optic without being influenced by our experiments.

workspaces/diff-engine/src/events/http_interaction.rs
```diff
@@ -0,0 +1,399 @@
...
+          let shape_hash_as_text = String::from_utf8(decoded_hash)?;
+          Ok(serde_json::Value::from(shape_hash_as_text))
+        });
+      match json {
+        Ok(json) => Some(json),
+        Err(json) => {
+          eprintln!("could not decode shape hash into a json value");
+          None
+        }
+      }
```
JaapRood, line 83: I added this implementation for shape hashes going from my naïve idea of what they were, but now looking at what the shape hashes actually are, this implementation is obviously totally borked and non-sensical. Looking at the implementation of the shape hash, it's protobuffs, so trying to read that as json or a utf-8 string is going to fail miserably.

devdoshi (approved):
Looks good to me, I don't believe publishing will be affected, and once we're ready to distribute the rust binary in published artifacts we can figure out the rest.

PR #351, opened 2020-09-14
Devops/bump script

trulyronak:
# What this does

This makes the release flow automatically bump each version. Whenever we want to release a new version of Optic, we need to make a new PR from `develop` to `release`, and give it a label. That label can either be `patch`, `minor`, or `major`. Upon choosing a label, a PR will automatically be generated, making the associated bump. We can then approve that change, and then we'll be good to go.

## Key Changes
- the `bump.js` script can now take in `patch`, `bump`, or `major` arguments (no argument defaults to `patch`), to do an intelligent bump of all packages. Past functionality to hard set the version is still there. (you can still do `yarn bump <arbitrary version number>`)
- added a new `bump.yml` workflow that triggers on any PR being opened between `develop` and `release`. Also triggers on any added label to said PRs. It will automatically comment with the PR containing the bump.


trulyronak:
@acunniffe this is what it looks like now
Screen Shot 2020-09-15 at 11 36 51 AM

PR #358, opened 2020-09-16
fixed bug where debian wasnt releasing

trulyronak:
Confirmed this fix works — https://github.com/opticdev/optic/runs/1125692261?check_suite_focus=true

`deployDebian/entrypoint.sh` required an input called `NPM_PACKAGE_NAME`, but this wasn't specified in the config file @ `deployDebian/action.yml`. Thinking this to be an extra added field, I removed that from the original input (which caused a failure, since that was a requirement to run. I have since made sure to update the config and actual release run file to include the required variables.

tl;dr: extra non required but needed parameter caused headaches/failures, added this parameter to be required in config and we're happy now

### DIFF
--- a/.github/deployDebian/action.yml
+++ b/.github/deployDebian/action.yml
@@ -16,6 +16,9 @@ inputs:
   PREFIX_NAME:
     description: "The prefix for the package"
     required: true
+  NPM_PACKAGE_NAME:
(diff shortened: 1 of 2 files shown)


trulyronak:
no idea why feature peek isn't working, bizzare

PR #363, opened 2020-09-17
adjusted all jobs to require npm

trulyronak:
Fixes slack bug where incorrect version is being commented, as well as bug where debian didn't comment the right version each time.

## The cause

All non-npm scripts rely on the latest npm package to release the latest version / determine the latest version number. Initially, all these jobs had relied on the release-npm package to succeed before running, but it would appear that I removed the `needs-npm` line, either for debugging, testing, or for some other non-relevant reason.

Nevertheless, by requiring npm to release and succeed successfully, this should ensure that everything works as expected in the future.


devdoshi (approved):

JaapRood:
Race condition squashed, nice one!

trulyronak:
woot last pr merged 🎉

PR #365, opened 2020-09-22
Prevent DiffViewer from leaking values to FullStory

JaapRood:
As the title describes, we really don't want interaction body values in our analytics: that data belongs to the users and should not be captured by us (in this way, at least).

**Note: this has only been in the wild for a few days and only affected the most recent CLI version which very few users had installed, we're going through and deleting logs now and notifying folks as required.**


acunniffe:
Public record: this has only been a problem for a few days, we're going through and deleting logs now.

devdoshi (approved):

PR #371, opened 2020-09-29
Migrate to a session based diff manager

JaapRood:
Currently, diffs of interactions vs spec happen on-demand, triggered from the UI. A lot of the logic for how to perform diffs and read the results were woven into the route handlers of the local `cli-server`. This is a first step in moving the diffing to the background and into a more reactive model.

Mostly an exercise in exploring the right interface and model, this PR introduces a `Session` construct for each local API running against optic, and a `SessionDiffs` manager, in charge of spawning and managing diff workers for said API. It does so through a generic `Diff` interface, exposing `DiffProgress` and `DiffQueries` interfaces to the route handlers to interact with the diffs with. Results are generally returned as `Readable` streams that can be piped straight through or manipulated. The existing `DiffManager` is renamed `OnDemandDiff` and implements this generic interface, putting it now fully in charge of matters like running the worker, persistence of results, etc.

Behaviour wise, nothing has really changed thus far. It's still the UI starting the diff that's the main trigger for a diff being performed for now. However, this PR puts us in a situation where we can implement a longer running, continuous type of diff, that be driven through other triggers. As long as the generic `Diff` interface is implemented, this can happen without changes to the route handlers or UI side.

### DIFF
--- a/workspaces/cli-server/src/routers/capture-router.ts
+++ b/workspaces/cli-server/src/routers/capture-router.ts
@@ -178,23 +178,14 @@ export function makeRouter(dependencies: ICaptureRouterDependencies) {
 
   router.get('/diffs/:diffId/stats', async (req, res) => {
     const { captureId, diffId } = req.params;
-    const diffOutputPaths = getDiffOutputPaths({
-      captureBaseDirectory: req.optic.paths.capturesPath,
-      captureId,
-      diffId,
-    });
-    try {
-      //@TODO: streamify
-      await lockfile.lock(diffOutputPaths.stats, { retries: { retries: 10 } });
-      const contents = await fs.readJson(diffOutputPaths.stats);
-      await lockfile.unlock(diffOutputPaths.stats);
-
-      res.json(contents);
-    } catch (e) {
-      res.status(404).json({
-        message: e.message,
-      });
+    const diffMetadata = diffs.get(diffId);
+    if (!diffMetadata) {
+      return res.json(404);
     }
+    const diffQueries = diffMetadata.manager.queries();
+
+    let stats = await diffQueries.stats();
+    res.json(stats);
   });
   ////////////////////////////////////////////////////////////////////////////////
 
@@ -264,4 +255,8 @@ function toJSONArray(
   return Readable.from(
     outputGenerator(objectTokenStream, tokenStream, ARRAY_ITEM_MARKER)
   ).pipe(jsonStringer());
+}
+
+function toJSONObject(): Duplex {
+  return chain([jsonDisassembler(), jsonStringer({ makeArray: true })]);
 }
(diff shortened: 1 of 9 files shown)


workspaces/ui/src/services/diff/LocalCliDiffService.ts
```diff
@@ -46,7 +46,7 @@ export class LocalCliDiffService implements IDiffService {
 
   async listUnrecognizedUrls(): Promise<IListUnrecognizedUrlsResponse> {
     const url = `${this.baseUrl}/undocumented-urls`;
-    const json = (await JsonHttpClient.getJson(url)).urls;
+    const json = await JsonHttpClient.getJson(url);
```
devdoshi, line 49: I suspect this might result in different behaviors when using the example data vs local cli, we'll have to make sure we don't break those.

workspaces/cli-server/src/routers/capture-router.ts
```diff
@@ -152,62 +147,29 @@ export function makeRouter(dependencies: ICaptureRouterDependencies) {
...
-        message: e.message,
-      });
+    const diffMetadata = diffs.get(diffId);
+    if (!diffMetadata) {
+      return res.json(404);
     }
+    const diffQueries = diffMetadata.manager.queries();
+
+    let diffsStream = diffQueries.diffs();
+    diffsStream.pipe(toJSONArray()).pipe(res).type('application/json');
```
devdoshi, line 158: is there a streamy way to wrap it in an object to preserve the original shape?

JaapRood: There totally should be and is, by prepending and appending [`stream-json` type tokens ](https://github.com/uhop/stream-json/wiki/Parser#stream-of-tokens) that describe this, but I bailed on that after 2 attempts of getting that to work right. That was while dealing some other issues too, so perhaps it's worth a revisit.

workspaces/ui/src/services/diff/LocalCliDiffService.ts
JaapRood: Ahh, I would have expected this `LocalCliDiffService` to only talk to the `cli-server`, and the `ExampleDiffService` to take care of the example data. If that's not the case, maybe we'll want to refactor things to be like that, as that would be pretty confusing 😅.

JaapRood: Or we can get the cli-server to adhere to the local API and figure out the wrapping of the streamed items!

workspaces/cli-server/src/routers/capture-router.ts
JaapRood: Found a way to implement this with and understandable API in https://github.com/opticdev/optic/pull/371/commits/60d6e55f883daca812f910fa688358c7d97438bb.

PR #372, opened 2020-09-30
Document url page

acunniffe:


workspaces/ui/public/index.html
```diff
@@ -46,7 +46,7 @@
   To create a production bundle, use `npm run build` or `yarn build`.
 -->
 <script>
-    window.opticAnalyticsEnabled = '%REACT_APP_ENABLE_ANALYTICS%' !== 'no';
+    window.opticAnalyticsEnabled = '%REACT_APP_ENABLE_ANALYTICS%' === 'yes';
```
devdoshi, line 49: x !== 'no' means enabled by default, opt-out with 'no'. x === 'yes' means disabled by default, opt-in.

PR #373, opened 2020-09-30
Learn your API page (draft)

acunniffe:
Ready for some light testing from others on the team :)


workspaces/ui/public/index.html
```diff
@@ -46,7 +46,7 @@
   To create a production bundle, use `npm run build` or `yarn build`.
 -->
 <script>
-    window.opticAnalyticsEnabled = '%REACT_APP_ENABLE_ANALYTICS%' !== 'no';
+    window.opticAnalyticsEnabled = '%REACT_APP_ENABLE_ANALYTICS%' === 'yes';
```
devdoshi, line 49: x !== 'no' means enabled by default, opt-out with 'no'. x === 'yes' means disabled by default, opt-in.

Is that the change you wanted to make?

acunniffe: For some reason the react template string doesn’t evaluate in yarn start-local so it wasn’t actually doing what it’s supposed to do. It’s on my list to investigate further before making this change.

PR #374, opened 2020-10-01
Feature/rust diff polymorphism

devdoshi:
This is a draft PR showing support for diffing the remaining polymorphic shapes (oneOf, nullable, optional). It should work, but deeper verification than the unit tests should be part of ensuring parity with the scala version.

### DIFF
--- a/workspaces/diff-engine/src/queries/shape.rs
+++ b/workspaces/diff-engine/src/queries/shape.rs
@@ -11,19 +11,15 @@ pub struct ShapeQueries<'a> {
 #[derive(Debug)]
 pub struct ResolvedTrail<'a> {
   pub core_shape_kind: &'a ShapeKind,
-  pub shape_id: &'a ShapeId,
+  pub shape_id: ShapeId,
 }
 
 impl<'a> ShapeQueries<'a> {
   pub fn new(shape_projection: &'a ShapeProjection) -> Self {
     ShapeQueries { shape_projection }
   }
-
-  pub fn list_trail_choices(
-    &self,
-    shape_trail: &ShapeTrail,
-    /*, parameter bindings */
-  ) -> Vec<ChoiceOutput> {
+  
+  pub fn list_trail_choices(&self, shape_trail: &ShapeTrail) -> Vec<ChoiceOutput> {
     let projection = &self.shape_projection;
     let root_node_index = projection.get_shape_node_index(&shape_trail.root_shape_id);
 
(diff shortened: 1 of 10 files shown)


workspaces/diff-engine/src/queries/shape.rs
```diff
@@ -60,21 +49,105 @@ impl<'a> ShapeQueries<'a> {
...
+                parent_trail: shape_trail.clone(),
+                additional_components: vec![ShapeTrailPathComponent::OptionalTrail {
+                  shape_id: shape_id.clone(),
+                }],
+                shape_id: shape_id.clone(),
+                core_shape_kind: core_shape_descriptor.kind.clone(),
+              }],
+              self.list_trail_choices(&trail),
+            ]
+            .concat()
```
JaapRood, line 119: This messing about with vecs makes one wonder if `list_trail_choices` should (optionally) accept an existing vec to push onto, could prevent a bunch of allocations. Definitely an optimization without proof it's necessary.

workspaces/diff-engine/src/queries/shape.rs
```diff
@@ -169,6 +306,43 @@ impl<'a> ShapeQueries<'a> {
...
+    projection
+      .graph
+      .neighbors_directed(*shape_node_index, petgraph::Direction::Incoming)
+      .filter_map(|neighbor_node_index| {
+        let neighbor = projection.graph.node_weight(neighbor_node_index).unwrap();
+        match neighbor {
+          Node::ShapeParameter(parameter_node) => {
+            let mut outgoing_edges = projection
+              .graph
+              .edges_connecting(*shape_node_index, neighbor_node_index);
```
JaapRood, line 328: I wonder if instead of going straight to `neighbours` and then use `edges_connecting` to get back to the edge, it wouldn't make more sense to walk the edges first and then follow the edge. Maybe another efficiency gain that's not really worth anything.

JaapRood (commented):
Looks good to me so far, with all the changes in the places where I would have expected them. I've highlighted a couple of nit-picks, but nothing essential or things that shouldn't be merged.

workspaces/diff-engine/src/queries/shape.rs
devdoshi: I was just lazy copy+pasting the singular one above - will update!

workspaces/diff-engine/src/queries/shape.rs
devdoshi: I think I can write it a better way now, will update!

devdoshi:
closed in favor of #412

PR #375, opened 2020-10-02
Coverage flag added to CLI

acunniffe:
Team [you-know-who] is really interested in our coverage ideas, specifically using them locally / in CI, not just in the Live Testing Dashboard. 

I was working on a demo for the meeting next week and realized that @JaapRood and @devdoshi did most of the work a few months back building that dashboard. only took around 90 mins to rig this up. 

This PR is very isolated, it won't breaking anything else so it should be safe to release. It does bring a lot of the spec processors from the frontend to the local CLI -- they're not factored for reusability as of yet -- just seemed hard to know exactly what we want to do w/ that bridge and felt silly to factor that w/o at least 1-2 other use cases to test the abstraction against. 

**How to test**: Pass the `--coverage` flag to the CLI when you start any Optic task. 

**Stability**: Our exact calculation for coverage may change as we evolve the feature. Depending on the % in CI isn't a good idea. The output files will be correct at report time, but may not reflect changes made the API specification (ie adding new endpoints).


workspaces/local-cli/src/shared/coverage.ts
```diff
@@ -0,0 +1,389 @@
...
+async function getReport(paths: IPathMapping, captureId: string): Promise<any> {
+  const capturesDirectory = path.join(paths.capturesPath, captureId);
+
+  const entries = await fs.readdir(capturesDirectory);
+
+  const coverageFiles = entries
+    .filter((x) => x.startsWith('coverage-'))
+    .map((x) => path.join(capturesDirectory, x));
+
+  const reportsToMerge = await Promise.all(
```
devdoshi, line 230: Can you explain the design here?

workspaces/local-cli/src/shared/coverage.ts
```diff
@@ -0,0 +1,389 @@
...
+  return sortBy(allEndpoints, (i) => i.fullPath);
+}
+
+async function getReport(paths: IPathMapping, captureId: string): Promise<any> {
+  const capturesDirectory = path.join(paths.capturesPath, captureId);
+
+  const entries = await fs.readdir(capturesDirectory);
+
+  const coverageFiles = entries
+    .filter((x) => x.startsWith('coverage-'))
```
devdoshi, line 227: this prefix should be extracted into a variable so it is linked to the other usages

workspaces/cli-shared/src/captures/avro/file-system/capture-saver-with-diffs.ts
```diff
@@ -80,6 +88,23 @@ export class CaptureSaverWithDiffs extends FileSystemAvroCaptureSaver {
...
+      .getCoverage(
+        this.shapesResolvers,
+        this.rfcState,
+        JsonHelper.jsArrayToSeq(items.map((x) => JsonHelper.fromInteraction(x)))
+      );
+
+    const asJs = opticEngine.CoverageReportJsonSerializer.toJs(report);
+
+    await fs.writeJson(
+      path.join(outputDirectory, `coverage-${batchId}.json`),
```
devdoshi, line 104: the prefix should be extracted out into a variable so the contract is clear.

workspaces/local-cli/src/shared/coverage.ts
```diff
@@ -0,0 +1,389 @@
...
+  const report = opticEngine.CoverageReportBuilder.emptyReport();
+  //merge in all the reports
+  reportsToMerge.forEach((i) => report.merge(i));
+
+  // const finalReport = opticEngine.CoverageReportJsonSerializer.toJs(report);
+
+  return report;
+}
+
+export function getCachedQueryResults(queries: any) {
```
devdoshi, line 249: I imagine this would be in @useoptic/domain-utilities if it is common across cli and web ui

workspaces/local-cli/src/shared/coverage.ts
```diff
@@ -0,0 +1,389 @@
...
+    //@ts-ignore
+    (requestId) => requests[requestId]
+  );
+  const requestsOnPathAndMethod = requestIdsOnPath.filter(
+    //@ts-ignore
+    (request) => request.requestDescriptor.httpMethod === method.toUpperCase()
+  );
+
+  const requestBodies = requestsOnPathAndMethod.map(
+    //@ts-ignore
```
devdoshi, line 333: Is there any reason for all these to be ignored?

devdoshi (changes requested):
Nice - PR description would benefit from a little feature overview / design description and testing instructions. Should users expect stability of any of the artifacts (cli output or coverage-* files?)

workspaces/local-cli/src/shared/coverage.ts
acunniffe: well the JSON serialized reports are saved alongside the interactions and diffs. In order to make them mergeable with the scala code I need to load them in, deserialize them and turn them into the scalajs obejcts.

workspaces/local-cli/src/shared/coverage.ts
acunniffe: I don't feel entirely comfortable factoring it out because it is slightly different than the other use case in the UI. It's meant to be used by this feature only for the time being so I'll remove the 'export' to make that clear.

workspaces/local-cli/src/shared/coverage.ts
acunniffe: no, I switched over to 'any'. We don't export those types from scala yet

devdoshi (approved):

PR #379, opened 2020-10-06
Support for test tasks

acunniffe:
**Problem**: During our off-cycle time, Lou and I have been thinking a lot about how to bring more sources of traffic into Optic. This should aid activation by giving people more to do once they're inside the tool. One traffic source, which many teams may come with, is their code tests, poly traffic, or postman collections, etc. Today it's non-trivial to use automated traffic with Optic tasks -- it is only doable today w/ two shells and some out-of-band knowledge for the traffic generator to know where the proxy runs. 

**Goal**: Make it possible to run automated traffic through Optic in a single shell, both locally and (eventually) in CI

**Solution**: Test tasks that are designed to run traffic against a service whose lifecycle Optic manages. For example:
- start task (`docker up`), test task (`newman run ... $OPTIC_PROXY`)
- start task (`npm start`), test task (`yarn run tests --host $OPTIC_PROXY`)

```
tasks:
  # The default task, invoke using `api run start`
  # Learn how to set up and use Optic at https://app.useoptic.com
  start:
    command: npm run server-start
    inboundUrl: http://localhost:3005
  test:
    command: "mocha test1.js"
    useTask: start
```

Screen Shot 2020-10-06 at 12 36 49 PM

**Things to review:** 
- This feels a bit 'jammed in' today, but I think it's mainly down to the dependency injection choices we've made (which make sense given the multiple clis). Open to other approaches :) 
- There's some unknown race condition with traffic saving we've never seen before. The traffic runs right away once the service is up and if we shut down the proxy before it saves it never hits the capture repository. With this in local-cli-task-runner it works, `return new Promise((resolve) => setTimeout(resolve, 500));`. Basically seems like the tail logs are liable to get lost if it's received within a few ms of triggering shutdown and cleanup of the proxy.

### DIFF
--- a/workspaces/cli-config/src/index.ts
+++ b/workspaces/cli-config/src/index.ts
@@ -151,8 +151,20 @@ function randomLowerBound(): number {
   return Math.floor(Math.random() * (max - min + 1)) + min;
 }
 
-export function isTestTask(aliasedTask: IOpticTaskAliased) {
-  return aliasedTask.useTask && aliasedTask.command;
+export function isTestTask(aliasedTask: IOpticTaskAliased): boolean {
+  return Boolean(aliasedTask.useTask && aliasedTask.command);
+}
+
+export function isRecommendedTask(aliasedTask: IOpticTaskAliased): boolean {
+  return Boolean(
+    aliasedTask.command && (aliasedTask.inboundUrl || aliasedTask.baseUrl)
+  );
+}
+
+export function isManualTask(aliasedTask: IOpticTaskAliased): boolean {
+  return Boolean(
+    (aliasedTask.inboundUrl || aliasedTask.baseUrl) && aliasedTask.targetUrl
+  );
 }
 
 export async function TaskToStartConfig(
--- a/workspaces/cli-config/src/index.ts
+++ b/workspaces/cli-config/src/index.ts
@@ -29,8 +29,9 @@ export interface ITestingConfig {
 
 export interface IOpticTask {
   command?: string;
-  baseUrl: string;
+  baseUrl?: string;
   proxy?: string;
+  useTask?: string;
   targetUrl?: string;
 }
 
@@ -95,6 +96,8 @@ export class TargetPortUnavailableError extends Error {}
 
 export class TaskNotFoundError extends Error {}
 
+export class RunsWithTaskNotFoundError extends Error {}
+
 export interface IOpticCaptureConfig {
   persistenceEngine: 'fs' | 's3';
   captureId: string;
@@ -148,12 +151,16 @@ function randomLowerBound(): number {
   return Math.floor(Math.random() * (max - min + 1)) + min;
 }
 
+export function isTestTask(aliasedTask: IOpticTaskAliased) {
+  return aliasedTask.useTask && aliasedTask.command;
+}
+
 export async function TaskToStartConfig(
   aliasedTask: IOpticTaskAliased
 ): Promise<IOpticTaskRunnerConfig> {
   const task = normalizeTask(aliasedTask);
 
-  const baseUrl = url.parse(task.baseUrl);
+  const baseUrl = url.parse(task.baseUrl!);
   const targetUrl =
     (task.targetUrl && url.parse(task.targetUrl)) ||
     (task.proxy && url.parse(task.proxy)); // TODO: add deprecation warning
@@ -199,8 +206,14 @@ export async function TaskToStartConfig(
 
 function normalizeTask(aliased: IOpticTaskAliased): IOpticTask {
   const baseUrl = aliased.baseUrl || aliased.inboundUrl;
-  if (!baseUrl)
+  if (!baseUrl && !aliased.useTask)
     throw new Error('Task definition must have baseUrl (aliases: inboundUrl)');
+
+  if (aliased.useTask && (baseUrl || aliased.targetUrl)) {
+    throw new Error(
+      'Task definitions with useTask should not include an inboundUrl or targetUrl '
+    );
+  }
 
   return {
     ...aliased,
--- a/workspaces/cli-config/src/index.ts
+++ b/workspaces/cli-config/src/index.ts
@@ -318,4 +318,10 @@ export interface ITestingConfig {
   authToken: string;
 }
 
+export enum Modes {
+  Recommended = 'Recommended',
+  Manual = 'Manual',
+  Test = 'Test',
+}
+
 export { parseIgnore, parseRule, IIgnoreRunnable };
--- a/workspaces/cli-shared/src/captures/avro/file-system/capture-saver.ts
+++ b/workspaces/cli-shared/src/captures/avro/file-system/capture-saver.ts
@@ -2,8 +2,6 @@ import Bottleneck from 'bottleneck';
 import fs from 'fs-extra';
 import path from 'path';
 import avro from 'avsc';
-//@ts-ignore
-import waitUntil from 'wait-until';
 import {
   IInteractionBatch,
   IGroupingIdentifiers,
@@ -24,8 +22,12 @@ export class CaptureSaver implements ICaptureSaver {
     maxTime: 500,
   });
 
-  private enqueuedForSave: number = 0;
-  private completed: number = 0;
+  private throttler: Bottleneck = new Bottleneck({
+    maxConcurrent: 10,
+    minTime: 1,
+  });
+
+  public savePromises: Promise<void>[] = [];
 
   private batchCount: number = 0;
 
@@ -50,12 +52,14 @@ export class CaptureSaver implements ICaptureSaver {
         batchId,
       };
       try {
-        await this.onBatch(
-          groupingIdentifiers,
-          batchId,
-          items,
-          outputDirectory
-        );
+        await this.throttler.schedule(async () => {
+          await this.onBatch(
+            groupingIdentifiers,
+            batchId,
+            items,
+            outputDirectory
+          );
+        });
       } catch (e) {
         console.error(e);
       }
@@ -98,22 +102,15 @@ export class CaptureSaver implements ICaptureSaver {
   }
 
   async save(sample: IHttpInteraction) {
-    this.enqueuedForSave++;
     // don't await flush, just enqueue
-    const promise = this.batcher.add(sample);
-    promise.finally(() => {
-      this.completed++;
-    });
+    await this.batcher.add(sample);
   }
 
   async cleanup() {
-    await new Promise((resolve) => {
-      waitUntil()
-        .interval(100)
-        .times(10)
-        .condition(() => this.completed === this.enqueuedForSave)
-        .done(resolve);
+    await new Promise((resolve, reject) => {
+      this.throttler.on('idle', resolve);
     });
+
     developerDebugLogger('stopping capture saver');
   }
 }
--- a/workspaces/cli-shared/src/captures/avro/file-system/capture-saver.ts
+++ b/workspaces/cli-shared/src/captures/avro/file-system/capture-saver.ts
@@ -2,6 +2,8 @@ import Bottleneck from 'bottleneck';
 import fs from 'fs-extra';
 import path from 'path';
 import avro from 'avsc';
+//@ts-ignore
+import waitUntil from 'wait-until';
 import {
   IInteractionBatch,
   IGroupingIdentifiers,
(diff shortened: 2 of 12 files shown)


LouManglass:
Initial thought: nice.

The coverage detail report breaks out all bodies like in the dashboard, so I can see not only which endpoints I'm covering but which response bodies/status codes I'm covering. Great to know I'm hitting an endpoint, but missing a potential edge case that we should test and for which we should get a graceful response.

The summary data is helpful: tell me how much, in aggregate, I've hit. I can go back to the details if I'm not at my expectations (we're still building our coverage, so we need a minimum of 80% for example. I can tell that right away as the last thing the output spits out). If I don't hit my expectations, I can dig into why and find my easy wins (oh, I documented this route and forgot to write some automated traffic tests. And hey, my tests didn't cover some of these cases we've documented. Cool, I'll just drop those in my Postman collection).

acunniffe:
@LouManglass  -- how do the test tasks feel in the optic.yml? Ergonomics? Can you see a good documentation story around them?

LouManglass:
> @LouManglass -- how do the test tasks feel in the optic.yml? Ergonomics? Can you see a good documentation story around them?

It's fine. It's just invoking another task, I wouldn't get fancy with it. Just leave it simple. It _shouldn't_ have to do much. If my testing harness is complicated, there's likely already a bash script or some other rigging that can be invoked with `command`.

Here's a walkthrough for posterity https://www.loom.com/share/57d6ebb34c984560a2725be73dc96528

workspaces/cli-config/src/index.ts
```diff
@@ -29,8 +29,9 @@ export interface ITestingConfig {
 
 export interface IOpticTask {
```
devdoshi, line 30: probably better to make this a type composed of 2 interfaces (something like type IOpticTask = IOpticStandardTaskDefinition | IOpticTestTaskDefinition). That way there will be fewer states

acunniffe: Yeah I can do that — it makes more sense now that we have all the uses cases in mind. 

Overall how’s the implementation look?

acunniffe:
@LouManglass this now works with api check and lists all tasks when you don’t specify one. Can you test out some use cases where useTask is manual, recommended and invalid?

workspaces/local-cli/src/shared/local-cli-task-runner.ts
```diff
@@ -73,7 +76,35 @@ export async function LocalTaskSessionWrapper(
   const captureId = uuid.v4();
   const runner = new LocalCliTaskRunner(captureId, paths);
   const session = new CliTaskSession(runner);
-  await session.start(cli, config, taskName);
+
+  const task = config.tasks[taskName];
+
+  if (!task) {
```
devdoshi, line 82: This is nice, I think you can handle the empty task list case too, in case they haven't specified any yet.

workspaces/local-cli/src/shared/spawn-process.ts
```diff
@@ -0,0 +1,24 @@
+import { spawn, SpawnOptions } from 'child_process';
```
devdoshi, line 1: not sure if there are subtle differences but I think you could have used command-session.ts

workspaces/local-cli/src/shared/verify/verify.ts
```diff
@@ -108,16 +107,45 @@ export async function verifyTask(
...
+          )
+        )
+      );
+
+      passedAll = await verifyTask(cli, foundTask!.useTask!, true);
+    }
+  }
+
   if (mode == Modes.Recommended) {
+    const startConfig = await TaskToStartConfig(foundTask!);
```
devdoshi, line 148: the ! all over the place makes me think that the types aren't right, unless you're running into some ts limitation

workspaces/local-cli/src/shared/local-cli-task-runner.ts
```diff
@@ -167,7 +199,29 @@ ${blockers.map((x) => `[pid ${x.pid}]: ${x.cmd}`).join('\n')}
...
+                colors.grey.bold(commandToRunWhenStarted)
+            )
+          );
+
+          await spawnProcess(commandToRunWhenStarted!, {
+            OPTIC_PROXY_PORT: taskConfig.proxyConfig.port.toString(),
+            OPTIC_PROXY_HOST: taskConfig.proxyConfig.host.toString(),
+            OPTIC_PROXY: `http://${taskConfig.proxyConfig.host.toString()}:${taskConfig.proxyConfig.port.toString()}`,
+          });
+          return new Promise((resolve) => setTimeout(resolve, 500)); // needs time to finish saving
```
devdoshi, line 217: command-and-proxy-session-manager.ts line 100 calls persistenceManager.cleanup(), which in this flow is the file-system/capture-saver-with-diffs.ts, which extends capture-saver.ts. the saas implementation handles the asynchronous calls to save() that batch after cleanup() is called. So, a less time-based and more event-based way would be to modify cleanup() so it stops receiving and waits for the batches to finish before resolving cleanup()

devdoshi:
Looks good overall. Looking forward to the possibility of live coverage and your todo-list ui that makes the report more actionable.

LouManglass:
> @LouManglass this now works with api check and lists all tasks when you don’t specify one. Can you test out some use cases where useTask is manual, recommended and invalid?

useTask suggested tasks works now, digging into a few alternative scenarios for base tasks.

LouManglass:
When running a failing task, we get the failure of the base task passed through. Makes sense, perhaps a bit unclear but I don't think adding overhead to say which task has a bad definition is worth it (maybe an hour of appetite tops).

```
$ apidev run start-fail 
    Error: Task definition must have baseUrl (aliases: inboundUrl)
$ apidev run use-fail
[optic] Running dependent task start-fail...
    Error: Task definition must have baseUrl (aliases: inboundUrl)
```

If the base task doesn't exist, it fails as expected with a clear message:
```
apidev run use-notask
[optic] Running dependent task arglebargle...
[optic] No task named arglebargle found in optic.yml
```

On to proxy definitions. Remember how we can't run a short-lived process like "docker up" with command, and we use Optic as a proxy then run our other commands in another window? **not anymore**

```
  start-proxy:
    targetUrl: http://localhost:3500
    baseUrl: http://localhost:3001
  use-proxy:
    useTask: start-proxy
    command: OPTIC_API_PORT=3500 npm run start
```

Note I needed the OPTIC_API_PORT specified there when I tested it. We could get past that if OPTIC_API_PORT is set to the port defined in targetUrl. If I don't set that, I get...

```
 apidev run use-proxy                                                                                                                                                                                         
[optic] Running dependent task start-proxy...
[optic] Review the API Diff at http://localhost:3000/apis/1/diffs
[optic] Optic is observing requests made to http://localhost:3001
[optic] Running test command npm run start

> ergast-docs-example@1.0.0 start /Users/lou/repos/example-ergast-project
> node lib/index.js

Server running on http://localhost:3001
```

which, what? That's weird.


Assuming we're running our infra ourselves (say, it's a behind an API gateway or I do some docker up magic on my time, or something something...I can spin up the proxy and run test traffic.

```
  start-proxy:
    targetUrl: http://localhost:3500
    baseUrl: http://localhost:3001
  use-proxy-test:
    useTask: start-proxy
    command npm run simulate-traffic
```

However, I don't get a coverage report until I break out of the process in this case. Note I see the end of the Newman report, then `^C`, then Optic coverage.

```
 total data received: 31.32KB (approx)                              │
├────────────────────────────────────────────────────────────────────┤
│ average response time: 406ms [min: 319ms, max: 704ms, s.d.: 119ms] │
└────────────────────────────────────────────────────────────────────┘
^C[optic] Observed Unexpected API Behavior. Click here to review: http://localhost:3000/apis/1/diffs/ed92301c-d554-42ad-b3d2-366c18e0511a
[optic] Calculating Coverage...... done


 API Coverage Report:
┌─────────────────────────────────────────────────────┬───────────────────────────────────┐
│ Endpoint                                            │ Coverage                          │
├─────────────────────────────────────────────────────┼───────────────────────────────────┤
│ GET /api/f1/:yr/1/constructors                      │ Requests (1/1) No Body ✔          │
```

acunniffe:
Thanks @LouManglass @devdoshi -- I'll address these things tomorrow 👍 . Lou I think proxy mode wasn't implimented properly -- should be quick to fix. FYI you can use $OPTIC_PROXY in your test tasks

```
use-proxy:
    useTask: start-proxy
    command: OPTIC_API_PORT=$OPTIC_PROXY npm run start
```

LouManglass:
> Thanks @LouManglass @devdoshi -- I'll address these things tomorrow 👍 . Lou I think proxy mode wasn't implimented properly -- should be quick to fix. FYI you can use $OPTIC_PROXY in your test tasks
> 
> ```
> use-proxy:
>     useTask: start-proxy
>     command: OPTIC_API_PORT=$OPTIC_PROXY npm run start
> ```

Oh, perfect. The sample API relies on `OPTIC_API_PORT || nnnn` so that's why I shimmed it this way, but ideally that could just be `command --port $OPTIC_PROXY` then and it looks like a "normal" command.

workspaces/local-cli/src/shared/local-cli-task-runner.ts
acunniffe: it's not obvious to me what I should await in that cleanup -- where's the example for agents?

devdoshi: workspaces/cli-shared/src/command-and-proxy-session-manager.ts:34 is where we save interactions. We never wait for those promises, so can add an implicit guarantee that cleanup() ensures interactions have been flushed, or perhaps better and more simply, the command-and-proxy-session-manager can add those promises to a structure like a task queue (e.g. using bottleneck) and wait for the queue to be empty when stopping

devdoshi: So you probably won't end up messing with cleanup, but here's what the agents use: workspaces/cli-shared/src/captures/avro/saas/capture-saver.ts

workspaces/cli-config/src/index.ts
devdoshi: Ah I missed these types changing around IOpticTaskAliased, so I might be misreading - but to clarify: the goal for me is to not have types with states that are representable but do not or should not occur. If that can still here, then I'd consider changing the types. Your functions below should probably be https://www.typescriptlang.org/docs/handbook/advanced-types.html#type-guards-and-differentiating-types type guards so typescript can narrow the types for you.

devdoshi: Once you do that, you should be able to remove a lot of the !s that seem to be unnecessary

workspaces/local-cli/src/shared/local-cli-task-runner.ts
acunniffe: @devdoshi check out the latest. I'm saving the promises and doing `promise.all()`. this might be memory unfriendly at high volumes though.

workspaces/cli-shared/src/captures/avro/file-system/capture-saver.ts
```diff
@@ -44,12 +52,14 @@ export class CaptureSaver implements ICaptureSaver {
         batchId,
       };
       try {
-        await this.onBatch(
-          groupingIdentifiers,
-          batchId,
-          items,
-          outputDirectory
-        );
+        await this.throttler.schedule(async () => {
```
devdoshi, line 55: I think you want to put const promise = this.onBatch(...) , and then schedule(() => promise) and then await promise. that way, the flushing done in onBatch proceeds immediately, and the recordkeeping of scheduled work happens on its own separately

workspaces/cli-shared/src/captures/avro/file-system/capture-saver.ts
```diff
@@ -21,6 +21,14 @@ export class CaptureSaver implements ICaptureSaver {
     maxSize: 20,
     maxTime: 500,
   });
+
+  private throttler: Bottleneck = new Bottleneck({
```
devdoshi, line 25: in this usage it's not serving as a throttler, just as a way to track pending batches, so it can be named better

workspaces/cli-shared/src/captures/avro/file-system/capture-saver.ts
```diff
@@ -21,6 +21,14 @@ export class CaptureSaver implements ICaptureSaver {
...
+    maxConcurrent: 10,
+    minTime: 1,
+  });
+
+  public savePromises: Promise<void>[] = [];
```
devdoshi, line 30: this can be removed now

PR #380, opened 2020-10-09
analytics fix -- fullstory/segment integration 

acunniffe:


LouManglass (approved):
Per pairing, review of behavior on call

PR #384, opened 2020-10-14
Learn API Page

acunniffe:
This PR contains a mostly implemented Learn API UX. You can watch it learn one of the sessions from a user with the largest captures yet in the team Slack. 

When batch learning takes place, the UX is blocked and a progress indicator ticks as more and more endpoints are learned. 

The contract between the app and the daemon is pretty straightforward -- given `events` `pathId` and `method` -> return all the new request/response bodies and the commands required to add these regions to the spec. 

I've mostly followed the diff-worker patterns, but changed IPC so the manager return Promise<result> -- seemed more inline with what we needed for this use case since incremental results per endpoint are meaningless 

The UX uses bottleneck to stage 4 inferences at a time, when each finishes, the progress bar updates, and the next scheduled job begins. It may make sense to increase or decrease that number, but hard to tune it that much since we're targeting end-user hardware.

I'm most interested in feedback on my async worker, IPC, and UX. 

Thanks!


JaapRood (commented):
I'm at the moment of wrapping up my day, but had a first initial glance at this and think it's important to highlight a possible issue with the overall approach.

It's great that you've tried to implement this in the same way as the current production diff pipeline. What's unfortunate is that as part of the Reactive Diffs pipeline we've just moved away from this, the first iteration of which landed in `develop` yesterday. There, I'm not talking about running against the new Rust Diff engine, but even just how the Scala based one works. During that work of moving away from it, we've identified multiple issues with the approach, especially around failure modes (current production suffers from silently failing diffs for bigger captures).

So I'm not quite sure how we'll want to continue with this. While the new approach isn't  complete, at a minimum we'll want this part of the pipeline to adhere to new interface we created for this in the last couple of weeks. Otherwise, this PR greatly inflates the amount of work necessary to get the new Diff pipeline to work, which is not a way I think projects should interact. That said, the reverse is also kind of true 🤷‍♂️.

acunniffe:
@JaapRood  -- great -- I'm willing to work with whatever pattern you guys think is best and fully expect my first Rust contributions will be the shape builder / infer poly stuff 👍. There's not enough for me to work off of in your comments, but happy to do some of the work to move towards something better if you have a pattern I can follow.

workspaces/cli-server/src/diffs/initial-body-manager.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  async run(config: IInitialBodyManagerConfig): Promise<ILearnedBodies> {
+    const outputPaths = getInitialBodiesOutputPaths(config);
+    const scriptConfig: IInitialBodiesProjectionEmitterConfig = {
+      captureId: config.captureId,
+      pathId: config.pathId,
+      method: config.method,
+      specFilePath: outputPaths.events,
+      captureBaseDirectory: config.captureBaseDirectory,
+    };
```
JaapRood, line 29: We adapted the signature of this in `run` (renamed `start`) to no longer accept the config, but instead construct it itself. Ultimately, if the task requires any persistence to compute it's results, that should be an implementation detail to the task, not something an API route handler should have to worry about. See the comment in the route-handler for more info.

Here's how we've currently approached that for the diffing task, effectively just moving the logic in here: 

https://github.com/opticdev/optic/blob/19386cab002eae2af7db3d8fe3969ce1248ff005/workspaces/cli-server/src/diffs/on-demand.ts#L44-L64

The initial config is assigned at moment of construction: https://github.com/opticdev/optic/blob/19386cab002eae2af7db3d8fe3969ce1248ff005/workspaces/cli-server/src/diffs/on-demand.ts#L32-L35

workspaces/cli-server/src/diffs/initial-body-manager.ts
```diff
@@ -0,0 +1,56 @@
...
+      });
+    });
+
+    return await result;
+  }
+
+  async stop() {
+    if (this.child) {
+      this.child.kill('SIGTERM');
+    }
```
JaapRood, line 54: If, as mentioned above, the event handlers match the lifecycle of the child process, calling stop on this learning task should resolve the results `Promise`, probably rejecting it for being interrupted and incomplete.

workspaces/cli-server/src/routers/capture-router.ts
```diff
@@ -99,6 +102,42 @@ export function makeRouter(dependencies: ICaptureRouterDependencies) {
...
+      const { events, pathId, method } = req.body;
+      const manager = new InitialBodyManager();
+      const outputPaths = getInitialBodiesOutputPaths({
+        captureBaseDirectory: req.optic.paths.capturesPath,
+        captureId,
+        pathId,
+        method,
+      });
+      await fs.ensureDir(outputPaths.base);
+      await Promise.all([fs.writeJson(outputPaths.events, events)]);
```
JaapRood, line 120: We moved away from spawning tasks right from API route handlers. Instead, we prefer all that logic to live somewhere inside the rest of the server process. One of the major reasons is that the task can very well outlive the request, but still needs to be properly managed. Even when that's not the case, it's nice to keep implementation details about the task, for the task.

With the diffs, we ended up writing a `Session` class that's attached to the `req` and exposes a method to start a new diff. https://github.com/opticdev/optic/blob/19386cab002eae2af7db3d8fe3969ce1248ff005/workspaces/cli-server/src/routers/capture-router.ts#L81-L100

Since the diff _is_ a longer running process, it then also exposes methods to query the results of the diff in   other route handlers.

https://github.com/opticdev/optic/blob/19386cab002eae2af7db3d8fe3969ce1248ff005/workspaces/cli-server/src/routers/capture-router.ts#L129-L139

JaapRood (changes requested):
I had a bit of a more detailed look at what we can do to make this new part of the pipeline fit in with the new approach we've formulated. Fortunately, it looks like it should be _mostly_ moving some code around, or copying approaches from the diff tasks ☺️.

Reading through the code, I realised this isn't really a diffing task as much as it's a spec updating task. Yes, it does diffing in order to update the spec correctly, but that's almost more of an implementation detail. Therefor, I don't think it has to implement the new `Diff`, `DiffQueries` and `DiffProgress` interfaces.

Firstly, if we can agree on the above, I'd recommend moving this out of the `src/diffs` folder and putting it somewhere that better personifies this. Since it's a long running task the most safe might just be `src/tasks`, where it can live until we find some better way to group it as similar tasks pop up.

Secondly, I think the responsibility of spawning and managing the lifecycle of the task should be moved from the route handlers to the new `Session` class. An instance of this class is attached to each request and provides all the context needed about the specific API the task is ran for. More detailed motivation you'll find in the inline comments. Here's an example of how a diffs are now started through the same means: https://github.com/opticdev/optic/blob/19386cab002eae2af7db3d8fe3969ce1248ff005/workspaces/cli-server/src/sessions.ts#L59-L69

Apart from some of these changes, I'm also wondering what the idea is behind persisting the results to disk. Generally, I'd expect persistence of this data to happen when the results can be queried later, or if a paused / interrupted task can later be resumed. However, here the results seem to be returned straight apes a result of running the task, with no further way to access them or tasks to be resumed. Perhaps during development you ran into other issues that caused you to choose persistence to disk instead of doing it in-memory?

workspaces/cli-server/src/routers/capture-router.ts
acunniffe: This is the only part I'd call out-- the response is the thing managing the lifecycle here. You start a request, it does work, it returns the result to the UX. The constract, as currently written does not have an affordance for outliving the HTTP call

acunniffe:
@JaapRood on board for copying over the new styles, this is helpful guidance. 

Also re: persistence, at first it made a lot of sense since I was copying code, but once I realized the IPC supported messaging back to parent, it got less useful. I'll remove it 👍

workspaces/cli-server/src/routers/capture-router.ts
JaapRood: Those lifecycles match and are tied, point taken. That leaves a bunch of other concerns that don't really belong in a route handler, like setting up the config that's related to the session context, piping the logging and errors to the right place (which isn't the response, leaking implementation details we'll want to avoid), etc.

acunniffe: Yeah, point 2, well taken. I'm checking out the new styles and taking a stab before happy hr.

acunniffe:
@JaapRood  -- should be good for another review. I like the new styles 👍  nice work

JaapRood (changes requested):
Thanks for all the work in trying to fit it all in. I think there might have been a miscommunication though :/ I didn't mean to suggest that the `OnDemandDiff` and your new `InitialBody` task share the same interface: they are both tasks, but a `Diff` is a way more specific type of Task.

I'm going to suggest reverting all the changes to the `OnDemandDiff`, the `Diff` interface and `SessionDiffs`: those are all specific to diffing and will instantly divert from what the `OnDemandInitialBody` needs. You can already see the indicators of that, with the methods for it throwing "unimplemented" errors.

The `Session` class can directly house and manage your new task, leaving the `SessionDiffs` to deal with the background diffing.

Hope that's fair!

acunniffe:
Cool that's a quick change :) on it

workspaces/ui/src/contexts/CaptureContext.js
```diff
@@ -75,7 +75,7 @@ class _CaptureContextStore extends React.Component {
           if (this.state.notificationChannel) {
             this.state.notificationChannel.close();
           }
-          // track('Completed Diff', {
+          // track('Completed IAsyncTask', {
```
JaapRood, line 78: Not sure if you left this intentionally, but might be a bit of revert back from `AsyncTask` -> `Diff` missed.

workspaces/ui/src/components/diff/v2/DiffReviewExpanded.js
```diff
@@ -152,7 +152,7 @@ export const DiffReviewExpanded = (props) => {
   const { interaction, interactionScala } = currentInteraction;
 
   const { method, path } = interactionScala.request;
-  // track('Display Diff in Behavior Page', props);
+  // track('Display IAsyncTask in Behavior Page', props);
```
JaapRood, line 155: Another bit of rename that you might want to revert as well, in case this comment is ever re-enabled!

workspaces/cli-server/src/routers/capture-router.ts
```diff
@@ -99,6 +94,75 @@ export function makeRouter(dependencies: ICaptureRouterDependencies) {
...
+          console.log('progress tick ', JSON.stringify(data));
+          if (!data.hasMoreInteractions && data.results) {
+            resolve(data.results);
+          }
+        });
+        initialBodyGenerator.events.on('exit', reject);
+        initialBodyGenerator.events.on('error', reject);
+      });
+
+      result.finally(() => initialBodyGenerator.stop());
```
JaapRood, line 126: This result Promise is going to leak the EventEmitter and possibly cause a race condition in rejecting a Promise that was already resolved.

I imagine it was introduced as to fit the `start` method of the `Diff` interface, which doesn't allow returning results directly. Since that interface is no longer implemented, I can imagine going back to that. If that poses some issue, perhaps using the `initialBodyGenerator.progress()` steam is a reasonable alternative, which does all the error-prone subscribing and unsubscribing from events, dealing with back-pressure, etc.

JaapRood (changes requested):
Awesome! Apart from some small issues (deriving results from event emitters is painful 😞) it's looking pretty good and in-line with the rest of our efforts 🙂.

JaapRood (approved):
Looks good to me, nothing to block merging this, especially as it lives behind a feature flag.

There's a couple of small things one could take on board if there was enough time, attention and motivation do so something about them:

- the `safeWrite` and `safeRead` use locks, while the new ignore file format allows for append-only writes, which makes both reading and writing safe without locks. Most important bit that makes this possible is that `fs.createReadStream`, which all fs functions use under the covers, reads a snapshot of the file at the moment of acquiring the file descriptor.
- I reckon the snapshots we're producing don't require all of the might and size of Jest, but could be just as easily accomplished with string-based snapshots. Maybe that doesn't matter, but I've been hit by Jest's mass and complexity too many times to use it as a go-to option 😅.

PR #385, opened 2020-10-14
Traverse and visit BodyDescriptor instead of JsonValue to allow for future optimisations

JaapRood:
One of the major optimisations we found (deduplicating array items and only visiting uniques) relies on a construct that can encode the structure of a body, rather than all values of it. Rather than only applying this optimisation where the inputs source allows for it (shape hashes), this PR changes it so that each type of source is converted into this one intermediate representation and always perform the optimisation.

This new construct is called `BodyDescriptor` (should probably be a `ShapeDescriptor`, but that name is already used to indicate specced shapes) into which `JsonValue` and `shape_hash::ShapeDescriptor` are converted. Like  `shape_hash::ShapeDescriptor` it encodes only the shape, not actual values for now. However, it is not limited by that, in that it _could_ hold some value information if available, but also metadata about how it maps back to the full body (like mapping of unique array items to all array items). Since `shape_hash::ShapeDescriptor` is codegened from a protobuf, retrofitting that seems like a bad idea.

### DIFF
--- a/workspaces/diff-engine/src/state/body.rs
+++ b/workspaces/diff-engine/src/state/body.rs
@@ -4,7 +4,7 @@ use serde_json::Value as JsonValue;
 
 #[derive(PartialEq, Clone, Debug)]
 pub enum BodyDescriptor {
-  Object(Vec<FieldDescriptor>),
+  Object(ObjectDescriptor),
   Array(ItemsDescriptor),
   String,
   Number,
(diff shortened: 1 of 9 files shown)


workspaces/diff-engine/src/state/body.rs
```diff
@@ -0,0 +1,146 @@
...
+  }
+}
+
+impl ItemsDescriptor {
+  pub fn into_all(self) -> impl Iterator<Item = BodyDescriptor> {
+    self.all_items.into_iter()
+  }
+
+  // @TODO: implement way to get unique items rather than all items
+  // pub fn unique(&self) -> Vec<(BodyDescriptor, Vec<usize>)> {}
```
devdoshi, line 77: So how is the performance until this changes?

devdoshi:
Nice, looks like a rusty version of JsonLike from the scala code. Reviewing on mobile but looks good

PR #386, opened 2020-10-14
Prevent visiting of duplicate array items when diffing Body with Shape

JaapRood:
This PR implements an optimisation around the visiting of array items of a body we're diffing. 

In our experiments, visiting array items is where most time is spent when diffing larger bodies. Conceptually, this makes sense, as this is the main unit of scale: it's rare to see thousands of fields for objects, it's common to see thousands or array elements. 

Fortunately, by nature of array semantics, multiple items often share the exact same structure. Taking the time to deduplicate them by comparing the structure by hashing it and perform a deep comparison, pays off very well. In one a particular debug capture taken from real-world usage, we saw a diff reduce from approx  `4m30s` to approx `0m10s`.

It uses Rust's built-in `Hash` and `Eq` traits to perform the comparison, reducing an array of items down to the unique ones and only storing those. A vector of indexes of the original elements is kept, so it can be mapped when writing results about the original `Body`. This yields the same performance improvement as from our experiments.

### DIFF
--- a/workspaces/diff-engine/src/shapes/traverser.rs
+++ b/workspaces/diff-engine/src/shapes/traverser.rs
@@ -84,9 +84,9 @@ impl<'a> Traverser<'a> {
           _ => unreachable!("expect body to be an array"),
         };
 
-        items.into_all().enumerate().for_each(|(index, item)| {
+        items.into_unique().for_each(|(item, indexes)| {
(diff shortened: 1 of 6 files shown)


JaapRood:
There's a couple of snapshot tests failing, which I'll have a closer look into whether that's correct behaviour. If so, it's likely these existing snapshots already cover the new behaviour implicitly, which might make writing additional tests to test it explicitly not immediately necessary.

PR #388, opened 2020-10-15
Add normalisation of diffs produced through new Rust diff engine

JaapRood:
So far in the Rust diff engine identical diffs recorded across multiple interactions are treated as separate diffs, rather than a single diff tagged with multiple interactions. These changes re-implements this.

This is a work in progress, I'll add more information as this PR makes it out of draft.

### DIFF
--- a/workspaces/cli-server/src/routers/capture-router.ts
+++ b/workspaces/cli-server/src/routers/capture-router.ts
@@ -160,30 +160,7 @@ export function makeRouter(dependencies: ICaptureRouterDependencies) {
 
     let diffsStream = diffQueries.diffs();
 
-    async function* normalizeDiffs(resultsStream: Readable) {
-      let pointersByFingerprint: Map<String, string[]> = new Map();
-      let diffs: [any, string][] = [];
-
-      for await (let [diff, pointers, fingerprint] of resultsStream) {
-        let existingPointers = pointersByFingerprint.get(fingerprint) || [];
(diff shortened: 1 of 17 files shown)


workspaces/cli-server/src/routers/capture-router.ts
```diff
@@ -135,7 +135,31 @@ export function makeRouter(dependencies: ICaptureRouterDependencies) {
...
+      for await (let [diff, pointers, fingerprint] of resultsStream) {
+        let existingPointers = pointersByFingerprint.get(fingerprint) || [];
+        if (existingPointers.length < 1) {
+          diffs.push([diff, fingerprint]);
+        }
+        pointersByFingerprint.set(
+          fingerprint,
+          existingPointers.concat(pointers)
+        );
+      }
```
JaapRood, line 152: Just realised this is going to fail against the non-Rust version of the diff engine, which doesn't include the fingerprint with each result. We can either move this normalisation to the `DiffQueries.diffs` implementation, where a fork already exists, or we can adapt this to yield the result untouched when there's no fingerprint. Not sure which has my preference yet, or if it even matters which one we choose.

JaapRood: Of course, I ended up doing both: https://github.com/opticdev/optic/pull/388/commits/cc43c8408c0b430251b5ecf5f4e8af7c1d08088c.

devdoshi (approved):

PR #393, opened 2020-10-20
Hotfix diffing of spec changes yet to be committed

JaapRood:
#371 inadvertently introduced a change in behaviour where only endpoints with a single diff could be handled by the UI. Endpoints with multiple diffs, involving the diffing of events that have not yet been committed, got stuck at the first diff.

This fixes the issue by once again accepting the spec events to diff from the API endpoint the UI uses to start a diff. There's a bunch of good reasons why we're trying to move away from that interface, but this use case went unconsidered in that so far. We should figure that out before re-enabling this.

### DIFF
--- a/workspaces/cli-server/src/diffs/index.ts
+++ b/workspaces/cli-server/src/diffs/index.ts
@@ -51,6 +51,7 @@ export interface DiffConfigObject {
   captureId: string;
   captureBaseDirectory: string;
   diffId: string;
+  events?: Array<{ [key: string]: any}>;
   endpoints?: Array<{ pathId: string; method: string }>;
   specPath: string;
 }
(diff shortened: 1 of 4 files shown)


acunniffe:
@JaapRood  -- this doesn't respect `additionalCommands` anymore. It's a pretty small change on the frontend to make it send its current Rfc Events (since it has those already) instead of base events + additional commands. 

Before we change the contract just wanted to verify that's what you had intended.

JaapRood:
> @JaapRood -- this doesn't respect `additionalCommands` anymore. It's a pretty small change on the frontend to make it send its current Rfc Events (since it has those already) instead of base events + additional commands.

We looked into this and discovered the client already didn't send `additionalCommands` anymore, but instead added the new events. While we could change that, we decided to leave it that way for now. The Rust diff worker doesn't apply commands yet, so it's the most straightforward way at this point in time.

JaapRood:
Testing this more, I can no longer reproduce the reported issue. Time to land it!

PR #394, opened 2020-10-20
Derive undocumented urls of diffs produced through new Rust diff engine

JaapRood:
The diffs UI retrieves undocumented urls as a separate resource, which must be built from detected diffs. So far, the Rust diff engine and its pipeline did not build this resource, making it impossible to add any new endpoints to the spec using it. These changes implement this part of the pipeline and provides the required resource.

More details on implementation to come as this makes it out of draft.


JaapRood:
This is still behind a feature flag, so moving forward with the merge. Any issues from review can be dealt with separately.

PR #395, opened 2020-10-21
Build and test the Diff engine as part of CI

JaapRood:
We'll want to make sure that our Rust code builds on all the targets we're after. The earlier we can get feedback on any of that not working out, the better. Doing so for each pull request and push to `develop` seems like a pretty good place for that feedback loop.


JaapRood:
@notnmeyer especially interested to hear what you have to say about this most basic setup. Rather than the example we've discussed before where everything was built on ubuntu, I took the approach of Github's support for native hosts. That should let `rustup` and `cargo` deal with all the specifics given it's host os. 

I'm sure this is fairly naive, but perhaps not a bad place to start. There's a bunch more fidelity `cargo build` gives us, like more specifically setting the target rather than just relying on the default choice for the host OS. Apparently, [there are limits to this](https://medium.com/@jondot/building-rust-on-multiple-platforms-using-github-6f3e6f8b8458), but I don't think we're worrying about any of those limits quite yet. 

The alternative would be running everything on `ubuntu-latest`, but that does make everything more complicated. Rust's official `cross` tool that can be used instead of cargo would cover most of the targets, but very essentially [its supported targets](https://github.com/rust-embedded/cross#supported-targets) is missing Darwin (seems mostly focused on embedded devices). Even if we do get things to cross compile for Darwin as well, I imagine code signing perhaps becoming an issue.

.github/workflows/build.yml
```diff
@@ -12,8 +12,39 @@ on:
...
+        with:
+          toolchain: stable
+      - name: 'Cache cargo dependencies'
+        uses: actions/cache@d1255ad9362389eac595a9ae406b8e8cb3331f16 # https://github.com/actions/cache/commits/v2
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            target
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
```
devdoshi, line 35: Looks good overall, but I'm not sure this is working. I re-ran the jobs just now and some failed. Also, would it need to be keyed by runner.os?

JaapRood: Yeah, the caching seems to fail, basically just copied https://github.com/actions/cache/blob/main/examples.md#rust---cargo. 

As far as I understand Cargo's cache includes compiled libraries, so yeah, we'd want to cache those separately for each OS.

notnmeyer:
a matrix build seems like a good starting place to me. what do you think about including platforms newer than latest (like macos-11.0, ubuntu-20.04) as a means to get out ahead of upcoming changes?

JaapRood:
> a matrix build across platforms seems like a good starting place to me. what do you think about including platforms newer than latest (like macos-11.0, ubuntu-20.04) as a means to get out ahead of upcoming changes?

Sounds good!

.github/workflows/build.yml
devdoshi: Yeah my mistake, definitely makes sense to key by os.

JaapRood:
If anyone else has a suggestion as why the caching of Cargo dependencies doesn't work, please let me know. If not, I reckon I'll remove it, as I've spent enough time, commits and pushes not getting it to work 😂.

notnmeyer (approved):
preemptive 👍, Jaap and I just sorted out the caching issue over a call

PR #397, opened 2020-10-23
Attach release binaries to draft releases

JaapRood:
We'll need some place to host the binary part of the new Diff engine, ready to be downloaded on install. The Github release seems to be the most straightforward spot. We're already updating a draft release on each push to develop: we should include building the binaries and attaching them to the release, ready for publishing at any time.

### DIFF
--- a/.github/workflows/release-drafter.yml
+++ b/.github/workflows/release-drafter.yml
@@ -27,7 +27,35 @@ jobs:
         with:
           name: optic_diff-macos-tarball
           path: optic_diff-macos-tarball
+      - name: Download linux tarball
+        uses: actions/download-artifact@c3f5d00c8784369c43779f3d2611769594a61f7a # https://github.com/actions/download-artifact/commits/v2
+        with:
+          name: optic_diff-linux-tarball
+          path: optic_diff-linux-tarball
+      - name: Download win64 tarball
+        uses: actions/download-artifact@c3f5d00c8784369c43779f3d2611769594a61f7a # https://github.com/actions/download-artifact/commits/v2
+        with:
+          name: optic_diff-win64-tarball
+          path: optic_diff-win64-tarball
       - name: Attach macOS tarball to release
+        uses: actions/upload-release-asset@e8f9f06c4b078e705bd2ea027f0926603fc9b4d5 # https://github.com/actions/upload-release-asset/commits/v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          asset_path: optic_diff-macos-tarball/optic_diff-macos.tar.gz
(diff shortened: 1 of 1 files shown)


JaapRood:
I got to the point that we're ready to upload the prepared tarballs and attach them to the release, but `release-drafter/release-drafter` action does not seem to be exposing it's outputs, leaving us with the required `upload_url` to make the uploads work. If anyone can see what I'm missing, you'll be my hero of the week. Probably requires reading release-drafter's source to figure out what's going on.

notnmeyer:
don't know how much of this is relevant, just dumping any thoughts,

* you're doing the referencing of outputs correctly, and `-` vs. `_` and `'` vs `"` is irrelevant for the expressions to evaluate
* https://github.com/release-drafter/release-drafter/tree/v5.12.1#configuration says,
    > Once you’ve added Release Drafter to your repository, it must be enabled by adding a .github/release-drafter.yml configuration file to each repository.

    that template isn't present in the pr.
* you're pinned to `release-drafter/release-drafter@06d4616a80cd7c09ea3cf12214165ad6c1859e67` which is v5.11.0. the most recent release is v.5..12.1 `release-drafter/release-drafter@3782ccd`. i dont see a changelog in their project to scan for relevant changes though.
* lastly, im questioning release-drafter is even doing anything in your runs,
    image
    i dont think its setting the outputs because i dont think its doing anything

notnmeyer:
okay! so i got a test of my own to work: https://github.com/opticdev/monorail/runs/1299618564?check_suite_focus=true

in that run release-drafter is producing some output (as in to stdout) and setting the output values as expected. so, im still not sure whats up with this workflow but im pretty positive at this point its release-drafter not doing anything.

JaapRood:
@notnmeyer you're totally right. Diving into the issues of release-drafter, there are numerous mentions about how it only works on 'push' actions, not on any other type of event. Here [a maintainer mentions how there's multiple PRs in flight to do something about it](https://github.com/release-drafter/release-drafter/issues/671#issuecomment-707853901). I can imagine why, only working with a subset of Github Workflow triggers seems oddly limiting and a mixing of concerns.

I added the pull-request for it the be easily testable, but I guess we'll have to go a bit further and relax the trigger as you mentioned.

JaapRood:
By running `on: push`, the release drafter now runs on my fork, so see [the actions tab there](https://github.com/JaapRood/optic/actions?query=workflow%3A%22Release+Drafter%22) for the results on that.

Unfortunately, those results don't look great: once the binary was attached properly, but subsequently we're getting this super helpful error:

Screenshot 2020-10-26 at 12 26 25

If I were to wage a guess, I'd say this is because this asset already exists on the release. The GitHub API _does_ support updating assets, but I think that officially marks the end for attachments to the release being the approach we'll want to take initially.

PR #402, opened 2020-10-26
Publish Rust binaries to S3 for each development version

JaapRood:
We'll need some place to host the binary part of the new Diff engine, ready to be downloaded on install. Publishing them to an S3 bucket we control should be a fairly straightforward and totally within our control. The first step in making the entire system work end-to-end is making sure we have the workflow ready to do this for the `develop` branch, so the installer can be developed against it (behind a feature flag).

### DIFF
--- a/.github/workflows/develop-binaries.yml
+++ b/.github/workflows/develop-binaries.yml
@@ -1,0 +1,108 @@
+# Build binaries and publish to S3 for every version of the `develop` branch.
+
+name: Develop binaries
+
+on:
+  push:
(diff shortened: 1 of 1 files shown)


JaapRood:
In trying to make this work from the consuming / downloading side, I realised that while we can produce these builds fairly easily, it's hard to reference them, decide which ones to use given the developer's local state. To keep that way simpler, we decided that locally built development versions will require having the Rust toolchain installed and building the diff engine from scratch. Given that, I'll close this PR and resurface the approach for publishing binaries for published versions of Optic.

PR #404, opened 2020-10-27
Download, install and run pre-built diff-engine binaries as part of npm / yarn install

JaapRood:
For the new Rust-powered diff engine, we don't want users to have to compile the binary on their machines. Instead of requiring them to have the Rust toolchain installed (like we require Node), we want to serve them pre-built binaries for whatever platform their running on. This PR is about the "client" side of that: making sure that installing `@useoptic/diff-engine` (as part of installing `@useoptic/cli`) downloads and installs the right binary as well.

The binary we download will be relate 1:1 to the version of the `diff-engine` Node package, allowing us to keep using `package.json` for version management.

To allow for local development of the diff-engine, we're assuming that making a local build (like using the `optic_build` command from `sourceme.sh`) is the signal for wanting to run a local build of the diff engine. Through that, `optic_build` with the Rust diff engine enabled through a feature flag will require the Rust toolchain to be present. Any convenience for development where that's not the case is beyond the scope of this PR.

When a pre-built binary for the current version is not available while the Rust diff engine _has_ been enabled, installation will fail, since it's the only sane behaviour possible at that point, with little chance of recovery.


JaapRood:
So far, I've implemented the installing of the binary from a local S3 mock, as well adapting the spawning of the engine process by looking for a pre-built binary, falling back to a local build instead. If we trigger the uninstall of the pre-built binary on a build, that should give us the desired behaviour for local development.

Note: in trying to implement this, I ported the very small amount of Typescript we had for spawning the diff engine, to vanilla Javascript. While pulling in some dependencies I was once again confronted with the pains of using JS written packages in TypeScript. Given that this is all limited glue-code, very little business logic, I could not see how wrangling that all into place would worth the cost at this point in time. Rather than configuring it to be non-strict, I took the opportunity to just remove the entire layer of indirection and write it straight up.

JaapRood:
All the main bits are there now, so it's time for a review. Note that I'm leaving this in draft since there aren't actually any pre-built binaries anywhere, and the `config.js` is still pointing towards my local S3 mock. I'd like to try and get the Github workflow that publishes these up and running and merge to before: other way around would lead to failed builds while running with the Rust feature flag enabled. If push would come to shove, an additional env variable to skip installing pre-built binaries when doing a local build can be conceived of as well, but for now I'd rather just not litter our scripts with that.

notnmeyer:
definitely agree about getting the artifacts publishing first

devdoshi:
Looks good. Maybe @LouManglass and @acunniffe  can weigh in on any tracking, etc. we should also have for failures in the wild

JaapRood:
> Looks good. Maybe @LouManglass and @acunniffe can weigh in on any tracking, etc. we should also have for failures in the wild

Lets consider that separately from this PR, don't want to make this about anything but getting a first version functional.

JaapRood:
Seems to work nicely against the now public binaries for v8.5.0, at least for macOS (not sure how we'll test the rest yet). Some failing early on did make me realise two more changes were needed:

- separate building of binaries from building of Typescript: while we want to build both in local development (`optic_build`), during release we're dealing with these two concerns separately (`optic_build_for_release`). Only reason why that hasn't appeared as an issue yet is through the build only happening when the Rust diff engine feature flag was disabled in CI/CD. Will amend #405 for release workflow to use `optic_build_for_release`.
- distinguish between installing dependencies in development and when consuming Optic. When consuming Optic, we want the install to fail when a pre-built binary for a given version isn't available: there's no other options left, no other way to recover from it and produce something that runs. While developing, though, that's different: we can locally build our own version. As soon as we make a local build, we uninstall any prebuilt, to make sure we're not running against a stale build (source might have changed). Relying on the packages not having been bumped yet (which you can't do, since bumping is it's  own PR that gets merged and causes more builds to happen) is way too fragile, so instead, we want to skip that install all-together. Since we _do_ need install to happen on `yarn install` hooks (for the consuming side), the best I could come up with for now is an env variable injected from `sourceme.sh` scripts.

JaapRood:
Should be ready to go, ready for final review @notnmeyer @devdoshi.

workspaces/diff-engine/lib/index.js
```diff
@@ -0,0 +1,136 @@
...
+  if (Fs.existsSync(binaryDir)) {
+    Rimraf.sync(binaryDir);
+  }
+
+  const archiveName = Path.basename(binaryDir);
+
+  const downloadStream = await fetchBinary(
+    `/v${Config.version}/${archiveName}.tar.gz`
+  );
+  if (downloadStream.statusCode === 404) {
```
notnmeyer, line 107: what if the response isn't a 200 or a 404?

JaapRood: Throws, which kills the process and fails the install, which is the only course of action left at that point.

notnmeyer: ah ok, missed that being handled elsewhere 👍

.github/workflows/deployDemo.yml
```diff
@@ -45,7 +45,7 @@ jobs:
           curl -sL "https://keyserver.ubuntu.com/pks/lookup?op=get&search=0x2EE0EA64E40A89B84B2DF73499E82A75642AC823" | sudo apt-key add
           sudo apt-get update
           sudo apt-get install sbt -y
-          . ./sourceme.sh && optic_build
+          . ./sourceme.sh && optic_build_for_release
```
notnmeyer, line 48: kind of a tangent but `. ./sourceme.sh` is super confusing syntax imo. i know this is correct here, but its somewhat ambiguous because you arent sure if sourceme.sh is executable and maybe its a typo and should be `../sourceme.sh `, etc. `. sourceme.sh` or `source ./sourceme.sh` are significantly clearer imo

notnmeyer (approved):
looks good to me

.github/workflows/deployDemo.yml
JaapRood: You're totally right. I'll be a good scout and clean that up.

PR #405, opened 2020-10-28
Upload prebuilt binaries as part of every Optic release

JaapRood:
We'll need some place to host the binary part of the new Diff engine, ready to be downloaded on install. To make sure those are always in place, we'll build and publish these binaries to S3 _before_ we publish the new Optic version to package registries.

We'll publish the binaries under Optic's main version, making sure there is always a 1:1 mapping between pre-built binaries and the Node.js code. 

Since Optic can be used from across platforms, we supply builds of different kinds. For now, rather than cross-compiling from a single host, we're taking the approach of building on the hosts native to the build target. The biggest reason for this is so we can keep using `cargo build` for now and avoid additional complexities. Additionally, there is also the compilation for macOS, which is not supported by the official [cross](https://github.com/rust-embedded/cross). While an open source alternative for that exists, its complexity we'd rather not deal with right now. Finally, staying to compilation native to the host, we're also keeping the doors open for any platform dependent signing that might be required.

To start of with, and without any additional info, we're electing to support the following subset of [Rust targets](https://doc.rust-lang.org/nightly/rustc/platform-support.html):

- 64-bit Windows 7+ (`x86_64-pc-windows-msvc`)
- 64-bit Linux (kernel 2.6.32+, glibc 2.11+) (`x86_64-unknown-linux-gnu`)
- 64-bit macOS (10.7+, Lion+) (`x86_64-apple-darwin`)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -13,12 +13,17 @@ on:
   push:
     branches:
       - release
+  # TODO: remove pull-request trigger, only for development of prebuilt binaries publishing
+  pull_request:
+    branches:
+      - develop
 
 jobs:
   # Deploys the current version to NPM, and also verifies that the version is correct in the process
   # All Jobs rely on this job because it verifies the version
   release-npm:
     runs-on: ubuntu-latest
+    if: false # TODO: re-enable actual release after developing publishing of pre-builts
     steps:
     - uses: actions/checkout@28c7f3d2b5162b5ddd3dfd9a45aa55eaf396478b # https://github.com/actions/checkout/commits/v2
       with:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,10 +19,124 @@ on:
       - develop
 
 jobs:
+  build-rust-binaries:
+    strategy:
+      matrix:
+        include:
+          - os: macos-latest
+            target: x86_64-apple-darwin
+            platform_name: macos
+            suffix: ''
+          - os: windows-latest
+            target: x86_64-pc-windows-msvc
+            platform_name: win64
+            suffix: .exe
+          - os: ubuntu-latest
+            target: x86_64-unknown-linux-gnu
+            platform_name: linux
+            suffix: ''
+    runs-on: ${{ matrix.os }}
+    env:
+      ARCHIVE_NAME: optic_diff-${{ matrix.platform_name }} # TODO: include version number?
+      FILE_NAME: optic_diff${{ matrix.suffix }}
+    steps:
+      - name: 'Set CARGO_HOME and RUSTUP_HOME'
+        run: |
+          echo "RUSTUP_HOME=$HOME/.rustup" >> $GITHUB_ENV
+          echo "CARGO_HOME=$HOME/.cargo" >> $GITHUB_ENV
+      - name: 'Checkout source'
+        uses: actions/checkout@28c7f3d2b5162b5ddd3dfd9a45aa55eaf396478b # https://github.com/actions/checkout/commits/v2
+      - name: 'Cache cargo registry'
+        uses: actions/cache@d1255ad9362389eac595a9ae406b8e8cb3331f16 # https://github.com/actions/cache/commits/v2
+        with:
+          path: |
+            ${{ env.CARGO_HOME }}/registry
+            ${{ env.CARGO_HOME }}/git
+          key: "${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}-vendor-v1"
+      - name: 'Cache build target'
+        uses: actions/cache@d1255ad9362389eac595a9ae406b8e8cb3331f16 # https://github.com/actions/cache/commits/v2
+        with:
+          path: target
+          key: "${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}-release-target-v1"
+      - name: 'Rust toolchain'
+        uses: actions-rs/toolchain@b2417cde72dcf67f306c0ae8e0828a81bf0b189f # https://github.com/actions-rs/toolchain/commits/v1
+        with:
+          toolchain: stable
+          profile: minimal
+          override: true
+      - name: 'Build'
+        uses: actions-rs/cargo@844f36862e911db73fe0815f00a4a2602c279505 # https://github.com/actions-rs/cargo/commits/v1
+        with:
+          command: build
+          args: --workspace --release --target=${{ matrix.target }}
+      - name: 'Flush Cargo cache to disk on macOS'
+        if: runner.os == 'macOS'
+        run: sudo /usr/sbin/purge
+      - name: 'Upload artifact'
+        uses: actions/upload-artifact@27bce4eee761b5bc643f46a8dfb41b430c8d05f6 # https://github.com/actions/upload-artifact/commits/v2
+        with:
+          name: ${{ env.ARCHIVE_NAME }}-build
+          path: target/${{ matrix.target }}/release/${{ env.FILE_NAME }}
+
+  publish-binaries:
+    strategy:
+      matrix:
+        include:
+          - platform_name: macos
+            suffix: ''
+          - platform_name: win64
+            suffix: .exe
+          - platform_name: linux
+            suffix: ''
+    runs-on: ubuntu-latest
+    needs: build-rust-binaries
+    env:
+      FILE_NAME: optic_diff${{ matrix.suffix }}
+      BUCKET: optic-packages
+    steps:
+      - uses: actions/checkout@28c7f3d2b5162b5ddd3dfd9a45aa55eaf396478b # https://github.com/actions/checkout/commits/v2
+      - uses: actions/setup-node@44c9c187283081e4e88b54b0efad9e9d468165a4 # https://github.com/actions/setup-node/commits/v1
+        with:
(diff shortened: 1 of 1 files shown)


JaapRood:
@notnmeyer it's the same approach we've talked about before, and everything seems to be running smoothly, but would really appreciate you checking that once over ☺️. If we're all happy I'll get rid of the develop triggers + disabling of release steps.

.github/workflows/release.yml
```diff
@@ -13,12 +13,140 @@ on:
...
+            suffix: .exe
+          - os: ubuntu-latest
+            target: x86_64-unknown-linux-gnu
+            platform_name: linux
+            suffix: ''
+    runs-on: ${{ matrix.os }}
+    env:
+      FILE_NAME: optic_diff${{ matrix.suffix }}
+    steps:
+      - uses: actions/checkout@28c7f3d2b5162b5ddd3dfd9a45aa55eaf396478b # https://github.com/actions/checkout/commits/v2
```
notnmeyer, line 42: this isn't current. latest is `a81bbbf8298c0fa03ea29cdc473d45769f953675 # v2.3.3`

https://github.com/actions/checkout/tree/v2.3.3

.github/workflows/release.yml
```diff
@@ -13,12 +13,140 @@ on:
...
+      - uses: actions/setup-node@44c9c187283081e4e88b54b0efad9e9d468165a4 # https://github.com/actions/setup-node/commits/v1
```
notnmeyer, line 43: same, most recent v1 sha is "56899e050abffc08c2b3b61f3ec6a79a9dc3223d # v1.4.4`

.github/workflows/release.yml
```diff
@@ -13,12 +13,140 @@ on:
...
+        with:
+          node-version: 12
+      - name: 'Determine version and archive name'
+        shell: bash
```
notnmeyer, line 47: does the shell need to be specified?

.github/workflows/release.yml
```diff
@@ -13,12 +13,140 @@ on:
...
+        run: |
+          OPTIC_VERSION=$(npm view @useoptic/cli version)
+          echo "OPTIC_VERSION=$OPTIC_VERSION" >> $GITHUB_ENV
+          echo "ARCHIVE_NAME=optic_diff-v$OPTIC_VERSION-${{ matrix.platform_name }}" >> $GITHUB_ENV
+      - name: 'Set CARGO_HOME and RUSTUP_HOME'
+        run: |
+          echo "RUSTUP_HOME=$HOME/.rustup" >> $GITHUB_ENV
+          echo "CARGO_HOME=$HOME/.cargo" >> $GITHUB_ENV
+      - name: 'Checkout source'
+        uses: actions/checkout@28c7f3d2b5162b5ddd3dfd9a45aa55eaf396478b # https://github.com/actions/checkout/commits/v2
```
notnmeyer, line 57: checking the code out again?

.github/workflows/release.yml
```diff
@@ -13,12 +13,140 @@ on:
...
+      - name: 'Cache cargo registry'
+        uses: actions/cache@d1255ad9362389eac595a9ae406b8e8cb3331f16 # https://github.com/actions/cache/commits/v2
```
notnmeyer, line 59: latest is `d1255ad9362389eac595a9ae406b8e8cb3331f16 # v2.1.2`

.github/workflows/release.yml
```diff
@@ -13,12 +13,140 @@ on:
...
+            ${{ env.CARGO_HOME }}/registry
+            ${{ env.CARGO_HOME }}/git
+          key: "${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}-vendor-v1"
+      - name: 'Cache build target'
+        uses: actions/cache@d1255ad9362389eac595a9ae406b8e8cb3331f16 # https://github.com/actions/cache/commits/v2
+        with:
+          path: target
+          key: "${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}-release-target-v1"
+      - name: 'Rust toolchain'
+        uses: actions-rs/toolchain@b2417cde72dcf67f306c0ae8e0828a81bf0b189f # https://github.com/actions-rs/toolchain/commits/v1
```
notnmeyer, line 71: v1.0.6

.github/workflows/release.yml
```diff
@@ -13,12 +13,140 @@ on:
...
+        with:
+          toolchain: stable
+          profile: minimal
+          override: true
+      - name: 'Build'
+        uses: actions-rs/cargo@844f36862e911db73fe0815f00a4a2602c279505 # https://github.com/actions-rs/cargo/commits/v1
```
notnmeyer, line 77: v1.0.3

.github/workflows/release.yml
JaapRood: Hmm, if only the PR showed the blame and commit message :) It does, as Powershell is default for Windows and does not like the exporting of the variable.

notnmeyer (approved):
looks mostly fine, just a couple nits overall. i stopped commenting each one, but i think itd be better if the github action sha pins were commented with the specific release tag rather than just 'v1'.

also the build has numerous deprecation warnings for set-path and set-env stuff, which is related to CVE-2020-15228. when i made a pass on everything a couple weeks back, i updated _our_ usage but there werent updated builds for all actions we used.

so, at least for actions you've added can you ensure that everything is pinned to the latest tag? bonus points if you want to update the the others in this workflow while youre at it, but in any case i filed https://github.com/opticdev/issues/issues/29 to clean up the rest of them

.github/workflows/release.yml
JaapRood: Addressed with https://github.com/opticdev/optic/pull/405/commits/7909b8a7582395c45ecc5ec348bc50789d35e428

.github/workflows/release.yml
JaapRood: Addressed with https://github.com/opticdev/optic/pull/405/commits/7909b8a7582395c45ecc5ec348bc50789d35e428

.github/workflows/release.yml
JaapRood: Addressed in https://github.com/opticdev/optic/pull/405/commits/7909b8a7582395c45ecc5ec348bc50789d35e428

.github/workflows/release.yml
JaapRood: Addressed in https://github.com/opticdev/optic/pull/405/commits/7909b8a7582395c45ecc5ec348bc50789d35e428

.github/workflows/release.yml
JaapRood: Addressed in https://github.com/opticdev/optic/pull/405/commits/7909b8a7582395c45ecc5ec348bc50789d35e428

.github/workflows/release.yml
JaapRood: Addressed in https://github.com/opticdev/optic/pull/405/commits/7909b8a7582395c45ecc5ec348bc50789d35e428

.github/workflows/release.yml
```diff
@@ -13,12 +13,140 @@ on:
...
+        with:
+          command: build
+          args: --workspace --release --target=${{ matrix.target }}
+      - name: 'Flush Cargo cache to disk on macOS'
+        if: runner.os == 'macOS'
+        run: sudo /usr/sbin/purge
+      - name: 'Upload artifact'
+        uses: actions/upload-artifact@27bce4eee761b5bc643f46a8dfb41b430c8d05f6 # https://github.com/actions/upload-artifact/commits/v2
```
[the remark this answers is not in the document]
JaapRood: Addressed in https://github.com/opticdev/optic/pull/405/commits/7909b8a7582395c45ecc5ec348bc50789d35e428

JaapRood:
@notnmeyer I ended up tagging them all to the specific release. Not only does it point to the specific tag, the Github UI shows a nice "latest release" badge, which should help verify we're on current.
Screenshot 2020-10-28 at 16 49 28

JaapRood:
@notnmeyer @devdoshi one last sanity check before we officially make this the new release workflow would be greatly appreciated ☺️

notnmeyer (approved):
👍

.github/workflows/release.yml
```diff
@@ -15,15 +15,137 @@ on:
...
+      FILE_NAME: optic_diff${{ matrix.suffix }}
+      BUCKET: optic-packages
+    steps:
+      - uses: actions/checkout@a81bbbf8298c0fa03ea29cdc473d45769f953675 # https://github.com/actions/checkout/releases/tag/v2.3.3
+      - uses: actions/setup-node@56899e050abffc08c2b3b61f3ec6a79a9dc3223d # https://github.com/actions/setup-node/releases/tag/v1.4.4
+        with:
+          node-version: 12
+      - name: 'Determine version and archive name'
+        run: |
+          OPTIC_VERSION=$(npm view @useoptic/cli version)
```
devdoshi, line 107: Not sure but at first glance is there a chicken and egg situation if the npm publish wants the binary to be published under the same version first? At least if we're using npm view as opposed to the version from the package.json?

JaapRood: The bump happens separately from this workflow. By the time we're hitting the release workflow, the version has already been bumped. This is only to resolve the version that's being released.

PR #406, opened 2020-10-30
Fix bug where prebuilt binaries would be published against currently published version

JaapRood:
@devdoshi was right in his review: the way pre-built binaries were versioned used the currently published version, rather than the version of the local package. As I was testing against an already published npm package, it still worked, but for new releases it would fail.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -43,7 +43,7 @@ jobs:
       - name: 'Determine version and archive name'
         shell: bash
         run: |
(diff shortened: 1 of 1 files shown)


JaapRood:
Don't really want to go through the hassle of attempting to run this workflow again. Instead, I tested it by copy-pasting the line of code and running it locally. Given it's simplicity, I reckon that's enough.

notnmeyer (approved):
makes sense and looks straighforword 👍

PR #408, opened 2020-10-30
Add a pathway for prereleases to be created and published

JaapRood:
We've built a release pipeline where part the building of binaries is part of the release + publishing to npm. During a new npm install, the pre-built binary must be able to be downloaded, or else it's install fails. Publishing a pre-release for a non-`latest` tag through this same pipeline is our best way of verifying this all works, while making sure nothing breaks for customers.

Due to [how semver treats prereleases](https://www.npmjs.com/package/semver#prerelease-tags) only users who explicitly opt-in to the pre-release will get it, and anyone who has will automatically get the full released version, eliminating the risk of users getting stuck on stale betas:

> If a version has a prerelease tag (for example, 1.2.3-alpha.3) then it will only be allowed to satisfy comparator sets if at least one comparator with the same [major, minor, patch]tuple also has a prerelease tag.
> 
> For example, the range >1.2.3-alpha.3 would be allowed to match the version 1.2.3-alpha.7, but it would not be satisfied by 3.4.5-alpha.9, even though 3.4.5-alpha.9 is technically "greater than" 1.2.3-alpha.3 according to the SemVer sort rules. The version range only accepts prerelease tags on the 1.2.3 version. The version 3.4.5 would satisfy the range, because it does not have a prerelease flag, and 3.4.5 is greater than 1.2.3-alpha.7.
> 
> The purpose for this behavior is twofold. First, prerelease versions frequently are updated very quickly, and contain many breaking changes that are (by the author's design) not yet fit for public consumption. Therefore, by default, they are excluded from range matching semantics.
> 
> Second, a user who has opted into using a prerelease version has clearly indicated the intent to use that specific set of alpha/beta/rc versions. By including a prerelease tag in the range, the user is indicating that they are aware of the risk. However, it is still not appropriate to assume that they have opted into taking a similar risk on the next set of prerelease versions.

Prereleases are only published to npm, since Homebrew and Debian registries don't work with semver in the same way. That's fine, since prereleases aren't for wide consumption in the first place. Requiring npm to use them should be fine.

This PR doesn't deal with trying to enable the feature flag: that's something we'll want to cover separately as we actually prepare the shipping of the new diff engine.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -159,12 +159,6 @@ jobs:
         node-version: 12
     - name: Install Dependencies and Build Optic
       run:  |
-        # sudo apt-get update
-        # sudo apt-get install apt-transport-https -y
-        # echo "deb https://dl.bintray.com/sbt/debian /" | sudo tee -a /etc/apt/sources.list.d/sbt.list
-        # curl -sL "https://keyserver.ubuntu.com/pks/lookup?op=get&search=0x2EE0EA64E40A89B84B2DF73499E82A75642AC823" | sudo apt-key add
-        # sudo apt-get update
-        # sudo apt-get install sbt -y
         source ./sourceme.sh && optic_build_for_release
     - run: echo "//registry.npmjs.org/:_authToken=${{ secrets.PUBLISH_NPM_OPTIC_BOT }}" > ~/.npmrc
     - run: node ./workspaces/scripts/publish.js
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -13,10 +13,6 @@ on:
   push:
     branches:
       - release
-  # TODO: remove this trigger after we're done developing the prerelease specifics
-  pull_request:
-    branches:
-      - develop
 
 jobs:
   build-rust-binaries:
@@ -149,9 +145,8 @@ jobs:
       is_prerelease: ${{ steps.release_type.outputs.is_prerelease }}
     steps:
     - uses: actions/checkout@a81bbbf8298c0fa03ea29cdc473d45769f953675 # https://github.com/actions/checkout/releases/tag/v2.3.3
-      # TODO: re-enable publishing the release branch before merging
-      # with:
-      #     ref: release
+      with:
+        ref: release
     - uses: actions/setup-node@56899e050abffc08c2b3b61f3ec6a79a9dc3223d # https://github.com/actions/setup-node/releases/tag/v1.4.4
       with:
         node-version: 12
@@ -183,7 +178,6 @@ jobs:
       - run: |
           echo "determined as not a prerelease" 
           echo "is_prelease=${{ needs.release-npm.outputs.is_prerelease }}"
-          exit 1 # TODO: remove this safety that's preventing actual releases from happening
       - uses: actions/checkout@a81bbbf8298c0fa03ea29cdc473d45769f953675 # https://github.com/actions/checkout/releases/tag/v2.3.3
         with:
           ref: release
@@ -201,9 +195,8 @@ jobs:
   trigger-brew-release:
     runs-on: ubuntu-latest
     needs: release-npm
-    if: needs.release-npm.outputs.is_prerelease != 'true'
+    if: needs.release-npm.outputs.is_prerelease == 'false'
     steps:
-      - run: exit 1 # TODO: remove this safety that's preventing actual releases from happening
       - name: Repository Dispatch
         uses: peter-evans/repository-dispatch@95531d6358eb144ac65db1d7f39af794979ea97e # https://github.com/peter-evans/repository-dispatch/releases/tag/v1.1.2
         with:
@@ -214,7 +207,6 @@ jobs:
 
   # Tells slack a release has happened
   notify-slack:
-    if: false # TODO: remove this safety that's preventing actual releases from happening
     runs-on: ubuntu-latest
     needs: release-npm
     name: Message Slack that Release has happened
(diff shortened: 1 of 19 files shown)


JaapRood:
@notnmeyer I'm pretty happy I've demonstrated with the last action run that if a prerelease version is published, the Debian and Homebrew versions are skipped. Would be great if you can sanity check that and verify the approach. If that's all good, I'd say we take it for a spin and actually let it publish a pre-release version to npm by re-enabling the publishing step. As is the point with prereleases, that shouldn't have any affect to customers unless they opt-in to it.

.github/workflows/release.yml
```diff
@@ -159,14 +169,31 @@ jobs:
...
     - run: echo "//registry.npmjs.org/:_authToken=${{ secrets.PUBLISH_NPM_OPTIC_BOT }}" > ~/.npmrc
     - run: node ./workspaces/scripts/publish.js
+      if: false
       env:
         OPTIC_PUBLISH_SCOPE: public
+    - name: Set is_prerelease output
+      id: release_type
+      run: |
+        OPTIC_VERSION=$(node -e "console.log(require('./workspaces/local-cli/package.json').version)")
+        if [[ $OPTIC_VERSION == *'-'* ]]; then
```
notnmeyer, line 179: so this is determining if this is a prerelease based on whether or not there's a hyphen in the version? i feel like that could be written to be more clear with a regex instead of globbing... maybe,

```
bash-3.2$ [[ "8.5.0" =~ (.+-.+) ]]; echo $?
[[ "8.5.0" =~ (.+-.+) ]]; echo $?
1

bash-3.2$ [[ "8.5.0-blah" =~ (.+-.+) ]]; echo $?
[[ "8.5.0-blah" =~ (.+-.+) ]]; echo $?
0
```

having written that and looking at it—meh, im cool with whatever. although i do wish there was more explicit check to make, like all prerelease versions contained "beta" or "prerelease" or something

.github/workflows/release.yml
```diff
@@ -159,14 +169,31 @@ jobs:
...
+        else
+          echo "v$OPTIC_VERSION is not a prerelease"
+          echo "::set-output name=is_prerelease::false"
+        fi
 
   # Creates debian package for users to install, hosted on s3
   release-deb:
     runs-on: ubuntu-latest
     needs: release-npm
+    if: needs.release-npm.outputs.is_prerelease != 'true'
```
notnmeyer, line 191: i cant help but feel like this is backwards. prereleases should be the default, and a public release is the less frequent case. this expressions reads better without negation, `if: needs.release-npm.outputs.is_release == 'true'`

.github/workflows/release.yml
```diff
@@ -159,14 +169,31 @@ jobs:
...
     steps:
+      - run: |
+          echo "determined as not a prerelease"
```
notnmeyer, line 194: similar to the above this reads to me as something that doesn't need negation. "determined to be a release" vs. "determined as not a prerelease"

notnmeyer (commented):
left a couple comments. conceptually this looks good though.

.github/workflows/release.yml
JaapRood: > having written that and looking at it—meh, im cool with whatever. although i do wish there was more explicit check to make, like all prerelease versions contained "beta" or "prerelease" or something

I went through the exact same thing, thought I'd just keep it simple. We're writing versions through a library that only returns valid semver. Anything that isn't semver would have failed the publish commands outright. The prerelease id (alpha, beta, rc, etc) is supposed to be a free string in semver, but is also allowed to be omitted, so adding an optional match for that in a regex doesn't actually yield any additional safety.

.github/workflows/release.yml
JaapRood: > prereleases should be the default, 

Yeah I reckon that makes sense as being the safer option.

.github/workflows/release.yml
JaapRood: Added in a 4 or 5 push debug cycle, will clean up!

JaapRood: Actually, not changing, as this will just be deleted as part of the debug code when we get rid of the `exit 1` safety.

.github/workflows/release.yml
JaapRood: Had a look at this, but it reading `is_release` is more confusing to me: a prerelease is a kind of release, and "public" release doesn't cover it either, as a prerelease is also done publicly. I've changed the logic ever so slightly to get rid of the negation (in https://github.com/opticdev/optic/pull/408/commits/6a3f17eb265e32b14681c0b9455b502d4661f436), but that's where I'm going to leave it.

.github/workflows/release.yml
notnmeyer: yeah, im cool with that 👍

JaapRood:
Alright, that lands us at the point where I'm going to attempt a publishing of `8.6.0-beta.0` through this development trigger. However, not on a Friday afternoon at 5:30pm 😅. First thing Monday!

JaapRood:
A couple of published prereleases on, `8.6.0-beta-3` got successfully published under the `beta` tag. As it turned out, npm install the `latest` blindly for new installs (running `npm install -g @useoptic/cli`), not caring about prerelease as when it does during upgrades / versioned installs. As a response, the `publish.js` script now extract the prelease id from a version (`alpha`, `beta`, `rc``, etc), falling back to `unstable` for untagged preleases (e.g. `8.7.0-1`)  and uses that to tag the publish, keeping `latest` in-tact to indicate the current stable version of Optic.

In the first releases, we were also confronted with the install script for `diff-engine` not working properly, missing a dependency it should have just installed prior to it (but apparently wasn't). For the purposes of this PR being about the release workflow, we'll just leave that disabled for now and deal with that separately.

notnmeyer (approved):

PR #414, opened 2020-11-03
Fix publishing of diff-engine

JaapRood:
As this is the second time this donkey hurts himself on this stone, I've opted to choose which files we choose to _ignore_ when publishing to NPM, rather than explicitly stating which make it through.


JaapRood:
Given the scope of the changes, I'm gonna go ahead an merge, relying on the CI builds as sanity checks there's no syntax errors.

PR #416, opened 2020-11-03
Allow feature flags to be enabled for prereleases and enable Rust diff engine for beta

JaapRood:
Prereleases are only of limited use unless we can enable features for them that have yet to land in the stable version. Feature flags are processed at build-time, so it must be enabled before we publish. At all times `develop` should remain releasable, with features not enabled in production releases until we explicitly opt-in to it.

This approach uses the tag in the prelease version (like `beta` in `8.6.0-beta.3`) to read an `.env.<tag>` file, and "stages" that right before making the release build that's published to npm. It should fail gracefully if the release is not a prerelease or wasn't tagged (`8.6.0-0` would also be a valid prerelease). The `.env.<tag>` file is then also included and published as part of the cli packages (`local-cli`, `agent-cli` and `ci-cli` to make sure that also at runtime, the feature flags are enabled.

Finally, to verify it all works, we enabled the Rust diff engine for prereleases with the `beta` tag.


JaapRood:
@notnmeyer @devdoshi 3 successful dry runs and local testing of the run-time feature flags make me pretty happy this is working as designed.

notnmeyer (approved):

PR #417, opened 2020-11-04
Stop publishing ARM build to Debian

JaapRood:
As we work with pre-built binaries but don't include ARM compilation yet, the ARM build for Debian would fail installation. Rather than going through the trouble of building the cross compilation workflows required for this (the current setup requires host-native builds), we will stop publishing the ARM build to Debian for now.

We can make this call as we have little proof of the ARM package being used, nor sufficient information on why it was included in the first place. We'll rely on customers or other indicators telling us that ARM deployments _are_ worth investing the development time for.

### DIFF
--- a/.github/deployDebian/entrypoint.sh
+++ b/.github/deployDebian/entrypoint.sh
@@ -26,6 +26,4 @@ echo "Installing Ruby"
 apt install ruby-full -y
 gem install deb-s3
 export PATH_TO_DEB_AMD="/tmp/optic/workspaces/local-cli/dist/deb/api_$(npm view $INPUT_NPM_PACKAGE_NAME version)-1_amd64.deb"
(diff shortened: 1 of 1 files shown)


LouManglass (approved):
This looks like it would stop the publish, I assume the build for Debian ARM is stopped elsewhere. I approve of the idea as well, as my understanding from skimming deb-s3 is that it won't remove existing packages, only update them as necessary. That means if anyone is using ARM builds, they can still pull what's already available and we can see if it's in use/how much.

JaapRood:
@LouManglass Good point! Fortunately, we never really made a special build for ARM: by virtue of running Node.js, which can run on ARM, it was compatible.

devdoshi:
@LouManglass per line 24 (oclif-dev pack:deb) -> https://github.com/oclif/dev-cli/blob/v1.19.4/src/commands/pack/deb.ts#L29 looks like if we wanted to stop it from producing anything but amd64 we'd have to make some config to tell it to only do amd64

PR #418, opened 2020-11-04
Fix bug where feature flags not present during install script of diff-engine

JaapRood:
Installing `8.6.0-beta.7` revealed that, while feature flags are shipped for use in the cli's, it was not present during the install. Since we're doing conditional installs (prevent diff engine from installing / failing installs while feature flag is disabled), we need that in place, too.


JaapRood:
Having to add another copying of the env file to a specific package in the workspace is probably reaching it's limit, where for the next one we want to automate doing it to all packages in a script like `publish.js` and `bump.js`. For our purposes now (getting this thing shipped), I propose we leave that till later.

devdoshi (approved):

notnmeyer (approved):

PR #420, opened 2020-11-05
Fixed 3 bugs from QA

acunniffe:
1. api spec no longer opened the UI
2. api start (on the first run w/ fresh daemon) wouldn't exit 
3. first runs of daemon no longer hang CLI


workspaces/cli-shared/src/command-and-proxy-session-manager.ts
```diff
@@ -123,6 +123,7 @@ class CommandAndProxySessionManager {
     await inboundProxy.stop();
     developerDebugLogger(`waiting for persistence manager to stop`);
     await persistenceManager.cleanup();
+    developerDebugLogger('waiting for cleanup');
```
devdoshi, line 126: this should go before line 125 right?

workspaces/local-cli/src/shared/local-cli-task-runner.ts
```diff
@@ -110,7 +110,7 @@ export async function LocalTaskSessionWrapper(
     await printCoverage(paths, taskName, captureId);
   }
 
-  return await cleanupAndExit();
+  return cleanupAndExit();
```
devdoshi, line 113: out of curiosity, does this make a difference?

devdoshi:
closed in favor of #433 and #444

PR #421, opened 2020-11-05
paths should be normalized so they do not include trailing slashes

devdoshi:


workspaces/diff-engine/src/queries/endpoint.rs
```diff
@@ -1,3 +1,4 @@
+use crate::events::SpecEvent::EndpointEvent;
```
JaapRood, line 1: This doesn't seem to be used.

workspaces/diff-engine/src/queries/endpoint.rs
```diff
@@ -6,6 +7,8 @@ use crate::state::endpoint::{
 use crate::HttpInteraction;
 use petgraph::graph::Graph;
 use petgraph::visit::EdgeFilteredNeighborsDirected;
+use serde_json::json;
```
JaapRood, line 10: If this is only used for tests, it should be imported inside the `test` module. Wouldn't want anyone to think json as any other place in the actual implementation of `EndpointQueries` 😅

workspaces/diff-engine/src/queries/endpoint.rs
```diff
@@ -16,9 +19,20 @@ impl<'a> EndpointQueries<'a> {
       endpoint_projection,
     }
   }
+
+  pub fn extract_normalized_path(interaction: &HttpInteraction) -> &str {
```
JaapRood, line 23: If we end up going with this, lets not make this a public function until something outside `EndpointQueries` has a good reason to call this from outside.

workspaces/diff-engine/src/queries/endpoint.rs
```diff
@@ -16,9 +19,20 @@ impl<'a> EndpointQueries<'a> {
...
+      &interaction.request.path[..interaction.request.path.len() - 1]
+    } else {
+      &interaction.request.path
+    }
+  }
   pub fn resolve_path(&self, interaction: &HttpInteraction) -> Option<PathComponentIdRef> {
+    let path = Self::extract_normalized_path(interaction);
     // eprintln!("{}", path);
-    let mut path_components = interaction.request.path.split('/');
+    let mut path_components = path.split('/');
```
JaapRood, line 35: Rather than solving this by string matching and manipulation, wouldn't we be better off skipping any last components that are empty, post-split? We do a skip for the first empty one, so it seems within expectation we might do the same for the last?

JaapRood (changes requested):

workspaces/diff-engine/src/queries/endpoint.rs
devdoshi: [] helper should be private

devdoshi: - [ ] helper should be private

workspaces/diff-engine/src/queries/endpoint.rs
devdoshi: - [ ] remove unused import that webstorm added

workspaces/diff-engine/src/queries/endpoint.rs
devdoshi: - [ ] move test dependency into test module

workspaces/diff-engine/src/queries/endpoint.rs
devdoshi: unless I'm missing something with Rust, it's more complex to track whether you are on the last item in the split iterator than to normalize it beforehand. Skipping the first item in an iterator is simpler afaik. In addition, the normalization has meaning as an extracted function. It is used in a few places in the scala and typescript code IIRC.

JaapRood: Fair enough, was probably thinking about that with JS brain where it would have been very straightforward. Not to say it would be very difficult in Rust, but must admit I can't instantly produce how to do that either!

JaapRood (approved):

PR #424, opened 2020-11-07
Learn API Mode was not cleaning up its child threads

acunniffe:
Child processes weren't exiting after documenting initial bodies. If you documented 50 endpoints, you'd have 50 new node processes that remained idle

Adding an explicit child.exit on cleanup works, but I don't know if it's the best approach. I know there are learnings from the other IPC work I may not be privy to

### DIFF
--- a/workspaces/cli-server/src/tasks/on-demand-initial-body.ts
+++ b/workspaces/cli-server/src/tasks/on-demand-initial-body.ts
@@ -97,6 +97,7 @@ export class OnDemandInitialBody {
       this.events.once('completed', onCompleted);
       function onCompleted(data: any) {
         cleanup();
+        child.kill(0);
         resolve(data.results);
       }
       const cleanup = () => {


workspaces/cli-server/src/tasks/on-demand-initial-body.ts
```diff
@@ -97,6 +97,7 @@ export class OnDemandInitialBody {
       this.events.once('completed', onCompleted);
       function onCompleted(data: any) {
         cleanup();
+        child.kill(0);
```
devdoshi, line 100: should it be kill()'d or exit()'d?

notnmeyer: i dont think there's an exit()? its this right, https://nodejs.org/api/child_process.html#child_process_subprocess_kill_signal

i think just `child.kill()` is correct, which sends a SIGTERM by default. i'm a little confused by sending it 0, because that doesn't map to a signal.

devdoshi: Oh you're right, I was looking at process, not child_process

devdoshi: @acunniffe if you remove the 0 it'll send the default SIGTERM

devdoshi:
closed in favor of #433 and #444

PR #429, opened 2020-11-10
rust diff engine should support the BaseShapeSet event

devdoshi:


JaapRood (commented):
Seems like a straightforward implementation to me! The one thing I wonder is we have to include any garbage collection of possibly orphaned nodes in the graph, like when changing an `$object` into some other base shape (orphaning the field nodes). If it's more a concern of being exhaustive rather than what's practically being used, perhaps we'll want to document it somewhere, like a comment explaining the made assumption.

PR #430, opened 2020-11-10
interaction iterator should stop when there are no more batches

devdoshi:
This ensures the frontend will receive a notification with hasMoreInteractions: false when there are no more interactions. If we add support for tailing a live capture's interaction stream, we will have to change this or make another interaction-iterator.


JaapRood (approved):
Seems like a bug well found! So if I'm getting this right, that last message with `hasInteractions: false` was never yielded in case the next file in the sequence couldn't be found?

workspaces/cli-shared/src/captures/avro/file-system/interaction-iterator.ts
```diff
@@ -90,7 +90,6 @@ export async function* CaptureInteractionIterator(
     }
     currentBatchId = currentBatchId + BigInt(1);
   }
-
```
JaapRood, line 93: lol

PR #435, opened 2020-11-12
use dbg instead of stderr when possible

devdoshi:
all output that should not be displayed to end users should use dbg! instead of eprintln! (at least as a short term measure). The release build of the cli will then spend less time flushing stderr


JaapRood (approved):
Great one! Let our development be slow, that's fine, but no need to push anything that to production. Eventually a proper logging can help us with getting user-end logs back to us, `eprintln!` is definitely not the way for that.

PR #437, opened 2020-11-13
Allow feature flags to be shipped for stable versions

JaapRood:
We want to be able to enable feature flags for stable releases as well. Mirroring the npm tag for stable releases, this PR amends the release flow to stage feature flags from `.env.latest`.

In addition, it puts a permanent mechanism for performing dry-runs in place, powered by the workflow wide `RELEASE_WORKFLOW_DRY_RUN` env variable. Having manually having to add / remove dry run conditionals constantly combined with the risk of getting it wrong, that seemed like a change well worth making.


JaapRood:
After multiple successful dry-runs for both preleases and stable releases (already cleaned up), I'm confident this should work as designed.

Screenshot 2020-11-13 at 13 35 15

.github/workflows/release.yml
```diff
@@ -8,8 +8,7 @@ env:
   BUCKET_NAME: optic-packages
   PACKAGE_NAME: api
   NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
-
-on:
```
notnmeyer, line 12: i think `on:` was accidentally removed

JaapRood: it was!

.github/workflows/release.yml
```diff
@@ -8,8 +8,7 @@ env:
   BUCKET_NAME: optic-packages
   PACKAGE_NAME: api
   NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
```
notnmeyer, line 10: what do you think about setting this on the steps that need it vs. setting global env?

JaapRood: 🤷‍♂️ indifferent!

notnmeyer (approved):

PR #440, opened 2020-11-23
New Interpreters + Diff UX

acunniffe:
This is a draft PR for a major re-work of the diff UX, optimized for speed, helping users better understand the context of each diff, and voice their intent to Optic. If Learn API mode is like 'git add -A' then the new diff UX is most akin to a nice git client that makes it easy to understand, stage, and discard changes. 

Also, as part of this PR, I've moved all the interpreters to Typescript backed up by Snapshot tests. 

Screen Shot 2020-11-23 at 8 49 22 AM

Limitation: The capability to run this against real captures will come online this week. Right now only debug sessions can use this UI

I'm looking for feedback on the UI, and a review of all domain logic that's been rewritten in typescript.


acunniffe:
Support for real captures is now working -- it follows the same async-task pattern we invoke in learn api mode

PR #441, opened 2020-11-24
Fix handling of stderr and stdout preventing diffing to hang on Windows

JaapRood:
Captures on Windows that were larger than about 200 interactions would hang and never finish.

The problem lied in that `stdout` and `stderr` streams not being consumed when the cli-server spawned them as children processes. On POSIX this was fine, as the I/O is asynchronous and the writes to the streams (not consumed, with content being buffered) didn't block. On Windows, however, where Node's pipes are blocking, once the buffers of the unconsumed streams filled up, blocked further execution.

We concluded that a reasonable fix was to perform the `ipc` over a separate file descriptor instead of `stdin` and explicitly ignoring or cosuming the rest for running managed scripts. That allows debug output to make it all the way down to the cli-server's output (and eventually do aggregated logging from a single point), but prevents tasks from getting blocked.

### DIFF
--- a/.github/workflows/regression-suites.yml
+++ b/.github/workflows/regression-suites.yml
@@ -1,0 +1,87 @@
+name: Test Regression Scenarios
+
+on: push
+
+jobs:
+
+  local-cli-regression-:
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ windows-latest, macos-latest, ubuntu-latest ]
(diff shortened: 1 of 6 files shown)


devdoshi (approved):

JaapRood:
Insta review approve through pair programming, @devdoshi I like it.

PR #445, opened 2020-11-30
Finish macOS workflows in some reasonable amount of time

JaapRood:
Multiple instances of waiting for a release to make it from a new version being pushed to `develop`, prompted me to look into why the workflows take such a long time to complete (I was waiting anyway!). macOS runners must be expensive, as Github seems to limit the amount that can run at any given time for the repo. We should treat that as the limited resource it seems to be, so we don't have to wait 30 minutes for workflows to eventually complete.

### DIFF
--- a/.github/workflows/regression-suites.yml
+++ b/.github/workflows/regression-suites.yml
@@ -9,7 +9,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ windows-latest, macos-latest, ubuntu-latest ]
-        interaction-count: [0, 1, 100, 1000, 10000]
+        interaction-count: [ 10000 ]
         rust-diff-engine: ["true"]
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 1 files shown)


JaapRood:
@devdoshi new macOS runners being spawned for multiple interaction capture sizes seemed the most obvious place to slim down our use of them. Given that memory pressure is the thing that triggers the issues / race-conditions in I/O that this suite seems to be testing for, sticking with only the largest seems like an easy and reasonable win. If we do want to run various capture sizes, we should probably do the as separate steps within a single workflow.

Perhaps it's worth integrating these regression tests as part of the `CI` workflow, which runs tests as well.

devdoshi:
@JaapRood This seems good for now. We should definitely revisit the scripts here to make them directly useful in CI.

devdoshi (approved):

PR #446, opened 2020-12-01
file system capture saver should save all interactions.

devdoshi:
The interface ICaptureSaver has 3 methods: init(), save(interaction), and cleanup(). We are currently using it in a few different usage scenarios, so the interface doesn't provide any guarantees about whether a given interaction has been saved when save(interaction) resolves or when cleanup() resolves. It requires cooperation between the caller and the saver to achieve the desired behaviors. In addition, the batching logic is getting difficult to follow, test, and maintain. In the future we should consider redefining the interfaces and reimplementing (perhaps with xstate representing the parallel states). For now, when cleanup() is called on the file system capture saver, it should wait until all the incoming interactions have been saved and then stop listening for more batches of interactions. This approach has issues as well, since it doesn't stop accepting save(interaction) calls. Let's discuss


workspaces/cli-shared/src/captures/avro/file-system/capture-saver.ts
```diff
@@ -101,13 +109,39 @@ export class CaptureSaver implements ICaptureSaver {
   }
 
   async save(sample: IHttpInteraction) {
+    this.interactionsReceivedCount++;
     // don't await flush, just enqueue
-    await this.batcher.add(sample);
+    this.batcher.add(sample);
```
devdoshi, line 114: batcher.add() returns a Promise that resolves once the request has been flushed to a "batch" event. Per https://github.com/SGrondin/bottleneck/blob/master/src/Batcher.coffee it resolves after every "batch" event handler has been called _synchronously_, so it doesn't wait for the async work we are doing.

JaapRood: Might be worth doing something with the rejection of the Promise  that `this.batcher.add` returns. Without us redirecting it, it'll be a cryptic uncaught rejection from `Batcher` that has lost trace of this `save` method. For these kinds of errors that can happen out of band emitting an `error` event in a `EventEmitter` the `CaptureSaver` is generally the way to go. If a consumer of `CaptureSaver` doesn't handle it, it'll cause a crash, which is exactly what you want unhandled errors to do.

workspaces/cli-shared/src/captures/avro/file-system/capture-saver.ts
```diff
@@ -101,13 +109,39 @@ export class CaptureSaver implements ICaptureSaver {
...
+    await new Promise((resolve, reject) => {
+      const poll = () => {
+        const interactionsReceivedCount = this.interactionsReceivedCount;
+        const interactionsSavedCount = this.interactionsSavedCount;
+        developerDebugLogger(
+          'waiting until interactionsReceivedCount matches interactionsSavedCount...',
+          interactionsReceivedCount,
+          interactionsSavedCount
+        );
+        if (interactionsSavedCount === interactionsReceivedCount) {
```
JaapRood, line 129: I'd consider `>=` here, from a fault-tolerance perspective in case there's some bug in arithmetic, but I might be just a bit too paranoid there.

JaapRood (commented):
This looks fine to me as a way to solve the issue of unsaved interactions given the existing `CaptureSaver`. Keeping and comparing counts smells a bit, but is a pragmatic solution while we think about rethinking this more generally. If you look at the function and possible states this could be in, the `WritableStream` concept fits really well, which gives us a good idea on how to stop writes as well: keep a flag that allows `ending -> end` transition.

workspaces/cli-shared/src/captures/avro/file-system/capture-saver.ts
devdoshi: I think if somehow the saved count advanced past the received count, we'd have some interesting issues at hand

PR #447, opened 2020-12-01
[Draft] explore local-cli diff error handling

devdoshi:
This PR explores what happens when the rust binary is missing or terminates with a non-zero exit code. It seems like there are a few issues stopping the error from reaching the frontend. One appears to be related to the SSE notifications channel not flushing the final write()s. Another is the daemon crashing due to uncaught errors. Another is execa's output seeming to be ignored


JaapRood (commented):
Nice exploration, it's good to know where we've got issues and how errors propagate, that's half the battle!

PR #448, opened 2020-12-01
Improve error handling during installing and running diff-engine

JaapRood:
We've seen multiple instances where the `diff-engine` failed with an error, due to both a runtime constraints or something that could be known at the moment of installation. However, all of these got swallowed, causing diffs to never finish or produce no results.

These changes aim to achieve two things:

- fail installation of the `diff-engine` package for when the downloaded binary is incompatible with the host, for example through missing dynamic libraries (we saw a missing C-runtime on Windows).
- throw errors occurring from spawning the `diff-engine` process through `spawn({ specPath })`. It's up to the consumer to decide how those should be handled in the context of spawning the diff engine.


JaapRood:
I've implemented the changes that achieve the two aims. 

For the runtime, I've introduced a `DiffEngineError`, that is thrown with information about what the nature of the failure was. The test for thrown runtime errors is still work in progress (extracted from our windows debugging code, needs a suite and case for working version as well). 

We test whether the binary is correctly compiled for the host is tested by running it right after extraction as with `--version`, as the last step of binary installation. Testing this in an automated way is tricky because of the end-to-end nature of downloading binaries. Fortunately, the functionality is also simple and straightforward enough that a manual test should do for now.

Here's a test error invoked by running the binary with a different command:
Screenshot 2020-12-01 at 13 11 38

And here's how we expect and hope installation to go by running `--version`:
Screenshot 2020-12-01 at 13 13 17

JaapRood:
This is now ready for review, landing 2 basic tests that verify spawning the diff engine works (snapshotting the diff results) and will throw errors upon crashes.

workspaces/diff-engine/tests/lib/spawn.js
```diff
@@ -0,0 +1,102 @@
...
+      ),
+    });
+
+    interactions.pipe(diff.input);
+    let outputLines = bufferLines(diff.output);
+    let errorLines = bufferLines(diff.error);
+
+    let diffs = (await outputLines).map((line) => JSON.parse(line));
+    diffs.sort(([, tagsA, fingerprintA], [, tagsB, fingerprintB]) =>
+      compareString(
```
devdoshi, line 36: any reason to not use a.localeCompare(b)?

JaapRood: Not at all, just me being used to lazy `sortBy` and forgetting the most elegant way of writing the compare function 😅. Will fix.

workspaces/diff-engine/tests/lib/spawn.js
```diff
@@ -0,0 +1,102 @@
...
+      let errorLines = bufferLines(diff.error);
+
+      let engineError;
+      try {
+        await diff.result;
+      } catch (err) {
+        engineError = err;
+      }
+
+      t.type(engineError, DiffEngineError);
```
devdoshi, line 75: if it's not too much trouble, can we snapshot the outputLines and errorLines for each test?

workspaces/diff-engine/lib/index.d.ts
```diff
@@ -8,4 +8,40 @@ export function spawn({
   input: Duplex;
   output: Duplex;
   error: Duplex;
+  result: Promise<void>;
 };
+
+export interface DiffEngineError extends Error {
```
devdoshi, line 14: the formatting seems messed up here but nbd

workspaces/diff-engine/lib/index.js
```diff
@@ -59,9 +59,15 @@ function spawn({ specPath }) {
   diffProcess.stdout.pipe(output);
   diffProcess.stderr.pipe(error);
 
-  return { input, output, error };
+  // get a clean result promise, so we stay in control of the exact API we're exposing
```
devdoshi, line 62: nice I had a feeling you'd want something like this instead of how I did it in my exploration

devdoshi (approved):
nice work!

devdoshi:
My thoughts for automated testing were to have some "dummy" binaries that simulated process.exit(1) etc. and injecting the path to that binary instead of using the prebuilt path. But I agree manual should be fine

workspaces/diff-engine/tests/lib/spawn.js
JaapRood: Tried, but too much trouble. I considered the same benefits, but already in development of tests ran into the noise and trouble that creates, as both make no guarantee on order of output, hence the parsing and sorting before. In addition to that, stderr yields debug output that's volatile (especially since we don't have a proper logging solution yet). 

In the end, I reckoned these tests are mostly about verifying the integration between the Node.js and Rust side, not exactly what output the diff engine produces. Of course, the two are related, but given the trouble of doing much more than what we're doing now, I found this to be a reasonable balance.

workspaces/diff-engine/lib/index.d.ts
JaapRood: Ahh, my editor was set to render tabs at 2 wide, so never noticed they got mixed. Github's 8 wide spaces to the rescue (do people even ever use that?). Fixing

JaapRood: Fixed in https://github.com/opticdev/optic/pull/448/commits/23c5c025e42ff8507fc20c1e50fc8fd9aa07c3d2

workspaces/diff-engine/tests/lib/spawn.js
JaapRood: Fixed in https://github.com/opticdev/optic/pull/448/commits/41d6f76cbc66e26f4ea554aa3e270ffe16ec0d80

JaapRood:
@devdoshi any idea why "Local CLI end-to-end tests" failed? Not really familiar with those, if it at all would be affected.

devdoshi:
> @devdoshi any idea why "Local CLI end-to-end tests" failed? Not really familiar with those, if it at all would be affected. A local end-to-end tests with `apidev` works without a hitch, if that counts for anything.

I removed the fauly job in the last PR I merged (#446), safe to ignore

PR #449, opened 2020-12-01
cli-server and cli-scripts should use sentry on the beta release channel

devdoshi:


devdoshi:
- [ ] created optic-local-cli project on Sentry
- [ ] set alerts to go to #sentry-errors slack channel
- [ ] verified alerts appear when running locally
- [ ] verified alerts appear when running prerelease build (need to merge and publish prerelease first)
- [ ] reviewed sentry alerts to get a sense of whether there's any sensitive data we should redact
- [ ] added additional tags representing the end user's context (os, node version, 

Some of the things I opted out of were:

- setting up sentry for typescript, which would require build process changes and uploading source maps.
- setting up the sentry express integration, which would require some changes to the control flow

devdoshi:
@LouManglass @JaapRood please help with the unchecked boxes in the preceding comment. test with 8.6.0-beta.21 when that is published

LouManglass:
Woo sentry!

workspaces/cli-server/src/sentry/index.ts
```diff
@@ -0,0 +1,33 @@
...
+      dsn: process.env.OPTIC__CLI_SERVER__SENTRY__DSN,
+      serverName: 'optic-cli-server',
+      environment:
+        process.env.OPTIC__CLI_SERVER__SENTRY__ENVIRONMENT || 'local',
+      release: process.env.OPTIC__NPM__VERSION || 'local',
+      tracesSampleRate: 1.0,
+    });
+    console.log('Sentry is enabled');
+  }
+}
```
JaapRood, line 19: Parsing environment variables to establish a config doesn't belong this deep into the codebase, but way closer to the edge of the program. Anything that's outside a `main` function or global `config` (that's immediately consumed by that `main` function) should deal with plain objects and values. Not doing so can make it incredibly opaque to reason about why code behaves the way that it does, hard to re-use anything, etc.

Once you take this logic out of here, I see little reason for why this needs to be an interface of a class and 2 implementations of it: a function that registers sentry based on whether a dsn was passed should be more than enough. Couple of lines, easier to parse and understand.

JaapRood (commented):
I think it's fine that this was merged to try and get it into a prerelease, but there's some issues we should return to and clean up.

workspaces/cli-server/src/sentry/index.ts
devdoshi: Sentry is meant to be loaded as early as possible from the main module like you say about configs, but I agree the envs can move to config so this module uses plain values. The interface is a remnant of me attempting to support the sentry express integration, which had more methods after init. I still think it is a good idea to create the separation for any 3rd party dependency and is overall simpler to maintain as an instance that callers control rather than a function with conditional effects

JaapRood: > The interface is a remnant of me attempting to support the sentry express integration, which had more methods after init. 

@devdoshi yeah, it seemed like something like that.

> I still think it is a good idea to create the separation for any 3rd party dependency and is overall simpler to maintain as an instance that callers control rather than a function with conditional effects.

Totally, I've been in projects where there were multiple remote-error tracking integrations at once, where we soon ended up with a generic `errorTracker` interface. Since it's mostly glue code that wasn't too hard to retrofit.

JaapRood:
I opened #450 to do said clean up.

PR #450, opened 2020-12-02
Create default and sanitised way for cli's to configure themselves

JaapRood:
A basic need of the various CLI's and related packages (like `cli-server` and `cli-scripts`) as we start to connect them to services, is the ability configure the use of those. Most of the time when developing Optic we will want to run against different instances of services, or not run against them at all. Because of this, the configuration often comes from the environment, which brings a bunch of extra risks and intricacies. In addition, there's a lot of the same information a lot of the packages need, like their version, what environment we're running (`development`, `production`), whether it's a prerelease, etc.

This PR introduces a default way to do this, by generating a `CliConfigObject` given a `package.json` and `process.env`. This object can then be extended to have package specific configuration attached as well.

Having used this setup and seen it used in a lot of other projects, I can vouch for how a single `Config` object like this encourages configuration to live near the edges of the process: often read by a `main` function and it's values propagated further down. That hygiene of keeping the environment close to the environment side of the process helps prevent surprising behaviour when trying to re-use models, when trying to test them etc.


JaapRood:
@devdoshi this touches some of the Sentry code you only wrote very recently. That would have been fine and functional and not worth doing this for alone, but it just reminded me of how this needed to be done at some point. At all projects where I didn't do this early, I've come to regret it, making it just harder to get into a hygienic and clean approach to configuration later. 

_How_ the config objects are constructed I don't really care too much about: whether that's through extending, or nicer utilities to parse env variables, cascading of files, etc. What you find here is the simplest approach I know off. To me, it's about the fact that we use a Config object constructed early on in the process in the first place.

workspaces/cli-shared/src/config.ts
```diff
@@ -0,0 +1,72 @@
...
+    envName,
+    env: {
+      development: envName === Env.Development,
+      production: envName === Env.Production,
+    },
+  };
+}
+
+export function isEnvTrue(val: string | undefined) {
+  return val === 'yes' || val === 'true' || val === '1' || val === 'y';
```
devdoshi, line 71: As an aside I'd strongly prefer we did not use anything except yes/no

devdoshi: *_ENABLED=yes|no|undefined/empty, and checking if it is yes means opt-in and checking if no means opt-out

workspaces/cli-server/src/config.ts
```diff
@@ -0,0 +1,24 @@
...
+} from '@useoptic/cli-shared';
+const pkg = require('../package.json') as PackageJson;
+
+const cliConfig = createCliConfig(pkg, process.env);
+
+export default {
+  ...cliConfig,
+
+  errors: {
+    sentry: cliConfig.env.production &&
```
devdoshi, line 16: This does make it slightly more difficult to enable sentry during development which I did find helpful. Will you be making the corresponding changes to cli-scripts?

workspaces/cli-server/src/config.ts
```diff
@@ -0,0 +1,24 @@
+import {
+  createCliConfig,
+  Env,
+  CliConfigObject,
+  isEnvTrue,
+  PackageJson,
+} from '@useoptic/cli-shared';
+const pkg = require('../package.json') as PackageJson;
```
devdoshi, line 8: I have a vague memory of issues using the package json in the past but if typescript doesn't complain and the package json is present in the published package then it should work.

workspaces/cli-shared/src/config.ts
JaapRood: we're already using `true` for feature flags as well, which is the main reason why I added this. Added `1` and `y` as that's super common in other Unixy / GNU tools, but happy to leave those out since they aren't used anywhere.

workspaces/cli-server/src/config.ts
JaapRood: Typescript complains if you use `import`, but require is just another function-call, so fine. Safer, if we didn't control the `package.json` would be to use [`read-package-json`](https://www.npmjs.com/package/read-package-json) which performs some validation steps, but since these packages won't even publish without a valid `package.json`, I didn't deem the necessary at this point.

workspaces/cli-server/src/config.ts
JaapRood: I've only ever found enabling sentry locally helpful in implementing the integration with sentry, and _have_ found it super annoying to find development errors to be tracked because it was accidentally enabled, so I optimised with that hard-won experience in mind 😅.

I actually don't think DSN and such has to come from the environment either, as it's often 1:1 mapped to the client, but it's there so 🤷‍♂️. Maybe if we want to clean up the .env files at some point.

And yes, going to apply this to `cli-scripts` as well! Forgot to mention that this is draft because I wanted to agree on the approach first, before spending the time to applying it fully.

JaapRood:
Happy where this has landed. One notable thing is how I've inlined the DSN straight in the config, with the reasoning that:

- with Sentry, everything environment related is injected through other means, with DSN staying the same
- there are no real scenarios' past initial implementation where it actually comes from the environment, so read from file either way
- this keeps the .env file mostly about feature flags, and buys us more time before that gets unwieldy.

workspaces/cli-scripts/src/config.ts
```diff
@@ -0,0 +1,25 @@
...
+} from '@useoptic/cli-shared';
+const pkg = require('../package.json') as PackageJson;
+
+const cliConfig = createCliConfig(pkg, process.env);
+
+export default {
+  ...cliConfig,
+
+  errors: {
+    sentry: cliConfig.env.production &&
```
devdoshi, line 16: Is the production check redundant if we're injecting whether it is enabled? I would prefer the SENTRY__ENABLED to be the source of truth rather than SENTRY__ENABLED * env

workspaces/cli-server/src/config.ts
```diff
@@ -0,0 +1,25 @@
...
+
+const cliConfig = createCliConfig(pkg, process.env);
+
+export default {
+  ...cliConfig,
+
+  errors: {
+    sentry: cliConfig.env.production &&
```
devdoshi, line 16: same comment re: redundant production check

devdoshi (approved):
looks good, but please address my comments.

workspaces/cli-scripts/src/config.ts
JaapRood: See [my original comment ](https://github.com/opticdev/optic/pull/450#discussion_r534184823)about why I decided against that:

> I've only ever found enabling sentry locally helpful in implementing the integration with sentry, and have found it super annoying to find development errors to be tracked because it was accidentally enabled, so I optimised with that hard-won experience in mind 😅.

PR #451, opened 2020-12-02
ensureDaemonStarted should verify pid exists and port is reachable

devdoshi:
When the lockfile is inaccurate, the daemon should be started fresh.

Tested on macos via:
```
$ export DEBUG=optic*
$ source sourceme.sh
$ apidev daemon:stop
$ apidev daemon:start
(might need to ^C here, or open a new terminal)
$ killall -9 node
$ apidev daemon:start
```
You can also compare the output of the final apidev daemon:start when you've done killall -9 node vs killall node which exists gracefully


devdoshi:
@LouManglass see 8.6.0-beta.24

JaapRood (commented):
Looks good to me! Was interested to see how the 2 new dependencies do their job and if those would result in any issues on Windows. Fortunately, there's no reason to think that they wouldn't!

PR #453, opened 2020-12-03
Propagate errors with Diff Worker instead of swallowing them

JaapRood:
Continuing the work of #448, we want to make sure that any errors thrown by the `diff-engine` are propagated by the `diff-worker-rust`, rather than being swallowed as they are currently.


JaapRood:
The logic of propagation itself was only a couple of lines, but since we've hurt our head against this multiple times already ( misdiagnosing crashes as diffs that hanged) this is worth taking the time to cover with a test. Given the nature of the diff worker, which accepts Avro captures and uses the file system actively to maintain it's state, that's a bit more involved than the average test. The main framework for getting the diff setup is there, but test itself still failing through, probably because the example interactions aren't generated correctly.

Apart from that, it's also worth considering how we can test whether it's reporting progress adequately. The diff-worker-rust uses `process.send` directly, which makes that tricky. It could be fixed by making `diff-worker-rust` work with a supplied `sendProgress` method, with the `emit-diff-projections-rust` script the place to turn that into IPC calls. That would allow stubbing more easily and for a cleaner test run. However, maybe verifying the sending of progress isn't really that interesting.

JaapRood:
We've ended up adapting the `DiffWorkerRust.run` to work as an async generator, because for the test we needed to be able to detect when the diff worker would be done. Before it was only run in it's own process with nothing after it, so therefor it could rely on Node not exiting while it was running because of the streams remaining open. A nice benefit is that we got to completely decouple the sending of progress as well, as we identified before.

Tests pass, but we _are_ seeing a mention of an uncaught promise rejection in the test output, despite it passing. Probably worth getting to the bottom off.

JaapRood:
Having look into why we're getting an `UncaughtPromiseRejectionWarning`, it's become evident that the async generator approach isn't going to work if we want to run without warnings, as the loop we're introducing prevents the rejection from being handled _before_ it's raised. We _do_ actually end up handling the rejection, as is evidenced by the `PromiseRejectionHandledWarning` that follows it soon after, referring to the initial warning by a `rejection id`.

So as far as error handling goes, we'll have to rely on some other mechanism to make sure we can propagate any errors swiftly.

JaapRood:
It's become my Friday "d*mn it this should be able to work" rabbit hole trying to `diffWorker.run` to work as an async generator and have it propagate _all_ errors from running, without any warnings about unhandled promise rejections / async handled promise rejections. However, I couldn't get it to behave the right way in that API without the code getting super ridiculous, so I ended up splitting the API up in more specific chunks that were easier to get right separately. 

The issue lied in the fact that once the `diffEngine` is spawned, we must instantly synchronously register a `catch` handler, as otherwise an early crash will race it and cause the warning (like in our test). While that's not necessarily the hardest, it's the fact that we need to do things *after* we spawn the diff engine that requires that error to throw and interrupt. The only real way to do that in an `async` function is by `await`ing that result promise, but then you can't do anything until the engine has crashed / shut down again.

A bigger refactor of how the `diffEngine` API (like setting it up before spawning it, so that can be last thing to hapen) could resolve all of these issues and still make the single `run` API work, however, that's not something I wanted to test while we're trying to actively stabilise what we have. One attempt did get close, but basically meant writing a complete custom `Promise`, so all work after diff-engine spawn can be raced with diff-engine completing or failing, which got really hairy, out of the world of `async / await` and just.. silly.

In the end, I went the pragmatic way:

- split up the API and have 2 separate ways for "sync" and "async" errors to propagate, rather than through rejecting a single Promise. This was done through an exposed `EventEmitter` on which an `error` event is emitted. If not handled by the consuming code, Node.js will crash, giving us the behaviour we want. That said, it _can_ be handled through `worker.events.on('err', ...)`
- separated starting the worker from reporting progress, as combining the APIs made it very confusing to know which errors would propagate through that function failing, and which wouldn't.

Ultimately it means that:
- any errors in start up of the worker will cause `await worker.start()` to reject.
- any errors in processing itself will cause `error` to be triggered on `worker.events`
- `worker.progress` will only reject for errors in the progress reporting, not the processing it's reporting on

Maybe at one point, like once we have time to change the `diff-engine` interface itself (as described above) we can get back to a unified way for errors to arise again, through means of an async generator. As a note to future self: a lot of the issues are about async generators not having any additional state, but once you add them.. it's basically a `Stream`, which has all the end, finishing, error states backed in and can be consumed as an async iterator. Perhaps the diff worker run should be expressed as `Stream`? This might haunt me this weekend 😅.

JaapRood:
We've spent more than enough time for this iteration, so with all tests passing (plus manual local tests), I'm merging this. Any follow-up improvements should be looked at separately.

PR #460, opened 2020-12-07
Give user instant feedback when trying to install diff-engine on unsupported platform

JaapRood:
Small bug, where we didn't interrupt the installation of diff-engine's prebuilt binaries when the platform turned out to be unsupported.


devdoshi (approved):

JaapRood:
Not _totally_ ready for merging yet as we're looking into whether we want to do some in addition to the message.

PR #461, opened 2020-12-07
Add support for installing the Rust Diff engine on Macs with an M1 (Apple Silicon) natively

JaapRood:
While there's a way to install optic with the new Rust diff engine on Macs with M1 (through running a x86 version of Node and relying on Rosetta 2 to translate), running a native binary would be far preferred. This PR provides this pre-built binary as part of the release workflow.

It uses the macOS 11 runner featuring the latest version of Xcode combined with the `1.49-beta.2` of the Rust compiler (we should set ourselves a reminder to move back to stable once 1.49 is stable). It then cross-compiles to `aarch64`, since there aren't any native runners yet.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -8,12 +8,15 @@ env:
   BUCKET_NAME: optic-packages
   PACKAGE_NAME: api
   NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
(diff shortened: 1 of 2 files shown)


JaapRood:
I've tested everything there is to test manually. Time to remove the safeties and give this a go, that's the only way we know whether it'll work all together.

PR #462, opened 2020-12-07
Sentry should record os/platform details

devdoshi:


workspaces/cli-scripts/src/errors.ts
```diff
@@ -21,4 +23,24 @@ export function trackWithSentry({
...
+    hostArch,
+    hostPlatform,
+    hostCpuCount,
+    hostMemorySize,
+  });
+
+  getOrCreateAnonId().then((id: string) => {
+    Sentry.setUser({
+      id,
+    });
```
JaapRood, line 44: Do we need to handle the error case here, or is it fine to raise an `UncaughtPromiseRejection`? Depends on `getOrCreateAnonId`, whether only rejections would be type / reference errors, or whether it actually does some form of I/O and has failure as natural part of it's existence, I guess?

JaapRood (commented):
Straightforward and how I would expect this to be done. Apart from some reflection on how a possible error is handled, ready to go.

workspaces/cli-scripts/src/errors.ts
devdoshi: We don't expect getOrCreate to fail under normal conditions. Since we're adding Sentry, Sentry should report the uncaught rejection.

JaapRood: Just checked: `getOrCreateAnonId` accesses the file-system, so we have to expect failure as part of normal operation if this is to be fault-tolerant. Since it's a failure in adding more context to error tracking, I'd say this might be one of those rare occasions where swallowing the error could be appropriate (but _only_ IO errors, of course, as any other puts the app in an undefined state).

devdoshi: Agreed it's nonessential functionality worth swallowing and optionally manually tracking with Sentry. However, an IO error would suggest that other things are likely to fail, if they haven't already, so I'd consider adding extra testing surface area counterproductive

JaapRood: Fair enough. No current appetite for a case closer to the edges, can agree with that!

PR #463, opened 2020-12-07
anonymous ids for sentry

acunniffe:
This PR introduces a stable anonymous id we can group sentry logs by. This should make it easier for users to ask us for help when diff engine issues are detected `api debug:identity` will print the IDs and they should correlate with what's in sentry

### DIFF
--- a/workspaces/local-cli/package.json
+++ b/workspaces/local-cli/package.json
@@ -43,7 +43,8 @@
     "tslib": "^1",
     "url-join": "^4.0.1",
     "uuid": "^8.0.0",
-    "which": "^2.0.2"
+    "which": "^2.0.2",
+    "human-readable-ids": "^1.0.4"
   },
   "devDependencies": {
     "@oclif/dev-cli": "^1",
--- a/workspaces/local-cli/package.json
+++ b/workspaces/local-cli/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@useoptic/cli",
   "description": "The Optic CLI",
-  "version": "8.6.0-beta.31",
+  "version": "8.6.0-beta.32",
   "author": "@useoptic",
   "bin": {
     "api": "./bin/run"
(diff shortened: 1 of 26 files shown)


workspaces/local-cli/src/shared/authentication-server.ts
```diff
@@ -97,6 +110,36 @@ export async function deleteCredentials() {
...
+
+export async function getOrCreateAnonId(): Promise<string> {
+  const storage: IUserStorage | undefined = await getCurrentStorage();
+
+  if (storage && storage.anonymousId) {
+    return storage.anonymousId;
+  } else if (storage) {
+    const storeValue = {
+      ...storage,
+      anonymousId: require('human-readable-ids').hri.random(),
```
devdoshi, line 130: 'human-readable-ids' should be imported at the top level

workspaces/local-cli/src/shared/authentication-server.ts
```diff
@@ -82,8 +85,18 @@ export async function ensureCredentialsServerStarted(
   };
 }
 
+function defaultStorage(): IUserStorage {
+  return {
+    idToken: undefined,
+    anonymousId: require('human-readable-ids').hri.random(),
```
devdoshi, line 91: 'human-readable-ids' should be imported at the top level

workspaces/local-cli/package.json
```diff
@@ -43,7 +43,8 @@
     "tslib": "^1",
     "url-join": "^4.0.1",
     "uuid": "^8.0.0",
-    "which": "^2.0.2"
+    "which": "^2.0.2",
+    "human-readable-ids": "^1.0.4"
```
devdoshi, line 47: if you run yarn install, it'll sort these dependencies. That will lead to less noise on diffs in the file.

workspaces/local-cli/src/shared/analytics.ts
```diff
@@ -27,16 +28,24 @@ const analytics = new Analytics(token, {
   flushAt: 1,
 });
 
+const anonIdPromise: Promise<string> = new Promise(async (resolve) => {
```
devdoshi, line 31: is this redundant?

acunniffe: this my pattern for making it run the first time only. 

Later on line 41,47 both await the resolution of this promise. Their parent function can be called many times throughout the lifecycle. It'll wait the first time until it has the token, then all future runs the promise is already resolved so it will go faster.

devdoshi: Isn't that the same as saying const x = getOrCreateAnonId(), and await x in both call sites?

devdoshi (approved):

JaapRood:
@acunniffe your PR describes what the effect is supposed to be, but can you say a bit more about how you implemented it? Reading this, but really confused by the naming of a module as `optic-rc`, why it's yet another config file we need, where it would live, etc.

PR #464, opened 2020-12-07
Local On-boarding flow

acunniffe:
This PR contains the new onboarding flow -- it starts immediately after init on` /apis/{number}/setup`.


workspaces/cli-client/src/spec-service-client.ts
```diff
@@ -51,12 +52,19 @@ export class Client implements ISpecService {
     );
   }
 
-  loadConfig(): Promise<{ config: IApiCliConfig }> {
-    return JsonHttpClient.getJson(
+  async loadConfig(): Promise<{ config: IApiCliConfig; configRaw: string }> {
+    return await JsonHttpClient.getJson(
```
devdoshi, line 56: awaiting should be up to the caller, this should just pass the promise through. Same for all of these, to be consistent.

workspaces/ui/scripts/fetch-docs.js
```diff
@@ -0,0 +1,19 @@
+const fetch = require('node-fetch');
```
devdoshi, line 1: @notnmeyer thoughts?
@acunniffe I think this should be done more statically

acunniffe: this is currently what we do for the saas site, but open to making it more static :)

LouManglass:
I would approve this based on function, given our call and the subsequent changesets I pushed up. That's separate from Dev's comments, above.

PR #465, opened 2020-12-07
Hide node deprecation warning on CLI commands

acunniffe:
There's an open issue in the latest mockttp (recently updated) where an intermediate deprecation warning is being emitted. https://github.com/httptoolkit/mockttp/issues/41

This change hides those to clean up the DX of the CLI.

### DIFF
--- a/workspaces/local-cli/bin/run.cmd
+++ b/workspaces/local-cli/bin/run.cmd
@@ -1,3 +1,3 @@
 @echo off
 
-node "%~dp0\run" %* --no-warnings
+node --no-warnings "%~dp0\run" %*
--- a/workspaces/local-cli/bin/run.cmd
+++ b/workspaces/local-cli/bin/run.cmd
@@ -1,3 +1,3 @@
 @echo off
 
-node "%~dp0\run" %*
+node "%~dp0\run" %* --no-warnings
--- a/workspaces/local-cli/bin/run
+++ b/workspaces/local-cli/bin/run
@@ -1,4 +1,4 @@
-#!/usr/bin/env node
+#!/usr/bin/env node --no-warnings
 require('../lib')
 .run()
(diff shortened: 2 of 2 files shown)


workspaces/local-cli/bin/run.cmd
```diff
@@ -1,3 +1,3 @@
 @echo off
 
-node "%~dp0\run" %*
+node "%~dp0\run" %* --no-warnings
```
devdoshi, line 3: node  --no-warnings "%~dp0\run" %*

JaapRood:
@acunniffe @devdoshi can any of shed some light on exactly when and how this mutes warnings? Can we be sure that we _do_ get warnings in local development?

LouManglass:
> can any of shed some light on exactly when and how this mutes warnings? Can we be sure that we _do_ get warnings in local development? Isn't `--no-deprecation` supposed to be [a thing](https://nodejs.org/api/cli.html#cli_no_deprecation) instead of `--no-warnings`?

In my testing, local development invocations do not return deprecation warnings. They're hidden in anything I've tried.

devdoshi:
There was an issue with this approach (see https://github.com/opticdev/optic/commit/9c5cd2d96445c5ffeecc6f6e62c4384e90fad574) but with that being the code you're running against, you can see why you won't see warnings. Easy to put it behind an environment variable, or use our developerDebugLogger, or ... for development, so let me know your thoughts @LouManglass @JaapRood

JaapRood:
Development should never hide warnings, that's what the warnings are there for: to make sure we move ourselves out of harms way and don't create way bigger issues down the line. It might be safe to ignore this particular warning, but there is 0 guarantee that's the case for the next one. Anything we merged that doesn't account for that should be reverted until that's addressed.

PR #467, opened 2020-12-08
Split diff-engine into cli and wasm compilable targets

JaapRood:
This is the first step in bringing the new Rust diff engine to WASM, so it can be used in all places the ScalaJS implementation is used in addition to the main diffing pipeline. It adds a `diff-engine-wasm` package and provides a build for use in both Node.js and a browser runtime. 

One important distinction to make is where in the toolchain the wasm build fits vs the CLI. While for the CLI we want to build an executable binary (and thus prebuild them for each platform), for the WASM we want it to happen as part of building the workspaces, in the same way we compile TypeScript. We can get away with this simplified setup as WASM is platform agnostic.

Further more, for the UI package it integrates with Webpack (and the Webpack dev server) to compile the WASM as part of the front-end development process. Any changes to Rust code will trigger new builds and changes to be reloaded in browser.

The interface provided is **a placeholder** extracted from earlier experiments to verify the integration. We'll adapt it as needed in future PRs.


JaapRood:
We've reached the point where the new `diff-engine-wasm` now compiles and can be used inside `ExampleDiffService`. Through a Webpack plugin integration, any changes in the underling Rust code triggers automatic rebuilds and refreshes in the browser.

Getting there was fairly tricky, with a bunch of moving pieces involved.

While I tried to avoid it, one of the initial crucial requirements that became clear is that we need a separate build of the package for use in Node and the browser. This stems from the mechanism of which the `WebAssembly` implementations of either differ in how the WASM is fetched and compiled. In Node.js this is a synchronous affair, allowing modules to be required like any other, but in the browser the WASM is fetched asynchronously in the background.

That difference becomes more pronounced once we ask Webpack to bundle and load it for us: the WASM module is not allowed to be part of the main entry point chunk and must be loaded async, the best way of which is to do through a dynamic import (calling `import()` which returns a `Promise<module>`). Choosing where to do that was tricky and one of compromise, especially once Typescript and resolving of types get involved. 

In the end, I made all the UI entrypoints require `ExampleDiffService` dynamically, as the services were already being constructed async anyway, and nothing more than construction happens (making it not matter we can't easily import any additional TS types from that module). That way, the `ExampleDiffService` can statically `import` the `DiffEngine` and have all the type checking work without issue.

JaapRood:
Having reviewed aligned on the chosen structure before with @devdoshi, I'm going to merge this as soon as tests pass. Since this is just the start of these changes, we should be able to address non-structural issues separately.

PR #470, opened 2020-12-09
added $PORT flag to API Check

acunniffe:
Going forward, the new Optic documentation is going to advise users to set up their APIs to listen on `$PORT` not `$OPTIC_API_PORT` -- although both will continue to be added to the environment. 

The reason being -- a lot of frameworks and a lot of APIs are already set up to look for that port. Lou has shifted the language towards, 'make sure your API starts of provided $PORT' and away from 'make you API look for this other port that Optic configures'. 

It's subtle, but it's a change worth making. @LouManglass  -- we tested this for start, but it wasn't in the API Check code before now. Can you make sure it works as expected and aligns with the new docs?


devdoshi:
Is there a list of frameworks and where they look for ports? I like the intent of the change, but in practice there might also be more opportunity for other things to interfere with less visibility since it is a more generic name. We can always provide both, right?

PR #471, opened 2020-12-09
Release new on-boarding flow

acunniffe:
This release continues in keeping Learn API + Rust diff engine out of @latest releases -- that's managed by the feature flags. 

Included in this release is the new on-boarding flow that Lou and I have been building in response to the learnings from 2 mo of real usage of the existing on-boarding. 

The guided wizard now lives in the dashboard on `/setup` and api init will open / guide the user through making their start alias on `/apis/{id}/setup`. 

This work pairs with documentation changes.


JaapRood:
Interesting how using the `patch` label it went from `8.6.0-beta.34` to `8.6.1`. Should have been `8.6.0` 🤔. Doesn't really matter, of course, just a bit odd.

PR #473, opened 2020-12-10
Fixes for setup page + check

acunniffe:
@LouManglass I was making a user on-boarding video last night and damn -- the empathy one gets from being forced to think about a camera and an audience. love it -- may actually start using the 'ol glass eye to stay more honest. 

The first thing I noticed is that the inboundProxy check exits with the wrong port in its feedback. Can't start on 3005 (then under) could not start on 3997. Fixed that

I also realized the information overload factor of the initial UI was really high. There's stuff to read above, there's code, there's stuff to the right. I realized we could probably cut down on the feedback until after the first check runs. 

Screen Shot 2020-12-10 at 8 11 27 AM


LouManglass (approved):
Nice sanding off some of the coarse edges here, looking good!

PR #474, opened 2020-12-10
Cleanup rough edges in API Check Flow

acunniffe:
Lou and I already found a few small copy/functionality changes we wanted to enhance to make the on-boarding flow easier for new users.  

I'm going to test a local build and put extra eyes on this to make sure there's nothing expected from the WASM work on develop that made it in. @JaapRood, I'm in no rush to release this (can go tomorrow). Wanted to check if you foresee any gotchas or can we consider that WIP stable for the next few releases we cut?


JaapRood:
@acunniffe reflecting on what could go wrong, I realised I might have broken the release workflow. While not being wired up to anything yet, the `diff-engine-wasm` now requires the Rust toolchain to be setup to build all packages before publishing, which I don't think I've done yet. I'll take a look at that right away and report back.

JaapRood:
@acunniffe false alarm, the release should work just fine without further fixes.

Turns out Github workflow runners come with the latest `stable` version of `rustc` and `cargo` installed, which is exactly what we're after. The `wasm-pack` module, which is used for handling the WASM builds with, automatically installs the `wasm32-unknown-unknown` target for cargo when it's not present, so everything just builds without issue!

PR #477, opened 2020-12-11
Run in-browser diffs with Rust diff engine

JaapRood:
There are various scenarios where diffs are run in-browser, feeding a spec and interactions and handling results in a mostly in-memory way. Mostly, this is for development / and private debug sessions. This PR allows that to be achieved against the same diff engine as we're on the cusp of releasing.

It implements an `ExampleDiff` that interacts with the WASM diff engine and using `async generators` to diff the example interactions one by on, yielding results over time and storing them. A shared instance of `ExampleDiff` is then inserted into `ExampleDiffService` and `ExampleCaptureService`, starting it through the normal `captureService.startDiff` like it happens for the local-cli versions of those services.

It intentionally **doesn't** hide this behind a feature flag. This is a PR in a set of ones that is supposed to _reduce_ the amount of technical debt, and especially in the ui package are we faced with lots of old and stale code. So, since this is mostly used for debugging and other non-core areas of the experience, going straight over to the Rust implementation is the desired step to take.

### DIFF
--- a/workspaces/diff-engine-wasm/src/lib.rs
+++ b/workspaces/diff-engine-wasm/src/lib.rs
@@ -23,17 +23,16 @@ pub fn diff_interaction(
   interaction_json: String,
   spec: &WasmSpecProjection,
 ) -> Result<String, JsValue> {
-  let TaggedInput(interaction, tags): TaggedInput<HttpInteraction> =
-    serde_json::from_str(&interaction_json).unwrap();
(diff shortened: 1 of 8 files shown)


JaapRood:
Since this is a first step in a longer process of refactoring, that also unblocks other projects, I'm going to merge this as soon as the tests pass. Reviews are still important, but will be taken on and resolved in future PR's. That way we can side step the refactoring rabbit hole that's endlessly deep and wide!

PR #480, opened 2020-12-16
Generalise diff results pipeline logic so it can be effectively composed

JaapRood:
Now the diff can be run with `diff-engine-wasm` we introduced a second place (the first being `OnDemandDiff` in `cli-server`) where we need to take the results from the diff engine and compute it's immediate derivatives: normalised diffs and unrecognised urls. That logic constitutes non-trivial business logic, so having it in two places isn't great.

The use of the `diff-engine-wasm` in the unit tests of the new interpreters has shown there is a third spot where this is required: right now it's consuming the `ExampleDiffService`, but what it really should be doing is implementing it's own way to generate diff results.

This PR implements find more generic way to express this type of logic, so it can be re-used in all these different places so all diff result derivatives can remain consistent, and these places can focus on how to feed it with spec + interactions and how to use the results.

It does so by implementing the logic and returning typed async iterators (`AsyncIterator<T>`), which can be used in any modern Javascript engine to express asynchronous traversals. It implements a couple of these operators for `DiffResults` and `UndocumentedUrls` streams. It also piggy backs on [`axax`](https://www.npmjs.com/package/axax) to provide a generic tool belt to compose streams together through the usual suspects like `map`, `filter`, `reduce`, `take`, etc.

The result is that both `OnDemandDiff` and `ExampleDiffService` can now express their pipelines in a couple of lines of code, rather than duplicating this business logic. The only thing they remain responsible for is where the stream of diff results come from, be it the file system,  an in-memory array of values, etc.


JaapRood:
As with the other PRs in this series, I'm going to merge this and take anything that comes up from review up for the next iteration of refactoring. The line at where we stop making improvements is pretty arbitrary, and landing this for @acunniffe is important, too!

PR #485, opened 2020-12-18
Fix bug where Webpack couldn't build and bundle diff-engine-wasm

JaapRood:
While all tests passed for #480, they don't cover the Webpack compilation! Sneaking in a tiny bit more clean up that I found stashed on my branch but forgot to pop as well.


JaapRood:
@acunniffe this fixes the issue you saw. Feel free to merge once all tests pass and unblock yourself!

PR #486, opened 2020-12-28
init: clarify next step when `optic.yml` exists.

LouManglass:
When optic.yml exists and you run `api init`, you get an error that the optic.yml file exists and init will not overwrite the existing structure. This change clarifies a next step: run `api spec` to open Optic and see the current documentation generated from the existing spec.


LouManglass:
Before:

```
$ api init 
_red_ This directory already has an **optic.yml** file.
```

After:

```
$ api init 
_red_ This directory already has an **optic.yml** file.
_yellow_ You can see the current documentation for this project with **api spec**.
```

PR #493, opened 2021-01-04
Fix bug where AsyncTools.lastBy would leak old results to new invocations

JaapRood:
This fixes a bug where to get the last stream items by some predicate wasn't pure per stream it was applied on. Instead, it would re-emit the last known entries for each stream it was ever called on. @acunniffe this fixes the issue you reported as regarding undocumented urls https://github.com/opticdev/optic/commit/3bffce1ebf90b6b997e358c22d309111e6b89bcb#r45483119 (turns out diffs more generally were affected).

The fix was fairly simple: make sure we use a new `Set` of seen keys for each stream, rather than using a single one per predicate. I added test coverage to reproduce the error and verify the fix.

Also: @acunniffe sneaked in a fix for some UI code which must have not gotten merged properly.

### DIFF
--- a/workspaces/cli-server/src/diffs/on-demand.ts
+++ b/workspaces/cli-server/src/diffs/on-demand.ts
@@ -240,13 +240,14 @@ export class DiffQueries implements DiffQueriesInterface {
   }
   undocumentedUrls(): Readable {
     if (fs.existsSync(this.paths.diffsStream)) {
-      let diffs = chain([
-        fs.createReadStream(this.paths.diffsStream),
-        jsonlParser(),
-        (data) => [data.value],
-      ]);
+      let diffResults = this.createDiffsStream(this.paths.diffsStream);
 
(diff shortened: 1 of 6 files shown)


JaapRood:
Almost ready to merge, just need to revert @acunniffe's [reversion](https://github.com/opticdev/optic/commit/3bffce1ebf90b6b997e358c22d309111e6b89bcb#r45483119) so that the new code is being used once again 😅.

acunniffe (approved):

acunniffe:
@JaapRood  -- looks good. I noticed a small UI issue (unrelated) I had to fix to test this properly in the UI. Looks like a variable declaration didn't merge properly last week. I believe that's why your previous commits failed.

JaapRood:
@acunniffe all green! It's my end of day, but feel free to merge in the meantime (as long as you're ready to revert if this somehow causes issues, although I can't see how). Otherwise, I'll do it first thing tomorrow!

PR #497, opened 2021-01-06
dump_capture_saver can now repeat samples to generate large captures

LouManglass:
Debugging tool now takes an optional value (default 1) to multiply the number of times it uses a sample in a dump capture conversion. This allows us to increase the size of a capture dump for performance testing without having to generate more traffic.


devdoshi (approved):
Nice! I have some stylistic feedback but nothing priority for a debug script

PR #498, opened 2021-01-07
introducing optic debug command

acunniffe:
Ahead of Optic 9's pending release, we want to provide users with a means to send us really large debug captures (much bigger than what the browser can prepare) and a way for larger teams who have adopted Optic to sanitize and share example traffic we can put in our test suites -- a big way to invest in the overall stability of the project.

If you run `api debug:capture <captureId>` Optic will prepare a sharable file that removes all bodies, and shape hashes json. I've also included Optic version, system stats (but nothing identifiable) and the optic.yml + ignore config. 

The only non-obvious part of this work is how to treat bodies for non-json thing ie -- a text/html response or the bytes for an image. Right now, they are just omitted, but we may want to fill them with some repeating bytes that compress well so we can simulate the memory/data load. Open to ideas and happy to defer this last choice @LouManglass @devdoshi @JaapRood

### DIFF
--- a/workspaces/local-cli/src/commands/debug/capture.ts
+++ b/workspaces/local-cli/src/commands/debug/capture.ts
@@ -77,7 +77,9 @@ export default class DebugCapture extends Command {
       events,
       configJson,
       ignoreRules,
-      interactions: cleanedInteraction,
+      session: {
+        samples: cleanedInteraction,
+      },
       opticVersion: pJson.version,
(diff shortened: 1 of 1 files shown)


JaapRood:
@acunniffe cool, useful tool! I'd defer the decision on non-json bodies and just omit them for now, and only revisit if we realise it's a problem.

PR #499, opened 2021-01-09
Speed up interaction iteration & learn workers 

acunniffe:
This PR addresses some new performance issues observed with exceedingly large captures, particularly those that contain a lot of non-api traffic ie (css, js, html, etc). 

The Capture Interaction Iterator is implemented in Node which means as we stream the interactions to various workers (diff, initial-bodies, trail-values), we're paying for garbage collection and the cost of serializing the avro files into our data structures. For a big webapp with an API on the same hostname, the breakdown of assets to API traffic can be in the order of 90/10. That means a lot of large bodies, loaded into memory, deserialized, then ignored, and GC'd. 

The changes within from @devdoshi allow the ignore filters to run before this expensive work takes place. Each interaction is partially deserialized, the path is checked, against the filter, and only fully deserialized should the filter return true. 

It's become noticeably faster to perform diffs, learn the bodies, and collect trail values. 

> **Caveat** -- there's a breaking change for past captures. None of our debug captures or debug scripts will be broken, but all captures from debugging data will have to be regenerated and users will need to recapture traffic. Since captures are transient we're not too worried about this, but we do need to figure out how best to communicate this to users. Probably makes sense to do it with the Optic 9 release later in the month. 

--- 

We also learned that Node's underlying file system access library (`libuv`) has max concurrent IO for file access which imposes constraints performance of concurrent reads. Note -- these maximums exist across node process boundaries. 5 independent diff or learner processes are all fighting for spots in the queue regardless of when they were started and what files they're accessing. 

After a lot of testing, we discovered that having the 5 API learners running in parallel is much less performant than having 1-2 workers due to less contention. The same is true for the diff, if you start 3 diffs at once, they all go slower, even if you have plenty of resources for the rust binaries. 

The fix that seems most workable and gets the learners to finish consistently is to limit the # of learn API workers to 1. In some scenarios this probably net-slower than iterating a few in parallel, but only at the smallest scales of traffic <10mbs. Even if it took a minute to learn you entire API, the workers won't hang, and they'll always complete without using all the disk read your physical machine can handle. 500 endpoints might take 3 mins, but you can do other stuff while it's happening and trust it to succeed. 

tl;dr This should work until the initial body learners get Rust-ified.


workspaces/cli-server/src/tasks/on-demand-initial-body.ts
```diff
@@ -72,7 +72,7 @@ export class OnDemandInitialBody {
     };
     const onExit = (code: number, signal: string | null) => {
       cleanup();
-      if (code !== 0) {
+      if (code !== 0 && code !== null) {
```
devdoshi, line 75: this function should be updated to have the correct argument types, if null is possible for `code`. It should check and log the signal argument or exit code. This function is duplicated in each of the tasks so those should be updated as well.

PR #501, opened 2021-01-11
Fix error propagation from diff-engine for errors occurring in diff compute

JaapRood:
While panics occurring on the main thread of the diff-engine were being propagated (e.g. panics in parsing the spec events stream), errors from worker tasks didn't propagated and cause the main thread to panic. Unless explicitly handled, we want these types of errors to stop further processing and make it all the way up to the UI, so the user can be informed about it.

This PR changes the diff-engine CLI's I/O and concurrency loop, keeping track of all tasks spawned and the `Err` these might in (through it panicking). As soon as any of these produces an error, the main thread panics and all further processing is suspended. We continue to monitor the correct completion of diff tasks when the stream of interactions has been exhausted.

More to follow about the Node.js side of this.


JaapRood:
@devdoshi while we're still in draft, it's probably wise for you to get acquainted with this new code and for it to be sanity checked. Feel free to ask any questions, or if it's all too much to digest we could sync up and go through the new mechanic.

workspaces/diff-engine/cli/src/main.rs
```diff
@@ -106,57 +109,105 @@ fn main() {
...
-          .expect("diffing of interaction should be successful");
-        //dbg!("got results");
-
-        if let Some((results, tags)) = results {
-          for result in results {
-            //dbg!(&result);
-            if let Err(_) = results_sender
-              .send(ResultContainer::from((result, &tags)))
+    loop {
+      let next_action = tokio::select! {
```
JaapRood, line 119: This running of `tokio::select!` in a loop is the core change of how we're diffing interactions one by one. We're doing multiple things asynchronously (reading interactions from `stdin`, computing the diff and writing results to `stdout`), this allows us to handle the next thing that's ready. Tokio polls these future at random, and in the case where that thing is happening on the same (main) thread (like the reading of interactions), the polling actually causes each future to progress. Spawning a task through Tokio causes progress to be made regardless.

workspaces/diff-engine/cli/src/main.rs
```diff
@@ -106,57 +109,105 @@ fn main() {
...
+                  // TODO: Find way to actually write error info
+                }
+              }
             }
+
+            drop(diff_task_permit);
+          });
+
+          // dbg!("Scheduling task", diff_tasks.len() + 1);
+          diff_tasks.push(diff_task);
```
JaapRood, line 176: This is a new essential part which allows the errors from the `diff_task` to be propagated. Before, we we'd just spawn the thread and let it run, with it's result never handled. Since these `select!`  handlers kind of work like a normal matching statement and are all handled on the same thread, we can write to `diff_tasks` while remaining accessible on the other branches as well.

workspaces/diff-engine/cli/src/main.rs
```diff
@@ -106,57 +109,105 @@ fn main() {
...
-            if let Err(_) = results_sender
-              .send(ResultContainer::from((result, &tags)))
+    loop {
+      let next_action = tokio::select! {
+        // scheduling a diff task upon new input
+        next_interaction = interaction_lines.next() => {
+          if let None = next_interaction {
+            // end of input, stop looping
+            break;
+          }
```
JaapRood, line 125: This the only place where we `break` out of this `select!` loop: when there's no more interactions to read on stdin, as it got closed. It's either that, or panicking.

workspaces/diff-engine/cli/src/main.rs
```diff
@@ -106,57 +109,105 @@ fn main() {
...
+        Some(res) = async {
+          if diff_tasks.is_empty() {
+            None
+          } else {
+            // dbg!("Waiting for a worker task to finish");
+            let (finished_diff_task_result, task_index, _) = future::select_all(&mut diff_tasks).await;
+            // dbg!("Task finished!", task_index);
+            diff_tasks.remove(task_index);
+
+            Some(finished_diff_task_result)
```
JaapRood, line 191: In case there are active `diff_tasks`, we create a new future that completes as soon as one of them completes. At that point we remove it from the list of diff tasks.

So what if a new `diff_task` is pushed to this Vec of `diff_tasks`? That's totally safe, as hat meant the other select branch matched, so this entire `async` block (a `Future`) will never be polled again, and so never make progress again, eliminating the chance of a race condition. The other way around works as well: if a diff task completes and we remove it from the vec (causing the items to the right of it to shift left), the branch that adds the task can't possibly make progress itself.

Wait.. 🤔. That also means that entire future of scheduling the interaction for diffing is dropped. So does that mean we can lose interactions if a diff task completes while it's being scheduled? 🔥 I think I identified a nasty bug here.

JaapRood: Looking at this some more, the bug is definitely there. Once we read the interaction from the input stream, we should only have synchronous code until it's moved to a task for processing, as other wise other futures in this `select!` get the chance to complete, and have the interaction be dropped.

In this case, there is one asynchronous action in between reading the interaction and moving it to the scheduled task where it's safe: acquiring the diffing permit (which limits the amount of diff tasks we can schedule, providing back-pressure). The fix for that seems fairly straightforward: acquire the permit _first_ and only then attempt to read from stdin. In case we _do_ acquire a permit but were still were waiting for a new interaction to be read while some another branch completes, the permit is dropped, which causes it to be released.

I'll give the a go!

JaapRood: Looking at it again: there was an issue with the order of acquiring permit vs reading interaction, but not the one I thought. 

There was _no_ risk of a diff task completing while waiting for a permit, as that happened in the _handler_ of the select branch: by the time that would execute, the other futures would have already been canceled. However, this meant the _entire_ main loop could be made to wait for a permit to become available. While not _quite_ a deadlock issue, we want to make sure this main select loop remains tight, probably with only synchronous handlers.

JaapRood: Resolved in https://github.com/opticdev/optic/pull/501/commits/3f0efa3ab8fa72a42c695494a0c4104294c2c370, by waiting for the permit to schedule a new diff task _before_ attempting to read an interaction. Makes the entire scheduling code synchronous, keeping the main loop tight.

workspaces/diff-engine/cli/src/main.rs
```diff
@@ -106,57 +109,110 @@ fn main() {
...
-      });
+      // panic when scheduling, diff tasks 
+      next_action.expect("essential worker task panicked");
     }
 
     drop(results_sender);
-    results_manager.await.unwrap(); // make sure the results manager is done flushing
+
+    let completing_diff_tasks = future::try_join_all(diff_tasks);
+    try_join!(completing_diff_tasks, results_manager).expect("essential worker task panicked");
```
JaapRood, line 215: If we've arrived at this point, `stdin` has been exhausted. While only awaiting `results_manager` _will_ prevent us from shutting down until all diff tasks are done (it won't complete until all `result_senders` have been dropped, of which each of a `diff_task` has one), there is _also_ the possibility of one of the diff tasks failing. Therefor, it's best to wait until they are all either completed, or one of these things fails.

JaapRood:
And that's the error propagation landed. Key was to handle daemon's in-process errors differently from errors in the spawned `diff-worker-rust`, by not crashing the daemon and actually propagating that to the UI side. @acunniffe the `debugger` statement in `CaptureContext` upon receiving an error is now triggered. For now it'll only be a generic 500-type error, which should be enough to instruct the user something didn't go right and to send us their `diff-engine-output.log` (saved to the diff folder, not included in the notification as it can be rather big). That should do for now, we can look at something more sophisticated separately.

With that, I'm pretty happy to call this ready for final review and merge!

JaapRood:
If checks pass, I'm happy for this be merged while I'm off for the night, especially as it's still behind a feature flag anyway. If not, and no other issues show up, I'll do so tomorrow morning.

JaapRood:
The one failing test seems to be unrelated, failing in #502 as well. Given that, I'm going ahead and merge this.

PR #503, opened 2021-01-13
Enable Rust diff engine for stable releases

JaapRood:
Exactly what it says on the tin!


JaapRood:
Made this draft for now, allowing us to do it some time soon, but also verify it's at least as stable as the ScalaJS implementation.

PR #506, opened 2021-01-14
Feature/capture diffing should use wasm

devdoshi:
We currently run a diff and coverage (which also runs a diff internally) on each captured batch of interactions. The diffing is meant to support notifying the user when there are diffs observed during `api start`/`api run *` and the coverage is only used when --collect-coverage is passed. This can have negative performance implications for large captures. This change allows users to specify --collect-diffs (default true) and --collect-coverage (default false) so they can choose whether this additional work is done. In addition, the diffing logic has been ported to use the wasm domain logic as part of the move away from our scalajs domain logic.


JaapRood (approved):
All good, ship it

PR #523, opened 2021-01-25
chore: update the demo build action

notnmeyer:
* `sync` with  `--delete` will remove any files in s3 that are not part of the set produced during the build
* invalidates the cloudfront dist after uploading files to s3


.github/workflows/deployDemo.yml
```diff
@@ -63,8 +65,9 @@ jobs:
             --acl=public-read \
             --metadata='{"Cache-Control":"no-store"}' \
             --cache-control=no-store
-          aws s3 cp ./demoRepo/workspaces/ui/build/ "s3://$BUCKET/" \
-            --recursive \
+          aws s3 sync ./demoRepo/workspaces/ui/build/ "s3://$BUCKET/" \
+            --delete \
             --exclude=index.html \
```
devdoshi, line 70: why do we exclude index.html btw?

notnmeyer: check out this section, https://github.com/opticdev/optic/blob/5258e839dd70563ec7d87d354ebb0958db881483/.github/workflows/deployDemo.yml#L63-L72

we first copy index.html and set a cache-control header, then we copy everything else _except_ the index.html. but now this has me wondering if this would delete index.html... but honestly, i dont even remember why we set cache-control in the first place....

notnmeyer: OH! we set cache-control and try and avoid unnecessary cloudfront invalidations. the header instructs the CDN to not cache it, and since the other resources are nonced, voila.

in any case, if we invalidate we can collapse that into one command—its what we wound up needing everywhere else.

devdoshi (approved):

PR #531, opened 2021-01-28
Use StreamExt and TryStreamExt to simplify diff-engine-cli's concurrency model

JaapRood:
The concurrency model for the diffing of interactions had gotten more complicated as we considered how to propagate and handle errors in the worker threads correctly. Through a `select!` mechanic, we manually tracked the result of each thread and raced it against all other concurrent processes. 

Turns out there's no need to do that manually anymore. `TryStreamExt::try_for_each_concurrent`, a trait Tokio turns out to implement now that `futures::Stream` is stable, does exactly what we want. It consumes items of a stream concurrently and allows for a limit, giving the back-pressure we're after. It also resolves to an error immediately if either the stream or predicate returns an error, another behaviour we're after.

And while it itself runs concurrently and not in parallel, there's nothing stopping us from spawning new tasks (and thus: threads) and forwarding the result of that.

Overall, this way of implementing makes the code a lot easier to understand and robust to change: a lot of the smaller subtle concerns and races are internal to the trait's implementation, where they belong.

### DIFF
--- a/workspaces/diff-engine/cli/src/main.rs
+++ b/workspaces/diff-engine/cli/src/main.rs
@@ -1,7 +1,7 @@
 use clap::{crate_version, App, Arg};
-use futures::future;
 use futures::try_join;
 use futures::SinkExt;
+use futures::{StreamExt, TryStreamExt};
 use num_cpus;
 use optic_diff_engine::diff_interaction;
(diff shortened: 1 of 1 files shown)


devdoshi (approved):
Looks good.

PR #551, opened 2021-02-03
use correct demo base url based on environment 

acunniffe:
We have both staging + production environments for the splash page and the demo. CSP rules are important to configure for security. Before these revisions, we were only ever pointing to demo.useoptic.com -- now the demo env matches the splash page  env.


notnmeyer (approved):
staging looks good

PR #555, opened 2021-02-04
Upgrade diff-engine to Tokio v1

JaapRood:
As it says on the tin. Tokio v1 will be maintained for at least 5 years, and v2 won't be released for at least 3 years. Required for some of the pending Rust work.

### DIFF
--- a/workspaces/diff-engine/cli/src/main.rs
+++ b/workspaces/diff-engine/cli/src/main.rs
@@ -71,16 +71,13 @@ fn main() {
 
   let spec_projection = Arc::new(SpecProjection::from(events));
 
-  let mut runtime_builder = tokio::runtime::Builder::new();
+  let mut runtime_builder = tokio::runtime::Builder::new_multi_thread();
   runtime_builder.enable_all();
-  runtime_builder.threaded_scheduler();
   if let Some(core_threads) = core_threads_count {
-    runtime_builder
-      .max_threads((core_threads as usize) * 2)
-      .core_threads(core_threads as usize);
+    runtime_builder.worker_threads(core_threads as usize);
   }
 
-  let mut runtime = runtime_builder.build().unwrap();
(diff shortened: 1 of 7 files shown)


workspaces/diff-engine/cli/src/main.rs
```diff
@@ -71,16 +71,13 @@ fn main() {
...
 
-  let mut runtime_builder = tokio::runtime::Builder::new();
+  let mut runtime_builder = tokio::runtime::Builder::new_multi_thread();
   runtime_builder.enable_all();
-  runtime_builder.threaded_scheduler();
   if let Some(core_threads) = core_threads_count {
-    runtime_builder
-      .max_threads((core_threads as usize) * 2)
-      .core_threads(core_threads as usize);
+    runtime_builder.worker_threads(core_threads as usize);
```
JaapRood, line 77: They got rid of the `max_thread` and `core_threads` because they were confusing. That they were!

JaapRood:
@notnmeyer any idea why the demo deploy didn't work? I'm exactly 0 caught up with the changes there.

JaapRood:
@notnmeyer gave me the green light in the a personal conversation: there's an issue with the demo release workflow and dealing with branch names that have forward slashes in them. Since this doesn't touch any of the WASM stuff, and thus the Demo, I'm happy to ignore and roll right on through.

notnmeyer:
```
task: docker build . -f Dockerfile_demo -t $DOCKER_REGISTRY/$REPO:$TAG
965
invalid argument "***.dkr.ecr.us-east-1.amazonaws.com/demo:chore/upgrade-tokio-rs-b121" for "-t, --tag" flag: invalid reference format
```

we'll need to fix the builds to replace characters that arent valid in docker image tags

devdoshi (approved):

PR #556, opened 2021-02-04
fix: sanitize git branch name for use in docker image tags

notnmeyer:
works on this branch,
image

only handles forward slashes for now, but easy to modify later.

this also replaces using  `Ilshidur/action-slack` with a task, `docker:slack-notifier`. doing the slack notification ourselves means that we can sanitize the tag within Task and not worry about exposing it back to the gha workflow for consumption in other steps.

### DIFF
--- a/.github/workflows/deployDemo.yml
+++ b/.github/workflows/deployDemo.yml
@@ -48,9 +48,7 @@ jobs:
           echo "DOCKER_REGISTRY=${{ steps.login-ecr.outputs.registry }}" >> $GITHUB_ENV
           echo "REPO=demo" >> $GITHUB_ENV
           echo "TAG=${BRANCH}-b${GITHUB_RUN_NUMBER}" >> $GITHUB_ENV
+          echo "USER=${GITHUB_ACTOR}" >> $GITHUB_ENV
       - run: task ui:demo:ci
-      - uses: Ilshidur/action-slack@689ad44a9c9092315abd286d0e3a9a74d31ab78a # v2.1.0
         env:
           SLACK_WEBHOOK: ${{ secrets.BUILD_BOT_SLACK_WEBHOOK_URL }}
-        with:
-          args: "{{ GITHUB_ACTOR }}: `{{ REPO }}:{{ TAG }}` was pushed to our Docker registry."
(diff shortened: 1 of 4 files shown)


.github/workflows/deployDemo.yml
```diff
@@ -47,7 +47,9 @@ jobs:
       - run: |
           echo "DOCKER_REGISTRY=${{ steps.login-ecr.outputs.registry }}" >> $GITHUB_ENV
           echo "REPO=demo" >> $GITHUB_ENV
-          echo "TAG=${BRANCH}-b${GITHUB_RUN_NUMBER}" >> $GITHUB_ENV
+          TAG=$(TAG=${BRANCH}-b${GITHUB_RUN_NUMBER} task docker:clean-tag)
```
notnmeyer, line 50: i really dont love this. the only reason to have set the "corrected" tag back to the gha env is to reference the artifact in the slack notification that follows. if we did the slack notification ourselves in Task we could drop this entirely.

notnmeyer:
marinating on this for a bit in case i come up with something better

bin/clean-docker-tag
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+set -eu
+# replaces character that aren't valid in Docker image tags with hyphens
+echo $1 | sed 's/\//-/g'
```
notnmeyer, line 4: 🍖

devdoshi (approved):

workspaces/ui/Taskfile.yml
```diff
@@ -16,18 +16,26 @@ tasks:
...
-    - docker push $DOCKER_REGISTRY/$REPO:$TAG
-    - |
-      if [[ "$TAG" =~ ^release-b ]]
-      then
-        docker tag $DOCKER_REGISTRY/$REPO:$TAG $DOCKER_REGISTRY/$REPO:latest
-        docker push $DOCKER_REGISTRY/$REPO:latest
-      fi
+    vars:
+      TAG:
+        sh: echo $(../../bin/clean-docker-tag $TAG)
```
notnmeyer, line 29: the var is set at the task level to not pollute the global scope. invoking any task through the top-level taskfile will load all included taskfiles and add global scope vars/env.

Issue #545, opened 2021-02-03
Support for websocket passthrough


dyc3:
I'm working on a project that has an API and also accepts websocket connections. Currently, the websocket connections don't get passed through and the console output gets spammed with
```
Error: socket hang up
    at connResetException (internal/errors.js:609:14)
    at Socket.socketOnEnd (_http_client.js:459:23)
    at Socket.emit (events.js:326:22)
    at endReadableNT (_stream_readable.js:1223:12)
    at processTicksAndRejections (internal/process/task_queues.js:84:21) {
  code: 'ECONNRESET'
}
```

acunniffe:
Hey @dyc3 I followed up on Discord. If you're interested in fixing it yourself, here are some places to look and instructions on getting the repo running. If it's quicker to provide a reproducible example we can work off of, please upload one here: 

here's where I'd look
https://github.com/opticdev/optic/blob/develop/workspaces/cli-shared/src/httptoolkit-capturing-proxy.ts

we use a debugging proxy called mockttp under the hood
https://github.com/httptoolkit/mockttp

developer setup instructions for the optic repo (let me know if you need help)
https://github.com/opticdev/optic/blob/develop/Developer-setup.md

Seems like this will help quite a few users so I'm around to help you get this all fixed

acunniffe:
And just for my curiosity -- your API serves both Rest endpoints and WebSockets?

dyc3:
That's correct, through the same express server. This is the repo I'm working with, for reference: https://github.com/dyc3/opentogethertube

acunniffe:
Makes sense, I'll be pulling this tonight and taking a look. If you find any more out today lmk here or on discord. I'll circle back with my own learnings

acunniffe:
Hey @dyc3  -- looks like an issue within Mockttp  -- I was able to reproduce it with a simpler setup 
https://github.com/acunniffe/repo-socket-issue

Chatting now with the maintainers and seeing if there's anything obvious they want us to try.

acunniffe:
@dyc3  -- there's a fix underway in mockttp, I'll track it here and share it back when it's ready -- hopefully, we can get it incorporated and tested within the next few days. Want to make sure it gets a solid QA pass since it touches the proxy

acunniffe:
@dyc3  -- we got some changes through at mockttp: [See working version](https://www.loom.com/share/688e7737352f4d3aa19cf376ea9cec3d)

[...]

PR #560, opened 2021-02-05
added websocket passthrough for forwarding proxy

acunniffe:
## Why
Many APIs have a combination of REST endpoints and some websockets. We don't  help teams document their WebSocket messages (yet), but those should at least be passed through the proxy

- Recent Issue: https://github.com/opticdev/optic/issues/545
- This has been around for a while and I remember it being reported before in a user session

## What
Tim from Mockttp released v 1.1.0 which allows us to forward websockets through the proxy. In previous versions, this worked in transparent mode, but not forwarding mode. 

## Validation
* [x] CI passes
* [x] Verified against https://github.com/acunniffe/repo-socket-issue
* [x] Verified against test projects -- default forwarding behavior for normal interactions not affected.

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -173,6 +173,15 @@
       "contributions": [
         "review"
       ]
+    },
+    {
+      "login": "pimterry",
+      "name": "Tim Perry",
+      "avatar_url": "https://avatars.githubusercontent.com/u/1526883?v=4",
+      "profile": "https://tim.fyi/",
+      "contributions": [
+        "code"
+      ]
     }
   ],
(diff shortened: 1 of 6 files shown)


dyc3:
I've tested it out on https://github.com/dyc3/opentogethertube and it works!

devdoshi (approved):

acunniffe:
Awesome @dyc3 I'll slice a release now. Looking forward to seeing an API spec for OpenTogetherTube -- let me know if you need any help. It's such a cool project I asked my sister to use it with me next week

PR #569, opened 2021-02-08
Report regression suite status to slack

notnmeyer:
## Why
we previously switched the regression suites to run less often, on `release` and on a schedule. we're not currently surfacing the results of these tests, which is predominantly relevant for the scheduled run.

## What
sends a pass/fail regression suite status message to slack

## Validation
* [x] test run works
image


notnmeyer:
looks like windows will need a conditional to handle the env var expansion, 

```
Run curl -X POST -H 'Content-type: application/json' --data '{"text":""}' ***
invalid_payload
```

devdoshi (approved):
Looks good. Might be worth adding the manual trigger? Also I love how much simpler and faster the slack notification probably is haha

notnmeyer:
> Might be worth adding the manual trigger?

once merged, https://github.com/opticdev/optic/pull/569/commits/82737eba667feb4ff9cdae4dbe6938b0f4dbff68 should give us a button in the ui, 
image

PR #574, opened 2021-02-09
Implement Spec Loader that can produce single event stream from spec changes distributed in multiple files

JaapRood:
## Why
This is a first step towards a vertical implementation of specs split-up into multiple files: being able to read all the files and stitch them back together. 

## What
This adds chunk splitting logic to `diff-engine` (which we should really consider renaming to `optic-core` or something similar), following a Git-commit-like linked-list approach, where each spec change points to it's parent. They're all read into memory at random, so a single in-order event stream can be assembled. 

As soon as there's more than a single spec change pointing to the same parent, it's considered a conflict and the assembly ends. Any automatic merging that might be possible (assuming partitioning by endpoint, for example), or projections that describe found conflicts, is out of the scope of this PR.

It guarantees:

- ordering of spec changes
- no reliance on filename 
- backwards-compatibility with existing `.optic/api/specification.json`

It's also intentionally not applied at _all_ places that read the spec from disk, only those for the main local workflow of authoring spec changes. A couple of other places, like generating the OpenAPI spec and retrieval by the GitHub Diffbot, are ignored for now.

All of this is enabled through a new feature flag `OPTIC_ASSEMBLED_SPEC_EVENTS`, which is set to `false` for all types of release for now. 

## Validation
* [x] CI passes
* [x] UI can render assembled spec
* [x] UI can render legacy spec
* [x] UI can trigger diffs and show results


workspaces/diff-engine/cli/src/main.rs
```diff
@@ -24,10 +24,28 @@ fn main() {
...
+        .short("d")
+        .takes_value(false)
+        .help("SPEC_PATH is expected to be a directory"),
+    )
+    .group(
+      ArgGroup::with_name("spec-type")
+        .args(&["use-spec-file", "use-spec-dir"])
+        .multiple(false)
+        .required(false),
+    )
```
JaapRood, line 48: This allows both the `assembly` and `diff` commands of the CLI to specify whether the passed in spec should be expected to be a single file, or a directory that needs to be assembled. Putting them together in an `ArgGroup` allows to make sure they're not used in a conflictingly.

workspaces/diff-engine/cli/src/main.rs
```diff
@@ -55,115 +83,170 @@ fn main() {
...
+      SpecPathType::DIR => {
+        let spec_chunk_events = streams::spec_chunks::from_api_dir(&spec_path)
+          .await
+          .expect("should be able to find spec event chunks in a folder");
+
+        streams::spec_events::from_spec_chunks(spec_chunk_events)
+          .await
+          .unwrap() // TODO: report on these errors in a more user-friendly way (like for single spec files)
+      }
+    };
```
JaapRood, line 119: No matter whether we're diffing or assembling, reading the spec from the filesystem is the same. Now inside of main Tokio runtime future, as the reading in both cases is now done with Async I/O.

workspaces/diff-engine/cli/src/main.rs
```diff
@@ -55,115 +83,170 @@ fn main() {
...
+  for spec_event in spec_events {
+    if let Err(_) = results_sink.send(spec_event).await {
+      panic!("could not stream event result to stdout"); // TODO: Find way to actually write error info
+    }
+  }
+  results_sink
+    .get_mut()
+    .write_u8(b']')
+    .await
+    .expect("could not write array start to stdout");
```
JaapRood, line 238: Writing valid JSON instead of JSONL this way was chosen to make implementation and backwards-compatibility easier. Given that downstream it's never really consumed partially, the main benefit of JSONL is lost anyway. That said, there's no reason why it couldn't support both through use of a flag, so we can move to JSONL in the future.

That would give us an opportunity to move some of this direct writing to stdout to something like `JsonArrayItemSink`, something I deemed outside of the scope of this PR (plenty of time was spent on writing these results as is).

workspaces/diff-engine/lib/index.js
```diff
@@ -16,6 +16,51 @@ function spawn({ specPath }) {
...
+
+  // surface any errors in execution on the output stream
+  assembleProcess.then(
+    (childResult) => {},
+    (childResult) => {
+      output.emit('error', new DiffEngineError(childResult));
+    }
+  );
+  return output;
+}
```
JaapRood, line 61: This is the addition to diff-engine's thin Node.js layer, allowing consuming code to call `readSpec({ specDirPath })` without having to worry about spawning binaries. Made sure to get the `stdio` right this time, with the results the only thing we want back-pressure and `stderr` being forwarded to the parent process. That simplified the return value a lot to, with a single stream being enough.

workspaces/diff-engine/src/events/endpoint.rs
```diff
@@ -1,13 +1,13 @@
...
+use serde::{Deserialize, Serialize};
 
 use crate::state::endpoint::{
   PathComponentId, RequestId, RequestParameterId, ResponseId, ShapedBodyDescriptor,
   ShapedRequestParameterShapeDescriptor,
 };
 
-#[derive(Deserialize, Debug)]
+#[derive(Deserialize, Debug, PartialEq, Serialize)]
 pub enum EndpointEvent {
```
JaapRood, line 11: All `SpecEvent` variants and their underlying structs became `PartialEq` (so we can express `specA == specB`, so far mostly useful in testing) and `Serialize` (so we can write the events to stdout).

workspaces/diff-engine/src/events/mod.rs
```diff
@@ -43,7 +47,7 @@ impl Event for SpecEvent {
 }
 
 impl SpecEvent {
-  pub fn from_file(filename: &str) -> Result<Vec<SpecEvent>, EventLoadingError> {
+  pub fn from_file(filename: impl AsRef<Path>) -> Result<Vec<SpecEvent>, EventLoadingError> {
```
JaapRood, line 50: Cool little bit I've learned: `impl AsRef<Path>` is used throughout `std::fs` and `tokio::fs` to express paths. Implemented from the most basic `&str` all the way up to `Path` and `PathBuf` itself.

workspaces/diff-engine/src/events/spec_chunk.rs
```diff
@@ -0,0 +1,332 @@
...
+use cqrs_core::Event;
+use serde_json;
+use std::convert::TryFrom;
+
+#[derive(Debug)]
+pub enum SpecChunkEvent {
+  Root(RootChunkEvent),
+  Batch(BatchChunkEvent),
+  Unknown(UnknownChunkEvent),
+}
```
JaapRood, line 11: The `Root` chunk is what represent the `specification.json` and is considered to be in-order and consistent with itself. The `Batch` chunk is what we consider the new spec change files will read as (more info on the logic below). `Unknown` is to make sure we have a way to detect where any new spec change files don't meet the required standard, and ignored in further assembly of the spec events.

workspaces/diff-engine/src/events/spec_chunk.rs
```diff
@@ -0,0 +1,332 @@
...
+  for event in intermediate_events {
+    if let SpecEvent::RfcEvent(RfcEvent::BatchCommitStarted(e)) = event {
+      return Err("Chunk cannot include nested BatchCommitStarted events");
+    } else if let SpecEvent::RfcEvent(RfcEvent::BatchCommitEnded(e)) = event {
+      return Err("Chunk cannot include nested BatchCommitEnded events");
+    }
+  }
+
+  Ok((batch_id, parent_id))
+}
```
JaapRood, line 165: This is where the logic that determines what a valid `Batch` chunk is. Made sure we have descriptive errors here, so we can debug it easily.

workspaces/diff-engine/src/events/spec_chunk.rs
```diff
@@ -0,0 +1,332 @@
...
+      Self::Root(RootChunkEvent {
+        id: String::from("root"),
+        name,
+        events,
+      })
+    } else {
+      match BatchChunkEvent::try_from((name, events)) {
+        Ok(batch_chunk) => Self::Batch(batch_chunk),
+        Err((msg, name, events)) => Self::Unknown(UnknownChunkEvent { name, events }),
+      }
```
JaapRood, line 103: For anything that's not the root chunk (so anything else but specification.json), we attempt to try and parse it as a `BatchChunkEvent` and otherwise store it as `UnknownChunkEvent`.

workspaces/diff-engine/src/events/spec_chunk.rs
```diff
@@ -0,0 +1,332 @@
...
+  }
+
+  #[test]
+  pub fn constructs_valid_batch_chunks() {
+    let valid_batch_events = serde_json::from_value::<Vec<SpecEvent>>(json!([
+      {"BatchCommitStarted": {"batchId": "batch-1", "parentId": "root", "commitMessage": "Add Request and Response for GET /todos" }},
+      {"RequestAdded": { "requestId": "request_1","pathId": "path_1","httpMethod": "GET" }},
+      {"ResponseAddedByPathAndMethod": { "responseId": "response_1", "pathId": "path_1", "httpMethod": "GET", "httpStatusCode": 200 }},
+      {"BatchCommitEnded": { "batchId": "batch-1" }}
+    ])).unwrap();
```
JaapRood, line 240: Here are tests covering all of the `Batch` parsing logic, using snapshots to verify each case produces the error we expect it to.

workspaces/diff-engine/src/streams/spec_chunks.rs
```diff
@@ -0,0 +1,89 @@
...
+        SpecChunkLoaderError::Other("Filename could not be converted to valid UTF-8")
+      })?;
+      let is_root = name == "specification.json";
+
+      chunks.push(SpecChunkEvent::from((name, is_root, spec_events)));
+    }
+  }
+
+  Ok(chunks)
+}
```
JaapRood, line 51: The SpecChunk stream is where all the logic lives around reading spec-change files from a directory and mapping them to `SpecChunkEvent` instances. Those paying attention will notice how this doesn't return a `impl Stream<Item=SpecChunkEvent>`, but that actually turned out be just tricky enough for me to decide it's something to consider in the future instead of now. 

However, it _does_ still belong in this module, as all of `streams` is not imported for `diff-engine-wasm`, which is important as Tokio + wasm isn't a love story yet. Reading files async from the file system this way is great, but requires that Tokio runtime, so best isolated away from the rest of the code.

workspaces/diff-engine/src/streams/spec_chunks.rs
```diff
@@ -0,0 +1,89 @@
...
+
+    let chunks = from_api_dir(String::from(api_dir_path.to_str().unwrap()))
+      .await
+      .unwrap();
+
+    assert_eq!(chunks.len(), 3);
+    dbg!(chunks
+      .iter()
+      .map(|chunk| (chunk.event_type(), chunk.name(), chunk.len()))
+      .collect::<Vec<_>>());
```
JaapRood, line 87: Decided not to snapshot here quite yet, basically as the spec-changes I've assembled by hand for this aren't both valid batches yet (which I do intend them to be). Going to revisit that once we can programmatically produce these spec-change files.

JaapRood:
@devdoshi @acunniffe I've added bits of review to the code to hopefully make it a bit easier to digest. I think it's probably worth keeping in mind that is behind a feature flag, and that it's a first step in a process where probably lots will change soon. Hoping to get this PR merged fairly swiftly because of that, using future PR's to address any issues / further refinements.

JaapRood:
I've double checked whether this was hidden behind feature flags, which are disabled in all published environments. Given that and the fact this is a first PR in a series of many, I'm going to merge this as soon as the tests pass. Any comments from review will be addressed in future PRs.

PR #576, opened 2021-02-09
Add Rust diff engine developer setup

svanderbleek:
## Why

Help new contributors get started with the Rust codebase

## What

Adding a section to the readme similar to existing sections for node

## Validation

setup worked for me and seems reliable


JaapRood:
Thanks @svanderbleek! Can confirm that from a technical point of view these instructions would get those new to Rust going, with the most important one (installing rustup) right at the top where it belongs. Visual Studio Code is probably also a decent editor to point to for those that don't have their favourites for Rust yet. 

I _would_ however suggest installing [rust-analyzer](https://marketplace.visualstudio.com/items?itemName=matklad.rust-analyzer) as the extension for it to support Rust. While still experimental, it's already a must with a project of even this modest size in both performance and features. Worth a shot, if you haven't already :)

LouManglass (approved):
Thanks for the update!

PR #580, opened 2021-02-10
PowerShell doc tip

jordaniza:
If you're using optic from within a non-bash environment (such as PowerShell), $PORT may need to be replaced with %PORT%

## Why

## What

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


LouManglass:
Hi @jordaniza,

Sorry for not following up earlier: we've been working on PR #582 , to resolve this issue once and for all. The goal is to avoid requiring folks to have to think about which platform they're on, at least as much as reasonably possible.

We'll be in touch, thanks for digging into this with us!

LouManglass (approved):
We didn't go with our other experiment for now, thanks again for the contribution @jordaniza . I'm going to merge this in and then add some changes later that I had stashed.

PR #582, opened 2021-02-10
[Experiment] use cross-spawn for optic tasks

acunniffe:
## Why
We've found a surprising mix of shells, operating systems on teams -- making it hard to share commands in their optic.yml files. We expected on the individual team level there'd be homogeneity. There is not. The lesson of the ages. 

## What
To make it easier to collaborate with people using different shells this PR introduces [cross-spawn](https://www.npmjs.com/package/cross-spawn) under the hood when you run `api run <task>`

In theory, this will allow commands in `optic.yml` files to be more portable/robust across environments so it's easier for teams to use the tool together. 

## Validation
* [ ] CI passes
* [ ] @LouManglass tests cross platform and known use cases, approves 
* [ ] @devdoshi approves 
* [ ] it's been 3 days -- Feb 13th

### DIFF
--- a/workspaces/cli-shared/src/command-session.ts
+++ b/workspaces/cli-shared/src/command-session.ts
@@ -1,7 +1,9 @@
-import { ChildProcess, spawn, SpawnOptions } from 'child_process';
+import { ChildProcess, SpawnOptions } from 'child_process';
 import { EventEmitter } from 'events';
 import treeKill from 'tree-kill';
 import { developerDebugLogger } from './index';
+//@ts-ignore
+import { spawn } from 'cross-spawn';
 
 export interface ICommandSessionConfig {
   command: string;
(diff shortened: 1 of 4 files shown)


acunniffe:
@notnmeyer -- this is an example of a spot where a PR build would really help. It'd be nice to test this in a low stakes way w/o having to have people build it themselves.

acunniffe:
@jordaniza -- curious what you think of this PR. 

Is portability of commands a desirable property? 
Are you comfortable with how https://www.npmjs.com/package/cross-spawn translates under the hood? 
When / How would this break? 


Here's the guide to pull and build locally https://github.com/opticdev/optic/blob/develop/Developer-setup.md

acunniffe:
This was fun to explore. we're putting to the side for now though

PR #588, opened 2021-02-11
feat: parallelize ui builds

notnmeyer:
## Why
the UI builds are some of the slower workflows because they build the world first. before digging into not building the world, we can build demo and apidocs in parallel.

## What
builds demo and apidocs in parallel

## Validation
* [ ] CI passes

### DIFF
--- a/workspaces/ui/Taskfile.yml
+++ b/workspaces/ui/Taskfile.yml
@@ -1,5 +1,7 @@
 version: "3"
 
+output: prefix
+
 tasks:
   clean:
     desc: Removes UI's build directories
--- a/workspaces/ui/Taskfile.yml
+++ b/workspaces/ui/Taskfile.yml
@@ -23,7 +23,6 @@ tasks:
     - task: ci:setup
     - task: ci:build
 
-
   #
   # docker namespace
   #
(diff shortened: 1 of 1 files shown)


notnmeyer:
running this a couple times to get a feel, but so far its only a minor improvement on the existing build times.

edit: yeah, doing this isnt appreciably faster than building serially

image

notnmeyer:
did some local testing

this branch,
```
➜ time TAG=(date +%s) task ui:clean workspaces:clean ui:ci
...
________________________________________________________
Executed in   95.40 secs   fish           external 
   usr time  282.06 secs  257.00 micros  282.06 secs 
   sys time   51.48 secs  819.00 micros   51.48 secs 
```

develop,
```
➜ time TAG=(date +%s) task ui:clean workspaces:clean ui:ci
...
Executed in  142.66 secs   fish           external 
   usr time  275.17 secs  200.00 micros  275.17 secs 
   sys time   47.44 secs  460.00 micros   47.44 secs 
```

so this looks more like what i expected. considering the vast time difference between local runs and CI i think its fair to say there's some resource contention happing in CI. so i think there's potential for this to be faster in CI but not sure we can do much to influence that at the moment.

notnmeyer:
meh

PR #589, opened 2021-02-11
Add test path for empty spec against default http interaction

svanderbleek:
## Why

Improve test coverage of diff_interaction by adding a test for a case that isn't currently covered, that of diffing an empty spec projection (from empty events) against a generic http interaction. What is covered already are spec projections that always have events.

## What

Add a default http interaction for ease of use as there is no way to create ArbitraryData due to its private fields. There is no need to manually create since None values for the fields are fine for the path under test, so a derive default will do. 

Add the test using the defaults for spec projection and http interaction against saved snapshot of result.

## Validation

* cargo test works and runs the test case
* the snapshot result looks correct to me

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ node_modules
 .task/
 *.swp
 .DS_Store
+.vscode
 
 #Added by cargo
 
--- a/workspaces/diff-engine/src/lib.rs
+++ b/workspaces/diff-engine/src/lib.rs
@@ -1,12 +1,12 @@
 #![allow(dead_code, unused_imports, unused_variables)]
 
-mod events;
-mod interactions;
-mod projections;
-mod protos;
-mod queries;
-mod shapes;
-mod state;
+pub mod events;
+pub mod interactions;
+pub mod projections;
+pub mod protos;
+pub mod queries;
+pub mod shapes;
+pub mod state;
 
 #[cfg(feature = "streams")]
 pub mod streams;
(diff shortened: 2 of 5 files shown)


.gitignore
```diff
@@ -5,6 +5,7 @@ node_modules
 .task/
 *.swp
 .DS_Store
+.vscode
```
JaapRood, line 8: Please add editor specific files to your local [global git ignore](https://stackoverflow.com/questions/7335420/global-git-ignore) list, so we can prevent this from becoming a pit.

workspaces/diff-engine/src/lib.rs
```diff
@@ -1,12 +1,12 @@
...
-mod queries;
-mod shapes;
-mod state;
+pub mod events;
+pub mod interactions;
+pub mod projections;
+pub mod protos;
+pub mod queries;
+pub mod shapes;
+pub mod state;
```
JaapRood, line 9: We've very intentionally not made all internal modules public, as that grows the API surface area to whatever consumes this crate. Instead, we like to be intentional about what we expose and keep internals just that: internal.

JaapRood (commented):
While this is code that compiles, I'm not sure if default implementations make any sense semantically. For example, an `HttpInteraction` with an empty string as it's method describes a variant we don't actually see out there in the wild, so forces the rest of the logic to then handle and support a use case that doesn't really exist.

PR #591, opened 2021-02-12
platform: version info

LouManglass:
## Why
Currently there's no insight into how well we support various platforms and our response is very reactive. It's also difficult to assess the scope of impact of platform-specific issues. More insight allows us to better prioritize response and be proactive to environmental patterns.

## What
platform, version, arch are now noted.

## Validation
* [ ] PR Rules pass (CI passes, review approved)
* [x] Manually tested by me


workspaces/ui/src/utilities/useAnalyticsHook.js
```diff
@@ -72,6 +73,9 @@ export function AnalyticsContextStore({ children, specService }) {
       const clientContext = {
         clientAgent: clientAgent,
         clientId: clientId,
+        //platform: os.platform(),
```
devdoshi, line 76: Should there be any changes to this file?

workspaces/cli-server/src/analytics.ts
```diff
@@ -19,6 +19,9 @@ import {getOrCreateAnonId} from "@useoptic/cli-config/build/opticrc/optic-rc";
 const packageJson = require('../package.json');
 
 const clientId = `local_cli_${packageJson.version}`;
+const platform = os.platform();
```
devdoshi, line 22: Looks good. See also: `workspaces/cli-scripts/src/errors.ts` for what we send to Sentry

workspaces/ui/src/utilities/useAnalyticsHook.js
LouManglass: Thanks, good catch. No. Reverted in upcoming push.

workspaces/cli-server/src/analytics.ts
LouManglass: Thanks, it appears that Sentry already grabs some of this, so I added the host's release information as well.

workspaces/cli-scripts/src/errors.ts
```diff
@@ -29,11 +29,13 @@ export function trackWithSentry({
   const hostPlatform = process.platform;
   const hostCpuCount = Os.cpus().length.toString();
   const hostMemorySize = Os.totalmem().toString();
+  const hostRelease = Os.release();
```
devdoshi, line 32: I forgot we've duplicated this code in `cli-server/src/errors.ts` as well - would you mind applying it there too?

devdoshi (approved):
Looks good - just one small comment left

devdoshi:
clear to merge!

workspaces/cli-scripts/src/errors.ts
LouManglass: Looks like that's already done, awesome!

PR #593, opened 2021-02-13
WIP initial interface for changelog projection endpoint queries

svanderbleek:
## Why

Want to answer added and changed endpoint queries over changelog projection

## What

ChangelogProjection
ChangelogQuery#added
ChangelogQuery#changed

## Validation

Would like feedback if this interface is a good or bad approach


devdoshi:
The skeleton looks good. As far as the particular approach for the projection representation (projection.changed, projection.added), that depends on your requirements/goal so I can't comment on that yet.

PR #594, opened 2021-02-15
domain conflicts should be surface-able

devdoshi:
## Why
When multiple developers make spec changes, merging can result in domain conflicts; commands that were valid in one git branch can yield events that would conflict with equally valid events from another git branch. This is a classic split-brain distributed systems scenario. We need to detect these conflicts so they can be surfaced to the developer.

## What
This PR introduces an additional projection that tracks how many internal ids correspond to one external spec concept. For example, if 2 concurrent edits add the same endpoint to the spec, they'll have different UUIDs but correspond to the same HTTP path and method. This PR only goes as far as endpoints, but can be extended as needed based on our conflict-resolution UX requirements. 

## Validation
* [ ] CI passes


workspaces/diff-engine/src/projections/conflicts.rs
```diff
@@ -0,0 +1,345 @@
+use crate::events::{EndpointEvent, Event, ShapeEvent, SpecEvent};
```
devdoshi, line 1: 🍖

workspaces/diff-engine/tests/domain-conflict-scenarios-generated.rs
```diff
@@ -0,0 +1,114 @@
+// DO NOT EDIT! This file was generated via workspace-scripts/diff-engine/generate-tests
+use insta::assert_debug_snapshot;
+use optic_diff_engine::{diff_interaction, HttpInteraction, SpecEvent, SpecProjection};
+use petgraph::dot::Dot;
+use serde_json::json;
+
+#[test]
+pub fn scenario_9949a8cc69a0063a70f5ae98672d8c23a1068a2b2d08f8bb7b9bbd968ec29f0d() {
```
devdoshi, line 8: The intent of these being content-addressable is so we can add useful descriptions separately (that reference the content address) and know whether the input data has changed since that would potentially invalidate the snapshot or description.

workspaces/diff-engine/tests/domain-conflict-scenarios-generated.rs
```diff
@@ -0,0 +1,114 @@
...
+  let interactions: Vec<HttpInteraction> = serde_json::from_str(&interactions_string).unwrap();
+  for interaction in interactions {
+    let label = format!("scenario_9949a8cc69a0063a70f5ae98672d8c23a1068a2b2d08f8bb7b9bbd968ec29f0d__interaction_{}__diffs" , interaction.uuid.clone());
+    let diffs = diff_interaction(&spec_projection, interaction);
+    assert_debug_snapshot!(label, diffs);
+  }
+}
+#[test]
+pub fn scenario_20e9bda8afc0279258c5ac2c01b82c437aa1976e1eba23cb92096f7434d9316c() {
+  let spec_file_path = "/Users/dev/work/optic/workspaces/diff-engine/tests/fixtures/domain-conflict-scenarios/events-20e9bda8afc0279258c5ac2c01b82c437aa1976e1eba23cb92096f7434d9316c.json";
```
devdoshi, line 36: I'll fix these paths

devdoshi:
To review the graph snapshots, I've been doing something like:
```
$ find . -iname "*graph*" | xargs -I % echo "x=%; tail -n +5 \$x | dot -Tpng > \$x.png" > x.sh
$ sh x.sh
$ open *.png
```
but in the meantime I'm designing a better snapshot approval flow

workspaces/diff-engine/src/projections/conflicts.rs
JaapRood: Looks good, exactly how and where I'd expect a projection like this to exist. The approach tracking node id's and domain id's hard to fault as well, in line with how the other projections work, too!

JaapRood (commented):
This is some legit good stuff, really showing off the power of the CQRS model as well as Rust being a useful place to implement it.

@devdoshi the only things I would nit pick are a couple debug hygiene things, like leftover `dbg!` statements. And perhaps I would question whether the test generation script warrants the introduction of TS to the package (vs vanilla JS, or even Rust) and whether it should live in `diff-engine/tests/fixtures` or something, but totally happy for you to consider that and then make the call however you see fit.

devdoshi:
@JaapRood I'm not sure where the best place for it is but it's fine for now and we'll be revisiting it soon as part of the testing story

PR #595, opened 2021-02-15
sourceme should not leak env

devdoshi:
closes https://github.com/opticdev/issues/issues/55
## Why
We observed that sourceme.sh scripts would leak the entire host env when the .env file is "empty" (only whitespace after skipping lines starting with `#`)

## What
The only change in behavior is when the .env file is "empty". Previously it would result in exporting the entire host env, and now it will emit a warning message

## Validation
* [x] CI passes
* [x] Verified manually when .env is empty
* [x] Verified manually when .env is missing

### DIFF
--- a/sourceme.sh
+++ b/sourceme.sh
@@ -8,10 +8,29 @@ alias apidev="OPTIC_DAEMON_ENABLE_DEBUGGING=yes OPTIC_DEVELOPMENT=yes OPTIC_UI_H
 alias apistage="OPTIC_DAEMON_ENABLE_DEBUGGING=yes $OPTIC_SRC_DIR/workspaces/local-cli/bin/run"
 alias cidev="$OPTIC_SRC_DIR/workspaces/ci-cli/bin/run"
 alias agentdev="$OPTIC_SRC_DIR/workspaces/agent-cli/bin/run"
+
+optic_export_env() {
+  ENV_FILE=$1
+  if [ ! -f "$ENV_FILE" ]
+  then
+    echo "could not find env file '$ENV_FILE'"
+    exit 1
+  fi
+
+  VARS_TO_EXPORT=$(grep -v '^#' "$ENV_FILE")
+  VARS_WITHOUT_WHITESPACE=$(echo $VARS_TO_EXPORT | sed s/\n// | sed s/\ //)
+  if [ "$VARS_WITHOUT_WHITESPACE" = "" ]
+  then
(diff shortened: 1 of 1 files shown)


sourceme.sh
```diff
@@ -8,26 +8,45 @@ alias apidev="OPTIC_DAEMON_ENABLE_DEBUGGING=yes OPTIC_DEVELOPMENT=yes OPTIC_UI_H
...
+  ENV_FILE=$1
+  if [ ! -f "$ENV_FILE" ]
+  then
+    echo "could not find env file '$ENV_FILE'"
+    exit 1
+  fi
+
+  VARS_TO_EXPORT=$(grep -v '^#' "$ENV_FILE")
+  VARS_WITHOUT_WHITESPACE=$(echo $VARS_TO_EXPORT | sed s/\n// | sed s/\ //)
+  if [ "$VARS_WITHOUT_WHITESPACE" = "" ]
```
notnmeyer, line 22: fine with this but `-z $BLAH` also works

devdoshi:
Now that the function is extracted, you may test its behavior by creating an arbitrary file and running something like:
```
$ echo "" > .empty.env
$ source sourceme.sh
$ optic_export_env .empty.env 
'.empty.env' is empty. You might want to put some variables there
```

```
$ rm .env
$ source sourceme.sh
$ optic_export_env .fake.env
could not find env file '.fake.env'
```

sourceme.sh
```diff
@@ -8,26 +8,45 @@ alias apidev="OPTIC_DAEMON_ENABLE_DEBUGGING=yes OPTIC_DEVELOPMENT=yes OPTIC_UI_H
...
+optic_export_env() {
+  ENV_FILE=$1
+  if [ ! -f "$ENV_FILE" ]
+  then
+    echo "could not find env file '$ENV_FILE'"
+    exit 1
+  fi
+
+  VARS_TO_EXPORT=$(grep -v '^#' "$ENV_FILE")
+  VARS_WITHOUT_WHITESPACE=$(echo $VARS_TO_EXPORT | sed s/\n// | sed s/\ //)
```
notnmeyer, line 21: not sure i'm following why we strip newlines?

notnmeyer (approved):
couple comments, but looks good to me

sourceme.sh
devdoshi: The grep can emit blank lines, but perhaps the echo will remove them. Perhaps the better check is to see if there are any alphanumeric characters at all?

PR #596, opened 2021-02-15
Fix sourceme leaking env when empty 2

devdoshi:
## Why

## What

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


devdoshi:
ignore

PR #599, opened 2021-02-16
Feature/changelog projection endpoint queries

svanderbleek:
## Why

Some queries and commands will not require the full graph, these can be answered with simpler data structures. This is a projection for answering questions about changes to endpoints, fields, etc starting with added endpoints.

## What

Added query can get list of added endpoints since a given time period.

## Validation

* cargo test passses 
* looks like it captures added endpoints in ergast and todos example spec


acunniffe:
@svanderbleek -- I sent ya an invite for office hrs! Dev and I will discuss this with your Friday

PR #602, opened 2021-02-17
feat: one-off npm package builds to s3

notnmeyer:
## Why
we want to be able to quickly build and share bugfix/testing/alpha/whatever builds without cutting a proper releases.

## What
adds tasks and a GHA workflow to produce a new set of NPM packages and upload them to S3 where they can be distributed outside of NPM's registry.

```
➜ task -l | grep release
* release:side-channel:run: 			Build a side-channel release for @useoptic/cli and publish it to s3
```

if `DRY_RUN=false` the packages will be uploaded to s3 and posted to our Slack releases channel. from there, they can be installed,

```
➜ npm install https://optic-side-channel-staging.s3.amazonaws.com/9.0.11-alpha.test/cli-9.0.11-alpha.test.tgz
...

➜ node_modules/@useoptic/cli/bin/run --help
API CLI from Optic. Document and test your APIs as you build them.
VERSION
  @useoptic/cli/9.0.11-alpha.test darwin-x64 node-v14.14.0
```

## Triggering the workflow

once merged to develop, there will be a "Run workflow" button where you're able to select the branch to build from and then make it so.

## Validation
* [x] CI passes, https://github.com/opticdev/optic/runs/1938524220?check_suite_focus=true
* [x] We've vetted the packages, https://opticteamworkspace.slack.com/archives/C01059SBJGK/p1613659633005600?thread_ts=1613589589.011700&cid=C01059SBJGK


taskfiles/release/side-channel/Taskfile.yml
```diff
@@ -0,0 +1,87 @@
...
+    cmds:
+      - task: bump-version
+        vars:
+          S3_HTTPS_URL: https://{{.BUCKET}}.s3.amazonaws.com
+          VERSION: "{{.VERSION}}"
+      - node workspaces/scripts/publish.js
+      - task: reset-working-tree
+
+  # reset working tree because package.json's and README.md's were updated with `yarn bump`
+  reset-working-tree:
```
devdoshi, line 81: is it simpler to just `git checkout -- .` ?

notnmeyer: this is only for us running it locally, in CI its fine. i intentionally tried to limit it to only the files being modified to avoid blowing up unrelated working changes you may have wanted to keep

devdoshi: gotcha yeah that makes sense

devdoshi (approved):
nice!

taskfiles/release/side-channel/Taskfile.yml
notnmeyer: and maybe orth mentioning, i was initially trying to glob with something like, `git checkout workspaces/*/package.json`. it worked in my shell, but wasnt working in task. so i fell back to this

PR #604, opened 2021-02-19
build tooling should have tools to make one-off artifacts simple

devdoshi:
## Why
Publishing artifacts to s3 requires modifying the package.json files to point to a s3 url. These scripts give the building blocks to achieve this in future work

## What
Nothing affecting the product changes. This adds scripts available to developer tooling

## Validation
* [x] CI passes


notnmeyer (approved):
👍

PR #606, opened 2021-02-19
feat: add missing parts of our side-channel release process

notnmeyer:
## Why
there's a couple things that the side-channel release process doesnt do

* package builds we dont load any feature flags
* diff-engine binaries are not built and distributed in a similar fashion as the npm packages

## What
* adds an optional input to the workflow disptach interface to select an env file, which is then used to set feature flags during builds.
* uploads diff-engine binaries to the side-channel bucket in the same location as the the release workflow
* cross-compiles the windows build on ubuntu so i can be liberated from having to think about windows build hosts
    * note that this means we're targeting `x86_64-pc-windows-gnu` target rather than `x86_64-pc-windows-msvc` 
* ci workflow looks like, 
     image

## Gotchas
* https://github.com/opticdev/optic/pull/606#discussion_r582248262

## Validation
* [x] CI passes
* [x] Packages poked at and vetted by humans

## Example Run
* the run: https://github.com/opticdev/optic/actions/runs/597016999
* the announcement: https://opticteamworkspace.slack.com/archives/C017H0H8M9A/p1614197337009000
* npm packages:
    ```
    ➜ aws s3 ls s3://optic-side-channel-production/9.0.11-alpha.b60/
    2021-02-24 10:47:21       4362 agent-cli-9.0.11-alpha.b60.tgz
    2021-02-24 10:47:21       5035 analytics-9.0.11-alpha.b60.tgz
    2021-02-24 10:47:21       3823 ci-cli-9.0.11-alpha.b60.tgz
    2021-02-24 10:47:21      50388 cli-9.0.11-alpha.b60.tgz
    2021-02-24 10:47:21       3219 cli-client-9.0.11-alpha.b60.tgz
    2021-02-24 10:47:21       8310 cli-config-9.0.11-alpha.b60.tgz
    2021-02-24 10:47:20       4895 cli-scripts-9.0.11-alpha.b60.tgz
    2021-02-24 10:47:21      24394 cli-server-9.0.11-alpha.b60.tgz
    2021-02-24 10:47:21      34952 cli-shared-9.0.11-alpha.b60.tgz
    2021-02-24 10:47:21       1880 client-utilities-9.0.11-alpha.b60.tgz
    2021-02-24 10:47:21       7143 diff-engine-9.0.11-alpha.b60.tgz
    2021-02-24 10:47:21     165644 diff-engine-wasm-9.0.11-alpha.b60.tgz
    2021-02-24 10:47:21       1340 saas-types-9.0.11-alpha.b60.tgz
    2021-02-24 10:47:21      14111 snapshot-tests-9.0.11-alpha.b60.tgz
    2021-02-24 10:47:21    9023858 ui-9.0.11-alpha.b60.tgz
    ```
* diff-engine:
    ```
    ➜ aws s3 ls s3://optic-side-channel-production/dists/optic_diff/v9.0.11-alpha.b60/
    2021-02-24 10:39:50    1918063 optic_diff-v9.0.11-alpha.b60-linux.tar.gz
    2021-02-24 12:08:40     974102 optic_diff-v9.0.11-alpha.b60-macos-aarch64.tar.gz <-- look at that timestamp :|
    2021-02-24 10:38:56     994601 optic_diff-v9.0.11-alpha.b60-macos.tar.gz
    2021-02-24 10:39:50    3736431 optic_diff-v9.0.11-alpha.b60-win64.tar.gz
    ```


Taskfile.yml
```diff
@@ -33,7 +33,16 @@ tasks:
     desc: Build Yarn workspaces
     deps: ["workspaces:setup"]
     cmds:
-      - yarn wsrun --stages --report --fast-exit --exclude-missing ws:build
+      - |
```
devdoshi, line 36: The requirement is to distribute the provided env to each workspace package that uses the env. Here's what is in the existing workflow:
```echo "OPTIC__NPM__VERSION=${{ needs.release_logic.outputs.version }}" >> "$FLAGS_FILE"
        if [[ -r $FLAGS_FILE ]]; then
          echo "Feature flags found for $RELEASE_CHANNEL, staging for build and runtime."
          cp $FLAGS_FILE .env
          cp $FLAGS_FILE workspaces/agent-cli/.env
          cp $FLAGS_FILE workspaces/ci-cli/.env
          cp $FLAGS_FILE workspaces/local-cli/.env
          cp $FLAGS_FILE workspaces/diff-engine/.env
          cp $FLAGS_FILE workspaces/ui/.env.production.local
        fi
```

I can make that change

workspaces/diff-engine/scripts/install.js
```diff
@@ -9,6 +8,8 @@ dotenv.config({
   path: envPath,
 });
 
+const { install } = require('../lib');
```
devdoshi, line 11: @JaapRood is this ok (changing the behavior by changing the require order so the env has been loaded) or would you do it a different way?

JaapRood: In general I'd say don't rely on the order of requires, but at the same time it's obvious to me that `dotenv` is what I would consider part of the program's "prelude". So yeah, it's okay, I wouldn't know of a better way (if any at all).

.github/workflows/release-side-channel.yml
```diff
@@ -35,7 +43,79 @@ jobs:
...
+      - name: Publish
+        env:
+          BUCKET: optic-side-channel-production
+        run: task release:side-channel:diff-engine
+
+  announce:
+    runs-on: ubuntu-latest
+    needs:
+      - npm_packages
+      - rust_binaries
```
notnmeyer, line 105: gha's macos-11.0 availability is still really bad. if we remove `needs:rust_binaries` we could report completion significantly faster most of the time. each diff-engine build matrix uploads its own builds, so by the time npm_packages is done, diff-engine binaries will already be uploaded with the exception of macos-aarch64.

notnmeyer: haha, pretty sure i should make this change: https://github.com/opticdev/optic/actions/runs/597016999

mac0s-11.0 job was queued for an hour and then randomly failed before starting :\

notnmeyer: for now, arm builds not being available for an hour or so after everything else will be a thing to keep in mind 🙃

notnmeyer:
as mentioned in basecamp, i cracked building aarch64 on x86 macos 10.15. im not going to include that in this PR though. once this lands, i'll follow up with a new PR

devdoshi (approved):
nice.

PR #607, opened 2021-02-19
Exp update notifier

LouManglass:
## Why

Notify users of updates to the local-cli, to keep new features flowing.

## What

- local-cli index.ts now includes update-notifier, and it is a new package dependency (don't forget the yarn install)

## Validation

* [ ] CI passes
* [ ] Verified in staging
* [ ] Review

### DIFF
--- a/workspaces/local-cli/package.json
+++ b/workspaces/local-cli/package.json
@@ -44,6 +44,7 @@
     "slugify": "^1.4.0",
     "strip-ansi": "^6.0.0",
     "tslib": "^1",
+    "update-notifier": "^5.0.1",
     "url-join": "^4.0.1",
     "uuid": "^8.0.0",
     "which": "^2.0.2"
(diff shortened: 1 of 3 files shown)


devdoshi (approved):
Looks good, was helpful to read https://www.npmjs.com/package/update-notifier on how to test it. Just curious if you checked out https://github.com/oclif/plugin-update (https://oclif.io/docs/releasing.html#autoupdater). This seems way simpler though so I like it.

LouManglass:
> Looks good, was helpful to read https://www.npmjs.com/package/update-notifier on how to test it. Just curious if you checked out https://github.com/oclif/plugin-update (https://oclif.io/docs/releasing.html#autoupdater). This seems way simpler though so I like it.

I did not,  gave it a quick perusal. Yeah, I think the current update-notifier solution is a bit simpler and I don't think we need much more/are constraining ourselves unnecessarily with this approach, and I'm inclined to stick with it as well (at least for now).

PR #608, opened 2021-02-20
Docs: Fix the yarn command in the Readme

batisteo:
## Why
Because `yarn add global` would install the `global` package.

## What
Just a tiny doc change.

## Validation
* [ ] CI passes
* [ ] Verified in staging


acunniffe:
Good catch @batisteo ! Thanks for this. I swear I mess this up every other time I do it.

PR #612, opened 2021-02-25
Implement persisting new spec changes as batch commits in their own files in Rust

JaapRood:
## Why
Once we started working on persisting spec changes in their own files, it became apparent pretty quickly that the command handlers for writing spec changes should live in Rust. That's where the core business logic is, where we can reason about spec chunks, the interface with spec files, etc. 

## What
This PR adds a new `commit` subcommand to the `optic_diff` binary, which is used by the `cli-server` to implement a new endpoint that accepts an array of commands and commit message to be saved as a new batch of spec changes, in its own file. The UI calls this endpoint with the generated commands based on the approved suggestions from the Diff Review UI.

The aim of this PR is to only implement a golden path use case (behind an opt-in feature flag), so we can establish a common base from which to address a bunch of known issues:

- Only the commands actively produced by the UI are accepted and handled.
- Works only for projects with existing specs (a `specification.json` considered as the root).
- A WASM interface to do this in browser (required for example / private / debug sessions) is still missing.

As with #574, these changes are disabled by default and hidden behind the `REACT_APP_OPTIC_ASSEMBLED_SPEC_EVENTS` (UI) and `OPTIC_ASSEMBLED_SPEC_EVENTS` (everything else) feature flags.

## Validation

- [x] Spec changes can be persisted to existing specs
- [x] Newly saved spec changes can be read by UI 
- [x] Newly saved spec changes can be diffed
- [x] Old way of persisting spec changes is the default, unless enabled through the feature flags.
- [x] CI passes


workspaces/diff-engine/src/commands/endpoint.rs
```diff
@@ -0,0 +1,1030 @@
...
+        )]
+      }
+
+      EndpointCommand::SetPathParameterShape(command) => {
+        validation.require(
+          !validation.path_component_is_root(&command.path_id),
+          "path id can not be root to set path parameter shape",
+        )?;
+        validation.require(
+          validation.path_component_id_exists(&command.path_id),
```
devdoshi, line 336: without looking at the scala implementation, seems like this is missing a check for whether it is a path parameter as opposed to a static path component. Can you summarize whether it was done 1:1 and any differences we might expect?

workspaces/diff-engine/src/commands/rfc.rs
```diff
@@ -0,0 +1,262 @@
...
+    let mut projection = HistoryProjection::from(initial_events);
+
+    let command: RfcCommand = RfcCommand::append_batch_commit(String::from("second commit"));
+
+    let new_events = projection
+      .execute(command)
+      .expect("new command should be applicable to initial projection");
+    assert_eq!(new_events.len(), 1);
+    // TODO: snapshot the results instead, once we figure out how insta redactions can control
+    // for the automatically generated batch id.
```
devdoshi, line 205: i'd follow the pattern we used in scala, injecting an idgenerator. for testing, use sequential ids; otherwise use uuidv4 or something else appropriate for the likelihood of collisions

workspaces/diff-engine/src/commands/shape.rs
```diff
@@ -0,0 +1,770 @@
...
+
+#[derive(Deserialize, Debug, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct AddShape {
+  pub shape_id: ShapeId,
+  pub base_shape_id: ShapeId,
+  pub name: String,
+}
+
+impl AddShape {
```
devdoshi, line 47: Is this for testing?

workspaces/diff-engine/src/events/shape.rs
```diff
@@ -214,3 +219,130 @@ impl Event for FieldRemoved {
...
+}
+
+impl From<shape_commands::AddShape> for ShapeAdded {
+  fn from(command: shape_commands::AddShape) -> Self {
+    Self {
+      shape_id: command.shape_id,
+      base_shape_id: command.base_shape_id,
+      name: command.name,
+      parameters: ShapeParametersDescriptor::empty_dynamic(),
+      event_context: None,
```
devdoshi, line 289: What is the plan for CommandContext/EventContext? We will definitely need something like it soon

workspaces/diff-engine/src/commands/rfc.rs
JaapRood: It's what I tried initially, but the Rust ownership model makes that pretty clunky. It's exactly the kind of thing Insta's redactions feature is for, I just couldn't figure out how its redaction paths work. Given the amount of code here I didn't feel like going any deeper into that instantly, but seems like the most straightforward way of dealing with this.

workspaces/diff-engine/src/events/shape.rs
JaapRood: Plan is to land this PR so we can then deal with details like this in parallel. More specifically for this case: accept `CommandContext` as part of commands coming from the UI (or generate our own for commands generated from the Rust domain itself), then map them near 1:1 to `EventContext` as part of the command -> event conversion.

workspaces/diff-engine/src/commands/shape.rs
JaapRood: Nope, used in handling of `AddPathParameter` command to add a shape for the new path parameter. Instead of creating an `Event` straightaway, I've opted to create command first and convert that to an event. That way  all the `From<Command> for Event` implementations remain the single source of truth on how stuff is mapped. Will especially come in handy once we start working with command + event contexts as well.

workspaces/diff-engine/src/commands/endpoint.rs
JaapRood: Port was done 1:1 with the Scala code open to the side, implementing all the same validation rules. Any omissions should be because we don't support certain commands or state descriptors anymore, like certain commands or the various shape providers found in some of the shape commands.

In this particular case, I can see how that validation is missing, but can also report that it's missing on the Scala side as well. https://github.com/opticdev/optic-core/blob/master/core/optic/shared/src/main/scala/com/useoptic/contexts/requests/RequestsAggregate.scala#L65-L70

workspaces/diff-engine/src/commands/rfc.rs
devdoshi: Having all of these transformations be deterministic during testing is worth it. I'm hesitant to invest in redactions because it adds friction to the testing process. snapshots should be a low friction thing and it's important to be able to reason about the output more clearly and there are a lot of ids you wouldn't want to all end up as `[uuid]`. I'm not clear on what is clunky about the id generator approach. As far as I can tell, it would be another member of the command handling aggregate's struct (I also note you didn't wrap the projections in a more semantic struct yet) and could be concisely represented as an iterator of strings.

workspaces/diff-engine/src/commands/shape.rs
devdoshi: ok that's helpful. I think it may change slightly if we make the id generator change

workspaces/diff-engine/package.json
```diff
@@ -18,6 +18,7 @@
   "types": "lib/index.d.ts",
   "dependencies": {
     "@iarna/toml": "^2.2.5",
+    "@useoptic/diff-engine-wasm": "9.0.11-beta.0",
```
devdoshi, line 21: not expecting it to change as part of this pr, but we should fix the dependencies here when we reorganize the packages too

workspaces/diff-engine/src/commands/rfc.rs
JaapRood: We're going to follow this up in a future PR. There's some friction from redactions in snapshots, but possibly also in the refactored version. We'll have to try and figure out what's best, but we'll leave that out of the scope of this first base one.

JaapRood:
@devdoshi not much has changed since your last review. Given the aim to land this and create a common version to further develop from, I'm mostly looking for a sanity check, especially to make sure the changes are hidden properly behind the feature flags. As long as we're still aligned on the general approach and architecture, we should leave any smaller details to follow-up pull requests ☺️.

JaapRood:
@devdoshi any idea why the node_packages check is failing?

devdoshi:
> @devdoshi any idea why the node_packages check is failing?

I believe it is because of diff-engine depending on diff-engine-wasm and diff-engine-wasm depending on diff-engine @JaapRood

JaapRood:
@devdoshi ahh you're totally right, thanks for the fix! I only quickly looked at the npm package dependencies, without taking into account that the Cargo dependency also exists. Your fix is a pragmatic one that will work for now, and we can fix it more structurally in the aforementioned renaming / restructuring.

JaapRood:
After careful and experimental review of this new behaviour being hidden behind a feature flag properly, and taking into the account the earlier reviewing that happened while in draft, I'm ready to merge this PR. Anything coming out of retroactive review can be addressed through follow-up PRs.

PR #615, opened 2021-02-27
Fix for optional field interpreters 

acunniffe:
## Why
Users reported an issue where optional fields, presented with an unmatched shape, would yield a diff, but never provide suggestions. 

The expected behavior is a suggestion when field: [A] is presented with B are to:

a) change the type to [B]
b) make the type of the field the union [A, B]

## What
There was a branch in the control flow that prevented optional fields to return suggestions, this has been mediated by creating a better separation of concerns in the interpreters 

## Validation
* [x] CI passes
* [x] Unit tests pass
* etc...


acunniffe:
Side channel build in progress: https://github.com/opticdev/optic/actions/runs/605636920

PR #613, opened 2021-02-25
feat[diff-engine]: build macos aarch64 from macos 10.15

notnmeyer:
## Why
macos-11.0 on GHA has really long queued times and has for more than a month. 

## What
* this compiles diff-engine's aarch64 build on x86 macos 10.15
* we can now wait until all diff-engine builds are complete before announcing the url in slack
 
## Validation
* [x] CI passes, https://github.com/opticdev/optic/actions/runs/600494220/workflow
* the aarch64 build runs on my M1 macbook air

PR #617, opened 2021-03-01
fix: update regerssion suites workflow to cross-compile diff-engine

notnmeyer:
## Why
in https://github.com/opticdev/optic/pull/613 i changes to the `diff-engine:build-macos` to enable cross-compiling aarch64 from macos 10.15 hosts. those changes broke the regression suite due to a couple missing env vars. https://opticteamworkspace.slack.com/archives/C017H0H8M9A/p1614599060000100

## Validation
* [x] https://github.com/opticdev/optic/actions/runs/611573338

### DIFF
--- a/.github/workflows/regression-suites.yml
+++ b/.github/workflows/regression-suites.yml
@@ -44,10 +44,8 @@ jobs:
           echo "RUSTUP_HOME=$HOME/.rustup" >> $GITHUB_ENV
           echo "CARGO_HOME=$HOME/.cargo" >> $GITHUB_ENV
       - name: Install test dependencies
-        shell: bash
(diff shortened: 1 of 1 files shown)


.github/workflows/regression-suites.yml
```diff
@@ -14,29 +15,27 @@ jobs:
       fail-fast: false
       matrix:
         platform:
-          - os: windows-latest
```
devdoshi, line 17: I think we still want to run the regression on windows, right?

notnmeyer: oh yeah, definitely. will fix 👍

devdoshi (approved):

PR #620, opened 2021-03-03
Allow persistence to a single or multiple spec change files through Rust

JaapRood:
## Why
We've identified the need to be able to ship Rust being in charge reading, mutating and persisting specs _without_ also shipping the splitting up of spec changes. The latter has further reaching implications on other parts of the system the roll-out of which needing to be tested, vetted and controlled carefully.

## What
This PR adds the ability for Rust to save new batch commits as part of the original `specification.json` when there aren't any additional spec change files yet (once there are, `specification.json` is assumed to be immutable). The `cli-server` uses this ability to toggle this ability, controlled through a feature flag.

## Validation
* [ ] CI passes
* [ ] New batches are appended to the specification file when splitting file feature is disabled.
* [ ] New batches can't be appended to the specification file once other spec change files exist.

### DIFF
--- a/workspaces/diff-engine/cli/src/main.rs
+++ b/workspaces/diff-engine/cli/src/main.rs
@@ -73,6 +73,15 @@ fn main() {
             .value_name("COMMIT_MESSAGE")
             .takes_value(true)
             .help("The commit message describing the commands as a whole"),
+        )
+        .arg(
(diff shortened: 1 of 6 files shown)


JaapRood:
Closing as I opened from the wrong fork, prevents the checks from passing.

PR #621, opened 2021-03-03
Oas generator should use listEvents path

acunniffe:
## Why
With changes to spec persistence in testing, it's important to make sure the OAS Generator can continue to read the proper event stream. By consuming events from spec service interface, as opposed to reading from the file system, we can ensure compatibility with the upcoming changes.  

## What
Using listEvents() instead of reading the events from the file system for OAS projection 

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


JaapRood (changes requested):
While this works, I reckon we can cut all that cruft of setting up a HTTP service and use the `diff-engine` interface directly, like the endpoint implementation does. Fewer things to go wrong, and probably less code. Let me know if you want me to have a go at that.

acunniffe:
@JaapRood  -- good point. I've ported over -- is there a preferred way / utility to turn the Readable stream into a string?

workspaces/local-cli/src/commands/generate/oas.ts
```diff
@@ -86,3 +90,11 @@ export async function emit(
...
   };
 }
+
+function getStream(stream: any) {
+  return new Promise((resolve) => {
+    const chunks: any = [];
+    stream.on('data', (chunk: any) => chunks.push(Buffer.from(chunk)));
+    stream.on('end', () => resolve(Buffer.concat(chunks).toString()));
+  });
+}
```
JaapRood, line 100: Given the circumstances of `OasProjectionHelper` needing to do its own serdes and only accepting a string, this is fine. However, with the use of `.on`, and no removing of listeners, this will leak memory (a bit). 

While you get get all fancy with EventEmitters, the thing you can do in Node now is use a `Readable` as an `AsyncIterator`, allowing you to write this and have it take care of all of that for you:

```ts
const chunks: any = [];
for await (let chunk of stream) {
  chunks.push(Bufer.from(chunk));
}
return Buffer.concat(chunks).toString()
```

As a bonus, also that also propagates the stream's errors nicely!

JaapRood (commented):
Nice! Identified a small memory leak with hopefully an easy enough fix.

JaapRood (commented):
Awesome, looking great! Would retroactively approve if that was something Github would let you do 😬.

PR #623, opened 2021-03-04
Allow persistence to a single or multiple spec change files through Rust

JaapRood:
## Why
We've identified the need to be able to ship Rust being in charge reading, mutating and persisting specs _without_ also shipping the splitting up of spec changes. The latter has further reaching implications on other parts of the system the roll-out of which needing to be tested, vetted and controlled carefully.

## What
This PR adds the ability for Rust to save new batch commits as part of the original `specification.json` when there aren't any additional spec change files yet (once there are, `specification.json` is assumed to be immutable). The `cli-server` uses this ability to toggle this ability, controlled through a feature flag.

## Validation
* [x] CI passes
* [x] New batches are appended to the specification file when splitting file feature is disabled.
* [x] New batches can't be appended to the specification file once other spec change files exist.
* [x] Split-off spec changes can be added to a single specification.json. 
* [x] Appending to the single specification file is the default behaviour


.env.beta
```diff
@@ -1,4 +1,5 @@
 OPTIC_ASSEMBLED_SPEC_EVENTS=false
+OPTIC_SPLIT_SPEC_EVENTS=false
```
JaapRood, line 2: New flag, as we still want the be able to turn the assembly (reading) and splitting (writing) on and off individually. In fact, `OPTIC_ASSEMBLED_SPEC_EVENTS` is actually only a flag for using the Rust implementation to read specs and handle commands, so perhaps a rename is in order.

workspaces/cli-server/src/routers/spec-router.ts
```diff
@@ -239,6 +239,7 @@ ${events.map((x: any) => JSON.stringify(x)).join('\n,')}
         const events = DiffEngine.commit(commands, {
           specDirPath: req.optic.paths.specDirPath,
           commitMessage: payload.commitMessage,
+          appendToRoot: !isEnvTrue(process.env.OPTIC_SPLIT_SPEC_EVENTS),
```
JaapRood, line 242: Using the flag here to keep reading from the env as close to the edge of the system as possible.

workspaces/diff-engine/src/commands/rfc.rs
```diff
@@ -188,7 +188,7 @@ mod test {
   pub fn can_handle_append_batch_command() {
     let initial_events: Vec<SpecEvent> = serde_json::from_value(json!([
       {"PathComponentAdded": {"pathId": "path_1","parentPathId": "root","name": "todos"}},
-      {"BatchCommitStarted": {"batchId": "batch_1","parentId": "root", "commitMessage": "initial commit"}},
+      {"BatchCommitStarted": {"batchId": "batch_1", "commitMessage": "initial commit"}},
```
JaapRood, line 191: Pragmatic (or perhaps lazy) way of verifying that batches without a `parentId` are assigned the previously ended commit (or root, if none at all). See the changes to `HistoryProjection`.

workspaces/diff-engine/src/projections/history.rs
```diff
@@ -68,33 +68,35 @@ impl HistoryProjection {
...
-        .expect("expected parent commit to have a corresponding node");
-
+    let parent_batch_id = parent_batch_id.unwrap_or_else(|| {
       self
-        .graph
-        .add_edge(parent_node_index, node_index, Edge::IsParentOf);
+        .get_commit_id(&self.find_last_batch_commit_index().unwrap())
+        .unwrap()
+        .clone()
+    });
```
JaapRood, line 76: This is an essential change to fix a bug I observed while testing the migration of single-file to multi-file spec. With it, the batch commits inside of `specification.json` also become part of this graph, which in turn allows batch chunks from other files to refer to these batch commits (instead of only being able to refer to the specification file globally, which is what the bug was). The "root" commit, for the purposes of this projection,  becomes basically every event before a first batch commit is observed. 

The added bonus of having the batches embedded in a specification.json part of this projection, is that they can now be walked and rendered in any views of this as well.

workspaces/diff-engine/src/projections/spec_events.rs
```diff
@@ -45,7 +46,7 @@ impl SpecAssemblerProjection {
       .root_chunk
       .ok_or_else(|| SpecAssemblerError::RootChunkRequired)?;
 
-    let mut current_chunk_id = root_chunk.id.clone();
+    let mut current_chunk_id = root_chunk.last_batch_id().clone();
```
JaapRood, line 49: Batch commits from outside chunks can append to the "specification.json" by referring to the last batch commit it contains.

devdoshi (approved):

PR #627, opened 2021-03-08
Write EventContext for each event with new Rust command handler

JaapRood:
## Why
We ported the command handling to the Rust domain implementation, but had so far ignored the writing of the `eventContext` for each event. We'll definitely want to make sure this contextual information gets captured for specs and achieve parity with the Scala implementation.

## What
The `commit` CLI command now accepts two additional arguments: `--client-id` and `--client-session-id`, which the endpoint to commit batch commands provides. Using a new `SpecCommandHandler` aggregate on the Rust side, the batch id and time of creation get added as well.

## Validation
* [ ] CI passes
* [x] Client id and Client session Id from Node make it into each new event's eventContext.
* [x] Time and batch information is persisted as part of each event context.
* [x] Resulsting spec is readable by both domain implementations.

### DIFF
--- a/workspaces/diff-engine/cli/src/commit.rs
+++ b/workspaces/diff-engine/cli/src/commit.rs
@@ -1,0 +1,185 @@
+use super::events_from_chunks;
+use chrono::Utc;
+use clap::{App, Arg, ArgMatches, SubCommand};
+use futures::StreamExt;
+use optic_diff_engine::streams;
+use optic_diff_engine::Aggregate;
+use optic_diff_engine::CommandContext;
+use optic_diff_engine::{RfcCommand, SpecCommand, SpecCommandHandler};
+use optic_diff_engine::{SpecChunkEvent, SpecEvent};
+use std::path::Path;
+use std::process;
+use tokio::io::{stdin, stdout};
+use uuid::Uuid;
+
+pub const SUBCOMMAND_NAME: &'static str = "commit";
+
+pub fn create_subcommand<'a, 'b>() -> App<'a, 'b> {
+  SubCommand::with_name(SUBCOMMAND_NAME)
+    .about("Commits results of commands received over stdin as a new batch commit")
+    .arg(
+      Arg::with_name("commit-message")
+        .short("m")
+        .required(true)
+        .value_name("COMMIT_MESSAGE")
+        .takes_value(true)
+        .help("The commit message describing the commands as a whole"),
+    )
+    .arg(
+      Arg::with_name("append-to-root")
+        .long("append-to-root")
+        .required(false)
+        .takes_value(false)
+        .help("Append new batch commit to the root spec file, instead of a new spec change file"),
+    )
+    .arg(
+      Arg::with_name("client-session-id")
+        .long("client-session-id")
+        .required(false)
+        .value_name("CLIENT_SESSION_ID")
+        .takes_value(true)
+        .default_value("unknown-session")
+        .help("The session id of the client requesting the commands to be committed"),
+    )
+    .arg(
+      Arg::with_name("client-id")
+        .long("client-id")
+        .required(false)
+        .value_name("CLIENT_ID")
+        .takes_value(true)
+        .default_value("anonymous")
+        .help("Unique id of the client the commands to be committed"),
+    )
+}
+
+pub async fn main<'a>(
+  command_matches: &'a ArgMatches<'a>,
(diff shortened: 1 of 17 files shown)


workspaces/diff-engine/cli/src/main.rs
```diff
@@ -6,19 +6,18 @@ use num_cpus;
...
 use std::cmp;
-use std::path::Path;
 use std::process;
 use std::sync::Arc;
 use tokio::io::{stdin, stdout};
 use tokio::sync::mpsc;
 
+mod commit;
+
 fn main() {
```
JaapRood, line 21: Moved the `commit` subcommand logic to it's own module, planning to do the same for `diff` as well. Logic for both is non-trivial and requires some command specific structs, so using `clap`'s ability to modularise seems like a good thing to get into at this point.

workspaces/diff-engine/cli/src/commit.rs
```diff
@@ -0,0 +1,185 @@
...
+        .help("The session id of the client requesting the commands to be committed"),
+    )
+    .arg(
+      Arg::with_name("client-id")
+        .long("client-id")
+        .required(false)
+        .value_name("CLIENT_ID")
+        .takes_value(true)
+        .default_value("anonymous")
+        .help("Unique id of the client the commands to be committed"),
```
JaapRood, line 51: With manual testing, instantly ran into requiring these contexts being a pain, so defined default values for them. We _can_ still require them at other layers if we want users to always provide them, but the CLI should only be developer used at the moment.

workspaces/diff-engine/src/commands/mod.rs
```diff
@@ -63,17 +68,105 @@ impl std::error::Error for SpecCommandError {}
...
+  fn execute_on(self, command_handler: &SpecCommandHandler) -> Result<Self::Events, Self::Error> {
+    let mut events = command_handler.spec_projection.execute(self)?;
+    let event_context = EventContext::from(command_handler.command_context.clone());
+    for event in &mut events {
+      event.with_event_context(event_context.clone());
+    }
+    Ok(events)
+  }
+}
+
```
JaapRood, line 169: Ended up only wrapping this on a surface-level right now and assigning the events with context once they have been created. I took at approach as I ran into the issue of having to propagate that command context to nested handlers and projections, which would have taken way more refactoring for a use case that we're not actually hitting at the moment (using command context in command handling).

Once we do require using the command context in the command handling logic, this should give us a good starting off point, regardless. Then, we'd also have something to actually test and measure whether how we've implemented it makes sense.

workspaces/diff-engine/src/events/mod.rs
```diff
@@ -102,3 +106,33 @@ impl From<avro_rs::Error> for EventLoadingError {
...
+        .created_at
+        .with_timezone(&Local)
+        .to_rfc3339(),
+    }
+  }
+}
+
+pub trait WithEventContext {
+  fn with_event_context(&mut self, event_context: EventContext);
+}
```
JaapRood, line 138: Trait implemented by each event so that we can set their event context generically.

workspaces/diff-engine/src/events/mod.rs
```diff
@@ -102,3 +106,33 @@ impl From<avro_rs::Error> for EventLoadingError {
...
+impl From<CommandContext> for EventContext {
+  fn from(command_context: CommandContext) -> Self {
+    Self {
+      client_id: command_context.client_id,
+      client_session_id: command_context.client_session_id,
+      client_command_batch_id: command_context.client_command_batch_id,
+      created_at: command_context
+        .created_at
+        .with_timezone(&Local)
+        .to_rfc3339(),
```
devdoshi, line 131: learned something new today
```
fn parse_from_rfc3339(s: &str) -> ParseResult<DateTime<FixedOffset>>[−]
Parses an RFC 3339 and ISO 8601 date and time string such as 1996-12-19T16:39:57-08:00, then returns a new DateTime with a parsed FixedOffset.

Why isn't this named parse_from_iso8601? That's because ISO 8601 allows some freedom over the syntax and RFC 3339 exercises that freedom to rigidly define a fixed format.```

devdoshi (approved):

workspaces/diff-engine/src/events/mod.rs
JaapRood: I learned the exact same thing about a week ago, after looking to `to_iso8601` 😅

PR #629, opened 2021-03-09
flags: implement transparent proxy as a flag

LouManglass:
## Why

Transparent proxy mode is currently enabled with an environment variable only. This work will bring it into the run configurations as a feature flag as well.

## What

- Local CLI task runner now accepts `--transparent-proxy` as a flag
- Documented on [docs site](https://o3c.info/docs/using/advanced-configuration/#transparent-proxy-mode)

## Validation
* [ ] CI passes
* [ ] Reviewed by Optic
* [ ] Draft revised to meet best practices


devdoshi (approved):
Looks good. What are your thoughts on it instead living inside the optic.yml task config?

workspaces/local-cli/src/shared/local-cli-task-runner.ts
```diff
@@ -97,6 +102,7 @@ export async function LocalTaskSessionWrapper(
     shouldCollectCoverage: flags['collect-coverage'] !== false,
     shouldCollectDiffs: flags['collect-diffs'] !== false,
     shouldExitOnDiff: flags['exit-on-diff'] !== false,
+    shouldTransparentProxy: flags['transparent-proxy'] !== false,
```
acunniffe, line 105: @LouManglass I'd confirm this logic works as intended and doesn't break people doing it the old way. 

Based on what I'm reading, it seems like anyone using the legacy ENV var will find it no longer works unless they use the flag. 

Undocumented feature, sure, so ok to change, but better if we don't have to break anyone.

LouManglass:
> Looks good. What are your thoughts on it instead living inside the optic.yml task config?

Seems like a win, though honestly, I could see a use for most flags to be settable in the yml.  "I always want this task to collect coverage". "This task always needs to exit 1 on a diff", and so on.

workspaces/local-cli/src/shared/local-cli-task-runner.ts
LouManglass: Yep, I broke that on line 250. Now it sets the variable properly when the flag is present, otherwise defaults to whatever the environment variable is (so if someone set it to `yes` in the shell, that should now carry forward). Loom to show that the flag and the environment variable work:

https://www.loom.com/share/4eaea81b9fad44beb06cf81a84244518

PR #633, opened 2021-03-11
Implement processing of spec commands for ExampleSpecService

JaapRood:
## Why
We've introduced a new `processCommands` method for the `SpecServiceClient`, which is the new interface to persist spec changes through the new Rust command handlers. However, we skipped implementing it for the `ExampleSpecService` implementation, disallowing this new way of saving spec changes to be used for debug / example / private sessions.

## What
This PR adds an `append_batch_to_spec` method to the WASM DiffEngine implementation, accepting commands describing the spec change and a commit message, returning the newly generated events, or throwing if there were any command handling errors.

In order to implement this, it extracts the logic for starting a batch, handling commands and committing it from the CLI interface, to a `AppendedBatch` struct and a convenient function to set this up. This way the CLI and WASM interfaces only detail the interface specifics, like how the I/O works.

## Validation
* [x] CI passes
* [x] Finalized changes from the Diff Review UI causes an updated spec in Documentation.

### DIFF
--- a/workspaces/diff-engine-wasm/engine/src/lib.rs
+++ b/workspaces/diff-engine-wasm/engine/src/lib.rs
@@ -1,8 +1,11 @@
 #![allow(dead_code, unused_imports, unused_variables)]
 
+use uuid::Uuid;
 use wasm_bindgen::prelude::*;
 
-use optic_diff_engine::{HttpInteraction, InteractionDiffResult, SpecEvent, SpecProjection};
+use optic_diff_engine::{
+  HttpInteraction, InteractionDiffResult, SpecCommand, SpecEvent, SpecProjection,
+};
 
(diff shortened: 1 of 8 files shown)


workspaces/diff-engine-wasm/engine/src/lib.rs
```diff
@@ -60,3 +64,36 @@ impl From<InteractionDiffResult> for ResultContainer<InteractionDiffResult> {
     Self(result, fingerprint)
   }
 }
+
+// Commit batch
+// ------------
+
+#[wasm_bindgen]
+pub fn append_batch_to_spec(
```
devdoshi, line 72: perhaps a better name is apply_commands

devdoshi (approved):
Looks good!

PR #636, opened 2021-03-12
command handlers should emit events for added contributions

acunniffe:
## Why
The new Rust command handlers need to implement the AddContributions command and emit the ContributionAdded event as users add descriptions throughout their specifications. 

## What
Simple command handler and snapshot tests have been written. As the contribution projection will be a simple lookup, there's no need to add validation rules this stage. You can add contributions even if some of the IDs they reference don't exist in the specification.  

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/diff-engine/src/commands/rfc.rs
+++ b/workspaces/diff-engine/src/commands/rfc.rs
@@ -40,14 +40,18 @@ impl RfcCommand {
   pub fn end_batch_commit(batch_id: String) -> Self {
     Self::EndBatchCommit(EndBatchCommit { batch_id })
   }
+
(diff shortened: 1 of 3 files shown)


devdoshi (approved):
looks good as long as the EventContext is still applied at some point

JaapRood:
Ah whoops, this got lost in a Friday afternoon notifications snafu. Sorry for the wait :(

JaapRood:
> looks good as long as the EventContext is still applied at some point

@devdoshi confirmed that it does, through the `SpecCommandHandler` that wraps it.

PR #637, opened 2021-03-12
Add cqrs_event macro to reduce Event trait implementation boilerplate

JaapRood:
## Why
The project uses a large amount of events, a number which is only set to grow. To allow us work within a CQRS methodology efficiently and correctly, this involves a lot of boilerplate for implementing `Event` and `WithEventContext` traits for both individual events as well as groups. On top of that, there's a bunch of `From<T>` trait implementations to make working with groups of events vs. individual events simpler.

Also, I've been studying declarative macros in Rust and needed something to apply the theory too 😅.

## What
This PR implements a `cqrs_event!` macro that handles these trait implementations for each event, as well as groups of events and applies it for all the `SpecEvent` variants.

The most basic use is a single event struct, where one provides the struct and it's &'static str event name
```rust
struct ExampleEvent {
  some_property: String,
  // struct is required to have an optional event_context
  event_context: Option<EventContext>,
}

cqrs_event!(ExampleEvent => "example_event_name");
```

A enum of nested events can also be expressed, as long as the variants and struct names match. In addition to
`Event` and `WithEventContext` traits, it also implements `From<Variant> for Group` for each variant.
```rust
struct ExampleEvent {
  some_property: String,
  // struct is required to have an optional event_context
  event_context: Option<EventContext>,
}

struct AnotherExampleEvent {
  other_property: u16,
  // struct is required to have an optional event_context
  event_context: Option<EventContext>,
}

enum EventGroup {
  ExampleEvent(ExampleEvent),
  AnotherExampleEvent(AnotherExampleEvent)
}

cqrs_event!(ExampleEvent => "example_event_name");
cqrs_event!(AnotherExampleEvent => "another_example_event");
cqrs_event!(EventGroup {
  ExampleEvent,
  AnotherExampleEvent
});
```

The two forms can also be combined, generating trait implementations for both the single events, as well as the group.
```rust
cqrs_event!(EventGroup {
  ExampleEvent => "example_event_name",
  AnotherExampleEvent => "another_example_event_name"
});
```

### Why not also the enum and struct definitions?

From what I've learned about declarative macros, things get very complicated if you want to generate the structs and enums as well. While it can works for the basics, it falls down as soon as you try to use things like derive macros, variant / field level derives, etc. 

On top of that, it makes navigating to the event definition elsewhere in the code a lot harder to understand, as it'll jump to the macro's code (if it works at all). Being able to Find / replace the basic constructs of a program is something you want to work without friction.

## Validation
* [ ] CI passes
* [x] Traits only, definitions of structs and enum still traversable.

### DIFF
--- a/workspaces/diff-engine/src/events/macros.rs
+++ b/workspaces/diff-engine/src/events/macros.rs
@@ -1,3 +1,4 @@
+/// Implements cqrs_core::Event and WithEventContext for struct or enum of structs
 #[macro_export]
 macro_rules! cqrs_event {
   // Single Event => event_name. Impls for a single event
--- a/workspaces/diff-engine/src/events/macros.rs
+++ b/workspaces/diff-engine/src/events/macros.rs
@@ -10,7 +10,7 @@ macro_rules! cqrs_event {
     }
 
     impl $crate::events::WithEventContext for $event_type {
-      fn with_event_context(&mut self, event_context: EventContext) {
+      fn with_event_context(&mut self, event_context: $crate::events::EventContext) {
         self.event_context.replace(event_context);
       }
     }
@@ -27,7 +27,7 @@ macro_rules! cqrs_event {
     }
 
     impl $crate::events::WithEventContext for $group_type {
-      fn with_event_context(&mut self, event_context: EventContext) {
+      fn with_event_context(&mut self, event_context: $crate::events::EventContext) {
         match self {
           $( $group_type::$event_type(evt) => evt.with_event_context(event_context)),+,
         }
--- a/workspaces/diff-engine/src/events/macros.rs
+++ b/workspaces/diff-engine/src/events/macros.rs
@@ -1,0 +1,52 @@
+#[macro_export]
+macro_rules! cqrs_event {
+  // Single Event => event_name. Impls for a single event
+  ($event_type: ty => $name: expr) => {
+    impl ::cqrs_core::Event for $event_type {
+      fn event_type(&self) -> &'static str {
+        $name
+      }
+    }
+
+    impl $crate::events::WithEventContext for $event_type {
+      fn with_event_context(&mut self, event_context: EventContext) {
+        self.event_context.replace(event_context);
+      }
+    }
+  };
+
+  // Group (enum) { Event }. Impls for a group of events
+  ($group_type: ident { $($event_type: ident),+ }) => {
+    impl ::cqrs_core::Event for $group_type {
+      fn event_type(&self) -> &'static str {
+        match self {
+          $( $group_type::$event_type(evt) => evt.event_type() ),+
+        }
+      }
+    }
+
+    impl $crate::events::WithEventContext for $group_type {
+      fn with_event_context(&mut self, event_context: EventContext) {
+        match self {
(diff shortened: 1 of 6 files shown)


JaapRood:
@devdoshi I've marked this draft so far, as I'm not 100% yet whether it's a good idea. Macros add a layer of indirection, and we might just want to leave all the trait implementations as an explicit part of the codebase. That said, it does clean up nicely, and sticking to only the boilerplate trait implementations seems like a reasonable scope. I'm happy if this dies right here and it was just an exercise in declarative macros for me. Lets discuss!

workspaces/diff-engine/src/events/macros.rs
```diff
@@ -0,0 +1,124 @@
...
+/// # use optic_diff_engine::cqrs_event;
+/// #
+/// struct ExampleEvent {
+///   some_property: String,
+///   // struct is required to have an optional event_context
+///   event_context: Option<EventContext>,
+/// }
+///
+/// cqrs_event!(ExampleEvent => "example_event_name");
+/// ```
```
JaapRood, line 17: I tried to make sure the macro is well documented, especially because their implementation is harder to read for those who haven't worked with them yet. By doing that, I learned that cargo compiles and runs all code blocks as tests as part of `cargo test`, making sure they stay up-to-date. Neat-o!

devdoshi:
It seems nice, would be cool to be able to say `cqrs_event!(ExampleEvent)` and have it use ExampleEvent as the string to output. In general I think I would like there to be the macro and the generated code as part of the repo so you can navigate easily.

JaapRood:
> It seems nice, would be cool to be able to say cqrs_event!(ExampleEvent) and have it use ExampleEvent as the string to output. 

Haven't figured out how to do that with declarative macros. Also wonder how one would distinguish between the second and third form (implementing traits for the  group of events from events who already have the traits vs. impl both at once).

> In general I think I would like there to be the macro and the generated code as part of the repo so you can navigate easily.

I couldn't figure out a way to do that, let alone an idiomatic one. One could `cargo install cargo-expand` so you can run `cargo expand`, which expands all macros and allows you to see what's written. That said, that's visibility into intermediate output, not something you can then use as source again.

That's exactly why I limited this to trait implementations, not the struct and enum definitions themselves.

devdoshi:
The `stringify!` macro converts an `ident` into a string if I understand it correctly. Applying that should remove the final bit of redundancy and then I'd be happy to merge this.

JaapRood:
@devdoshi that solves converting an ident to a string, but this still remains:

> Also wonder how one would distinguish between the second and third form (implementing traits for the group of events from events who already have the traits vs. impl both at once).

devdoshi:
It seems like we only need one form, right? Otherwise, I'm not clear on the issue

JaapRood:
We need at least two, to differentiate between a group of groups (`SpecEvent`, enum of enums) and a group of events (like `EndpointEvent`, `RfcEvent`, etc.). Afaik, we can't detect whether the `ident` describes a struct or an enum, so the `EventIDent => "event_name"` form aims to detect a struct impl, and `GroupIdent => EventIdent` to detect the enum case.

Perhaps the docs need more work to communicate this?

PR #638, opened 2021-03-15
using target as hostnameFilter

acunniffe:
## Why
In transparent proxy mode, users want to filter down to a specific hostname. 

## What
Now, whatever `targetUrl` is set will be used as a hostname filter. 

## Validation
* [ ] CI passes
* [x] Verified in staging

### DIFF
--- a/workspaces/cli-config/src/index.ts
+++ b/workspaces/cli-config/src/index.ts
@@ -130,6 +130,7 @@ export interface IOpticApiInterceptConfig {
 
 export interface IOpticTaskRunnerConfig {
   command?: string;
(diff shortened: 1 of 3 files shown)


acunniffe:
https://github.com/opticdev/optic/actions/runs/655456310

devdoshi (approved):
assuming ci finishes, approved

LouManglass:
Usability tests, testing new default ignore rules, tested bad ignore rules (then realized that was a separate branch and tested that there instead, whoops), and verified new `--transparent-proxy` flag and behavior for this PR. All looking good! Much cleaner.

PR #643, opened 2021-03-17
(diff-engine): Allow events module to be accessed outside the crate

jshearer:
## Why
For my work on the proxy, I need to create and serialize `HttpInteraction` objects. @JaapRood didn't expose them originally to make sure the surface area didn't explode, but in this case it seems pragmatic to expose them.

## What
I made `HttpInteraction` public, along with all of its dependent types. I also added a `Serialize` derive to teach serde how to write these out as json

## Validation
Not sure about this one. There are some tests, and they appear to pass.

### DIFF
--- a/workspaces/diff-engine/src/lib.rs
+++ b/workspaces/diff-engine/src/lib.rs
@@ -1,7 +1,7 @@
 #![allow(dead_code, unused_imports, unused_variables)]
 
 mod commands;
-mod events;
+pub mod events;
 mod interactions;
 mod projections;
 mod protos;
--- a/workspaces/diff-engine/src/lib.rs
+++ b/workspaces/diff-engine/src/lib.rs
@@ -1,7 +1,7 @@
 #![allow(dead_code, unused_imports, unused_variables)]
 
 mod commands;
-pub mod events;
+mod events;
 mod interactions;
 mod projections;
 mod protos;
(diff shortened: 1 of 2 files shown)


workspaces/diff-engine/src/lib.rs
```diff
@@ -1,7 +1,7 @@
 #![allow(dead_code, unused_imports, unused_variables)]
 
 mod commands;
-mod events;
+pub mod events;
 mod interactions;
```
JaapRood, line 5: I wondered if we could be more surgical than exposing every single event struct, bu thinking about it I think you're right about just doing the whole lot. Commands and Events are the I/O types to getting anything accomplished, so we might as well accept that and move on.

JaapRood (approved):

JaapRood:
@jshearer I'll leave the timing and honour of merging this up to you :)

devdoshi (approved):

workspaces/diff-engine/src/lib.rs
jshearer: @JaapRood so Rust question: I had to make this module public for _anything_ inside it to be accessible. My impression is that you still have to mark structs/traits as `pub` for them to be public.

Is there a way to surgically do that without making it public?

JaapRood: There is! What I've seen in other Rust codebases and adopted for `diff-engine` is the approach of declaring inline modules and re-exporting: https://github.com/opticdev/optic/blob/develop/workspaces/diff-engine/src/lib.rs#L29-L34.

jshearer:
@JaapRood I changed this to re-export specific types from `events::http_interaction`. Still look good?

JaapRood (approved):
Nice and surgical ☺️.

PR #644, opened 2021-03-18
dependency bumps

LouManglass:
## Why

Responding to yarn/npm audits.

## What

- package upgrades
- identifying breaking changes
- Usability test run

## Validation
* [ ] CI passes
* [ ] Usability test
* [ ] Review

## Notes

As of this PR...

- global-request-logger doesn't seem to have gotten an npm registry update recently, but it has some activity in GitHub. I'll dig into that tomorrow.
- mockttp latest does not resolve a few downstream warnings, it looks like that may be coming soon though. I can play with that more tomorrow as well.

Status of outstanding items subject to change as I dig in further.


LouManglass:
Looks like `global-request-logger` just needs a registry update, [issue 27 opened](https://github.com/meetearnest/global-request-logger/issues/27). There's global-json-request-logger as well, which appears to track global-request-logger v1.0.0, though that doesn't resolve everything and we might as well get the full version bump.

devdoshi:
@LouManglass We actually don't use global-request-logger anywhere, so we can safely remove it as a dependency. I'm not sure the best way to audit what dependencies we aren't using but we can make an issue for that. 

re: analytics-node - I'm trusting your usability tests also check analytics
re: mockttp - I saw some things being deprecated that we were using around the rules we use for proxying, so as long as that works it should be safe to assume it hasn't been removed yet
re: react-dev-utils, we are only using it to open the browser on `api spec` etc. so as long as that works we're good there

devdoshi (approved):
approved pending testing results per my comments

LouManglass:
Cleaned up global request logger, have a few more things to do per Dev's comments...and I busted my ui in the process, so I'm doing a clean check on my work as well.

LouManglass:
That should get us most of the way there, I'll do another round of testing and make sure I cover the particular cases Dev outlined but so far it's looking pretty good.

LouManglass:
[] re: analytics-node - Check analytics
[] re: mockttp -we'll need to follow up with the team there anyway, I'll look into opening a PR with the appropriate updates. They're low score issues, I don't believe it affects how we use it, and it only affects availability of the application (triggering the bug crashes, no leaks) so it's not critical to get done in this batch.
[] re: react-dev-utils, api spec works great.

PR #645, opened 2021-03-19
Move learning of initial request and response bodies to Rust

JaapRood:
## Why
While most of the diffing already happens in Rust, the part that's involved with learning new bodies for undocumented endpoints is still in Scala, with the performance penalty that comes with it. Just was with the diffs, for reasonably sized projects this can take many minutes, while porting it to Rust took that back down to seconds.

## What
Implements a new initial-bodies worker that calls to Rust instead of Scala.

A new `learn` command is added to the Rust CLI. This can be used from the Node.js side through `learnUndocumentedBodies`, which accepts the spec and interactions that are to be learned, returning a readable stream of endpoint bodies. 

A new worker task uses this to implement the learning of an interaction per endpoint and method, just like the Scala version.

Like the Scala version, this implementation causes a re-diffing of the interaction. However, it's structured for that to change and go with a more efficient pipeline, where bodies are learned immediately after diff results come available from their first production. While changes on that scale are beyond the scope of this work, the path to it is open.

Processing the interactions (diffing + learning their shapes) happens in a similar concurrency model as the diffing, with multiple threads diffing interactions concurrently, sending results to a thread dedicated to results aggregation. If need be, results aggregation can be split up between multiple threads in the future, probably by partitioning by endpoint. For now, that's considered beyond the scope of this PR.

## Validation
* [x] CI passes
* [x] Hidden behind a feature flag by default
* [x] No ScalaJS used on server + domain side.
* [x] Works as drop-in replacement for the UI client.


JaapRood:
Continuing on the work of @devdoshi and @acunniffe, the new `analyze_undocumented_bodies` now returns useful `BodyAnalysisResult`s. There's now also a new `learn` subcommand for the Diff Engine CLI, for which I've drafted the consuming of interactions from `stdin` and analysing them on their own thread.

Next up, is taking the results from each individual interaction and aggregating them per interaction trail. From there, we should be able to figure out exactly per body what it's shape should be and what commands we should generate. Once we have those, it's a matter of writing those back over stdout in the format the consumer expects.

JaapRood:
We're getting close to something that should work end-to-end, for primitive bodies. A slight reframe happened where `fn analyze_undocumented_bodies() -> BodyAnalysisResult` is now `fn observe_body_trails() -> TrailObservationsResult` to better fit in with the rest of the codebase and give us a place to implement the logic of mapping those observations to the commands that should be generated.

Some of that basic command generation has been added as well, be it only for primitive types thus far. To verify this and the concurrency model works at least somewhat, I'll look at doing the last bit of hook-up there next. From there, I'll look at adding support for more complex types, like lists, arrays, nullables, etc.

JaapRood:
The version I just pushed outlines our approach for making sure the commands generated from observations are in the correct order, as well as allowing us to reason about what the root shape id must be. Shape commands must be in order as the command handlers make sure the spec stays referentially and causally complete (e.g. a List can't point to a ListItem shape that doesn't exist). The root shape must be known so it can be registered as representing the body for a request / response.

The key to our approach for this logic is that we sort the json trails so we generate shape prototypes and their commands for **leafs first**, to then recurse inwards. That allows us to guarantee that the primitive shapes referenced by their parents (arrays, objects, etc) have been created already. It also allows us to reason that the last shape generated, is also the root shape.

To implement this sorting, `JsonTrail` now implements `Ord` and `PartialOrd`, which compare the _normalized_ trail [lexicographically](https://doc.rust-lang.org/std/cmp/trait.Ord.html?search=#lexicographical-comparison). Once reversed, that gets us exactly the sorting guarantees we're after, causing descendants to be ordered before their ancestors.

---

Next up, I'll take the last couple of steps on the CLI side to make sure we've got all the pieces we need to produce the output types as expected. Once I've got that confirmed, I'll address the fact that we have no tests to the confidence that our business logic works correctly (including the approach above). I'll look at getting that covered in some basic ways, especially around the more tricky bits like command ordering.

JaapRood:
I fixed and managed to manually verify the CLI `learn` interface accepts interactions on stdin and produces learned bodies aggregated by endpoint on stdout. Knowing that's all hooked up correctly, I'll move on to verifying the bodies and commands that get generated are correct for primitives, then move on to complex types.

JaapRood:
Generating commands for primitive bodies and one-off primitive bodies has been implemented and verified to be correct. Now for the array and object body types.

JaapRood:
We're now generating the correct commands for most array bodies! Array bodies are now being observed by their trail, shape prototypes are generated, which are subsequently turned into commands.

This is where the processing of the leaf shapes became important and we differ from the Scala implementation. Rather than generating shape prototypes recursively, we make sure the item shape of a list is guaranteed to already have been generated. For the command generation, this same guarantee simplifies the command generation as well, only having to add the commands for the list specifically, not its item.

To provide that guarantee and the shape prototype process access to the prototypes thus far generated, the approach to make this happen changed a bit. With the functional approach coming hard to follow, I was happy to rely on Rust' for loops being able to be used on any iterator, keeping it straightforward to read.

--- 

The one type of array which isn't generated correctly yet, are empty ones! At the moment we panic if an array doesn't have an item type, but that's of course totally possible. I'll look at that next, and then move on to object bodies.

JaapRood:
Empty arrays and objects are now supported by the learner. 

In case of the empty arrays, it was a matter of observing whether an array was empty and recording that as part of the results. During the shape prototype generation this allows us to only look up (and panic when failing to) the item's shape prototype.

For the objects, the two most significant changes are:

1. The tracking of the various field sets out there. This was originally modelled with a `HashSet<HashSet<String>>`, but that doesn't work in Rust as `HashSet` doesn't implement the `Hash` trait. Looking into it, that actually makes sense, as there's not actually a generic way to generate that hash, as iterating over a set's items is not necessarily stable (that's what ordered sets are for). For now, I've solved that by keeping a `Vec` of field sets instead, and managing a bit of merging manually. It's not the world's efficient way of doing it, but also definitely not most inefficient. It is, however, fairly straightforward, so a good candidate for now while we monitor it's performance impacts.
2. Replacing the `ShapePrototype::FieldOfObject` variant a `FieldPrototypeDescriptor` struct, as field and shapes are 2 different concepts. That showed up in trying to wire it all up, with lots of extra matches to deduce what we already know: `ShapePrototype::ObjectWithFields` holds descriptions of fields, not anything else.

--- 

With support for primitives, one-offs, arrays and objects now complete, I'll cover the last polymorphic types: optional and nullable shapes.

JaapRood:
In the version I just pushed optional and nullable shapes are now supported, one-off shapes now have support for non-primitives and the shape generation has been cleaned up more generally.

A couple of differences with the Scala side came up and were addressed:
- optional shapes are not really shapes of themselves, but more of a parameter of an object field (for arrays or other types, optional shapes would simply not be present). While the same commands are generated for optional types, they are internally represented as part of the `FieldPrototypeDescriptor` by turning its `value_shape_id` field into an `Option<String>`. That context simplified the matching and command generation significantly.
- since we're guaranteed to generate commands for leafs first, one collection of commands is enough to get them into the right order. That allowed the removal of "initiate" vs. "describe" commands and with that, a whole bunch of additional merging logic and allocations. Simpler, cleaner, yay ☺️.

---

With those changes, we should have complete parity on the initial bodies learner from a compute perspective. Time to hook the new CLI interface for it up to the Node.js and UI side!

JaapRood:
We now have an alternative version of the POST /initial-bodies capture endpoint which, when enabled with `OPTIC_RUST_INITIAL_BODY_LEARNER=true`, uses the Rust CLI to generate the learned bodies response for the requested endpoint. This is done through the `InitialBodyWorkerRust` (in cli-shared), which applies and endpoint filter to with in with the endpoint's current model and parses the output from the Rust side, ready to be used.

A couple of things to note on the implementation:

- on the `cli-server` side we instantiate this new worker straight in the daemon process, rather than spawning a separate node.js script like what happens for the ScalaJS version. All the compute and fallible logic is already in a different process by virtue of running through the Rust CLI, and that extra layer has mostly only caused a really hard to debug situation for other places we ported to the Rust CLI diff engine. Especially since we're likely to change how this processing is hooked up in the first place (hanging straight of diffs as post-processing), not having to implement the custom IPC stuff really simplified the integration.
- the logic for filtering the endpoint still relies on ScalaJS, as that way of querying the spec through the WASM domain interface doesn't exist yet. We'll want to look at that before we merge this (as the opportunity we're changing with this is to get rid of ScalaJS).

---

Now that we have learned bodies making it this far down, it's time to make sure the client can deal with it. Currently, there's some errors being thrown, most likely because the previous implementation of the endpoint has types that lie and actually return something of a different shape. I'll look at that next.

JaapRood:
The UI can now accept and render the suggested bodies for endpoints with undocumented bodies. However, approving and finalising the changes indicated something was missing, with the bodies not actually appearing in the documentation. Diffing the previous and new implementation's generated commands, showed two things:

1. Missing add request / response commands and command tying the shape to it.
2. Missing add request / response commands for requests or responses that didn't have a body. Turns out, a documentation of a response doesn't mean much without the request that caused it also having been specced.

The first have been covered: after figuring out whether this should be a client concern, the decision was made to keep the client dumb and have them be generated, too. 

The second is what I'll look at next. It took a little bit to find the right approach, but eventually we landed on generating body analysis results even in case of there being no body, so we have a `BodyAnalysisResult` with the right `BodyAnalysisResultLocation` to trigger the generation of the necessary events.

JaapRood:
A lot of digging later and attempts later, the learn CLI command now produces the correct commands for the adding of missing requests and responses. There's also proper support for requests and responses with empty bodies / no content-type. 

Multiple factors made it hard to see what the correct implementation would be and how to achieve it, not the least of which the Scala implementation generating these commands when they actually shouldn't (always includes AddRequest commands even though the request already exists, but probably just doesn't execute them), but also a failing assumption about when the diffs that power this are emitted. 

I've manually tested this behaviour directly against the CLI for now. I've scaffolded how we might integrate test this entire learn command, since some of the logic lives there, but decided that it's better to first get this thing working vertically first, as well as getting rid of the Scala code in the endpoint filtering, before providing automated test coverage of the I/O interface. I'll look those two things next.

JaapRood:
With a couple of more fixes around dealing with unknown shapes properly, the initial body learner is now successful with everything I can throw at it. I've also amended the WASM interface of the diff engine to allow for a non-Scala implementation of the endpoint filter that's included in the initial bodies worker, getting rid of the last bit of Scala.

As expected, since it involves re-diffing interactions against the spec, the performance jump is noticeable as well, with at least an order of magnitude improvement.

--- 

With this now working end-to-end without Scala involvement in the pipeline, I'll get into clean up and merging mode! While I work on a couple of smaller details (like using something more elegant than UUID's for ids in the spec), @devdoshi now would be a good time for an additional review.

JaapRood:
Having cleaned most of this up, it's time for review. @devdoshi it's quite the behemoth, but fortunately it's written in Rust, so must at least be semi-sane. Where a second pair of eyes would be useful is on making sure the feature flag is, so we can safely QA the feature with the rest of the team post merge.

workspaces/diff-engine/src/commands/mod.rs
```diff
@@ -188,7 +190,8 @@ impl AggregateCommand<SpecProjection> for SpecCommand {
         };
 
         let shape_added_event = ShapeEvent::from(ShapeCommand::add_shape(
-          String::from("$string"),
+          Uuid::new_v4().to_hyphenated().to_string(),
```
devdoshi, line 193: do you want to use the same id generator everywhere?

devdoshi (approved):
awesome work @JaapRood  there's a lot going on and it's verbose as usual with Rust but very clean and easy to follow if you know what it's supposed to do. I just had the one comment about id generation, and there are a few stale code comments throughout the PR.

workspaces/diff-engine/src/commands/mod.rs
JaapRood: Would make sense, wouldn't it? Especially with a trait generically named `SpecIdGenerator`. I'll get that cleaned up and consistent.

JaapRood:
> awesome work @JaapRood there's a lot going on and it's verbose as usual with Rust but very clean and easy to follow if you know what it's supposed to do. I just had the one comment about id generation, and there are a few stale code comments throughout the PR.

I'm going to be a good scout and do a short pass of clean up before I leave this site. Might add a bit of documentation on the top-level functions as well, just to make it clear what it's place in the bigger system is, give people a chance to understand despite the verbosity.

workspaces/diff-engine/src/commands/mod.rs
JaapRood: I'm going to cover this one in a separate PR, as it involves a restructuring of the command handling (moving all of `SpecProjection` implementation of `AggregationCommand` to the `SpecCommandHandler`). Totally worth doing, but given this PR has already become massive with main goal of killing Scala.

JaapRood:
This is at least a bit better, hopefully gives someone an entry in figuring out how things might have to work. Screenshot 2021-04-12 at 13 11 00

PR #646, opened 2021-03-24
feature: Implement s3 ingest capture saver

jshearer:
## Why
Slurp saves `IHttpInteraction`s in batches to an S3 bucket. We want to load that data into your local Optic instance. This PR implements the S3 ingest capture saver that does that

So far so good!

Opens the UI and contains the right knowledge from the captures! 
Screen Shot 2021-03-24 at 4 03 41 PM

And we have a progress bar
 

https://user-images.githubusercontent.com/4368270/112524972-ca9ee380-8d76-11eb-99ec-81bec6e5d571.mov

## TODO:
- [x] Better progress bar
- [x] Integrate with the existing CLI 
- [x] Launch daemon and open browser upon completion
- [ ] Add option to limit total size of files ingested. E.g to 1gb or something.

### DIFF
--- a/workspaces/cli-shared/src/captures/avro/file-system/ingest-s3-capture-saver.ts
+++ b/workspaces/cli-shared/src/captures/avro/file-system/ingest-s3-capture-saver.ts
@@ -7,11 +7,19 @@ import path from 'path';
 import { parser as jsonlParser } from 'stream-json/jsonl/Parser';
 import { map as streamMap, flatMap } from 'axax';
 import { IHttpInteraction } from '@useoptic/domain-types';
+import { fromNodeStream } from 'axax/es5/fromNodeStream';
 //@ts-ignore
 import AWS from 'aws-sdk';
 import { CaptureSaver } from './capture-saver';
 import { Token as ContinuationToken } from 'aws-sdk/clients/s3';
 import { parseIgnore } from '@useoptic/cli-config/build/helpers/ignore-parser';
+import { flatten, map, pipe } from '@useoptic/diff-engine-wasm/lib/async-tools';
+
+async function* fromReadable<T>(r: Readable): AsyncIterable<T> {
+  for await (const chunk of r) {
+    yield chunk;
+  }
+}
 
 export async function ingestS3({
   bucketName,
@@ -76,13 +84,13 @@ export async function ingestS3({
   let batchCount = 0;
   let totalBatches = 0;
 
-  async function* bucket_keys(
+  async function* bucket_key_chunks(
     cursor?: ContinuationToken
-  ): AsyncIterable<string> {
+  ): AsyncIterable<string[]> {
     const objects = await s3
       .listObjectsV2({
         Bucket: bucketName,
-        MaxKeys: 1000,
+        MaxKeys: 100,
         Prefix: prefix,
         ContinuationToken: cursor,
       })
@@ -95,14 +103,10 @@ export async function ingestS3({
 
     customBar.setTotal(totalBatches);
 
-    for (const batch of batch_files) {
-      yield batch;
-      batchCount++;
-      customBar.update(batchCount);
-    }
+    yield batch_files;
 
     if (objects.IsTruncated) {
-      yield* bucket_keys(objects.NextContinuationToken);
+      yield* bucket_key_chunks(objects.NextContinuationToken);
     }
   }
 
@@ -115,9 +119,31 @@ export async function ingestS3({
       .createReadStream();
   }
 
+  // We convert this to Readable and back so we can get the nice buffering functionality
+  // that readable provides, even for objectMode streams.
+  let key_chunk_readable = Readable.from(bucket_key_chunks(), {
+    objectMode: true,
+    highWaterMark: 5,
+  });
+  let iterable_again = fromReadable<string[]>(key_chunk_readable);
+
   let continuous_readable: Readable = Readable.from(
-    flatMap(key_to_readstream)(bucket_keys())
+    pipe(
+      // Starting out with AsyncIterable<string[]> from `bucket_key_chunks()` which yields string[], not string
+      // Flatten it into AsyncIterable<string>, while also updating the progress as a side-effect
+      flatMap(async function* (items: string[]) {
+        for await (const item of items) {
+          batchCount++;
+          customBar.update(batchCount);
+          yield item;
+        }
+      }),
+      // Go from an iterable of bucket keys to an iterable of readstreams,
+      // Then flatten that to a single stream of string chunks
+      flatMap(key_to_readstream)
+    )(iterable_again)
   );
+
   let parser = continuous_readable.pipe(jsonlParser());
   let obj_stream = streamMap((obj: { value: IHttpInteraction }) => obj.value)(
     parser
--- a/workspaces/cli-shared/src/captures/avro/file-system/ingest-s3-capture-saver.ts
+++ b/workspaces/cli-shared/src/captures/avro/file-system/ingest-s3-capture-saver.ts
@@ -50,20 +50,20 @@ export async function ingestS3({
 
   console.log(`Looking in s3://${bucketName}${prefix} in region ${region}, ${endpointOverride ?? ''}`);
 
-  const listAll = async (token?: ContinuationToken): Promise<S3Object[]> => {
+  const listAll = async (token?: ContinuationToken): Promise<string[]> => {
     const objects = await s3.listObjectsV2({
       Bucket: bucketName,
-      MaxKeys: 100,
+      MaxKeys: 1000,
       Prefix: prefix,
       ContinuationToken: token
     }).promise();
     if (objects.Contents) {
-      console.log(`Found ${objects.Contents.length} capture files to ingest.`);
+      let obj_keys = objects.Contents.map(o=>o.Key).filter((o): o is string => !!o);
       if (objects.IsTruncated) {
         let rest = await listAll(objects.NextContinuationToken);
-        return objects.Contents.concat(rest);
+        return obj_keys.concat(rest);
       } else {
-        return objects.Contents;
+        return obj_keys;
       }
     } else {
       console.log("No object contents?");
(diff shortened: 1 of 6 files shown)


jshearer:
@acunniffe suggests streaming s3 keys instead of fetching all at once, as well as streaming the actual contents

devdoshi:
@jshearer I'd definitely stream the s3 keys as opposed to await listAll(). your getObject().createReadStream is probably sufficient for streaming the actual contents.

workspaces/local-cli/src/commands/ingest/s3.ts
```diff
@@ -0,0 +1,74 @@
...
+
+      const daemonState = await ensureDaemonStarted(lockFilePath, Config.apiBaseUrl);
+
+      const apiBaseUrl = `http://localhost:${daemonState.port}/api`;
+      const paths = await getPathsRelativeToConfig();
+      const cliClient = new Client(apiBaseUrl);
+      cliClient.setIdentity(await getUser());
+      const cliSession = await cliClient.findSession(paths.cwd, null, null);
+      const uiBaseUrl = makeUiBaseUrl(daemonState);
+      const uiUrl = `${uiBaseUrl}/apis/${cliSession.session.id}/review/${captureId}`;
```
jshearer, line 52: @devdoshi like this?

jshearer:
@devdoshi after talking a bit more with @acunniffe, we realized that if we stream keys then we won't have useful progress bars, and that'll be sad. When we genuinely get to a place of having hundreds of MBs of keys, we'll stream them then. Does that seem good?

jshearer:
@acunniffe what should happen when we exceed the max file size to download? Just stop ingesting and finish up? Error out?

devdoshi:
> @devdoshi after talking a bit more with @acunniffe, we realized that if we stream keys then we won't have useful progress bars, and that'll be sad. When we genuinely get to a place of having hundreds of MBs of keys, we'll stream them then. Does that seem good?

I think the progress bars are still meaningful, however both the total and progress will increase as we get more keys and more downloads, right? I'm all for deferring the implementation, but it seems like early users will still have 1000s of interactions and your implementation is already pretty close to what would be needed

jshearer:
Mhmmmmm, didn't think about increasing the max on the progress bar. That's a good idea. Will do!

workspaces/cli-shared/src/captures/avro/file-system/ingest-s3-capture-saver.ts
```diff
@@ -0,0 +1,161 @@
...
+}: {
+  bucketName: string,
+  region?: string,
+  captureId: string,
+  pathPrefix?: string,
+  endpointOverride?: string
+}) {
+  let { capturesPath: capturesBaseDirectory, opticIgnorePath, configPath } = await getPathsRelativeToConfig();
+
+  let ignoreRules = await new IgnoreFileHelper(opticIgnorePath, configPath).getCurrentIgnoreRules();
```
acunniffe, line 31: In response to @devdoshi -- @jshearer, there is a better facade for this. 

You got ignore rules correctly on 31, use them like this

```
const {shouldIgnore} = parseIgnore(ignoreRules)
shouldIgnore(method, url) /// true
```

workspaces/cli-shared/src/captures/avro/file-system/ingest-s3-capture-saver.ts
```diff
@@ -0,0 +1,161 @@
...
+      customBar.update(objectsHandled);
+    }
+  }
+
+  let interactionCount = 0;
+
+  const oboePromose = new Promise<void>((resolve, reject) => {
+    oboe(passThrough)
+      .node('!', (row: IHttpInteraction) => {
+        if (!parsedRules.some(r => r.shouldIgnore(row.request.method, `${row.request.host}${row.request.path}?${row.request.query}`))) {
```
acunniffe, line 104:
```
shouldIgnore(row.request.method, row.request.path)
```
^ ignore rules are for paths, not for urls. It's confusing as written/typed.

workspaces/cli-shared/src/captures/avro/file-system/ingest-s3-capture-saver.ts
```diff
@@ -0,0 +1,161 @@
...
+      objectsHandled++;
+      customBar.update(objectsHandled);
+    }
+  }
+
+  let interactionCount = 0;
+
+  const oboePromose = new Promise<void>((resolve, reject) => {
+    oboe(passThrough)
+      .node('!', (row: IHttpInteraction) => {
```
JaapRood, line 103: We're already using [`stream-json`](https://www.npmjs.com/package/stream-json) for streaming json parsing, so lets not introduce oboe as well.

workspaces/cli-shared/src/captures/avro/file-system/ingest-s3-capture-saver.ts
```diff
@@ -0,0 +1,161 @@
...
+}
+
+async function pipeAsync(tap: Readable, sink: Writable) {
+  return new Promise((resolve, reject) => {
+    // end: false is important -- it prevents the stream from ending when one source emits an `end` event
+    tap.pipe(sink, { end: false })
+    tap.on("end", resolve)
+    tap.on("error", reject)
+  })
+}
```
JaapRood, line 161: This leaks memory the way the promise resolvement is implemented with event handlers here (no clean up of event listeners). You'll want to use `once` instead of `on` and also have a cleanup function removing the registered listeners upon 'end' or 'error', before resolving or rejecting. 

It also smells a bit, but possible through how it's named: connecting a pipe is a synchronous action, streams already an asynchronous construct. To "pipe" something asynchronously, when it's then actually done synchronously, seems non-sensical.

It seems what you're trying to do is create an 'StreamEnded' promise for the input stream, maybe something along those lines makes more sense.

JaapRood: Considering this further, I think the smell goes beyond the name: having an existing stream and adding more and more sources to it is non-idiomatic. Generally, the approach with building stuff with streams is:

1. construct the pipeline synchronously
2. run until the stream at the tail end of the pipeline finishes, or any of them error out.

JaapRood: Considering this further, I think the smell goes beyond the name: having an existing stream and adding more and more sources to it is non-idiomatic. Generally, the approach with building stuff with streams is:

1. construct the pipeline synchronously
2. run until the stream at the tail end of the pipeline finishes, or any of them error out.

\Considering this further, I think the smell goes beyond the name: having an existing stream and adding more and more sources to it is non-idiomatic. Generally, the approach with building stuff with streams is:

1. construct the pipeline synchronously
2. run until the stream at the tail end of the pipeline finishes, or any of them error out.

In this case, I would do this by:
1. create a stream of objects (from listing them)
2. map these objects to object read streams
3. merge these object read streams to a single read stream
4. Map the rows in the read streams to interactions that can be saved
5. map the interactions to the capture saver

Realising that Streams are Async Generators and Async Iterators now is the key to construct a pipeline like that in a reasonable way. Converting readable node streams to those early can make all that logic happen in a structured and typed way. See the `async-tools` and `streams` here for some ideas of how that's done: https://github.com/opticdev/optic/tree/develop/workspaces/diff-engine-wasm/src.

Also more than happy to pair with you on this, if it's worth addressing. The marriage between Node.js Streams and async / await / Promise has been a bit rough, until Async Generators and Async Iterators came to the scene, and it's not really widely talked about (I reckon since most Node engineers that would care are stuck in legacy stuff 😅).

JaapRood (commented):
Seems mostly fine, with some annoying little bits that come from the rough marriage between `async / await` and Node Streams. A lot of the code we wrote earlier that ended up looking like this, with promises created from streams by juggling event emitters, was made a lot nicer once we started dealing with Node Streams as Async Generators and Async Iterators (which they now implement by default). It might be worth looking into cleaning things up like that.

Happy to do some pairing with you if you're interested in doing so: it took me a lot of effort and time to land this stuff in a somewhat reasonable direction, so happy to prevent others from having to go through that same pain 😬.

workspaces/cli-shared/src/captures/avro/file-system/ingest-s3-capture-saver.ts
jshearer: Done!

workspaces/cli-shared/src/captures/avro/file-system/ingest-s3-capture-saver.ts
jshearer: Done!

workspaces/cli-shared/src/captures/avro/file-system/ingest-s3-capture-saver.ts
jshearer: Done!

jshearer: FWIW, I wasn't introducing oboe -- it's already in there. But refactored as per our call to use stream-json and be way nicer

workspaces/cli-shared/src/captures/avro/file-system/ingest-s3-capture-saver.ts
jshearer: Refactored as per our call -- muuch nicer now

jshearer:
@JaapRood / @devdoshi our call today was suuuuper helpful. I think it's written much more cleanly now (and maybe it doesn't even leak memory? 😅). One question: in the following video, you'll notice that, the object fetcher waits to fetch another chunk of keys until every single one of the existing ones is pushed through. This is exactly as intended, and clearly backpressure is working (yay!). But, given that the speed of light is finite, among other things, this introduces an unfortunate delay when a full chunk of keys needs to be fetched, blocking the parsing pipeline. 

Is there a way to introduce a buffer in between these asynciterators, such that (a few | one) additional values are read and buffered, ready to be sent immediately upon request? This would let the whole pipeline run along at full speed until the end, as opposed to pausing as it does now.

https://user-images.githubusercontent.com/4368270/112525774-9d9f0080-8d77-11eb-946d-c0bdea6982b6.mov

jshearer:
Ok, so I tried a thing that lets you load batches of keys concurrently with parsing, but there's no backpressure -- it loads keys as fast as it can, independently with parsing so it's probably not best. Thoughts on how to do this the "right way" @JaapRood?

```typescript
async function* bucket_keys(
    cursor?: ContinuationToken
  ): AsyncGenerator<string> {
    const objects = await s3
      .listObjectsV2({
        Bucket: bucketName,
        MaxKeys: 1000,
        Prefix: prefix,
        ContinuationToken: cursor,
      })
      .promise();

    let remaining;
    if (objects.IsTruncated) {
      remaining = bucket_keys(objects.NextContinuationToken);
    }

    let batch_files =
      objects.Contents?.map((c) => c?.Key).filter((o): o is string => !!o) ??
      [];
    totalBatches += batch_files.length;

    customBar.setTotal(totalBatches);

    let iters = [from(batch_files)];
    if(remaining){
      iters.unshift(remaining);
    }

    for await (const batch of merge(...iters)) {
        yield batch;
        batchCount++;
        customBar.update(batchCount);
    }

    if(remaining){
      yield* remaining;
    }
  }
```

https://user-images.githubusercontent.com/4368270/112537941-aac2ec00-8d85-11eb-8463-164ad571226e.mov

devdoshi:
I think you can prefetch the next key list as soon as the current key list has resolved. that resulting promise should not be awaited on in the current stream unit of work, but the promise should be cached such that the next stream unit of work is either a cache hit or waits for that same promise.

JaapRood:
> Ok, so I tried a thing that lets you load batches of keys concurrently with parsing, but there's no backpressure -- it loads keys as fast as it can, independently with parsing so it's probably not best

`merge`, if that's the operator from `axax`, will race the various iterators against each other, letting them all yield at least a single value before back-pressure is applied. Back-pressure comes from the down stream, which can only respond and provide that back-pressure once at least a single unit is written to it (since that's what calls it and yields control).

If you want these batches to be fetched from in series, you'll have to write it that way! Through either a map, or a `for await`, etc. You could even get fancy and run a couple of them concurrently up to a limit, but those are the more tricky operators to get right.

If you're worried about incurring overhead from network round-trips, going back to a Node Readable stream is your friend. It has all of the complicated additional logic that comes from optimistically filling a buffer, while maintaining correct back-pressure, drain / flushing behaviours, etc. That's definitely where they still have added benefit over a plain async generator / iterator, and are easily implemented through using `Readable.from(async_generator)`.

jshearer:
@JaapRood Very helpful. Now we do this:


https://user-images.githubusercontent.com/4368270/112887076-bde4fd00-90a0-11eb-9d2f-0775a5937681.mov

jshearer:
I'd say this is ready to merge now.

PR #652, opened 2021-04-01
switch macos aarch64 build to stable rustup

notnmeyer:
closes https://github.com/opticdev/issues/issues/39


JaapRood (approved):
All good, required compilation target has been part of stable since 1.49 and current stable is 1.51.

PR #660, opened 2021-04-07
9.0.19

acunniffe:
## Why
- Address user issue reported today about regression in oas-generator

## Validation
* [x] CI passes
* [x] Verified in staging
* etc...


workspaces/cli-config/tests/conflict-specs/ReadSpecification.test.ts
```diff
@@ -2,10 +2,17 @@ import * as path from 'path';
 import {resolveConflictsIfPresent} from "../../src/helpers/read-specification-json";
 // @ts-ignore
 import fs from "fs-extra";
+import {getSpecEventsFrom} from "../../build/helpers/read-specification-json";
```
devdoshi, line 5: seems like this conflicts with the first import, it should be one or the other; presumably src

LouManglass:
[Usability tests](https://github.com/opticdev/optic/blob/develop/Usability-tests.md) pass on the side channel build. Digging into checks and comments.

workspaces/cli-config/tests/conflict-specs/ReadSpecification.test.ts
LouManglass: REsolved.

PR #663, opened 2021-04-12
Move learning of diff trail affordances to Rust

JaapRood:
## Why
While most of the diffing already happens in Rust, the part that's involved with learning new affordances for shapes that caused diffs is still in Scala. These affordances power the UI where a user can choose to update their spec due to differences in fields. Since we're trying to get rid of Scala, it's important that we port this over

## What
We're introducing a new worker that calls a new Learn command in the Rust diff engine CLI. 

The `learn` CLI command now has a second `subject` implemented, `shapes-diffs-affordances`. This subject also requires `--tagged-diff-results` to point to the diff results that should be learned trails for. It then accepts interactions over `stdin` and learns from those one by one, for all results to be aggregated and returned over `stdout`.

The UI sends the `diffId` along with it's request for learning trail values, which allows the new Node.js worker to look up diff-results from disk. This replaces serialising and sending the diffs from the client to the server, which was an unnecessary overhead that's no longer required due to the shift of logic from UI -> backend.

A minimal amount of "re-diffing" is performed for each incoming interaction, basically to determine which interactions match a documented request or response body. From there, the qualifying interaction bodies are fully traversed and mapped, to then be matched against the trails relevant to the diffs. In the future, a more focused traverser could be implemented that only traverses the trail relevant to diffs, but so far, that kind of optimisation does not seem necessary.

Processing the interactions happens in a similar concurrency model as the diffing, with multiple threads diffing interactions concurrently, sending results to a thread dedicated to results aggregation. If need be, results aggregation can be split up between multiple threads in the future, probably by partitioning by endpoint. For now, that's considered beyond the scope of this PR.

## Validation
* [x] CI passes
* [x] Hidden by a feature flag by default (`OPTIC_RUST_SHAPE_DIFF_AFFORDANCES_LEARNER`).
* [x] No ScalaJS involved on server + domain side.
* [x] Works as a drop-in replacement for the UI client. (small caveat: UI client adjusted to POST `diffId` to endpoint).


JaapRood:
I'm still in the middle of drafting the compute approach, but the outlines are taking shape:

- diffs, tagged by their interaction pointers, are accepted through passing a file path from where they can be read.
- interactions tagged by their interaction pointers are accepted over stdin and iterated over, generating shape affordances for each. By matching the pointers to those in the diffs input, we can look up the diffs and prevent those from having to be recomputed. Especially as this concerns shape diffs, the most expensive kind, and affordances are requested _based_ on an existing diff, this approach. makes sense.
- interaction trail affordances are to be aggregated and merged by diff.
- output results (affordances per diffs, per interaction pointers) should be able to be constructed.

JaapRood:
As updated in my previous comment, I've got most of the approach drafted (compiling, but not yet tested or necessarily complete), including the concurrency model. 

For the concurrency model, like the rest of the commands, we're choosing to spawn a new task for each incoming interaction, sending results of the analysis to a single aggregation thread, which keeps the results and merges interactions touching the same diffs.

Notable: this is the first command implementation where the current API spec isn't required, indicating how the learning of initial bodies shouldn't do either.

---

Next up, I'll look at the last part of the Rust CLI: serialising the results and writing them to stdout. From there I'll do some basic testing to verify it basic I/O works and outputs look sane, to then integrate it with the Node.js / UI side.

JaapRood:
The learning of subcommands now seems to work correctly, with an automated test making sure there's no panics and a manual test verifying the outputs look sane (up for further scrutiny as this runs against the Node.js worker / UI).

A couple of decisions had to be made regarding deserialising inputs and serialising outputs. 

On the input side, I've implemented a custom deserialiser for `TaggedInput<T>` as to allow but ignore any additional items in array that represents it's internal tuple. This enhancement allows the diff results as generated by the Rust diff engine to be piped straight back in.

On the output side, I opted for writing the results as JSON new lines, as a tuple (array) where each result is tagged by it's diff hash. While that means we need to do a little bit of work on the Node.js side to transform this to a `Map<DiffHash, Result>`, it's worth staying consistent in API's over (it's the same as the diff engine results are pushed), as well to keep the door open for this being used in a more efficient, event-sourced / streaming-oriented pipeline.

---

Next, I'll start to hook this up to the Node.js side!

JaapRood:
While drafting the new trail values worker I realised that the serialized diffs, as sent over by the UI, wouldn't actually be enough to prevent re-computation of the diffs. While that was okay for the initial body learner, this re-computation would involve shape diffing, which is a lot more expensive, so something we'd want to avoid.

The main issue was that these serialised diffs aren't tagged with interaction pointers. These pointers are essential to allow the diffs to be looked up for every interaction that's visited by the learner, rather than having to re-compute it. The easiest backwards compatible fix for this (we'll ship this initially behind a feature flag, so both implementations should work) was to include the diff session id as part of the request to learn the trail values. Diff results are stored per this identifier, should passing it along is enough for the `cli-server` to read the diffs from disk itself, where they are stored with interaction pointers!

There was an additional issue where the hash each serialised diffs was keyed by not being the fingerprints generated by the Rust diff engine. That would make the Rust diff engine return results that the consumer would no longer be able to identify. Fortunately, that was a straightforward fix as well, by making the `ParsedDiff` not generate its own `diffHash` but using the one provided.

---

Now that those issues have been found and addressed, I'll wire up the route handler and new trail values worker and get this working end-to-end.

JaapRood:
An alternative implementation of the `POST **/trail-values` endpoint is now available behind the opt-in `OPTIC_RUST_SHAPE_DIFF_AFFORDANCES_LEARNER` flag. It uses the new `ShapeDiffAffordancesWorker` to leverage the new Rust learner and learn the shape diff affordances.

While results are now being produced, and while those make it all the way to the UI, they aren't correct yet. Next up, I'll debug the Rust implementation by comparing it's output to the Scala implementation's.

JaapRood:
With the changes pushed I fixed a bit of a head scratcher: the learning results yielded came back with different diff-fingerprints, despite those hashes being generated seconds earlier by the same codebase and through the same code.

Looking at it more closely, I realised that `JsonTrail`'s `normalized` field was being skipped in both serialisation and deserialisation, making each `ShapeDiffResult` hash to something different once rehydrated, while other diffs variants saw now change (didn't contain json trails!). 

The easiest fix would have been including that field in the serdes, however, it was a bit of derived state created as a result of a naive optimisation in ordering json trails (which we implemented for the initial bodies learner in #645). Making that part of the external API would not be great.

Instead, I opted to refactor both ordering and hashing (which were dependents of this `normalized` field) to work on any type of json trail and letting its consumers of the ordering and hashing traits decide whether they needed a normalised json trail or not. The only consumer where this actually seemed to matter was in the `ShapeDiffResult` hashing. With the help of some new snapshot tests and [cargo-expand](https://crates.io/crates/cargo-expand) I managed to replace `ShapeDiffResult`'s derived `Hash` trait implementation with an identical manual one, and then moved the requirement for that to happen with a normalised json trail to there. 

The result is that the diff fingerprints are correct again and the learner being able to expected fingerprints.

---

While that's fixed, there still seems to be a big difference in the output returned by the Scala and Rust implementations, where I suspect we're not returning the trail values at the leaves of the json trails, but instead its parents. I'll look at that next.

JaapRood:
Now that the overall shape is right, and we have some of the same affordances learned, it stands out that instances where learned values are 'missing' in interactions are not detected correctly. To blame here is the way we decide for what diffs and so which json trails we are trying to learn. There needs to be some matching between these two, as thousands of interactions might be consumed, all of which for endpoints that didn't concern the diffs we are trying to learn for.

While diffs give us the interaction pointers for the interactions that **triggered** the diffs, the output requires also to learn from compliant interactions. For example, in a diff for a newly added field, the UI wants to be able to show that this new field was also missing a bunch of times.

A first attempt to match incoming interactions by interaction trails failed to consider that interaction trails of diffs aren't fully qualified; interaction trails _can_ hold all the information to match an endpoint's request / response pair, but doesn't in order to keep diffs relatively compact. This approach _could_ still work if for each diff interaction trail we resolved it to become fully qualified, but that would involve a bunch of new traversing / querying code.

More straightforward for the implementation is to use the Interaction DiffTraverser and match on it's `Match*` result variants, which in fact does all the aforementioned traversing and querying of the spec for us. The one down side is that this code is setup to do this for an interaction, instead of diffs. Interactions are unbounded in size where Diffs aren't, but we're considering that a premature performance optimisation for now.

I'll look at the latter approach, next.

JaapRood:
We've got the Rust side returning correct results, with the UI rendering and using them correctly! To make this work, I took the approach outlined in the previous update: run each interaction through the endpoint differ to finding matching request / response id's and use those to map against the diffs that should be learned for.

While comparing raw output values, I actually think it helped identify a bug in the current learner. In a local example project through running against the Scala worker, one possible value for a diff is seen as "missing", even though the response doesn't match status codes. This causes the response body for a 400 to be used in the learning of a 201 response body. @acunniffe it would be great if you could confirm that this is indeed a bug, and not me misinterpreting 
Screenshot 2021-04-19 at 18 23 15

By using the same matching algorithm as the diff process, the new Rust learner doesn't exhibit this same bug, and doesn't suggest `rating` should be able to be missing in a 201 response: 
Screenshot 2021-04-19 at 18 21 24.

---

Having reached this milestone I'm going to test the implementation against a couple more test projects. I'll also rewire and enable the smoke test again (which I had to disable while making the changes described above).

acunniffe:
@JaapRood I can definitely hop in and do some of my own testings -- in general, I agree, this was a bug, and it sounds similar to something a few folks reported last month that I couldn't quite track down. 

Good two-for-one here

JaapRood:
I did a fair bit more of verification work here against my local test projects and can no longer break this. Time to get this merged.

JaapRood:
@devdoshi given our efforts of brining stuff into Optic v10, I'm going to go ahead and merge this, so I can continue on with that. I've double checked the feature flags, which means that even in the event of this being released as-is, it shouldn't have any effect. Still keen for your review, and we can deal with any feedback in a new PR / as part of moving the swarms over.

PR #666, opened 2021-04-13
Gather changes for shapes, requests, and responses

smizell:
Add resolvers to get changes for each shape ID.


workspaces/graph-lib/src/shapes-graph/index.ts
```diff
@@ -289,7 +319,7 @@ export class GraphQueries {
     throw new Error(`unexpected node.type`);
   }
 
-  wrapList(type: NodeType, nodes: Node[]): NodeListWrapper {
+  wrapList(type: NodeType | null, nodes: Node[]): NodeListWrapper {
```
smizell, line 322: I'm adding a TODO to move away from this at some point soon.

devdoshi (approved):
looks good to me, including your future plans. it's a little hard to review without the test cases/snapshots because there are a lot of edge cases but overall looks right

PR #668, opened 2021-04-20
enable `pass-exit-code` flag for tasks + test tasks

acunniffe:
## Why [Responding to User Issue]
Users trying to run their tests in CI with Optic, have asked if it's possible to pass through the exit code from their task. When Optic does not find diffs, but the tests fail for other reasons, Optic will swallow that failure and produce false positives in CI (yikes!).  

## What
We added a `--pass-exit-code` flag that will make `api` exit with the status code from a test task or a regular task the user ran when provided. 

## Validation
* [x] CI passes
* [x] Verified locally with example projects
* [ ] Works on Windows? different shells?

### DIFF
--- a/workspaces/local-cli/src/shared/spawn-process.ts
+++ b/workspaces/local-cli/src/shared/spawn-process.ts
@@ -22,3 +22,26 @@ export async function spawnProcess(
     });
   });
 }
+
+export async function spawnProcessReturnExitCode(
+  command: string,
+  env: any = {}
+): Promise<number> {
+  const taskOptions: SpawnOptions = {
+    env: {
+      ...process.env,
+      ...env,
+    },
+    shell: true,
+    cwd: process.cwd(),
+    stdio: 'inherit',
+  };
+
+  const child = spawn(command, taskOptions);
+
+  return await new Promise((resolve) => {
+    child.on('exit', (code) => {
+      resolve(code || 0);
+    });
+  });
+}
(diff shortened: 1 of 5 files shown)


acunniffe:
Pre-release build https://github.com/opticdev/optic/actions/runs/767109800

LouManglass (approved):

LouManglass:
> The only change I'd make is clarifying that in the help, which I'll do. I'll also add to the CLI article on the docs site and push that.

- https://o3c.info/docs/using/cli-commands/#run
- 
```
api run --help
run tasks from your optic.yml

USAGE
  $ api run [TASKNAME]

OPTIONS
  -c, --collect-coverage
  -d, --collect-diffs
  --exit-on-diff          When a capture session ends, if a diff is detected Optic will exit with exit code 1. This takes priority over pass-exit-code.
  --pass-exit-code        Passes through the exit code from your task (or dependent task). exit-on-diff overrides this when a diff is detected.
  --transparent-proxy
```

devdoshi (approved):
Looks good, nice additions from @LouManglass 

But - should the default be to pass through the exit code?

workspaces/local-cli/src/shared/spawn-process.ts
```diff
@@ -22,3 +22,26 @@ export async function spawnProcess(
...
+    },
+    shell: true,
+    cwd: process.cwd(),
+    stdio: 'inherit',
+  };
+
+  const child = spawn(command, taskOptions);
+
+  return await new Promise((resolve) => {
+    child.on('exit', (code) => {
```
devdoshi, line 43:
```code <number> The exit code if the child exited on its own.
signal <string> The signal by which the child process was terminated.
The 'exit' event is emitted after the child process ends. If the process exited, code is the final exit code of the process, otherwise null. If the process terminated due to receipt of a signal, signal is the string name of the signal, otherwise null. One of the two will always be non-null.```

so I'm not sure when it would exit with a signal in CI but then that's another opportunity for us to erroneously emit exit code 0

PR #669, opened 2021-04-20
Land Optic10 Workspaces

devdoshi:
## Why

## What

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


JaapRood:
FYI: when I `yarn install` this locally (with yarn v2.4.1) I get an almost completely regenerated `yarn.lock` file. Not sure if that's expected at this point, but thought to report in case it wasn't.

PR #674, opened 2021-04-21
feature: Implement cloud spec viewer

jshearer:
## Why
This introduces the cloud build variant of `ui-v2`. The goal is to provide a way for users to visually review the spec diffs generated by the gitbot. 

## What
Parameterize the entrypoint for `ui-v2`, and create another entrypoint that exposes the routing scheme `/public_specs/:specId`. When a user hits this url, a read-url is fetched from the [backend](https://github.com/opticdev/monorail/pull/231), and the contents of the spec is provided to `ui-v2`. 

## Issues
- [x] Doesn't work without events?
- [x] Does this actually need different behavior than the existing code? Aren't we trying to render diffs here, whereas by default the ui renders... whole specs? Or something?
- [x] I copied most of the CI stuff here from what I could find in `ui`. From what I can tell, this is just set up to build for one target: `specs.useoptic.com`. I'd love if we could build and deploy to staging first (`specs.o3c.info` maybe?), and then to master when ready. I'm calling with @notnmeyer to talk about this today, so hopefully we can figure this out.

## Validation
* [x] CI passes
* [x] Verified in staging


Taskfile.yml
```diff
@@ -78,13 +78,12 @@ tasks:
     cmds:
       - yarn wsrun --stages --report --fast-exit --exclude-missing ws:clean
 
-  spec:build:
-    dir: workspaces/ui
+  spec_viewer:build:
```
jshearer, line 81: Should I be... nuking the `ui` building code here?

workspaces/ui-v2/Taskfile.yml
```diff
@@ -0,0 +1,85 @@
...
+  #
+  # cloud_spec_viewer namespace
+  #
+  cloud_spec_viewer:build:
+    desc: Build cloud_spec_viewer site
+    cmds:
+      - task: :workspaces:build
+      - >
+        BUILD_PATH=build-cloud_spec_viewer
+        REACT_APP_API_URL=https://api.useoptic.com
```
jshearer, line 49: I... _think_ this might be the place to start adding staging/prod builds? I really only very tenuously understand how all of this fits together.

workspaces/ui-v2/src/spectacle-implementations/cloud-viewer.tsx
```diff
@@ -0,0 +1,163 @@
+import * as React from 'react';
```
jshearer, line 1: This is basically copy-pasted from `spectacle-implementations/local.tsx`, but I swapped out the spec fetching code. Would be good to generalize this, but I don't want to conflict with ongoing work. @devdoshi might be worth talking about the right way to generalize this?

workspaces/ui-v2/src/spectacle-implementations/cloud-viewer.tsx
```diff
@@ -0,0 +1,163 @@
...
+      let spec = await contentReq.json();
+      return spec;
+    };
+    const [example, opticEngine] = await Promise.all([
+      loadExample(),
+      import('@useoptic/diff-engine-wasm/engine/browser'),
+    ]);
+    return {
+      events: example.events,
+      samples: example.session.samples,
```
jshearer, line 52: So.. per our convo @devdoshi, I tried this with `samples: []` since you said that the gitbot won't be providing samples, but the ui seems to be unhappy with that. Is this just a "limbo" thing, or is there something worth thinking about here?

Taskfile.yml
devdoshi: on our call we discussed changing the webpack paths.js to also parameterize the output directory from the webpack build. in this previous incarnation, it was used to build them in succession into the same output directory, so it needed to be cleaned in between

workspaces/ui-v2/config/paths.js
```diff
@@ -50,14 +50,19 @@ const resolveModule = (resolveFn, filePath) => {
...
+
 // config after eject: we're in ./config/
 module.exports = {
   dotenv: resolveApp('.env'),
   appPath: resolveApp('.'),
   appBuild: resolveApp(buildPath),
   appPublic: resolveApp('public'),
   appHtml: resolveApp('public/index.html'),
-  appIndexJs: resolveModule(resolveApp, 'src/entry-points/local/index'), //@TODO: parameterize this
+  appIndexJs: resolveModule(resolveApp, entrypoints[process.env.ENTRYPOINT || "local"]),
```
devdoshi, line 65: I would namespace the env with OPTIC__ => OPTIC__WEBPACK_ENTRY_POINT or something like that

workspaces/ui-v2/src/entry-points/cloud/TopLevelRoutes.tsx
```diff
@@ -0,0 +1,20 @@
...
+import WelcomePage from '../../optic-components/onboarding/WelcomePage';
+
+export default function TopLevelRoutes() {
+  return (
+    <Switch>
+
+      <Route
+        strict
+        //@TODO: centralize this path pattern
+        path="/public_specs/:specId"
```
devdoshi, line 13: I believe in the past we've used `-` as a delimiter in urls as opposed to `_`. I believe that is more of an SEO thing but also consistency

workspaces/ui-v2/src/entry-points/cloud/TopLevelRoutes.tsx
```diff
@@ -0,0 +1,20 @@
...
+import CloudViewer from '../../spectacle-implementations/cloud-viewer';
+import WelcomePage from '../../optic-components/onboarding/WelcomePage';
+
+export default function TopLevelRoutes() {
+  return (
+    <Switch>
+
+      <Route
+        strict
+        //@TODO: centralize this path pattern
```
devdoshi, line 12: the intent of the comment here is to centralize all the root paths in an object map so it's more discoverable to developers

workspaces/ui-v2/src/spectacle-implementations/cloud-viewer.tsx
devdoshi: yep once we have the next few we'll see how they converge/diverge

workspaces/ui-v2/src/spectacle-implementations/cloud-viewer.tsx
devdoshi: can you send me a screencast of the issue?

workspaces/ui-v2/src/spectacle-implementations/cloud-viewer.tsx
```diff
@@ -0,0 +1,163 @@
...
+
+  constructor(
+    public readonly opticContext: IOpticContext,
+    public samples: any[],
+  ) {
+    this.spectaclePromise = makeSpectacle(opticContext);
+  }
+
+  async fork(): Promise<IBaseSpectacle> {
+    const opticContext = await InMemoryOpticContextBuilder.fromEventsAndInteractions(
```
devdoshi, line 105: there's another builder that doesn't use interactions, "fromEvents", that you may wish to use here, or make a new similar one

workspaces/ui-v2/Taskfile.yml
devdoshi: yep I believe the build process produces the docker artifact and publishes it to our container registry. we can then deploy that into our staging or production k8s per instructions I can send you in our slack

notnmeyer: post convo: optichub has an example of setting REACT_APP_* stuff based off branches as an environment proxy. here (or rather, where this was copied from), its being set in the taskfile because we only produced prod builds for UI _and_ there's no _deployment_ ci in monorail.

because i didnt want the optic repo to talk to our k8s cluster, the CI workflow in this repo builds and pushes to ECR.

in monorail, we define the helm chart (wrapped with taskfile), but there's no corresponding CI workflow there to do the deployment for you. the ui stuff is somewhat unique in the sense that it doesnt have automatic deployments set up. this part of the code base was slow moving enough it never became a thing we needed. can be changed for sure.

devdoshi:
@jshearer just in case, https://github.com/opticdev/optic/blob/develop/.github/workflows/build-ui.yml#L59 is the gha/CI entrypoint where the process starts. 

> Does this actually need different behavior than the existing code? Aren't we trying to render diffs here, whereas by default the ui renders... whole specs? Or something?

So, largely it should be the same as the local-cli ui, however there should be no concept of captures or diffs as noted in our Notion card, hence samples being empty and using the fromEvents builder. There are some additional requirements in the card that will also require some changes. This shared spec viewer should be primarily for sharing specs and being able to use the changelog. The ui should naturally leverage the changelog history present in the spec events, via the links we craft in the gitbot that indicate the reference point in the history for the changelog to be computed relative to

Taskfile.yml
jshearer: Sorry, I meant: these changes remove support for building the existing UI. Is that what I should be doing?

devdoshi: oh sorry. so there were previously 3 ui builds: the one used by the local-cli, the one used for the shared specs viewer, and the one used by the demo ui. This workflow only handles the shared specs and demo builds. The build used by local-cli was actually produced in the ui/package.json just before publishing, so that's not affected by these changes. (However, currently ui-v2 is not running that build, but that's a separate issue you don't need to worry about)

devdoshi: until we allocate resources for getting parity with the demo build, we should just leave the deployed one running and not do anything with ui-v2

workspaces/ui-v2/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM library/nginx:alpine
+ARG APP
```
notnmeyer, line 2: if we're only building one project from this code for now you could hardcode the value build directory on the next line. the intention here was originally to disambiguate the various builds produced from the UI code, but still allow apidocs and demo to share the docker image bits.

workspaces/ui-v2/Dockerfile
```diff
@@ -0,0 +1,5 @@
...
+COPY build-${APP}/ /usr/share/nginx/html/
+COPY config/nginx/ /etc/nginx/
```
notnmeyer, line 4: we add an nginx config (https://github.com/opticdev/optic/tree/develop/workspaces/ui/config/nginx) that sets security headers. it should largely just a drop-in, but give the CSP header a read in case there's anything to add or remove.

workspaces/ui-v2/Taskfile.yml
```diff
@@ -0,0 +1,87 @@
...
+      TAG:
+        sh: echo $(../../bin/clean-docker-tag $TAG)
+    cmds:
+      - task: docker:push
+        vars:
+          REPO: cloud_spec_viewer
+      - task: :docker:slack-notifier
+        vars:
+          TAG: "{{.TAG}}"
+          REPO: cloud_spec_viewer
```
notnmeyer, line 78: if you're pushing to a new repo, you need to create the repo first. just need to add the desired name to [this](https://github.com/opticdev/monorail/blob/master/projects/terraform/production/ecr/variables.tf) and open a pr.

workspaces/ui-v2/Dockerfile
jshearer: I do think that eventually we'll want multiple builds here: demo is eventually going to come from this, for example. So I'd rather leave it parameterized I think

Taskfile.yml
jshearer: Rebasing ontop of `swarm-1` resolves this: `ui` is gone

workspaces/ui-v2/config/paths.js
jshearer: Done

workspaces/ui-v2/src/entry-points/cloud/TopLevelRoutes.tsx
jshearer: This comment came with the copy-paste from `entry-points/local`. Is it not relevant here?

workspaces/ui-v2/Dockerfile
jshearer: Pushed!

workspaces/ui-v2/Taskfile.yml
jshearer: Excellent, love this process :D 

https://github.com/opticdev/monorail/pull/237

notnmeyer (approved):

acunniffe:
@jshearer -- I rebased to bring this up to the latest and added the config to hide various capabilities we established from #681

jshearer:
@acunniffe nice! Do you want me to take a look at the warnings that are causing the build to fail? https://github.com/opticdev/optic/runs/2450137625#step:9:359 or are you working on that? They're blocking deploying this so I can test the new [gitbot comments!!](https://github.com/opticdev/monorail/pull/242#issuecomment-827796301)

jshearer:
nvm I fixed them

acunniffe:
> nvm I fixed them

sorry about those!

workspaces/ui-v2/src/entry-points/cloud/TopLevelRoutes.tsx
devdoshi: no it's more relevant the more entry points there are. there should ideally be one place (like an object/map) where a developer can come and see the different entry points and ensure they don't collide, etc.

workspaces/ui-v2/src/spectacle-implementations/cloud-viewer.tsx
jshearer: Used it! :)

jshearer:
Just rebased ontop of `swarm-1`. Going to deploy and test out one more time, then merging!

jshearer:
Works

PR #676, opened 2021-04-22
Implement Rust learners for InMemorySpectacle

JaapRood:
## Why
We're moving to Spectacle, and we've got new Rust learners, separately. We'll want Spectacle to start using the Rust learners, with the first step being the in memory implementation doing so.

## What
Exposes function to the diff-engine-wasm interface to perform the learning and yield back the results. Extracts aggregation logic for both learners into projections so they can be re-used between interfaces.

## Validation
* [ ] CI passes
* [ ] Add more of these validation checks before going out of draft

### DIFF
--- a/workspaces/diff-engine-wasm/engine/Cargo.lock
+++ b/workspaces/diff-engine-wasm/engine/Cargo.lock
@@ -42,7 +42,7 @@ dependencies = [
  "digest",
  "libflate",
  "num-bigint",
- "rand",
+ "rand 0.4.6",
  "serde",
  "serde_json",
  "strum",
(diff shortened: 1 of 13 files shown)


JaapRood:
I've got the addition to the WASM interface compiling, now need to hook it up to Spectacle.

@devdoshi finally getting to this, but with my time off next week looming, I'm sticking to the Rust side of things first, as I reckon that's where I've got the most leverage. I'm also happy to give adding this to Spectacle a go, but won't do that until the WASM stuff is in. Maybe we could pair on the actual Spectacle stuff tomorrow (Friday), so I get a nice introduction to this and we can get this done swiftly!

PR #678, opened 2021-04-22
Save + load contributions in Optic GUI

niclim:
## Why

We want users to be able to edit contributions to their endpoints

## What
I'm going to split this into a separate PR as supporting the diff page is going to be slightly more involved. The diff page is not going to use the EditContributionsStore (since the diff page can hold pending endpoints) - the places that I need to edit are annotated with `@nic`. 

- Refactored hook usage in `/src/optic-components/documentation/Contributions.tsx` and `/src/optic-components/documentation/MarkdownBodyContributions.tsx` - all were effectively the same hook making mutations to the staging / edit store. Also removed the debounce as it provide much faster UI feedback (and given we aren't making external API calls + we aren't doing anything _too_ computationally expensive, we shouldn't need the debounce)
- Updated the PendingContributions store since it didn't handle the case where ids were different but the description was the same.
- Updated `useSpectacleCommand` to return a fn that returns a promise - since mutations are almost always going to be user triggered actions (never on render cycles)
- Removed lookupContribution - instead attached contribution loading to the field it's related to
- Fix diff viewing logic on ReviewEndpointDiffPage 
- only generated snapshots for the rust code - @devdoshi is going to handle the js snapshot generation (as there's a few other things tangential to fix there as well)

## To follow this PR
- Split the text field components for the diff + documentation page. The current components rely on being rendered in a certain context, and the diff + documentation pages have different actions they take on a input update. I'm planning on splitting the component logic from the hook / context logic.
  - This PR is here https://github.com/opticdev/optic/pull/682

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


acunniffe:
@niclim -- I really like 

> Updated useSpectacleCommand to return a fn that returns a promise - since mutations are almost always going to be user-triggered actions (never on render cycles)

I am working on the Save Changes functionality on the Optic 10 diff page and was about to look at doing something similar. This is great.

niclim:
FYI @acunniffe - I fixed a bug on the ReviewEndpointDiffPage which wasn't rendering the diffs - see https://github.com/opticdev/optic/pull/678/commits/e9eb28148787d4600189b182cb19546688e1056c - in case you were planning on working / fixing that bug

workspaces/ui-v2/src/optic-components/hooks/edit/Contributions.tsx
```diff
@@ -27,58 +26,101 @@ interface IContribution {
...
+
+  const stagePendingContribution = useCallback(
+    (
+      newId: string,
+      newContributionKey: string,
+      newValue: string,
+      initialValue: string
+    ) => {
+      const newContribution =
+        newValue !== initialValue
```
niclim, line 69: So here's an interesting bug. initialValue is set outside of this store, but the Contributions store mutates the value - what happens when we send a new contribution is that our UI is out of sync from spectacle (and we'll need to trigger a refetch) otherwise we have a stale initialValue (which causes a very minor bug when you edit, and reedit the same field).

How we fix this is either:
- We have a global store that stores the state of the app so that data + updates are in a single place (this could be something like redux, xstate for the entire app or apollo)
- Force a rerender (since the useEffect hook can't tell that data is stale)

Ultimately this is a pretty minor issue though but I think having a global store does simplify + reduce these stores of issues

workspaces/spectacle/src/index.ts
```diff
@@ -110,10 +114,23 @@ export interface IOpticContext {
...
+  | {
+      loading: false;
+      error: true;
+      data: null;
+    }
+  | { loading: false; error: false; data: T };
+
 export interface IBaseSpectacle {
-  query(options: SpectacleInput): Promise<any>;
+  query(options: SpectacleInput): Promise<AsyncStatus<any>>;
```
niclim, line 131: We can change query + mutate to be generic and pass through their generic to AsyncStatus - will update types in a separate PR since I want to keep this PR (somewhat) focused and not digress too much

workspaces/ui-v2/public/example-sessions/todos-partial.json
```diff
@@ -1785,6 +1785,94 @@
           "createdAt": "2020-10-20T20:52:31.794Z"
         }
       }
+    },
+    {
+      "BatchCommitStarted": {
```
niclim, line 1790: I'm not sure if there's a better way to generate commands (I just copy pasted from above and generated these by hand which seems to have worked)

workspaces/ui-v2/src/optic-components/documentation/ContributionGroup.tsx
```diff
@@ -58,7 +59,9 @@ function createFlatList(
       });
     }
     if (shape.asArray) {
-      contributions.push(...createFlatList(shape.asArray.shapeChoices, depth + 1));
+      contributions.push(
```
niclim, line 62: Prettier formatting here - I'm going to look into adding prettier on commit (on staged files) since it's easy to either not have prettier set up to format on save, or to forget to prettier something

workspaces/ui-v2/src/optic-components/PromptNavigateAway.tsx
```diff
@@ -0,0 +1,25 @@
+import * as React from 'react';
+import { FC, useEffect } from 'react';
+import { Prompt } from 'react-router';
+
+type PromptNavigateAwayProps = {
+  shouldPrompt: boolean;
+};
+
+export const PromptNavigateAway: FC<PromptNavigateAwayProps> = ({
```
niclim, line 9: There's two types of prompts (browser based and javascript based)
Browser prompt (closing the tab / page unload (typing a new url into the browser + navigating))
Screen Shot 2021-04-27 at 10 12 49 AM

React router navigation (clicking the docs/diff tab)
Screen Shot 2021-04-27 at 10 12 56 AM

workspaces/spectacle/src/index.ts
niclim: Ok, this actually requires more work to update the typing (since some error keys have non-boolean values - i.e. error types) - reverting for now

workspaces/spectacle/src/index.ts
```diff
@@ -110,6 +114,19 @@ export interface IOpticContext {
 
 ////////////////////////////////////////////////////////////////////////////////
 
+export type AsyncStatus<T> =
```
devdoshi, line 117: do you know a better pattern for this?

workspaces/ui-v2/public/example-sessions/todos-partial.json
devdoshi: typically we like to dump it from the ui state, so it is transitively/inductively valid

workspaces/ui-v2/src/optic-components/PromptNavigateAway.tsx
devdoshi: nice

workspaces/ui-v2/src/optic-components/PromptNavigateAway.tsx
```diff
@@ -0,0 +1,25 @@
...
+  shouldPrompt,
+}) => {
+  useEffect(() => {
+    window.onbeforeunload = shouldPrompt ? () => true : null;
```
devdoshi, line 13: my only nit is that this is destructive to any other handler

workspaces/spectacle/src/index.ts
niclim: So this pattern allows us to check `!loading` and `!error` - > which typescript will assert and know that data is _non nullable_ `T` - this is a similar pattern that react-use uses https://github.com/streamich/react-use/blob/master/src/useAsyncFn.ts#L5-L25 which I think provides us better type safety.

niclim: I think I need to modify this definition (to be more similar to react-use since my definition here doesn't include error types)

workspaces/ui-v2/src/optic-components/pages/diffs/AddEndpointsPage.tsx
```diff
@@ -177,6 +177,8 @@ export function DocumentationRootPageWithPendingEndpoints(props: any) {
                       method: endpoint.method,
                       pathId: endpoint.id,
                     })}
+                    // @nic todo make pendingendpointrow component
+                    purpose=""
```
devdoshi, line 181: - [ ] lingering todo?

workspaces/ui-v2/src/optic-components/pages/diffs/AddEndpointsPage.tsx
```diff
@@ -211,6 +213,8 @@ export function DocumentationRootPageWithPendingEndpoints(props: any) {
                       method: endpoint.method,
                       pathId: endpoint.pathId,
                     })}
+                    // @nic todo make pendingendpointrow component
```
devdoshi, line 216: - [ ] lingering todo?

workspaces/ui-v2/src/optic-components/pages/diffs/EndpointDocumentationPane.tsx
```diff
@@ -52,6 +52,8 @@ export function EndpointDocumentationPane({
         id={endpointId}
         contributionKey="purpose"
         defaultText="What does this endpoint do?"
+        // @nic todo
+        initialValue=""
```
devdoshi, line 56: - [ ] lingering todo?

workspaces/ui-v2/src/optic-components/pages/diffs/ReviewEndpointDiffPage.tsx
```diff
@@ -158,7 +137,11 @@ export function ReviewEndpointDiffPage(props: any) {
                 <DiffLinks
                   shapeDiffs={shapeDiffs.results}
                   setSelectedDiffHash={(hash: string) => {
-                    setSelectedDiffHash(hash);
+                    const hashIndex = shapeDiffs.results.findIndex(
```
devdoshi, line 140: why is an index-based approach better than the hash-based approach here?

devdoshi (changes requested):
looks good, aside from some very small comments!

devdoshi (approved):

workspaces/spectacle/src/index.ts
devdoshi: cool. yeah i also notice we've duplicated it a bit in a few files

workspaces/ui-v2/src/optic-components/pages/diffs/AddEndpointsPage.tsx
niclim: This is actually addressed in the follow up PR already https://github.com/opticdev/optic/pull/682 - it's just in there incase there's someone who view this before the linked PR has landed

workspaces/ui-v2/src/optic-components/pages/diffs/ReviewEndpointDiffPage.tsx
niclim: So it's slightly simpler in this case - the interactions on this page are:
- on mount, we want to set the current index to 0 (which is possible undefined if there are no selected hashes)
- go forward or back in current index
- go to a specific hash

Using a current index, and then computing the selected diff is a simpler way to model the state (in either scenario we have to compute either or) and it makes the initial state setting simple (since we'll always start on currentIndex=0 instead of starting on the selectedDiffHash which is possibly undefined). There's isn't a great catchall reason here other than, it takes less code + less derivation to store index over selected hash

Ideally we don't compute new state + listen to state using useEffect (instead we use useMemo or useCallback to derive state)

workspaces/ui-v2/src/optic-components/PromptNavigateAway.tsx
niclim: Hmm great catch - unfortunately I don't have a great solution for this, there's only a single onbeforeunload handler and htere's no guarantee that there'll be any ordering of cleanups (i.e. if I have two promptNavigateAways - they could remove each other in no deterministic order)

I think this is not a likely scenario since onbeforeunload is really only used for prompting - and we'd hopefully only have 1 prompt mounted at a time, but I'll add a comment here noting that this is destructive

workspaces/ui-v2/src/optic-components/pages/diffs/ReviewEndpointDiffPage.tsx
devdoshi: gotcha. yeah I didn't catch the original implementation in the diff, but as long as the assumptions of stability around the diff results holds then it is fine with me. Typically if I have a dropdown with next/prev, I'll set the next/prev when I render or change the current item so it's not computed on click

workspaces/ui-v2/src/optic-components/PromptNavigateAway.tsx
devdoshi: i'm pretty sure i've seen a library that provides a component and context supporting a stack of these types of handlers but agree it's not a likely scenario and would probably be easy to find if it collided. and we can just leave a note, i typically use //@GOTCHA or //@BUG or //@REFACTOR to mark these types of things. a quick workaround might be to check if it has a value, and console.warn that it is being overwritten? but i'm happy with it as-is, just wanted to bring up the observation and I know i've had uis where there are multiple things being edited at once

niclim: Ah I just merged this, but that's a great call about the console.warn - i'll add that into the following PR here https://github.com/opticdev/optic/pull/682

PR #681, opened 2021-04-26
Include convention for configuring various entrypoints + featureflags

acunniffe:
## Why
Our UI workspace is used in various contexts (ie Entrypoints), some of them run in the cloud, some local. At the same time, our old codebase had `REACT_APP...` feature flag scattered all over depending on what we were testing out that month. We've lost track of these in the past, and centralization feels helpful. 

## What
I added a Context that includes `config` and `featureFlags` as well as types for both. It took 10 mins to pull this together and show an example of how it could be used. 

I imagine that anything we pass in from CI with `REACT_APP....` is mapped onto the config context in one central place to keep things organized. 

Is there a better tool or convention for this? I'm open and happy to help implement it.


workspaces/ui-v2/src/optic-components/hooks/config/AppConfiguration.tsx
```diff
@@ -0,0 +1,42 @@
...
+  config: IAppConfigurations;
+};
+
+const AppConfigurationContext = React.createContext<OpticAppConfig | null>(
+  null
+);
+
+export const AppConfigurationStore = (props: {
+  config: OpticAppConfig;
+  children: any;
```
niclim, line 27: React nodes can be typed as `React.ReactNode` - or just importing `import { ReactNode } from 'react'`

Or alternatively you could type `AppConfigurationStore` as a `React.FC` - with a prop input in the type - see https://github.com/opticdev/optic/pull/678/files#diff-3389b405e926d5b93ec3fde35293c19a3453b75ad659c8e7576e51bce0d95882R5-R9 for an example

workspaces/ui-v2/src/optic-components/hooks/config/AppConfiguration.tsx
```diff
@@ -0,0 +1,42 @@
+import React, { useContext } from 'react';
+
+interface IFeatureFlags {}
+
+interface IAppConfigurations {
+  navigation: {
+    showDocs?: boolean;
+    showChangelog?: boolean;
+    showDiff?: boolean;
```
niclim, line 9: nit: It might be nice to make these non-optional (removing the `?`) since we can cooerce undefined -> false. Makes downstream checks only need to look for false / undefined (usually not an issue though)

niclim (commented):
I might be missing some context but I think this makes sense. If there's different configuration requirements based on where it's running - how about hard coding the configuration variants (and then passing in the env config through `REACT_APP...`)?

e.g. local -> has showDocs + changelog but cloud instances? (these are wrong since I'm not sure what the actual permutations are)

workspaces/ui-v2/src/optic-components/hooks/config/AppConfiguration.tsx
acunniffe: Good call -- I've updated these.

acunniffe:
> I might be missing some context but I think this makes sense. If there's different configuration requirements based on where it's running - how about hard coding the configuration variants (and then passing in the env config through `REACT_APP...`)?
> 
> e.g. local -> has showDocs + changelog but cloud instances? (these are wrong since I'm not sure what the actual permutations are)

@niclim I wanted to avoid something like a `mode` flag since we're only going to add more feature flags and config options as we go. Seems like making it easy to mix-match will serve us well. At the top of each entry point, I imagined mapping any config from `REACT_APP_...` flags. This could be used for things like analytics tokens, api base path, or feature flags.

niclim:
> > I might be missing some context but I think this makes sense. If there's different configuration requirements based on where it's running - how about hard coding the configuration variants (and then passing in the env config through `REACT_APP...`)?
> > e.g. local -> has showDocs + changelog but cloud instances? (these are wrong since I'm not sure what the actual permutations are)
> 
> @niclim I wanted to avoid something like a `mode` flag since we're only going to add more feature flags and config options as we go. Seems like making it easy to mix-match will serve us well. At the top of each entry point, I imagined mapping any config from `REACT_APP_...` flags. This could be used for things like analytics tokens, api base path, or feature flags.

I see - I'm not sure right now what sort of FFs or config we'd need (and how complex they would be) - my only concern comes from wanting to reduce the configuration permutations. The more variations of our app, the harder it is to verify the correctness of the app (i.e. testing all the variations / maintaining all combinations) and it'd be harder for us to maintain + support all the variants. If we had a mode, we would constraint it to a set number of configurations rather than all possible permutations dependent on how you run the command (especially if this is user injectable configuration). 

Configuration (such as api keys, base path, analytics) makes sense here though

workspaces/ui-v2/src/optic-components/hooks/config/AppConfiguration.tsx
devdoshi: agreed

workspaces/ui-v2/src/optic-components/hooks/config/AppConfiguration.tsx
```diff
@@ -0,0 +1,42 @@
...
+}) => {
+  return (
+    <AppConfigurationContext.Provider value={props.config}>
+      {props.children}
+    </AppConfigurationContext.Provider>
+  );
+};
+
+export function useFeatureFlags() {
+  return useContext(AppConfigurationContext)?.featureFlags!;
```
devdoshi, line 37: seems like we have some extra punctuation in a lot of places (for example the ! here), not sure where to start auditing that, but maybe walk through the ?s and !s (not just in this changeset) and the related types to make sure they can be stated more definitively

devdoshi: seems like it should be useContext(X)! if the assumption is that it will always be there

devdoshi (changes requested):
I forgot @JaapRood  was not going to weigh in this week. It's certainly fine for now if you address the comments, and we can revisit later once we've enumerated some expected variations between the different builds

workspaces/ui-v2/src/optic-components/hooks/config/AppConfiguration.tsx
niclim: One way that we could handle this is do something like...
```
export function useFeatureFlags() {
    const value = useContext(AppConfigurationContext)
    if (!value) { throw new Error("useFeatureFlags could not find AppConfigurationContext")
   return value
}
```

This way value is asserted that we have context, as well as if this is misused, the error is caught much quicker in the development cycle (rather than useFeatureFlags returning undefined instead of a boolean)

acunniffe: Done @niclim  -- but I subbed for `invariant` lib

acunniffe:
> I forgot @JaapRood was not going to weigh in this week. It's certainly fine for now if you address the comments, and we can revisit later once we've enumerated some expected variations between the different builds

I added a Twist thread to make sure we come back to this next week

workspaces/ui-v2/src/optic-components/hooks/config/AppConfiguration.tsx
niclim: That works - the only other consideration here is typescript can infer from value + error thrown blocks but can't infer through functions (i.e. you still need the `!` punctuation with `invariant` but dont need it with a if -> throw error block)

PR #682, opened 2021-04-27
Implement endpoint name contributions in diff flow

niclim:
## Why

This PR turned out a lot larger than I thought it would - there's a fair amount of refactoring involved here for several reasons annotated in the comments. Happy to discuss / revisit + hold of if there's anything controversial

Currently we have multiple ways of generating commands, and components that are coupled to the way they generate commands (useContributions, etc). The diff components need to talk to the XState store and the documentation components need to talk to the contributions store. These are rendered in different contexts, but the current components are context reliant. I think moving further towards a distinction between page level components (which have more context reliant state / specific page logic) vs composable shared components (such as text fields / endpoint name renderes, etc) will help us iterate faster by allowing better reuse of code.

A structure for the FE that I think might help solve this issue is:
- App + app level context
  - This could be a global store, or a single place to derive CQRS commands from (as well as ways to trigger refetches)
  - Spectacle implementation + bridges between the different contexts here
  - Any "globally" used components + hooks (see `useDebouncedFn` or `useStateWithSideEffect`)
- Pages (see `ReviewEndpointDiffsPage`)
  - Anything that is page specific (like components or hooks used / specialized more for that specific context) - this likely will use global components / hooks

I'd like to get feedback from @JaapRood when he's back as well and see what his thoughts are but I think to get this work done this was somewhat necessary to implement a small part (twist thread to follow!)

## What's still left to do
- Loading state for diff page
- Spectacle mutation -> figure out how to retrigger a gql requery (i can think of a couple of hacky solutions but they aren't great)
- Handling PathParameters + contributions

## What
- Implemented diff name contributions (which uses the xstate store instead of the contributions store)
- Some refactors + splitting up components such that they don't rely on specific context stores. 

Screen Shot 2021-04-28 at 4 14 03 PM
Screen Shot 2021-04-28 at 4 14 25 PM
Screen Shot 2021-04-28 at 4 14 54 PM

## Other notes:
- On the undocumented endpoints page - there seems to be some race condition where sometimes the page loads and there is no endpoint (this could be for a number of reasons, such as useEffects / useMemos with missing dependency arrays, state resolving and not being deterministic) - this is something I'll need to follow up and track down. I've seen it resolve and not resolve (even without my changes)
see
Screen Shot 2021-04-28 at 11 28 00 AM

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


workspaces/ui-v2/src/optic-components/pages/diffs/AddEndpointsPage.tsx
```diff
@@ -163,23 +167,40 @@ export function DocumentationRootPageWithPendingEndpoints(props: any) {
             {pendingEndpointsToRender.map(
               (endpoint: IPendingEndpoint, index: number) => {
                 return (
-                  <EndpointRow
+                  <ListItem
```
niclim, line 170: Removed EndpointRow since it uses _mostly_ precomposed components (and that there were slight variations - e.g. changelog + different children). I think it's potentially easier / clearer to remove this layer of abstraction + it means I don't need to build a `DiffEndpointRow` and a `DocumentationEndpointRow`

workspaces/ui-v2/src/optic-components/common/EditableTextField.tsx
```diff
@@ -0,0 +1,96 @@
...
+  defaultText?: string;
+  variant: TextFieldVariant;
+};
+
+export enum TextFieldVariant {
+  SMALL,
+  REGULAR,
+}
+
+export const EditableTextField: FC<EditableTextFieldProps> = ({
```
niclim, line 21: So this isn't very generic / reusable - but I figured the best way to start creating reusable components is to copy + paste + create variants until we know what a common pattern is (or when we want to implement some consistency around our UI / design). Basically I figured copy paste while just hard coding different variants until it becomes too crazy seems like a good middle ground here

workspaces/ui-v2/src/optic-components/common/EndpointName.tsx
```diff
@@ -0,0 +1,51 @@
+import React from 'react';
```
niclim, line 1: Moved from `/src/optic-components/documentation/EndpointName.tsx` - this is effectively a "common" component that we can use wherever we need to render an endpoint name

workspaces/ui-v2/src/optic-components/diffs/render/ApproveAll.tsx
```diff
@@ -33,6 +33,9 @@ export default function ApproveAll(props: { disabled?: boolean }) {
   };
 
   const handleSave = () => {
+    if (shapeDiffs.loading) {
```
niclim, line 36: This shouldn't be reachable (since we shouldn't reach this flow before having shapes to approve) - but if it does get reached - do we have a way of sending errors (or sentry alerts)?

workspaces/ui-v2/src/optic-components/diffs/render/AskForCommitMessage.tsx
```diff
@@ -29,14 +37,28 @@ export default function AskForCommitMessage(props: { hasChanges: boolean }) {
 
   const pendingEndpointsCount = pendingEndpoints.filter((i) => i.staged).length;
 
-  const handleSave = () => {
-    //@todo decide if every endpoint (changes/pending) gets a batch commit id.
+  const handleSave = async () => {
+    // TODO figure out how to attach a commit message to match id
```
niclim, line 41: So there's a commit message in the UI, but I'm not sure where this should be used? I don't see anywhere in mutations / projections where we can attach this to?

workspaces/ui-v2/src/optic-components/diffs/render/AskForCommitMessage.tsx
```diff
@@ -29,14 +37,28 @@ export default function AskForCommitMessage(props: { hasChanges: boolean }) {
...
+        }
+      }
+      `,
+      variables: {
+        commands,
+      },
+    });
+    // If there are no batch commits (first commit) - link to the just created commit
+    history.push(changelogPageRoute.linkTo(lastBatchCommitId || batchCommitId));
+    // @nic TODO we need to trigger a new full page refresh - or trigger a refetch of data through out queries
```
niclim, line 61: Since we redirect + our gql query hasn't retriggered, we can either do a full page refresh (i.e. window.location.push) or figure out a way to programatically retrigger fetches (there's some hacky ways, and some less hacky ways that might require more effort)

workspaces/ui-v2/src/optic-components/documentation/PathParameters.tsx
```diff
@@ -40,7 +40,7 @@ export function PathParametersViewEdit(props: PathParametersViewEditProps) {
             name={param.pathComponentName}
             shapes={[alwaysAString]}
             depth={0}
-            // @nic todo
+            // TODO - implement when query params are added
```
niclim, line 43: I believe @acunniffe said there was a new resolver to use here? If so, I can update this in this PR (otherwise I'll follow up with this)

workspaces/ui-v2/src/optic-components/hooks/diffs/LearnInitialBodiesMachine.ts
```diff
@@ -166,7 +166,7 @@ export function recomputeCommands(ctx: InitialBodiesContext): any[] {
     commands.push(
       AddContribution(
         getEndpointId({ method: ctx.method, pathId: ctx.pathId }),
-        'PURPOSE',
+        'purpose',
```
niclim, line 169: I'm not 100% here, but I believe this should be lowercase (spectacle / rust seems to accept lowercase for me)

workspaces/ui-v2/src/optic-components/hooks/diffs/SharedDiffContext.tsx
```diff
@@ -166,6 +167,22 @@ export const SharedDiffStore = (props: SharedDiffStoreProps) => {
...
+        type: 'SET_ENDPOINT_NAME',
+        id,
+        command,
+      }),
+    setPendingEndpointName: (id: string, name) => {
+      const pendingEndpoint = context.pendingEndpoints.find((i) => i.id === id);
+      if (!pendingEndpoint) {
+        return console.error(`Could not find pending endpoint with id ${id}`);
+      }
+      pendingEndpoint.ref.send({ type: 'STAGED_ENDPOINT_NAME_UPDATED', name });
```
niclim, line 181: Here we delegate to the other state machine which is used to generate the pending endpoint commands - there's better type safety we can add here though. We need to trigger a recompute commands here in the top level xstate store though

workspaces/ui-v2/src/optic-components/hooks/util/useDebouncedFn.ts
```diff
@@ -0,0 +1,9 @@
+import { useMemo } from 'react';
+import debounce from 'lodash.debounce';
+
+export const useDebouncedFn = <T extends (...args: any) => any>(
```
niclim, line 4: There's an existing useDebounce hook somewhere in the code - but I think it's used to debounce changes to a state object - in this case we want a fn to only run in a certain instance (and we need to wrap it in a useMemo otherwise the identity between renders change and it doesn't actually debounce

workspaces/ui-v2/src/optic-components/pages/diffs/ReviewEndpointDiffPage/ReviewEndpointDiffPageContainer.tsx
```diff
@@ -0,0 +1,50 @@
...
+    context.results.trailValues
+  );
+
+  //
+  // const newBodyDiffs = useNewBodyDiffInterpretations(
+  //   endpointDiffs.newRegionDiffs,
+  // );
+
+  return !endpoint || shapeDiffs.loading ? (
+    // @nic todo add in this loading state
```
niclim, line 39: @acunniffe do you know if we have a generic / standard loading state we can throw in?

workspaces/ui-v2/src/optic-components/pages/diffs/ReviewEndpointDiffPage/ReviewEndpointDiffPageContainer.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+import { useEndpointDiffs } from '../../../hooks/diffs/useEndpointDiffs';
+import { useShapeDiffInterpretations } from '../../../hooks/diffs/useDiffInterpretations';
+import { useSharedDiffContext } from '../../../hooks/diffs/SharedDiffContext';
+import { useEndpoint } from '../../../hooks/useEndpointsHook';
+import { SpectacleContext } from '../../../../spectacle-implementations/spectacle-provider';
+
+import { ReviewEndpointDiffPage } from './ReviewEndpointDiffPage';
+
+export const ReviewEndpointDiffContainer: FC<
```
niclim, line 13: I split this page component into multiple components for a couple of reasons.

We want to set initial state on the initial render for the initial contribution. If the endpoint is `null` - i.e. still loading, we don't actually inject the old contribution.

As a result, I updated this to have a loading state + shard this into multiple pages in an attempt to:
- handle typing and loading states more upfront (so that we don't silently swallow errors and force us to have stronger data constraints)
- Hopefully split up concerns so that this component is easier to digest what it does

workspaces/ui-v2/src/optic-components/pages/diffs/ReviewEndpointDiffPage/ReviewEndpointDiffPage.tsx
```diff
@@ -0,0 +1,249 @@
...
+
+type ReviewEndpointDiffPageProps = {
+  endpoint: IEndpoint;
+  spectacle: IForkableSpectacle;
+  shapeDiffs: IInterpretation[];
+  method: string;
+  pathId: string;
+};
+
+export const ReviewEndpointDiffPage: FC<ReviewEndpointDiffPageProps> = ({
```
niclim, line 52: I _usually_ prefer to have 1 component per file (but I try not to be dogmatic about it) - I didn't do this here since I'm not sure what eveyone's thoughts are around having component "folders" for each page (or subpages)

workspaces/ui-v2/src/optic-components/documentation/PathParameters.tsx
niclim: Ah this isn't query parameters - this is path parameters - I'll update this - but I'm going to again open a separate PR to tackle this work

acunniffe: sorry for typing the wrong thing in that comment

workspaces/ui-v2/src/optic-components/hooks/diffs/LearnInitialBodiesMachine.ts
acunniffe: Yes. It should be. We may want to make a constants file somewhere so we’re not reusing the string directly

workspaces/ui-v2/src/optic-components/diffs/render/AskForCommitMessage.tsx
acunniffe: @devdoshi probably has good thoughts on how we should approach this. 

We definitely want the UI to work without a forced refresh so our example sessions and demo sessions that run in the browser continue functioning as expected. A full refresh here would nuke their state w/o some local storage or other persistence

niclim: Agreed - we definitely want to not full refresh, however since we use CQRS we inherently need to trigger requeries on all our gql things. 

My initial thoughts are we could expose something from the spectacle provider to "refetch all" / invalidate all queries (which is the most correct since any request could modify any query) - and if that proves challenging (or a lot of work) there are slightly hackier ways we can do this as well

workspaces/ui-v2/src/optic-components/pages/diffs/AddEndpointsPage.tsx
acunniffe: I think this makes sense, as well as the architecture choices motivating it.

workspaces/ui-v2/src/optic-components/diffs/render/AskForCommitMessage.tsx
acunniffe: looks like we still need to expand `applyCommands(commands: [JSON]): AppliedCommandsResult` to handle the commitMessage

it was hardcoded in spectacle mutation. 
```
      applyCommands: async (parent: any, args: any, context: any) => {
        const batchCommitId = uuidv4();
        const events = await opticContext.specRepository.listEvents();
        try {
          const newEventsString = opticContext.opticEngine.try_apply_commands(
            JSON.stringify(args.commands),
            JSON.stringify(events),
            batchCommitId,
+            'proposed changes'
          );
```

workspaces/ui-v2/src/optic-components/hooks/diffs/SharedDiffContext.tsx
acunniffe: > there's better type of safety we can add here though

It's been on my maintenance list to figure out how to get the [event / schema typings to pass through](https://xstate.js.org/docs/guides/typescript.html#using-typescript) -- it hasn't been able to infer as much as I was promised. Happy to pair on it sometime

acunniffe (approved):

acunniffe:
Looks good @niclim  -- I tried things out locally too and can report that functionally, this does what I'd expect, and code-wise, good improvements that I'm glad we got into the UI early. 

We should definitely bring @JaapRood in next week to get some principles written down for a "how we frontend" session that we can share among ourselves, outside contributors, and employees.

workspaces/ui-v2/src/optic-components/diffs/render/AskForCommitMessage.tsx
niclim: Ahh i see - I'll also need to add a commit message into the documentation contribution flow. I'll also address this in a follow up PR

workspaces/ui-v2/src/optic-components/hooks/diffs/SharedDiffContext.tsx
niclim: I think we need to pass in the `ref` type which is the pendingEndpoint state machine? I'll need to spend some time digging through and understanding the xstate typescript typing though

workspaces/ui-v2/src/optic-components/common/EditableTextField.tsx
JaapRood: Yes! No abstraction > the wrong abstraction, so when not sure, copying code is the way to go :)

PR #685, opened 2021-04-27
refactor: Include `pathId` in response from `generateEndpointChanges`

jshearer:
## Why
The gitbot needs `pathId` in order to generate the deep links correctly

## What
This exposes the new `pathId` field on `EndpointChange` to the consumers of `generateEndpointChanges`

### DIFF
--- a/workspaces/changelog/src/index.ts
+++ b/workspaces/changelog/src/index.ts
@@ -39,6 +39,7 @@ export async function generateEndpointChanges(
           change {
             category
           }
+          pathId
           path
           method
         }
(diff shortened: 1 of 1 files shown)


jshearer:
@devdoshi will merging this trigger a side-channel build? If so, how can I use that in the gitbot? Do I have to... edit `yarn.lock` or something?

devdoshi:
> @devdoshi will merging this trigger a side-channel build? If so, how can I use that in the gitbot? Do I have to... edit `yarn.lock` or something?

@jshearer merging will not trigger a side-channel build, those are only manually triggered (and you would specify swarm-1 as the target branch, once you merge)

jshearer:
Ah, gotcha. Thanks!

smizell (approved):

PR #687, opened 2021-04-28
Add the gitbot comment header image to public hosting

jshearer:
This change will have useoptic.com host the included gitbot comment header.

@acunniffe sent me a transparent image, but it still looks fairly dark when used in github's official dark theme. Thoughts?

Screen Shot 2021-04-28 at 12 30 13 PM


LouManglass:
> This change will have useoptic.com host the included gitbot comment header.
> 
> @acunniffe sent me a transparent image, but it still looks fairly dark when used in github's official dark theme. Thoughts?
> 
> Screen Shot 2021-04-28 at 12 30 13 PM

I too am a dark theme user (though I just switched over to "dark dimmed" for a bit). I suspect this looks better on "default light", and to that end...I'm not sure there's a good single image compromise for all themes.

I'll let folks chime in, and if we decide to ship as-is, I can deploy it.

acunniffe:
changelog

Let's bring back the one with the background, I don't think it's super abrasive for dark mode or this one with lighter colors 

changelogpng

Tiny choices, make it in under a min @jshearer 

We should use a better file path for this, since we're planning on it being there for a long time and who knows if /website will stay in here. maybe /assets/changelog_header.png

jshearer:
I like the transparent one with lighter colors! `/assets/changelog_header.png` also works for me.

jshearer:
Wrt hosting, I _think_ we're probably okay to just link straight to github's `raw.github.com` for this, unless there's a reason not to? For example: https://raw.githubusercontent.com/opticdev/optic/72fb22810216fa4925fa696688c5ebf848feb19c/assets/changelog_header.png (actually will be https://raw.githubusercontent.com/opticdev/optic/master/assets/changelog_header.png once merged)

PR #688, opened 2021-04-28
Update response resolver for requests query

smizell:
This updates the HttpRequest.responses resolver to return only responses related to the parent request.


devdoshi (approved):
Looks good - if it's awkward/confusing we can add more nodes/edges to the graph to make the relationships more straightforward

PR #689, opened 2021-04-29
Add prettier or rustfmt on commit

niclim:
## Why

It's easy to forget to run the formatter (rust or prettier) - adding this on commit will make this something we don't need to nit about.

## What

- This will only run on files that you have changed during in your commits (incremental runs)
- This uses `prettier` and `rustfmt` and are required in order to run / commit

To run husky or to get the githooks installed we have to run `yarn run husky install` - otherwise nothing will happen when you commit. Added this to the task file
`task workspaces:setup`
Screen Shot 2021-04-28 at 5 59 04 PM

`git commit`
Screen Shot 2021-04-28 at 5 58 37 PM

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


devdoshi (approved):
what are your thoughts on the CI side of this?

niclim:
@devdoshi so I'm not sure CI is the right place for this, once this reached CI, there's already commits that have been generated - so precommit is a better place for this to live. However we _might_ want to add in a CI check to see that it is formatted using prettier / rustfmt (as well as linting steps) but i think that comes later once this has been running for a while

PR #690, opened 2021-04-29
Expose shape -> commands over WASM bridge

acunniffe:
## Why
Old Optic domain was still powering the part of the code that was used to generate commands from examples. That logic is pretty straightforward, but it's still a foot in the past, and we want everything using Rust going forward. Now it does

## What
I had to update the interface of `into_commands` to accommodate these changes. Since we want to be able to create a new set of commands that can be applied to any id/trail, we need to also return the `new_shape_id`, that is, the ID of the top-most shape created. This will be a key utility for design-by-examples as well as the diff use cases. 

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


workspaces/diff-engine/src/learn_shape/result.rs
```diff
@@ -56,13 +60,19 @@ impl TrailObservationsResult {
...
+    let mut new_shape_id: Option<String> = None;
 
     for json_trail in sorted_trails.into_iter().rev() {
       let trail_values = self.values_by_trail.remove(&json_trail).unwrap();
 
       let shape_prototype =
         trail_values.into_shape_prototype(id_generator, &shape_prototypes_by_trail);
 
+      // collect the id of the first shape prototype we create
+      if new_shape_id.is_none() {
```
acunniffe, line 72: @devdoshi is there a better Rust way of doing this?

devdoshi: we preallocate the vec of shape_prototypes so you could clone it from the last vec element after the iteration is done, instead of checking it every iteration

devdoshi: Also, should it be the first or the last?

workspaces/ui-v2/src/lib/shape-diffs/build-inner-shape.ts
```diff
@@ -37,24 +37,30 @@ export function builderInnerShapeFromChoices(
...
   const innerShapeIds = Array.from(targetKinds).map((i) => {
     const foundShapeId = Object.entries(allowedKindsById).find(
-      ([key, shape]) => shape === i,
+      ([key, shape]) => shape === i
     );
 
     if (foundShapeId) {
       return foundShapeId[0];
     } else {
+      //@aidan remove this after using the proper trail learners from Rust
```
devdoshi, line 46: is this still pending?

devdoshi (approved):
approved assuming the newShapeId is correct

workspaces/diff-engine/src/learn_shape/result.rs
```diff
@@ -828,6 +854,7 @@ mod test {
     let collections_results =
       collect_commands(collections_observations.into_commands(&mut test_id_generator));
     assert!(collections_results.0.is_some());
+    assert!(collections_results.0.is_some());
```
JaapRood, line 857: Dupe!

workspaces/diff-engine/src/learn_shape/result.rs
```diff
@@ -745,6 +765,7 @@ mod test {
     let nullable_primitive_results =
       collect_commands(nullable_primitive_observations.into_commands(&mut test_id_generator));
     assert!(nullable_primitive_results.0.is_some());
+    assert!(nullable_primitive_results.0.is_some());
```
JaapRood, line 768: Should be checking the `.2`?

JaapRood, line 768: Should be asserting on `.2` instead?

workspaces/diff-engine/src/learn_shape/result.rs
```diff
@@ -754,6 +775,7 @@ mod test {
     let nullable_object_field_results =
       collect_commands(nullable_object_field_observations.into_commands(&mut test_id_generator));
     assert!(nullable_object_field_results.0.is_some());
+    assert!(nullable_object_field_results.0.is_some());
```
JaapRood, line 778: Should be asserting on `.2` instead?

workspaces/diff-engine/src/learn_shape/result.rs
```diff
@@ -763,6 +785,7 @@ mod test {
     let nullable_array_item_results =
       collect_commands(nullable_array_item_observations.into_commands(&mut test_id_generator));
     assert!(nullable_array_item_results.0.is_some());
+    assert!(nullable_array_item_results.0.is_some());
```
JaapRood, line 788: Should be asserting on `.2` instead?

workspaces/diff-engine/src/learn_shape/result.rs
```diff
@@ -772,6 +795,7 @@ mod test {
     let nullable_one_off_results =
       collect_commands(nullable_one_off_observations.into_commands(&mut test_id_generator));
     assert!(nullable_one_off_results.0.is_some());
+    assert!(nullable_one_off_results.0.is_some());
```
JaapRood, line 798: Should be asserting on `.2` instead?

workspaces/diff-engine/src/learn_shape/result.rs
```diff
@@ -782,6 +806,7 @@ mod test {
       collect_commands(only_null_observations.into_commands(&mut test_id_generator));
     assert!(only_null_results.0.is_some());
     assert_valid_commands(only_null_results.1.clone());
+    assert!(only_null_results.0.is_some());
```
JaapRood, line 809: Should be asserting on `.2` instead?

workspaces/diff-engine/src/learn_shape/result.rs
```diff
@@ -819,6 +844,7 @@ mod test {
     let primitive_results =
       collect_commands(primitive_observations.into_commands(&mut test_id_generator));
     assert!(primitive_results.0.is_some());
+    assert!(primitive_results.0.is_some());
```
JaapRood, line 847: Should be asserting on `.2` instead?

workspaces/diff-engine/src/learn_shape/result.rs
JaapRood, line 857: Should be asserting on `.2` instead?

JaapRood (commented):
Nice one @acunniffe, this looks reasonable and fits right in. I did find some test assertions that don't seem correct, might be worth the quick fix.

PR #691, opened 2021-04-29
Add in commit messages on saving / accepting diffs through spectacle

niclim:
## Why

Users want to be able to add commit messages to their changes

## What

Talking with @acunniffe I think it might not make that much sense to include commit messages for contributions (since contributions likely tie to versions) - for now we'll add in commit messages since contributions are part of our versioning + batch commits, but in the future we should revisit this.

## Screenshots!
Screen Shot 2021-04-29 at 3 00 33 PM
Screen Shot 2021-04-29 at 3 01 08 PM

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


workspaces/ui-v2/src/optic-components/diffs/contexts/SimulatedCommandContext.tsx
```diff
@@ -27,14 +27,15 @@ export function SimulatedCommandStore(props: SimulatedCommandStoreProps) {
...
-  applyCommands(commands: $commands) {
+mutation X($commands: [JSON], $commitMessage: String) {
+  applyCommands(commands: $commands, commitMessage: $commitMessage) {
     batchCommitId
   }
 }
         `,
         variables: {
           commands: props.previewCommands,
+          commitMessage: '',
```
niclim, line 38: I'm not entirely sure why / when this runs (seems to run on the new diff page) - but this needs to be shimmed since we aren't automatically adding a new commit message

workspaces/ui-v2/src/optic-components/common/CommitMessageModal.tsx
```diff
@@ -0,0 +1,78 @@
+import React, { ChangeEvent, FC, useState } from 'react';
```
niclim, line 1: Pulled out from `/src/optic-components/diffs/render/AskForCommitMessage.tsx` and pulled out the generic UI part (since this is now used in two contexts)

acunniffe (approved):

PR #692, opened 2021-04-30
Feature/add path parameters

niclim:
## Why

## What

Note that this PR is branched off of https://github.com/opticdev/optic/pull/691 - as it requires updating the commit message diff modal (number of pending changes)

- Updated to use the fullPathWithParameterizedNames (i.e. `/todo/{todoId}` instead of just `/todo/{}`
- Add ability to add contributions to the path on the review diff page and the documentation page
- Fixed a regression on the ReviewEndpointDiffPage (not showing the highlighted location

### Docs page editing
Screen Shot 2021-04-30 at 12 47 54 PM

### Diff page editing
Screen Shot 2021-04-30 at 12 48 12 PM

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


workspaces/ui-v2/src/optic-components/pages/diffs/ReviewEndpointDiffPage/ReviewEndpointDiffPage.tsx
```diff
@@ -106,6 +103,11 @@ export const ReviewEndpointDiffPage: FC<ReviewEndpointDiffPageProps> = ({
           <EndpointDocumentationPane
             method={method}
             pathId={pathId}
+            highlightedLocation={
```
niclim, line 106: I realized I had a regression here when I split up this component - it now is fixed on the ReviewEndpointDiffPage... It should look something like this:
Screen Shot 2021-04-30 at 12 41 28 PM

workspaces/ui-v2/src/optic-components/common/EditableTextField.tsx
```diff
@@ -16,6 +16,7 @@ type EditableTextFieldProps = {
 export enum TextFieldVariant {
   SMALL,
   REGULAR,
+  FIELDORPARAM,
```
niclim, line 19: This is really hard coded but I think this is a start to consolidate the usages of textfields - ideally we have a somewhat consistent style of input field that can be adapted to different containers.

I think these text fields are _basically_ identical apart from a couple of interactions / styles - we can consolidate them later though

workspaces/ui-v2/src/optic-components/diffs/render/AskForCommitMessage.tsx
```diff
@@ -29,6 +27,9 @@ const useStagedChangesCount = () => {
       (diffHash) => diffHashToEndpoint[diffHash]
     ),
     ...Object.keys(context.choices.existingEndpointNameContributions),
+    ...Object.values(context.choices.existingEndpointPathContributions).map(
+      (pathContribution) => pathContribution.endpointId
```
niclim, line 31: There's a bit of juggling / passthrough for data here - I think here is where a global store would really help to simplify these data derivations (such that you have a single source of truth - especially since our endpoints / paths / shapes are modeled as a graph)

The FE here cares about which endpoints were modified (which is not super trivial to do with the current structure)

workspaces/ui-v2/src/optic-components/diffs/render/AskForCommitMessage.tsx
```diff
@@ -74,7 +75,6 @@ export default function AskForCommitMessageDiffPage(props: {
     });
     // If there are no batch commits (first commit) - link to the just created commit
     history.push(changelogPageRoute.linkTo(lastBatchCommitId || batchCommitId));
-    // @nic TODO we need to trigger a new full page refresh - or trigger a refetch of data through out queries
```
niclim, line 77: Turns out I was wrong and I think this happens already!

acunniffe (approved):

PR #695, opened 2021-04-30
side-channel: ui-v2

LouManglass:
## Why

- Add in the [prepare](https://docs.npmjs.com/cli/v6/using-npm/scripts#prepare-and-prepublish) task following `ui` workspace to build interface for packages.

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/ui-v2/package.json
+++ b/workspaces/ui-v2/package.json
@@ -18,7 +18,8 @@
     "build": "build:local",
     "test": "node scripts/test.js",
     "ws:clean": "echo 'cleaning' ",
-    "ws:build": "echo 'build'"
+    "ws:build": "echo 'build'",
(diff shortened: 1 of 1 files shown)


devdoshi:
@LouManglass thoughts on how to keep CI='' for side-channel builds but enabled for production builds?

LouManglass:
> @LouManglass thoughts on how to keep CI='' for side-channel builds but enabled for production builds?

How do we do that today with the `ui` workspace? Follow-up, what are the growing pains there - we don't need to keep to this pattern, but I'd like to have more discussion around what would work better for us if I'm going to weigh in on a change. Happy to have a call on it, if that's more efficient.

LouManglass:
We actually don't need this anyway, due to the work that came in yesterday. Closing.

PR #697, opened 2021-04-30
Fix some react warnings

niclim:
## Why

The react warnings were annoying me, so I wanted to fix them.

## What

Fixed two types of warnings (well 3 depending on your definition of fix)
- `Function components cannot be given refs` - this is because of how material ui expects Menu and MenuItem to be rendered (it passes in refs) - so the direct children are cloned and injected refs. The easy fix here is to just change the component structure slightly
- `Failed prop-type` - change to use determinate instead of static
- `Can't perform a React state update on an unmounted component` - this is because we are setting state after the component unmounts due to a remount (?) - (not entirely sure about the interactions here) - we can "fix" this by throwing in a shouldCancel block. But the better solution here is to load up all data in a provider and pass it down using a hook - this somewhat fits into the idea of a global state - I think this is somewhat worth tackling to simplify and not have to handle loading / error states when any `useSpectacleQuery` hook is used so I'll lean towards using that as a fix.

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/ui-v2/src/optic-components/changelog/ChangelogDropdown.tsx
+++ b/workspaces/ui-v2/src/optic-components/changelog/ChangelogDropdown.tsx
@@ -58,9 +58,6 @@ export function ChangesSinceDropdown(props: any) {
       <ToggleButton
         value="check"
         selected={false}
-        // onClick={() => {
-        //   isEditing ? save() : setEditing(true);
-        // }}
         size="small"
         className={classes.button}
         onClick={handleClick}
@@ -79,9 +76,8 @@ export function ChangesSinceDropdown(props: any) {
         style={{ marginTop: 20 }}
       >
         {allBatchCommits.map((i, index) => (
-          <BatchCommitMenuItem
-            batch={index === 0 ? { ...i, commitMessage: 'Latest' } : i}
-            key={index}
+          <MenuItem
+            key={i.batchId}
             onClick={() => {
               if (index === 0) {
                 history.push(documentationPageLink.linkTo());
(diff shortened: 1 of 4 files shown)


workspaces/ui-v2/src/optic-components/changelog/ChangelogDropdown.tsx
```diff
@@ -79,62 +76,57 @@ export function ChangesSinceDropdown(props: any) {
         style={{ marginTop: 20 }}
       >
         {allBatchCommits.map((i, index) => (
-          <BatchCommitMenuItem
-            batch={index === 0 ? { ...i, commitMessage: 'Latest' } : i}
-            key={index}
+          <MenuItem
+            key={i.batchId}
```
niclim, line 80: We shouldn't actually use the index as a key - the way react works is that it needs an id that maps deterministically to an item - using the index could lead to bugs where we rerender the wrong item (i believe)

workspaces/ui-v2/src/optic-components/hooks/useCapturesHook.tsx
```diff
@@ -8,35 +8,43 @@ export function useCaptures(): {
   error?: any;
 } {
   const capturesService = useContext(CapturesServiceContext)!;
-  const [captures, setCaptures] = useState<ICapture[] | null>(null)
+  const [captures, setCaptures] = useState<ICapture[] | null>(null);
```
niclim, line 11: prettier strikes again

devdoshi (approved):
looks good - not sure how those index keys made it in!

PR #698, opened 2021-04-30
Fix webpack module resolution alias

niclim:
## Why

It looks like this was attempted but it wasn't fully set up

## What

- Update webpack config + stuff to handle <src> alias
- Also organize package.json into dependencies + dev dependencies

## Validation
I tested `yarn start` and `yarn build:local` and it seems to have worked fine resolving the `<src>` alias


workspaces/ui-v2/config/modules.js
```diff
@@ -65,28 +65,7 @@ function getWebpackAliases(options = {}) {
...
-    };
-  }
-}
-
-/**
- * Get jest aliases based on the baseUrl of a compilerOptions object.
- *
- * @param {*} options
- */
-function getJestAliases(options = {}) {
```
niclim, line 78: So this isn't currently used, and I don't believe jest actually runs through webpack (if it does, it'd use the same config). Jest runs off node and i _think_ it uses it's own module dependency resolution (which I specified in `jest.config.js`)

Deleting this to try to avoid confusion (webpack config is already confusing :/)

workspaces/ui-v2/jest.config.js
```diff
@@ -0,0 +1,44 @@
...
+  },
+  transformIgnorePatterns: [
+    '[/\\\\]node_modules[/\\\\].+\\.(js|jsx|mjs|cjs|ts|tsx)$',
+    '^.+\\.module\\.(css|sass|scss)$',
+  ],
+  modulePaths: [],
+  moduleNameMapper: {
+    '^react-native$': 'react-native-web',
+    '^.+\\.module\\.(css|sass|scss)$': 'identity-obj-proxy',
+    '^<src>/(.*)$': '<rootDir>/src/$1',
```
niclim, line 25: Copied out of package.json and added this line to line up with src

workspaces/ui-v2/tsconfig.json
```diff
@@ -4,6 +4,10 @@
     "jsx": "react",
     "outDir": "build",
     "resolveJsonModule": true,
-    "rootDir": "src"
+    "rootDir": "src",
+    "baseUrl": ".",
```
niclim, line 8: I could only get this to work when i specified _both_ the tsconfig + webpack resolve.alias fields. I'm not sure why we need to double specify this and it'll be confusing since we'll need to add this in two locations if we want to update stuff???

workspaces/ui-v2/package.json
```diff
@@ -15,65 +15,35 @@
     "start": "node scripts/start.js",
     "build:local": "env OPTIC_WEBPACK_ENTRYPOINT=local node scripts/build.js",
     "build:cloud": "env OPTIC_WEBPACK_ENTRYPOINT=cloud node scripts/build.js",
-    "build": "build:local",
+    "build": "yarn run build:local",
```
niclim, line 18: I don't believe this command is directly called, but it wasn't working for me so i updated to what i think it should call

workspaces/ui-v2/config/modules.js
```diff
@@ -65,28 +65,7 @@ function getWebpackAliases(options = {}) {
...
-  if (!baseUrl) {
-    return {};
-  }
-
-  const baseUrlResolved = path.resolve(paths.appPath, baseUrl);
-
-  if (path.relative(paths.appPath, baseUrlResolved) === '') {
-    return {
-      '^src/(.*)$': '<rootDir>/src/$1',
+      '<src>': paths.appSrc,
```
niclim, line 68: I called this `<src>` but if there's a more preferred convention I'm happy to change it - the only other convention is `@` but that seems more appropriate for packages

devdoshi:
looks good, we should probably do that in many of our workspace packages!

niclim:
> looks good, we should probably do that in many of our workspace packages!

I think the ui-v2 package is the only one that uses webpack and the other packages are fairly small. I'm going to add this to the TODO list but couldn't figure out how to get this added into the tsconfig globally - a quick attempt to add this into the global tsconfig was not successful

niclim:
I'm going to merge this in - I believe the failing build is from an upstream commit and looks like it'll be addressed this this PR? https://github.com/opticdev/optic/pull/700#issuecomment-831376205

PR #700, opened 2021-05-03
added a jsontrail to the into_commands signature

acunniffe:
## Why
An assumption was baked into the into_commands function: we always want root_shape_id to be the id of the shape created at json trail [] -- in reality, sometimes we want a subtrail to the ID that gets returned. 

## What
Jaap and I paired and updated the into_commands signature to allow you to specify which trail you cared about and wanted to get the generated shape ID for. 

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/diff-engine/src/learn_shape/result.rs
+++ b/workspaces/diff-engine/src/learn_shape/result.rs
@@ -757,7 +757,6 @@ mod test {
       nullable_primitive_observations.into_commands(&mut test_id_generator, &JsonTrail::empty()),
     );
     assert!(nullable_primitive_results.0.is_some());
-    assert!(nullable_primitive_results.0.is_some());
     assert_valid_commands(nullable_primitive_results.1.clone());
     assert_debug_snapshot!(
       "trail_observations_can_generate_commands_for_nullable_bodies__nullable_primitive_results",
@@ -767,7 +766,6 @@ mod test {
     let nullable_object_field_results = collect_commands(
       nullable_object_field_observations.into_commands(&mut test_id_generator, &JsonTrail::empty()),
     );
-    assert!(nullable_object_field_results.0.is_some());
     assert!(nullable_object_field_results.0.is_some());
     assert_valid_commands(nullable_object_field_results.1.clone());
     assert_debug_snapshot!(
@@ -779,7 +777,6 @@ mod test {
       nullable_array_item_observations.into_commands(&mut test_id_generator, &JsonTrail::empty()),
     );
     assert!(nullable_array_item_results.0.is_some());
-    assert!(nullable_array_item_results.0.is_some());
     assert_valid_commands(nullable_array_item_results.1.clone());
     assert_debug_snapshot!(
       "trail_observations_can_generate_commands_for_nullable_bodies__nullable_array_item_results",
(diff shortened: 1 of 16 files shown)


workspaces/diff-engine/src/learn_shape/result.rs
```diff
@@ -762,8 +753,9 @@ mod test {
 
     let mut test_id_generator = TestIdGenerator::default();
 
-    let nullable_primitive_results =
-      collect_commands(nullable_primitive_observations.into_commands(&mut test_id_generator));
+    let nullable_primitive_results = collect_commands(
+      nullable_primitive_observations.into_commands(&mut test_id_generator, &JsonTrail::empty()),
+    );
     assert!(nullable_primitive_results.0.is_some());
     assert!(nullable_primitive_results.0.is_some());
```
devdoshi, line 760: seems like there are a bunch of duplicates like this

devdoshi (approved):
Looks good, please address my 1 comment and make sure all the snapshots are generated, updated, and correct @acunniffe @JaapRood

workspaces/diff-engine/src/learn_shape/result.rs
```diff
@@ -862,14 +858,45 @@ mod test {
...
+    };
+
+    let mut test_id_generator = TestIdGenerator::default();
+
+    let collections_results =
+      collect_commands(collections_observations.into_commands(&mut test_id_generator, &json_trail));
+    assert!(collections_results.0.is_some());
+    assert_valid_commands(collections_results.1.clone());
+    assert_debug_snapshot!(
+      "trail_observations_can_generate_for_non_root_json_trails__collection_results",
```
devdoshi, line 891: i don't see a snapshot for this

workspaces/diff-engine/src/learn_shape/result.rs
JaapRood: Echo! Leftover from #681, and we've gotta be good scouts :)

JaapRood (approved):
We paired on this, so no surprise that I think this approach is the right one. Echoing the clean up mentioned by @devdoshi.

JaapRood:
@acunniffe if you don't get to this by my tomorrow morning, I'm happy to do the clean up, so no worries if you have more pressing matters to attend to.

workspaces/diff-engine/src/learn_shape/result.rs
acunniffe: fixed

acunniffe:
Thanks for the help @JaapRood and @devdoshi  -- still getting my paws dirty with the rust test rigs. cleaned things up!

JaapRood:
A lot of digging (thx @jshearer, @notnmeyer, @devdoshi) revealed:

* new snapshots to these same tests had been added in develop in the 12 hours between the creation of the PR and the first correct version of it (where tests should have passed)
* Github behind the scenes merges the PR commit against the current target branch and runs workflows against that (full credit to @jshearer for figuring this part out)
* That caused these new snapshots to be introduced, which were instantly outdated as the PR changed the code's behaviour, causing the snapshot tests to fail.

It was fixed by rebasing this branch onto the latest develop and updating the newly introduced snapshots.

PR #701, opened 2021-05-03
Fix position sticky on TwoBodyColumn

niclim:
## Why
The twobody column now sticks!
Screen Shot 2021-05-03 at 10 41 38 AM

## What

So this is caused by [ancestor elements having the overflow property set](https://www.designcise.com/web/tutorial/how-to-fix-issues-with-css-position-sticky-not-working#checking-if-an-ancestor-element-has-overflow-property-set). I also tried adding a specific height here to fix this problem, but it wouldn't work for me. I removed the overflow properties here - I'm not sure what the original intent for these properties were but I played around with this with those properties and it seems to work as intended without them? I think it should be fine without the NavigationRoute ones since those are on the top level - so overflows will be handled by the browser (?)

I also fixed the lint errors which was causing the `ui-v2` pipeline to fail (perhaps we should run this on PR build + block PR merges if the build fails)?

rust binary build is failing - potentially will be fixed in this PR https://github.com/opticdev/optic/pull/700#issuecomment-831376205 - happy to also fix the changes there

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/ui-v2/src/optic-components/documentation/TwoColumn.tsx
+++ b/workspaces/ui-v2/src/optic-components/documentation/TwoColumn.tsx
@@ -23,7 +23,6 @@ const useStyles = makeStyles((theme) => ({
     display: 'flex',
     flexDirection: 'row',
     alignItems: 'flex-start',
-    overflow: 'auto',
     width: '100%',
   },
   left: {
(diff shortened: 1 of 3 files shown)


JaapRood (approved):
@niclim I haven't had time to review whether these layouts have been setup correctly (which can be tricky!), but as long as it's flex box all the way down, what you've done _should_ work. Let me know if you think a closer review is warranted :)

Also: the Rust tests failing is totally unrelated to any of this and is being addressed elsewhere, so feel free to ignore.

PR #703, opened 2021-05-03
Add default text to new endpoints page

niclim:
## Why
Without the default text here, it looks bare - updating the textfields in this to have default helper text

## What
Screen Shot 2021-05-03 at 12 42 49 PM

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


devdoshi (approved):
looks good. I wonder if we can give more thought to the various states, since I don't think they should all be red and yelling at me. Empty | Unchanged | Changed | UnchangedButMyParentChanged?

niclim:
> looks good. I wonder if we can give more thought to the various states, since I don't think they should all be red and yelling at me. Empty | Unchanged | Changed | UnchangedButMyParentChanged?

Agreed, I think empty should just have the default text - the thing we will want to distinguish between is `Unchanged` and `Changed` - I'm going to sit on this for a little and think what might best distinguish these two states

PR #704, opened 2021-05-03
Add types to spectacle query + mutate

niclim:
## Why

Incremental type hardening to have more useful typescript inferences - starting with spectacle / asyncness

## What
- Consolidated `AsyncStatus` - it's now a type only used in the ui-v2 workspace (since it's really for handling UI state for loading / errors
- Made query + mutate on spectacle generic (and passed down relevant generic types where required)
- Converted one or two usages of these to pass in types (but left the rest to any)

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


workspaces/spectacle/src/index.ts
```diff
@@ -144,23 +144,14 @@ export interface IOpticContext {
 
 ////////////////////////////////////////////////////////////////////////////////
 
-export type AsyncStatus<T> =
```
niclim, line 147: Moved out into `ui-v2/src/types` since spectacle doesn't actually contain async state (it just provides a promise that has data or an error)

workspaces/ui-v2/src/optic-components/diffs/render/AskForCommitMessage.tsx
```diff
@@ -57,28 +57,32 @@ export default function AskForCommitMessageDiffPage(props: {
 
   const handleSave = async (commitMessage: string) => {
     const commands = context.simulatedCommands;
-    const {
-      data: {
+    try {
```
niclim, line 60: Added very rudimentary error catch blocks - I think we will want to have some generic error handlers (and perhaps toasts or something) but it doesn't look like we have any existing patterns.

Added this in here since there's no type safe way to swallow exceptions from the spectacleCommand function - but this is also unideal

workspaces/spectacle/src/index.ts
```diff
@@ -144,23 +144,14 @@ export interface IOpticContext {
...
-  | {
-      loading: false;
-      error: true;
-      data: null;
-    }
-  | { loading: false; error: false; data: T };
-
 export interface IBaseSpectacle {
-  query(options: SpectacleInput): Promise<any>;
+  query<Result, Input = {}>(
```
niclim, line 148: I feel like there should be a way to make these typescript generics reusable (since they are used in a bunch of places) but I'm not sure if this is possible / how to do this

JaapRood (commented):
Looks reasonable to me and a useful step forward, but leaving the approve over to @devdoshi , since I've not really worked with this new code yet and he is the original author :)

devdoshi (approved):

PR #705, opened 2021-05-03
Fix next diff interactions when reviewing diffs

niclim:
## Why

Fixing interactions when reviewing diffs

## What
- On an approve or an ignore, the next unhandled diff should be selected
- On a navigate back to review diffs page with a partially reviewed endpoint, a user should land on the newest available diff

These are both handled by `getNextIncompleteDiff` - this will return the first diff that has _not_ been completed (a user could jump around, the simpleist logic is to just find the first one in shapeDiffs)

See demo here:
https://www.loom.com/share/161bc3dd3391415392d191d4fdfac8c8

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/ui-v2/src/optic-components/pages/diffs/CapturePage.tsx
+++ b/workspaces/ui-v2/src/optic-components/pages/diffs/CapturePage.tsx
@@ -191,7 +191,12 @@ function DiffCaptureResults() {
                 disableRipple
                 button
                 key={index}
-                onClick={handleChangeToEndpointPage(i.pathId, i.method)}
+                onClick={() =>
+                  !done && handleChangeToEndpointPage(i.pathId, i.method)
+                }
+                style={{
+                  cursor: done ? 'default' : 'pointer',
+                }}
               >
                 <EndpointName
                   leftPad={3}
--- a/workspaces/ui-v2/src/optic-components/pages/diffs/CapturePage.tsx
+++ b/workspaces/ui-v2/src/optic-components/pages/diffs/CapturePage.tsx
@@ -145,9 +145,8 @@ function DiffCaptureResults() {
 
   const [handled, total] = handledCount;
 
-  // const history = useHistory();
-
-  const handleChangeToEndpointPage = (pathId: string, method: string) => () => {
+  const handleChangeToEndpointPage = (pathId: string, method: string) => {
+    console.log(diffForEndpointLink.linkTo(pathId, method));
     history.push(diffForEndpointLink.linkTo(pathId, method));
   };
   const handleChangeToUndocumentedUrlPage = () => {
--- a/workspaces/ui-v2/src/optic-components/pages/diffs/CapturePage.tsx
+++ b/workspaces/ui-v2/src/optic-components/pages/diffs/CapturePage.tsx
@@ -146,7 +146,6 @@ function DiffCaptureResults() {
   const [handled, total] = handledCount;
 
   const handleChangeToEndpointPage = (pathId: string, method: string) => {
-    console.log(diffForEndpointLink.linkTo(pathId, method));
     history.push(diffForEndpointLink.linkTo(pathId, method));
   };
   const handleChangeToUndocumentedUrlPage = () => {
--- a/workspaces/ui-v2/src/optic-components/pages/diffs/ReviewEndpointDiffPage/RenderedDiffHeader.tsx
+++ b/workspaces/ui-v2/src/optic-components/pages/diffs/ReviewEndpointDiffPage/RenderedDiffHeader.tsx
@@ -1,0 +1,81 @@
+import React, { FC, useState } from 'react';
+
+import { Collapse, IconButton, Typography } from '@material-ui/core';
+import { ArrowLeft, ArrowRight, Menu as MenuIcon } from '@material-ui/icons';
+
+import { EndpointName } from '<src>/optic-components/common';
+import { DiffHeader } from '<src>/optic-components/diffs/DiffHeader';
+import { IEndpoint } from '<src>/optic-components/hooks/useEndpointsHook';
+import { IInterpretation } from '<src>/lib/Interfaces';
+
+import { DiffLinks } from './DiffLinks';
+
+export type RenderedDiffHeaderProps = {
+  endpoint: IEndpoint;
+  shapeDiffs: IInterpretation[];
+  currentIndex: number;
+  setCurrentIndex: React.Dispatch<React.SetStateAction<number>>;
+};
+
+export const RenderedDiffHeader: FC<RenderedDiffHeaderProps> = ({
+  endpoint,
+  shapeDiffs,
+  currentIndex,
+  setCurrentIndex,
+}) => {
+  const [showToc, setShowToc] = useState(false);
+
+  return (
+    <DiffHeader
+      name={
+        <EndpointName
+          method={endpoint.method}
+          fullPath={endpoint.fullPath}
+          leftPad={0}
+        />
+      }
+      secondary={
+        <Collapse in={showToc}>
+          <DiffLinks
+            shapeDiffs={shapeDiffs}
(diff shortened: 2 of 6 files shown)


workspaces/ui-v2/src/optic-components/pages/diffs/CapturePage.tsx
```diff
@@ -145,9 +145,8 @@ function DiffCaptureResults() {
 
   const [handled, total] = handledCount;
 
-  // const history = useHistory();
-
-  const handleChangeToEndpointPage = (pathId: string, method: string) => () => {
```
niclim, line 150: I'm actually extremely confused why this was working before - this is a double fn that wasn't called but apparently it was...????

workspaces/ui-v2/src/optic-components/pages/diffs/ReviewEndpointDiffPage/RenderedDiffHeader.tsx
```diff
@@ -0,0 +1,81 @@
...
+import { DiffLinks } from './DiffLinks';
+
+export type RenderedDiffHeaderProps = {
+  endpoint: IEndpoint;
+  shapeDiffs: IInterpretation[];
+  currentIndex: number;
+  setCurrentIndex: React.Dispatch<React.SetStateAction<number>>;
+};
+
+export const RenderedDiffHeader: FC<RenderedDiffHeaderProps> = ({
```
niclim, line 20: Pulled out into a separate component (from `ReviewEndpointDiffPage.tsx `)

workspaces/ui-v2/src/optic-components/pages/diffs/ReviewEndpointDiffPage/ReviewEndpointDiffPage.tsx
```diff
@@ -41,9 +42,9 @@ const useRedirectForDiffCompleted = (shapeDiffs: IInterpretation[]) => {
 };
 
 type ReviewEndpointDiffPageProps = {
-  endpoint: IEndpoint;
+  endpoint: RenderedDiffHeaderProps['endpoint'];
```
niclim, line 45: Inverted this such that we are inheriting off the component we are importing (to avoid circular imports)

workspaces/ui-v2/src/optic-components/pages/diffs/ReviewEndpointDiffPage/ReviewEndpointDiffPage.tsx
```diff
@@ -70,30 +73,56 @@ export const ReviewEndpointDiffPage: FC<ReviewEndpointDiffPageProps> = ({
...
-  );
-  const [currentIndex, setCurrentIndex] = useState(0);
+  const [currentIndex, setCurrentIndex] = useState(getNextIncompleteDiff());
   const [previewCommands, setPreviewCommands] = useState<any[]>([]);
+  const renderedDiff = shapeDiffs[currentIndex];
 
   return (
     <TwoColumnFullWidth
       left={
-        allDiffsHandled ? (
```
niclim, line 85: So I realized that this view was never rendered, given that we _always_ redirect when all the diffs are handled

workspaces/ui-v2/src/optic-components/pages/diffs/ReviewEndpointDiffPage/ReviewEndpointDiffPage.tsx
```diff
@@ -71,29 +74,66 @@ export const ReviewEndpointDiffPage: FC<ReviewEndpointDiffPageProps> = ({
...
-  const allDiffsHandled = shapeDiffs.every(
-    (shapeDiff) =>
-      shapeDiff?.diffDescription?.diffHash &&
-      isDiffHandled(shapeDiff.diffDescription.diffHash)
-  );
-  const [currentIndex, setCurrentIndex] = useState(0);
+  const getNextIncompleteDiff = (recentlyCompletedDiff?: string): number => {
+    for (let i = 0; i < shapeDiffs.length; i++) {
+      const shapeDiff = shapeDiffs[i];
+      if (!shapeDiff.diffDescription) {
```
niclim, line 80: There's a bunch of things we can do, since it appears that this is always _set_ and never undefined (at least when it reaches this block of code) but for now, we'll just deal with the implicit checks - it looks like there's some usages of this where it's undefined (in which case we should probably have different types)

workspaces/ui-v2/src/optic-components/pages/diffs/CapturePage.tsx
```diff
@@ -191,7 +189,12 @@ function DiffCaptureResults() {
                 disableRipple
                 button
                 key={index}
-                onClick={handleChangeToEndpointPage(i.pathId, i.method)}
+                onClick={() =>
+                  !done && handleChangeToEndpointPage(i.pathId, i.method)
+                }
+                style={{
+                  cursor: done ? 'default' : 'pointer',
+                }}
```
niclim, line 197: Since we autodirect back to the capture page, i updated this to not allow navigation to the review page - otherwise you navigate and you get navigated back immediately

Screen Shot 2021-05-03 at 5 18 25 PM

workspaces/ui-v2/src/optic-components/pages/diffs/ReviewEndpointDiffPage/ReviewEndpointDiffPage.tsx
```diff
@@ -71,29 +74,66 @@ export const ReviewEndpointDiffPage: FC<ReviewEndpointDiffPageProps> = ({
   });
   const endpointId = getEndpointId({ method, pathId });
 
-  const allDiffsHandled = shapeDiffs.every(
-    (shapeDiff) =>
-      shapeDiff?.diffDescription?.diffHash &&
-      isDiffHandled(shapeDiff.diffDescription.diffHash)
-  );
-  const [currentIndex, setCurrentIndex] = useState(0);
+  const getNextIncompleteDiff = (recentlyCompletedDiff?: string): number => {
```
niclim, line 77: So this is slightly less ideal, when we trigger this function, the xState store hasn't updated yet so the subsequent `isDiffHandled` call returns false (i think it uses useState under the hood) - this is likely a problem with any state management solutions. The other thing we could try do a `setTimeout(setCurrentIndex, 0)` which should defer this call until xState updates, but this is hacky.

workspaces/ui-v2/src/optic-components/pages/diffs/ReviewEndpointDiffPage/ReviewEndpointDiffPage.tsx
```diff
@@ -71,29 +74,66 @@ export const ReviewEndpointDiffPage: FC<ReviewEndpointDiffPageProps> = ({
...
             setCurrentIndex={setCurrentIndex}
           />
-        )
+          <DiffCard
+            key={renderedDiff.diffDescription?.diffHash}
+            updatedSpecChoices={(choices) => {
+              setPreviewCommands(renderedDiff.toCommands(choices));
+            }}
+            diffDescription={renderedDiff.diffDescription!}
+            handled={isDiffHandled(renderedDiff.diffDescription!.diffHash)}
```
niclim, line 117: This isn't used in the `DiffCard` component but I think it would be worthwhile to display to users that this is diff has been handled. The user would have to accept a diff, and then press back to see the old diff (something that I don't believe should be _that_ common).

There's a few other optimizations here such as:
- Setting the input state to what the user had already added in the diff review
- Disabling the input until a new change was done
- Displaying that the diff has already been handled - but more changes can be done
- Being able to review + change an already `Done` diff.

I don't think this is super common, but would be nice polish to add - I also think that this would be easier to implement with a global state that modeled core data (and then things were derived from that)

acunniffe (approved):

PR #706, opened 2021-05-04
updated links in local-cli

acunniffe:
## Why
We've changed the internal routing structure of Optic's UI

## What
- I updated the links in local-cli
- factored them into a constants file so future changes are simpler than the 15 mins this set of changes took

## Validation
* [ ] CI passes
* [ ] Verified in staging


LouManglass (approved):
Looks reasonable and correct.

PR #709, opened 2021-05-04
Fix typing after merge

niclim:
## Why

Fixes the ui-v2 build

## What

I missed this because of:
- the latest run of the ci on the latest commit passes - this PR https://github.com/opticdev/optic/pull/704
- something on develop gets merged
- the merge is still valid (i.e. i can still press merge without having the latest develop changes merged into my branch)

Any thoughts on blocking PRs on:
- failing builds
- requiring PRs to be up to date before merging?

We should also run the ui-v2 build on PRs

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


niclim:
@devdoshi I'm going to merge this since the current UI build is breaking - but see questions above (around blocking merges without passing builds, and running ui-v2 on PR branches)

PR #710, opened 2021-05-04
fixed critical diff bugs

acunniffe:
## Why
Our diff use cases example session was failing, I spent time this afternoon debugging it and getting it ready for prime time. There's still one bug, documented on twist, for @JaapRood to take on in the AM. 

## What
- Handled the case where an undocumented field was at the root object
- Handled optional fields with diffs 
- Updated to a correct version of shape hash

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/cli-shared/package.json
+++ b/workspaces/cli-shared/package.json
@@ -40,7 +40,7 @@
     "ora": "^4.0.4",
     "proper-lockfile": "^4.1.1",
     "semver": "^7.3.4",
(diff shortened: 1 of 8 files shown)


devdoshi (approved):
Do we have snapshot tests covering these? I thought we did

acunniffe:
> Do we have snapshot tests covering these? I thought we did

yeah they had scala deps in them so I have them stashed. Now that trail learners works I'll start to bring them back.

PR #711, opened 2021-05-04
Allow for user enter press on input fields

niclim:
## Why
Allow users to press enter on input fields to go to the next stage (this is usually opening the commit modal)

## What

### Documentation page
pressing enter in one of these pages will open the commit modal
Screen Shot 2021-05-04 at 4 26 38 PM
Screen Shot 2021-05-04 at 4 26 57 PM

Commit modal will submit the message
Screen Shot 2021-05-04 at 4 26 46 PM

### Diff page
Review diffs page - will open the commit modal
*NOTE - I don't actually think this is the right place for contributions (for request / response fields), or even that we should commit here - it may be potentially confusing to have two different workflows for adding contributions? Perhaps having a streamlined diff review process (simplified) where it walks you through what to do? This way a user isn't going to "leave" the diff process half way through, which it feels like currently they can quite easily do so

Screen Shot 2021-05-04 at 4 28 48 PM

Unrecognized url - will stage the endpoint
Screen Shot 2021-05-04 at 4 29 29 PM

Pending + existing endpoints - will open commit modal
Screen Shot 2021-05-04 at 4 29 51 PM

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/ui-v2/src/optic-components/hooks/edit/Contributions.tsx
+++ b/workspaces/ui-v2/src/optic-components/hooks/edit/Contributions.tsx
@@ -1,16 +1,19 @@
 import * as React from 'react';
 import { FC, useCallback, useContext, useState } from 'react';
-import { useSpectacleCommand } from '../../../spectacle-implementations/spectacle-provider';
-import { AddContribution } from '../../../lib/command-factory';
-import { useStateWithSideEffect } from '../util';
+import { v4 as uuidv4 } from 'uuid';
+import { useSpectacleCommand } from '<src>/spectacle-implementations/spectacle-provider';
+import { AddContribution } from '<src>/lib/command-factory';
+import { useStateWithSideEffect } from '<src>/optic-components/hooks/util';
 
 export const ContributionEditContext = React.createContext({});
 
 type ContributionEditContextValue = {
   isEditing: boolean;
+  setEditing: React.Dispatch<React.SetStateAction<boolean>>;
+  commitModalOpen: boolean;
+  setCommitModalOpen: React.Dispatch<React.SetStateAction<boolean>>;
   save: (commitMessage: string) => void;
   pendingCount: number;
-  setEditing: (isEditing: boolean) => void;
   stagePendingContribution: (
     id: string,
     contributionKey: string,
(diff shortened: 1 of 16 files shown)


workspaces/ui-v2/src/optic-components/pages/diffs/CapturePage.tsx
```diff
@@ -266,63 +266,6 @@ function DiffCaptureResults() {
   );
 }
 
-/*
```
niclim, line 269: This looks like a duplicate of above

workspaces/ui-v2/src/optic-components/hooks/edit/Contributions.tsx
```diff
@@ -106,15 +111,30 @@ export const ContributionEditingStore: FC<ContributionEditingStoreProps> = (
         try {
           await spectacleMutator({
             query: `
-            mutation X($commands: [JSON], $commitMessage: String) {
-              applyCommands(commands: $commands, commitMessage: $commitMessage) {
+            mutation X(
```
niclim, line 114: One variant of code that was missing some other updates here https://github.com/opticdev/optic/blob/develop/workspaces/ui-v2/src/optic-components/diffs/render/AskForCommitMessage.tsx#L65

workspaces/ui-v2/src/optic-components/hooks/util/useRunOnKeypress.ts
```diff
@@ -0,0 +1,35 @@
+import { KeyboardEvent, useCallback } from 'react';
+
+export const useRunOnKeypress = <T extends (...args: any) => any>(
```
niclim, line 3: This is something we could use for the future if we wanted to better support power users (keyboard shortcuts to accept, etc)

devdoshi: This looks cool, but I would prefer if we stuck to standard web form ux, unless that is way more complicated. I do see the value of supporting future shortcuts, but for now I'd focus on just making sure people don't have to unnecessarily switch between keyboard and mouse throughout the ux. Thoughts?

devdoshi (approved):
Looks good, let's discuss the 2 threads (native form capabilities, and whether it makes sense to allow editing contributions in the diff ux) in twist

PR #713, opened 2021-05-05
revert some changes that were overwritten

devdoshi:
## Why
Our ui needs access to the opticEngine, which needs to be imported asynchronously. Previously we were duplicating the logic and adding complexity to the data loading responsibilities

## What
- remove duplicate `OpticEngineStore`s
- use `useOpticEngine` hook
- restore simpler logic that was overwritten

## Validation
* [x] trust since there are no tests covering it

### DIFF
--- a/workspaces/ui-v2/src/spectacle-implementations/cloud-viewer.tsx
+++ b/workspaces/ui-v2/src/spectacle-implementations/cloud-viewer.tsx
@@ -17,6 +17,7 @@ import {
   OpticAppConfig,
 } from '../optic-components/hooks/config/AppConfiguration';
 import { AsyncStatus } from '<src>/types';
+import { useInMemorySpectacle } from '<src>/spectacle-implementations/public-examples';
 
 const appConfig: OpticAppConfig = {
   featureFlags: {},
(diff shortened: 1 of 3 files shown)


workspaces/ui-v2/src/spectacle-implementations/cloud-viewer.tsx
```diff
@@ -145,14 +139,78 @@ class CloudInMemorySpectacle
...
-  throw new Error('copy me from public-examples.tsx when ready');
+): AsyncStatus<CloudInMemoryBaseSpectacle> {
+  const opticEngine = useOpticEngine();
+  const [spectacle, setSpectacle] = useState<CloudInMemoryBaseSpectacle>();
+  const [inputs, setInputs] = useState<{
+    events: any[];
+    samples: any[];
+  } | null>(null);
+
+  useEffect(() => {
```
niclim, line 165: Does opticengine ever change? If not, it might be simpler to consolidate these useEffects (such that await loadDependencies -> opticContext load from that)? I see this being useful only if opticEngine changes (which it doesn't look like the current implementation does)

niclim (approved):

workspaces/ui-v2/src/spectacle-implementations/cloud-viewer.tsx
devdoshi: The goal is to have the initial state loaded async, and from then on, update so there's a new spectacle instance every time the events change. I imagine we'll revisit this as part of our ui state management cleanup

PR #714, opened 2021-05-05
Pruning dead code

devdoshi:
## Why
removing dead code

## What
nothing should change

## Validation
* [ ] CI passes


acunniffe (approved):

niclim (approved):
These are my favorite sorts of PRs

JaapRood (commented):
Best review of this would be by trying and seeing if anything is gone. I've worked against this and haven't missed anything thus far!

PR #716, opened 2021-05-05
Fix infinite load on example

niclim:
## Why

useEffect strikes again! But this was causing infinite loading loops since the dependency changes
Screen Shot 2021-05-05 at 2 09 07 PM

## What
Memoizing `task` since it's used in this https://github.com/opticdev/optic/blob/develop/workspaces/ui-v2/src/spectacle-implementations/public-examples.tsx#L161

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


devdoshi (approved):
oops :( thanks for fixing!

PR #719, opened 2021-05-06
Add prompt away to diff workflow

niclim:
## Why
It's not clear where a user can navigate to and from without losing their data. On the diff flow, it's a little odd since it's a multi stage flow but if you accidently navigate away from this, you lose your data :(

I think there's probably something we can reimagine for handling new diffs (or even better, thinking about what our edit flow is and if it'd encapsulate the diff flow)

## What

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


workspaces/ui-v2/src/optic-components/diffs/render/AskForCommitMessage.tsx
```diff
@@ -103,16 +104,20 @@ export default function AskForCommitMessageDiffPage(props: {
...
+          onClose={() => setCommitModalOpen(false)}
+          onSave={handleSave}
+          dialogText={`You have added ${pendingEndpointsCount} new ${
+            pendingEndpointsCount === 1 ? 'endpoint' : 'endpoints'
+          } and updated ${changedEndpointsCount} existing ${
+            changedEndpointsCount === 1 ? 'endpoint' : 'endpoints'
+          }.`}
+        />
+      ) : (
+        <PromptNavigateAway shouldPrompt={props.hasChanges} />
```
niclim, line 119: Since we navigate away programatically, we need this to not be mounted

workspaces/ui-v2/src/optic-components/pages/docs/EditContributionsButton.tsx
```diff
@@ -53,7 +53,10 @@ export function EditContributionsButton() {
       <CommitMessageModal
         open={commitModalOpen}
         onClose={() => setCommitModalOpen(false)}
-        onSave={save}
+        onSave={(commitMessage: string) => {
```
niclim, line 56: This is for the document version of the commit message modal - interestingly we need to not automatically close the diff modal, otherwise what happens is that it remounts the PromptNavigateAway modal and since the history.navigate is outside of the react render cycle this prompt still shows up

workspaces/ui-v2/src/optic-components/common/PromptNavigateAway.tsx
```diff
@@ -6,6 +6,15 @@ type PromptNavigateAwayProps = {
   shouldPrompt: boolean;
 };
 
+const isDiffsPage = (pathname: string) => /diffs/i.test(pathname);
```
niclim, line 9: This can cause false positives, but this is a good enough test for most cases

acunniffe (approved):

acunniffe:
Looks good.

PR #724, opened 2021-05-06
Prevent shape commands / events to be generated for learned shape affordances not selected by the user

JaapRood:
## Why
As #720 was fixed, @acunniffe highlighted how commands could be generated for shapes that the user disabled in the UI: 

> the way the UI lets users 'choose' the changes they're making is by mutating the affordances. We only want to mutate the root one.

> imagine a case where you have 'string' and 'object', then deselect 'object'. It will only create the string.

## What
Upon the generation of commands we now only include commands for those that make sense given the generated shapes. This goes in two directions:

- descendant shapes only have commands generated for them if their parents can host them (e.g. are an object or list)
- ancestor shapes are only included if they are part of the root for which commands are being generated

It's implemented by looking up the root shape prototype (intermediate type from observation result -> commands) and recursively walking the tree it represents to find all reachable json trails and filtering all shape prototypes by it. The implementation is covered by multiple snapshot tests, including the original failing one that highlighted the bug.

## Validation
* [x] CI passes
* [x] No longer generates superfluous shape commands

### DIFF
--- a/workspaces/diff-engine/src/learn_shape/result.rs
+++ b/workspaces/diff-engine/src/learn_shape/result.rs
@@ -94,8 +94,12 @@ impl TrailObservationsResult {
     let included_trails: HashSet<_> =
       shape_prototypes_trails(root_shape, &shape_prototypes_by_trail).collect();
 
-    let commands =
-      shape_prototypes_to_commands(shape_prototypes).map(|command| SpecCommand::from(command));
+    let commands = shape_prototypes_to_commands(
+      shape_prototypes
+        .into_iter()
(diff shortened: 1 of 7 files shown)


JaapRood:
@acunniffe so far this is a test to verify the current affordances -> command generator yields the extraneaous shape commands in the scenario laid out by you. Will see now if there's something we can reasonable do to prevent this from happening. Have some ideas that I will look at next.

devdoshi:
Is this needed still @JaapRood

JaapRood:
@devdoshi yeah it is! Got sidelined with providing correctness frist, but high on the list of clean up!

devdoshi (approved):

PR #731, opened 2021-05-06
Fix/busted diff engine build

notnmeyer:
## Why
diff-engine builds are broken on the latest stable rust,

```
Error:  --> workspaces/diff-engine/src/protos/generated/shapehash.rs:9:4
  |
  | #![rustfmt::skip]
  |    ^^^^^^^^^^^^^
  |
  = note: `#[deny(soft_unstable)]` on by default
  = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
  = note: for more information, see issue #64266 <https://github.com/rust-lang/rust/issues/64266>
```

## What
Updates protobuff, and fixes a subsequent warning produced by the upgrade

## Validation
* [ ] CI passes

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -689,24 +689,24 @@ dependencies = [
 
 [[package]]
 name = "protobuf"
-version = "2.17.0"
+version = "2.23.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cb14183cc7f213ee2410067e1ceeadba2a7478a59432ff0747a335202798b1e2"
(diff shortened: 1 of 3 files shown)


jshearer:
I think you need to also regenerate the file that had the issue in it, right? Not sure how to do that

jshearer:
Oh, it goes through a `build.rs` file, maybe it'll happen automatically...

notnmeyer:
the generated files are gitignored. building locally produced a modified file 👍

jshearer (approved):
Looks like this fixed it!

devdoshi (approved):

PR #733, opened 2021-05-06
spectacle: override default body size to allow for 'large' bulk-learned specs

LouManglass:
Per discussion

## Why

- 'Large' specifications may have enough data when bulk-learning to exceed the default body size limit.

## What

- Added a limit override to the body parser for spectacle

## Validation
* [ ] CI passes
* [ ] Verified in staging


acunniffe (approved):

devdoshi (approved):

JaapRood:
For anyone finding this in the future thinking about raising this limit even further: it's time to consider reading the body as a stream, rather than buffering it all into memory.

PR #738, opened 2021-05-07
Fix generating of empty affordances for diffs of array items

JaapRood:
## Why
For shape diffs that described a diff in a non-zero-index array item, no affordances would be generated, nor would interaction pointers be generated for the right place. This was caused by incorrect normalising of `JsonTrails` and made the Diff review UI render empty diffs + offer no remedies.

## What
Instead of normalising the `JsonTrail`s while we analyse the interaction, we move it to the result aggregation stage. This allows the shape diff affordances learner to access the original json trails it needs to generate the right data structure and allow the Diff review UI to highlight the right rows, while allowing the undocumented endpoints learner to normalise them right off the bat.

## Validation
* [x] CI passes
* [x] Shape diffs for array items now render in the Diff review UI
* [x] In the Diff review UI, the right array index is highlighted per example.
* [x] Provide coverage of correct normalisation behaviour for both the learner output projections.

### DIFF
--- a/workspaces/diff-engine/src/learn_shape/mod.rs
+++ b/workspaces/diff-engine/src/learn_shape/mod.rs
@@ -47,7 +47,7 @@ mod test {
 
     let body: Option<BodyDescriptor> = Some(BodyDescriptor::from(object_body));
 
-    let result = observe_body_trails(body);
+    let result = observe_body_trails(body).normalized();
 
     let root_shape_trail = JsonTrail::empty();
     let root_shape_result = result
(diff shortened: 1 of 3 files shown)


JaapRood:
I'd like to cover this moving of the normalisation logic in some automated tests for the two learner output projections, but have already manually verified the implementation is correct. If anyone is waiting on this to get merged, I'm happy to provide that coverage through an additional PR as well.

JaapRood:
Got the existing tests to pass and covered the difference in trail normalisation strategies between the undocumented bodies learner and shape diff affordances learner.

@devdoshi @acunniffe ready to be merged, awaiting your review.

PR #741, opened 2021-05-07
Sort endpoint responses by status code

niclim:
## Why

## What

Response are now sorted by status code

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


workspaces/ui-v2/src/optic-components/hooks/useEndpointBodyHook.ts
```diff
@@ -94,22 +94,27 @@ export function useEndpointBody(
...
+              pathId: request.pathId,
+              method: request.method,
+              changes: response.changes,
+              description: request.requestContributions.description || '',
+            };
+          }
+        );
       }
     );
+    const sortedResponses = [...responses].sort(
```
niclim, line 113: This should affect both the documentation page as well as the pending endpoint page (they both use this hook)

acunniffe (approved):

PR #747, opened 2021-05-10
Use relative link for changelog/documentation dropdown

niclim:
## Why
A user looking at a specific endpoint, navigating to a different page would want to stay on the same endpoint page

## What

Users on the endpoint doc page clicking on the changelog dropdown will now stay on the same page (just with a different changelog version)

I've used a somewhat brittle regex here - what's probably more ideal is to split up the changelog + documentation pages and restructure react-router routes so that they can detect the current path prefix - ideally this should be extensible to any documentation / changelog route we add in the future as well.

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


devdoshi (approved):
This seems fine as a temporary fix, but I'd vote for the proper fix. Perhaps you can use the router state to strip the prefix instead of the regex? Or instead of the component providing links, it can cause a state change indicating the desired batch commit id?

PR #748, opened 2021-05-10
shape-diff tests

acunniffe:
## Why
Using the Rust diff engine + WASM in the UI, and interpretations tests will help us ship better / more stable software

## What
Tests are coming back :) but they don't all pass yet. 

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


workspaces/ui-v2/src/lib/shape-diff-dsl-rust.ts
```diff
@@ -135,7 +137,10 @@ export class Actual {
     private shapeTrail: IShapeTrail,
     public jsonTrail: IJsonTrail
   ) {
-    this.trailAffordances = learnedTrails.affordances;
+    this.trailAffordances = learnedTrails.affordances.filter((i) => {
```
acunniffe, line 140: @JaapRood @LouManglass @devdoshi this is why the choices presented for some of the diffs were incorrect. Optic was not filtering these in the right spot. You want the `trailAffordances` filtered when you are presenting options to update it, but you want all of them when you're building the nested shapes.

workspaces/ui-v2/src/lib/__tests/diff-helpers/fixture.ts
```diff
@@ -1,282 +1,164 @@
...
+    input.diffs[0].asShapeDiff(universe.currentSpecContext)!,
+    trailValues,
+    universe.spectacleQuery,
+    universe.currentSpecContext
+  );
+
+  return {
+    preview,
+    trailValues,
+    commands: [], //preview.toCommands(preview.updateSpecChoices!),
```
acunniffe, line 114: @devdoshi  -- if you wanted to make the tests generate commands / prompt that WASM issue, here's where that happens

acunniffe:
@devdoshi -- double-checked all the snapshots and added Spectacle path queries.

JaapRood:
@acunniffe one thing I just realised in #752 that might be important to you with these tests: the orders in shape diff affordances are not stable. Both the `affordances` and collections of interaction pointers (like `was_missing`, `was_missing_trails`) have undefined ordering due to their use of `HashMap + HashSet` internally. I've already seen some of my snapshots fail because of that. Not sure how it impacts you exactly, but it might be important for you to know. Nobody wants flaky tests 😅.

We _are_ considering options to make the results stable, but since that would highly likely have a performance impact, I don't really want to do that until we get performance back into a reasonable window. Especially because it doesn't affect correctness of the result, with the stable ordering just being for snapshotting, it's something we'll want to at least delay before we commit.

workspaces/ui-v2/src/lib/stable-path-batch-generator.ts
```diff
@@ -0,0 +1,33 @@
+import { pathStringToPathComponents } from '<src>/optic-components/hooks/diffs/LearnInitialBodiesMachine';
```
devdoshi: @acunniffe is this WIP?

devdoshi:
@acunniffe please task workspaces:clean, task workspaces:build, and regenerate the snapshots; when I run them locally i see changes in the snapshots and I'm not sure how to verify them. I'm looking into the id generation issue right now, but you should be able to switch it to "sequential" instead of "random" to unblock yourself.

acunniffe:
> We are considering options to make the results stable, but since that would highly likely have a performance impact, I don't really want to do that until we get performance back into a reasonable window. Especially because it doesn't affect correctness of the result, with the stable ordering just being for snapshotting, it's something we'll want to at least delay before we commit.

It seems far simpler for me to pre-process the arrays and sort them. The order shouldn't matter to the UI logic, so I wouldn't encode it in the responses from the diff engine.

PR #749, opened 2021-05-10
Add loading state onto bulk edit load flow

niclim:
## Why

The selection + bulk mode has an async portion to this we aren't taking into account. If a user saves changes before these endpoints are learnt, the endpoints don't actually get added. To solve this, we are adding in a loading state with a lock out.

## What
Screen Shot 2021-05-10 at 3 43 23 PM

There's some kinda annoying stuff with xstate where there's some hacky stuff to get certain bits of state to update due to how the subscriptions / refs work. I think again this could be modified with modeling state in a different way (probably a fair amount of work here). Perhaps I'm also missing a much easier way to do this (couldn't figure out how to pull out / reference the `loading` promise on the [LearnEndpoint loading function](https://github.com/opticdev/optic/blob/e818d75bf3457d3256c176248e4bd5ff1082a8aa/workspaces/ui-v2/src/optic-components/hooks/diffs/LearnInitialBodiesMachine.ts#L89)

To follow this PR - there's work to be done to speed up learning of a large number of endpoints. Due to how node works, there's slowness with processing in parallel and it's faster to do this in series

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/ui-v2/src/optic-components/pages/diffs/AddEndpointsPage/AddEndpointsPage.tsx
+++ b/workspaces/ui-v2/src/optic-components/pages/diffs/AddEndpointsPage/AddEndpointsPage.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useMemo, useState } from 'react';
+import React, { FC, useEffect, useMemo, useState } from 'react';
 import {
   useDiffReviewPagePendingEndpoint,
   useEndpointPageLink,
@@ -13,6 +13,7 @@ import { TwoColumnFullWidth } from '../../../layouts/TwoColumnFullWidth';
 import { DiffHeader } from '../../../diffs/DiffHeader';
 import {
   Box,
+  CircularProgress,
   Divider,
   List,
   ListItem,
@@ -41,6 +42,7 @@ import {
   PendingEndpointNameField,
 } from './EndpointNameEditFields';
 import { useAnalytics } from '<src>/analytics';
+import { useActor } from '@xstate/react';
 
 export function DiffUrlsPage(props: any) {
   const urls = useUndocumentedUrls();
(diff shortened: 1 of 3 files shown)


workspaces/ui-v2/src/optic-components/pages/diffs/AddEndpointsPage/AddEndpointsPage.tsx
```diff
@@ -158,6 +160,54 @@ export function DiffUrlsPage(props: any) {
...
+const PendingEndpointItem: FC<{
+  endpoint: IPendingEndpoint;
+}> = ({ endpoint }) => {
+  const diffReviewPagePendingEndpoint = useDiffReviewPagePendingEndpoint();
+  const changelogStyles = useChangelogStyles();
+  const classes = useStyles();
+  const history = useHistory();
+  const { forceRerender } = useSharedDiffContext();
+
+  const [state]: any = useActor(endpoint.ref);
```
niclim, line 172: To subscribe to changes here, we need to use the `useActor` hook. However in the `Save Changes` button, we need to group together all the states of all pending endpoints. Due to how hooks work, we can't actually do a `for endpoint, useActor loop`

This code below effectively _recreates_ what happens in react if the endpoint ref was _not_ transparent and did cause rerenders.

It's also possible there is an easier way to do this (the only "real" way I could think through here was to remodel the state + pull up the isLoading state into the sharedDiffStore and access it there).

niclim: I think you are correct - however it's not deterministic / ordered since this would loop over pendingEndpoints - which is an array that changes as we add more pendingEndpoints.

niclim:
I'm going to put this PR on hold + potentially close this in a bit, we might not need this - I'm planning on building a slightly different UI flow that has a more consolidated loading state (similar to the optic 9 learning modal)

niclim:
I'm closing this - but if you are following along, i found a non-hacky way to do what I was doing by using [sendUpdate](https://xstate.js.org/docs/guides/actors.html#sending-updates). Turns out reading documentation is useful - separate PR to follow up with this

PR #750, opened 2021-05-11
Force local cli learn undocumented bodies to run sequentially

niclim:
## Why

When running optic against the local cli server, there's file contention with learning multiple instances

## What

Throttle learnUndocumentedBodies so that we don't get file read contention. 

Queueing a bunch of requests here against the cli server (i think the 500s are a different bug - returned as `expected to receive a learning result`)
Screen Shot 2021-05-11 at 11 40 46 AM

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/spectacle-shared/src/local-cli/client.ts
+++ b/workspaces/spectacle-shared/src/local-cli/client.ts
@@ -21,6 +21,8 @@ import {
   IValueAffordanceSerializationWithCounterGroupedByDiffHash,
 } from '@useoptic/cli-shared/build/diffs/initial-types';
 
+let throttlerPromiseSingleton = Promise.resolve();
+
 export class LocalCliSpectacle implements IForkableSpectacle {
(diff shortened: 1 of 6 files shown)


workspaces/spectacle-shared/src/local-cli/client.ts
```diff
@@ -21,6 +21,8 @@ import {
   IValueAffordanceSerializationWithCounterGroupedByDiffHash,
 } from '@useoptic/cli-shared/build/diffs/initial-types';
 
+let throttlerPromiseSingleton = Promise.resolve();
```
devdoshi, line 24: this should be class-local state, but I'd still recommend using something like bottleneck

niclim:
@devdoshi I've made updates to move the throttling logic to the cli-server - there's going to be a follow up to this to update the bulk mode flow UI

PR #752, opened 2021-05-11
Fix bugs where shape diff affordances projection was incorrectly detecting missing fields

JaapRood:
## Why

~~We were seeing diffs about missing fields that lived inside arrays, but without the right line being highlighted, or not marked as missing. A bunch of other side effects regarding missing fields were seen as well, most likely originating from this same broken logic for detecting them.~~ We are seeing multiple bugs where missing fields are either detected false positively or false negatively.  Turns out the logic added in #751 was too naive, with the symptoms of this problem first looking at an issue in normalisation of resulting trails.

## What
Reimplements the logic around detecting what trail affordances we might not have observed, but were expecting. Given the diff's json trail, we look up each expected parent trail and from there determine all the trails we expect to have been observed. When they weren't we register a missing trail.

## Validation
* [x] CI passes
* [x] Includes affordances for missing fields nested in objects
* [x] Includes affordances for missing fields nested in arrays
* [x] Doesn't identify bodies observed as only objects to be recorded as missing
* [x] Generates correct json trails for `was_missing_trails`

### DIFF
--- a/workspaces/diff-engine/src/learn_shape/result.rs
+++ b/workspaces/diff-engine/src/learn_shape/result.rs
@@ -325,6 +325,7 @@ impl TrailValues {
       && !self.was_null
       && !self.was_array
       && !self.was_empty_array
+      && !self.was_object
   }
 
   pub fn insert_field_set(&mut self, field_set: FieldSet) {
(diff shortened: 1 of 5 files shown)


JaapRood:
While this has a test that shows this problem, the body that contains the diff is _way_ too big to effectively debug this. Given that, the next step would be to create a way smaller body that exhibits the same issue.

JaapRood:
Took this back to draft, as there's a couple of use cases where missing fields are not detected correctly yet. Might as well get them all in one go!

JaapRood:
@devdoshi @acunniffe  All good to go, covering all the known bugs having to do with missing fields. Feel free to merge if this passes review while I'm off!

devdoshi (approved):

PR #753, opened 2021-05-11
OPTIC_RUST_DIFF_ENGINE flag should be removed as it is the default now

devdoshi:
## Why
We don't need the flag anymore

## What
The change should be transparent. Remove it from your local .env if necessary

## Validation
* [x] CI passes
* [ ] Verified in staging


Developer-setup.md
```diff
@@ -27,62 +27,56 @@ It's a good idea to have a design in place beforehand. It doesn't have to be fan
...
 
 - Clone the [Optic repository](https://github.com/opticdev/optic) to your development environment with Git.
-- If necessary, install Yarn and Task. On MacOS, this can be done via Homebrew `brew install yarn`. Otherwise, check out the **Resources** section for more information.
+- If necessary, install Yarn and Task. You can not use NPM; we are using Yarn workspaces. On MacOS, this can be done via Homebrew `brew install yarn`. Otherwise, check out the **Resources** section for more information.
 - Navigate to the Optic repository. You should be on the `develop` branch by default.
 - Navigate to the project root, and run `task`. This provides aliases for managing the development lifecycle of Optic. Note, this must be run for every terminal instance, and you may find it convenient to add this to your `.profile` or equivalent so future terminal invocations will be ready to go.
 - Install dependencies and build the code with `task workspaces:build`. Yarn will install dependencies for all of the workspaces present in the project.
 
+## Evironment Variables
```
niclim, line 35: Evironment -> environment

workspaces/diff-engine/package.json
```diff
@@ -14,6 +14,13 @@
     "url": "ssh://git@github.com/useoptic/optic-package.git",
     "directory": "workspaces/diff-engine"
   },
+  "files": [
+    ".env",
+    "scripts",
+    "lib",
+    "Cargo.toml",
+    "cli/Cargo.toml"
+  ],
```
JaapRood, line 23: Diff-engine uses a `.npmignore`, that's why this wasn't in there. We had _multiple_ borked publishes because of this opt-in list instead of opting out. If that needs to be switched (I still wouldn't, but 🤷‍♂️), we should at least get rid of the other, as having both is way worse than choosing whichever one. 

Also, why does the npm package need cargo files published?

devdoshi: @JaapRood uses the cargo.toml to extract the binary name

PR #754, opened 2021-05-11
Commands should use client and session ids

devdoshi:
## Why
currently we're sending placeholder values for the clientId (think userId) and clientSessionId (which lets us distinguish between client/ui instances). @acunniffe should we be using analytics ids for the clientId?

## What
This should be done before releasing, so people's event streams have meaningful values for clientId and clientSessionId

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


JaapRood (approved):
Reviewing just the client + session id parts, as it seemed there's some bleed here from other PRs: looking good.

 > @acunniffe should we be using analytics ids for the clientId?
 
That's also my recollection on how we reasoned that to be used, and how it was implemented in the first time around.

acunniffe:
> @acunniffe should we be using analytics ids for the clientId?

@devdoshi -- I think these become most valuable when users can see who changed what / given IDs they understand. Our analytics ids are random and won't mean anything to the users. For now I think it's fine, but I would like to see it move to Optic cloud IDs or something from the user's `git config` that can be recognized by teammates

devdoshi:
closed in favor of #759

PR #755, opened 2021-05-11
Specify rustfmt.tml on precommit lint-staged

niclim:
## Why

Rustfml tries to infer edition (when it's run in different contexts) - this causes certain contexts to fail on runs. Lint-staged also doesn't automatically pick up these changes so we need to point the lint-staged command to pick this up.

I'm also not sure how we specify editions in cargo.toml files, etc. This currently unblocks git commits but ideally we have a centralized place to trigger these

The problematic file was `workspaces/diff-engine/tests/shape-diff-use-cases.rs` which had `async await` - which i tried formatting through `yarn run lint-staged` and it formatted

## What

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
       "yarn prettier --write"
     ],
     "**/*.rs": [
-      "rustfmt"
+      "rustfmt --config-path ./rustfmt.toml"
     ]
   }
 }
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,1 +1,2 @@
+edition = "2018"
 tab_spaces = 2 # spaces per tab, held at 2 as that's what most of the code was written in originally


JaapRood (approved):
Thanks for unblocking this @niclim!

I think we're running into trouble because we're using Cargo workspaces, and actually multiple of those so wasm + binary builds can be built separately. That most likely muddles the water.

That said, defining the `edition` in the rust format config file is fine. Everything is on the same `edition` nor is a new value for that in the pipeline any time soon. I'm pretty sure it's to keep the pre-2018 Rust working with the current Cargo, which is when a lot changes happened that are pretty core to the language, that shouldn't be changing again.

Of course it _can_ change, but I reckon we can just deal with that then.

jshearer:
Looks like the 2021 edition won't have much of any breaking changes, especially big structural ones like traits https://blog.rust-lang.org/2021/05/11/edition-2021.html

jshearer (approved):
Thanks @niclim! I also got bit by not being able to commit lol

JaapRood:
Thanks for linking that post @jshearer, very useful and educational :)

PR #760, opened 2021-05-12
feature: Implement coverage calculation

jshearer:
Fundamentals of coverage computation. See here for more details:

https://www.notion.so/useoptic/Optic-in-CI-01e647cc49504835a0e2750764d097b6#ddb03c2041dc4bf3bceb6773791b8067


workspaces/local-cli/src/shared/coverage-2.ts
```diff
@@ -0,0 +1,129 @@
...
+import * as DiffEngine from '@useoptic/diff-engine-wasm/engine/build';
+
+// This concept came from the Scala codebase's "Concern" concept
+// https://github.com/opticdev/optic-core/blob/master/core/optic/shared/src/main/scala/com/useoptic/coverage/CoverageConcerns.scala
+// JS/TS doesn't have the kind of value-types or case-classes that scala does,
+// but we can emulate them like this. I think...
+
+// o.O
+const createDataClass = <P extends Array<any>>(func: (...args: P)=>void): (...args: P)=>string => {
+  return (...args: P) => `${func.name}-${args.map(arg=>JSON.stringify(arg)).join("-")}`
```
devdoshi, line 16: did you get this from someplace in particular? I think it would be better to JSON.stringify(args) than to stringify each one separately. but looks fine either way

workspaces/local-cli/src/shared/coverage-2.ts
```diff
@@ -0,0 +1,129 @@
...
+        const [response_id, response] = responses[0];
+        recordEvent(TotalForResponse(response_id), is_diff, coverage_map);
+
+        if(response.body){
+          recordEvent(TotalForPathAndMethodAndStatusCodeAndContentType(path_id, interaction.request.method, interaction.response.statusCode, response.body.http_content_type), is_diff, coverage_map)
+        } else {
+          recordEvent(TotalForPathAndMethodAndStatusCodeWithoutBody(path_id, interaction.request.method, interaction.response.statusCode), is_diff, coverage_map)
+        }
+      } else {
+        // More than 1 response... Same as case below. Shouldn't happen I don't think?
```
devdoshi, line 118: These things are technically possible when there are "domain conflicts" but safe to throw an error at each of these spots since it violates the invariants

workspaces/local-cli/src/shared/coverage-2.ts
jshearer: I make it up. Can just `JSON.stringify` and be done with, it just feels... weird to use JSON strings as keys in a map.

workspaces/local-cli/src/shared/coverage-2.ts
jshearer: Okay great, so at least the places where I didn't know what logic to implement are the cases that we should never reach. Sweet!

workspaces/local-cli/src/shared/coverage-2.ts
devdoshi: I like the combo of function.name as namespace + serializing its args. ideally it would also keep the keys, which is more evolvable, but certainly fine if it's an isolated report that doesn't get transferred anywhere. anything else you do would be making up some random serialization algorithm so it's safest to stick with json

workspaces/local-cli/src/shared/coverage-2.ts
devdoshi: er...for a given (path_id, method), you can have 0+ bodies, aka 1+ requests each with a distinct content-type.

Similarly for a given (path_id, method, status_code) you can have 0+bodies aka 1+ responses each with a distinct content type. so I think you should check if length === 0 => error, otherwise, for each accumulate the counts.

workspaces/local-cli/src/shared/coverage-2.ts
jshearer: Fair enough. I was thinking about evolvability and versions and stuff, but with the keys that's kind of a moot point. Will change

jshearer:
Ok @devdoshi I changed it to resolve a single request/response by passing in the content_type. Does that look better?

workspaces/local-cli/src/shared/coverage-2.ts
```diff
@@ -0,0 +1,126 @@
...
+const TotalForPathAndMethodAndContentType = createDataClass(function TotalForPathAndMethodAndContentType(pathId: String, httpMethod: String, requestContentType: String) {})
+const TotalForPathAndMethodAndStatusCodeWithoutBody = createDataClass(function TotalForPathAndMethodAndStatusCodeWithoutBody(pathId: String, httpMethod: String, httpStatusCode: number) {})
+const TotalForPathAndMethodAndStatusCodeAndContentType = createDataClass(function TotalForPathAndMethodAndStatusCodeAndContentType(pathId: String, httpMethod: String, httpStatusCode: number, responseContentType: String) {})
+const TotalForRequest = createDataClass(function TotalForRequest(requestId: string) {})
+const TotalForResponse = createDataClass(function TotalForResponse(responseId: string) {})
+
+type ICoverageMap = {[key: string]: number};
+
+const recordEvent = (evt: string, is_diff: boolean, coverage_map: ICoverageMap) => {
+  const key = `${evt}-${is_diff?'diff':'nodiff'}`;
```
devdoshi, line 34: i think you should initialize 2 coverage maps, and have the caller provide either the diff one or the non-diff one, as opposed to adding another layer of keys

jshearer: I like that

devdoshi:
> Ok @devdoshi I changed it to resolve a single request/response by passing in the content_type. Does that look better?

yup!

workspaces/local-cli/src/shared/coverage-2.ts
```diff
@@ -0,0 +1,231 @@
...
+
+  for await (let { coverage: individual_coverage, is_diff } of coverages) {
+    for (const [k, v] of Object.entries(individual_coverage)) {
+      let chosen_map = is_diff ? combined_diff_map : combined_non_diff_map;
+      chosen_map[k] = (chosen_map[k] ?? 0) + v;
+    }
+  }
+}
+
+type BodyDescriptor = {
```
devdoshi, line 123: Is this used anywhere?

workspaces/local-cli/src/shared/coverage-2.ts
```diff
@@ -0,0 +1,231 @@
...
+      };
+    }
+  })(captureIterator);
+
+  let combined_diff_map: ICoverageMap = {};
+  let combined_non_diff_map: ICoverageMap = {};
+
+  for await (let { coverage: individual_coverage, is_diff } of coverages) {
+    for (const [k, v] of Object.entries(individual_coverage)) {
+      let chosen_map = is_diff ? combined_diff_map : combined_non_diff_map;
```
devdoshi, line 117: Looks like this can move into the outer loop

devdoshi (commented):
Looks good!

jshearer:
Now we have fancy table formatting! 
Screen Shot 2021-05-14 at 2 58 19 PM

workspaces/local-cli/src/shared/coverage-2.ts
jshearer: Done!

workspaces/local-cli/src/shared/coverage-2.ts
jshearer: Yep! It's used to provide type info for the JSON returned by WASM. Here for example: https://github.com/opticdev/optic/pull/760/files#diff-9a692dec003501b8a06e88fabf0837c6a2e44cc90c8cc0856d16a2e24b92c20aR505

Sorry for the messier diff, I deleted the old coverage file and renamed `coverage-2` back to `coverage`, so the diff is worse now :(

workspaces/local-cli/src/shared/coverage-2.ts
jshearer: Done

workspaces/local-cli/src/shared/coverage-2.ts
jshearer: Done!

workspaces/local-cli/src/shared/coverage-2.ts
jshearer: Thanks!

workspaces/cli-shared/src/captures/avro/file-system/capture-saver-with-diffs.ts
```diff
@@ -103,25 +102,6 @@ export class CaptureSaverWithDiffs extends FileSystemAvroCaptureSaver {
       );
     }
 
-    if (this._config.shouldCollectCoverage) {
```
devdoshi, line 106: You should clean up init() as well; that should allow you to get rid of the useoptic/domain dependency from this file

jshearer: Love it. Done.

PR #761, opened 2021-05-12
Update diff bulk mode flow

niclim:
## Why

Updated the diff bulk mode flow - this is pretty similar to Optic 9's flow with:
- collapsing while you name / match path patterns
- displaying a modal to learn endpoints (using checkbox selections)

## What
Main changes
- Undocumented urls now collapse _as_ you start naming path parameters (similar to optic 9)
- Added checkbox functionality to select endpoints
- Show progress loader while learning endpoints

Auxillary changes:
- Refactor AddEndpointsPage
  - created local component / hooks / utils for addendpoints page - we can treat these pages as modules, and as they become more generic they can get "promoted" across the app - making a change can ideally be localized to a single folder location (unless there's shared components / hooks that need to be updated)
  -  removed a bunch of useEffects that could be replaced with derived state
  - updated how wipPatterns is updated such that it's only updated on user input
- allow click interaction on entire row for undocumented urls (so the user doesn't need to aim for an icon)
- Sort response status codes for learning endpoints
- Removed deep-equal, this seems equivalent to lodash.isEqual

Side notes
- There's a fair amount of excessive renders - and that's due to how we've modeled our stores (i.e. a single change in the sharedDiffContext triggers rerenders in every component that uses `sharedDiffContext`) - we would either need to split up these into different contexts, or use something that more smartly memoizes this (like the react-redux connect component)
  - e.g. An update for `wipPatterns[endpoint1]` triggers an update in sharedDiffContext which rerenders effectively everything on the page, but we really only need to update components that reference `wipPatterns[endpoint1]`

### Screenshots!
Screen Shot 2021-05-12 at 3 16 41 PM

Screen Shot 2021-05-12 at 3 17 04 PM

Screen Shot 2021-05-12 at 3 17 58 PM

## Validation
The best way to test this is to pull down this and run it against a local cli instance


workspaces/diff-engine/tests/fixtures/shape-diff-use-cases/deeply nested fields inside of arrays.json
```diff
@@ -1,2637 +1,2571 @@
 {
```
niclim, line 1: Ok so during my merge commit, prettier picked this up since it's considered a staged commit - this should be identical just format changes

workspaces/ui-v2/package.json
```diff
@@ -44,7 +44,6 @@
     "camelcase": "^6.1.0",
     "classnames": "^2.2.6",
     "css-loader": "4.3.0",
-    "deep-equal": "^2.0.5",
```
niclim, line 47: We have lodash.isEqual which i believe is equivalent

workspaces/ui-v2/src/optic-components/pages/diffs/AddEndpointsPage/AddEndpointsPage.tsx
```diff
@@ -37,128 +26,131 @@ import { useChangelogStyles } from '../../../changelog/ChangelogBackground';
...
-      />
-    );
-  }
+  const unmatchedUrlLengths = undocumentedUrls.length;
+  const visibleUrls = undocumentedUrls.filter((url) =>
+    url.path.startsWith(searchQuery)
+  );
+  const bulkSelectedUrls = undocumentedUrls.filter((url) =>
+    selectedUrls.has(url.path + url.method)
+  );
```
niclim, line 57: This is slightly confusing - the reason we need this is we want to show:
- The total number of urls we have that are uncollapsed
- The visible urls (i.e. things with a filter applied)
- All the urls you've selected (excluding any urls that fall under a matched group, BUT including urls that have been filtered by the url filter)

workspaces/ui-v2/src/optic-components/pages/diffs/AddEndpointsPage/AddEndpointsPage.tsx
```diff
@@ -37,128 +26,131 @@ import { useChangelogStyles } from '../../../changelog/ChangelogBackground';
...
+                  {({ height, width }: any) => (
+                    <FixedSizeList
+                      height={height}
+                      width={width}
+                      itemSize={47}
+                      itemCount={visibleUrls.length}
+                      itemKey={(index, data) => {
+                        const item = data.undocumentedUrls[index];
+                        return item.method + item.path;
+                      }}
```
niclim, line 94: So this was the magic piece that we needed to add if the list of urls changes. This allows FixedSizeList to figure out when something is changed, so we don't bleed / reference state between renders + the list changing. 1 more point towards reading documentation

workspaces/ui-v2/src/optic-components/pages/diffs/AddEndpointsPage/AddEndpointsPage.tsx
```diff
@@ -37,128 +26,131 @@ import { useChangelogStyles } from '../../../changelog/ChangelogBackground';
...
+    visibleUrls,
+    selectedUrls,
+    setSelectedUrls
+  );
 
   return (
-    <TwoColumnFullWidth
-      left={
-        <>
-          <DiffHeader name={name}>
```
niclim, line 101: Pulled out into `AddEndpointDiffHeader`

workspaces/ui-v2/src/optic-components/pages/diffs/AddEndpointsPage/components/BulkLearnModal.tsx
```diff
@@ -0,0 +1,150 @@
...
+};
+
+const LearningProgress: FC<{
+  pendingEndpointIds: Set<string>;
+  closeModal: BulkLearnModalProps['closeModal'];
+}> = ({ pendingEndpointIds, closeModal }) => {
+  const { stageEndpoint, context } = useSharedDiffContext();
+  const refStageEndpoint = useRef(stageEndpoint);
+
+  // @nic TODO handle when learning fails! - currently the modal just gets stuck :(
```
niclim, line 31: There's a couple of endpoints that I've seen fail, that should be fixed but separately, a failure should not block a user from continuing here - I'm going to follow this up in a separate PR

workspaces/ui-v2/src/optic-components/pages/diffs/AddEndpointsPage/components/BulkLearnModal.tsx
```diff
@@ -0,0 +1,150 @@
...
+      return { pattern, method };
+    }
+  );
+  const [learningInfo, setLearningInfo] = useState<{
+    pendingEndpointIds: Set<string>;
+  } | null>(null);
+
+  const learnEndpoints = () => {
+    const pendingEndpointIds = endpointsAsPatterns.map(({ pattern, method }) =>
+      documentEndpoint(pattern, method)
```
niclim, line 95: This dumps all the pending endpoints onto the learn worker - in the local-cli instance this will be slow, but we've already implemented a separate PR to bottleneck this https://github.com/opticdev/optic/pull/750

workspaces/ui-v2/src/optic-components/pages/diffs/AddEndpointsPage/hooks/useCheckboxState.ts
```diff
@@ -0,0 +1,42 @@
...
+) => {
+  const checkboxState = visibleUrls.every((url) =>
+    selectedUrls.has(url.path + url.method)
+  )
+    ? CheckboxState.Checked
+    : visibleUrls.every((url) => !selectedUrls.has(url.path + url.method))
+    ? CheckboxState.NotChecked
+    : CheckboxState.Indeterminate;
+
+  const toggleSelectAllCheckbox = () => {
```
niclim, line 23: When we have:
- indeterminate / all checked -> unselect all selected in the current view
- not checked -> select all in the current view 

This means we have to take into account what the current visible urls are (i.e. the checkbox behavior should not add / remove if we've got a search term in here)

workspaces/ui-v2/src/optic-components/hooks/diffs/LearnInitialBodiesMachine.ts
```diff
@@ -114,6 +121,7 @@ export const newInitialBodiesMachine = (
                 assign({
                   allCommands: (ctx) => recomputeCommands(ctx),
                 }),
+                sendUpdate(),
```
niclim, line 124: This is the magic thing from xstate that avoids the hacky thing that i was doing here https://github.com/opticdev/optic/pull/742

acunniffe (approved):

PR #762, opened 2021-05-13
Add `was_empty_array` and `was_empty_array_trails` to `InteractionAffordances` so UI can reason about not rendering empty array examples

JaapRood:
## Why
In the Diff Review UI we'll want to distinguish between examples that have just an empty array, or an array with actual values. While both make valid examples of an array, one with values takes preference over an empty one, as its more informative to the user. For the UI to be able to reason about this, the Shape Diff Affordances learner must tell it which interactions contains empty arrays.

## What
The result of the Shape Diff Affordance learner now has a `was_empty_array` and `was_empty_array_trails`, which are subsets of `was_array` and `was_array_trails` respectively. 

It also includes a bug fix where the `was_x_trails` would only ever include a single json trail per interaction pointer, which instantly stood out given the relation ship between arrays and empty arrays.

## Validation
* [ ] CI passes
* [x] `was_empty_array` and `was_empty_array_trails` are included in the ShapeDiffAffordance learner results
* [x] subset / superset relationship between arrays and empty arrays holds

### DIFF
--- a/workspaces/cli-shared/src/diffs/initial-types.ts
+++ b/workspaces/cli-shared/src/diffs/initial-types.ts
@@ -42,6 +42,7 @@ export interface IAffordanceInteractionPointers {
   wasArray: string[];
   wasObject: string[];
   wasMissing: string[];
+  wasEmptyArray: string[];
   wasStringTrails: { [key: string]: IJsonTrail[] };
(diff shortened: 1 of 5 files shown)


acunniffe:
The frontend now uses these new affordances and prioritizes showing array trails with contents

Screen Shot 2021-05-13 at 8 58 07 AM

PR #765, opened 2021-05-13
new endpoints should share path lineage

acunniffe:
## Why
Optic needs to build an accurate path graph for all the endpoints in your API. Optic 10 was creating new path components for every endpoint, that did not share a common lineage. 

## What
Now all paths will share a common lineage that accends from the leaves up to the 'root' path component. 

This was a little more complicated than initially expected. In optic 9, we knew all the new path patterns we had to support since all endpoints were documented in the same batch. Optic 10 lets you unstage endpoints and read/change paths. This means changes to one pending endpoint has the potential to have side effects on the pathId that should be used by another. 

Rather than rebuilding the commands for each endpoint when other endpoints are added/deleted, this implementation uses separate path commands for previewing new endpoints. When it comes time to save all the endpoints, a shared set of path commands and pathIds compatible with the entire batch of new endpoints is used. 

In the future, we should consider if we want the learn_undocumented_bodies to know about pathId -- because it returns commands to the UI, not just affordances, I have to rewrite those commands before applying them to the top-level spectacle. This isn't hard, but does make one scratch their head. 

 
## Validation
* [x] CI passes
* [x] Unit tests for stable path matcher
* [x] Video https://www.loom.com/share/f16efd84616a4ada804e758bcfb55b11
* etc...


workspaces/spectacle/src/in-memory/index.ts
```diff
@@ -261,7 +261,7 @@ export class InMemoryDiffService implements IOpticDiffService {
         this.dependencies.opticEngine.learn_undocumented_bodies(
           spec,
           interactionsJsonl,
-          'sequential'
+          'random'
```
acunniffe, line 264: @devdoshi  -- Whenever we document new endpoints in an example session they are getting ShapeID conflicts. Is changing this right? Can it be parameterized?

devdoshi: They shouldn't get conflicts with the random one and it should be random by default. For the sequential one we would some extra state

workspaces/ui-v2/src/optic-components/hooks/diffs/LearnInitialBodiesMachine.ts
```diff
@@ -174,14 +190,31 @@ export function recomputeCommands(ctx: InitialBodiesContext): any[] {
...
-        getEndpointId({ method: ctx.method, pathId: ctx.pathId }),
+        getEndpointId({ method: ctx.method, pathId }),
         'purpose',
         ctx.stagedEndpointName
       )
     );
   }
 
-  return commands;
+  return commands.map((i) => {
```
acunniffe, line 200: this is the part that feels funcky. I have to rewrite the pathIds to get this to work with the global commands

devdoshi:
I suspect if you carried the parentPathId + metadata -> pathid map around, which you should be able to do naturally with xstate, then you wouldn't have to do the weirdness. We can pair on it tomorrow

niclim (commented):
I think this will work, but it seems odd - what do you think of doing something like:
- Passing in the existing `context.pendingEndpoints` array into [here](https://github.com/opticdev/optic/blob/a97b04ddd114fd5bdeb4ec2fb88d82d1a4f350d8/workspaces/ui-v2/src/optic-components/hooks/diffs/SharedDiffState.ts#L117-L123)
- Passing that pending endpoints into [pathToCommands](https://github.com/opticdev/optic/blob/develop/workspaces/ui-v2/src/optic-components/hooks/diffs/LearnInitialBodiesMachine.ts#L90)
- Updating pathToCommands to look up in pendingEndpoints if there's any path that matches [here?](https://github.com/opticdev/optic/blob/develop/workspaces/ui-v2/src/optic-components/hooks/diffs/LearnInitialBodiesMachine.ts#L230) - it might be easier to generate a path nested hashMap / graph to look up what the pathIds should be

devdoshi:
@acunniffe and I discussed this and we'll merge as-is for now and make the fixes soon when we make some other related state management changes

PR #771, opened 2021-05-15
Removed OPTIC_UI_HOST env var

bojan88:
Hi, and thank you for developing such a nice tool.
I want to get involved a bit, so I'm running `develop` branch on one of my projects

## Why
I noticed wrong UI URL when I start my project, and I'm suspecting this was left out when you migrated to `ui-v2`.
Feel free to close this PR if you think this is not the right way to fix it, or let me know if you want me to update it.

## What
Just a dev command fix and cleanup

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/sourceme.sh
+++ b/sourceme.sh
@@ -4,7 +4,7 @@ export OPTIC_SRC_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && p
 echo "Optic development scripts will run from $OPTIC_SRC_DIR"
 export OPTIC_DEBUG_ENV_FILE="$OPTIC_SRC_DIR/.env"
 
-alias apidev="OPTIC_DAEMON_ENABLE_DEBUGGING=yes OPTIC_DEVELOPMENT=yes OPTIC_UI_HOST=http://localhost:3000 OPTIC_AUTH_UI_HOST=http://localhost:4005 $OPTIC_SRC_DIR/workspaces/local-cli/bin/run"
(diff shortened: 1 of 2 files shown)


devdoshi:
Hi @bojan88 - thanks! We'd love to talk more - please reach out to us via https://calendly.com/opticlabs/maintainer-office-hours?month=2021-05 (via https://useoptic.com/docs/community/) if you'd like to chat face to face.

We use `apidev` in conjunction with 
```
cd workspaces/ui-v2
yarn start:local
```
which exposes the ui on localhost:3000 as expected by `apidev`

If you don't have the development ui started, you may wish to try `apistage` instead of `apidev`. Does that make sense? Please let me know if there was a different issue.

bojan88:
Hi @devdoshi, thanks for explaining.
The problem before this change is that if you run `apidev start`, it outputs something like `[optic] Review the API Diff at http://localhost:3000/apis/1/diffs` which is not true. In fact `apidev` is serving at `34444`, and for someone trying to run `apidev`, and not knowing that, it can lead to some confusion, and turn away possible contributors.

If you are starting `yarn start:local` from another terminal session, you'll get `http://localhost:3000` from webpack, and in that case I see the change from this PR as valid.
Please feel free to close this PR if I'm wrong, or there's another use case that I'm not aware of. Now that I figured out what's going on, I don't find this problematic anymore, but for anyone just starting with this project, it can be a problem.

devdoshi:
I think you are pointing out a valid issue, however I would make a different change since there are 2 valid developer flows, one who is only iterating on the local-cli and one who is iterating on the ui. I'll look at updating the docs and I'll mention you when I make the change so you can review! Thanks again

bojan88:
Thanks @devdoshi. Looking forward for the docs update

PR #772, opened 2021-05-15
Stop cli-server on exit

bojan88:
## Why
I noticed that cli-server is not being stopped on `api` process exit, and I guess that's not intentional.

## What
Just a bug fix, I hope.

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


devdoshi:
Hi @bojan88 - thanks for the contribution! Per our understanding of how developers are using Optic, they may have multiple API projects running at once, and they may stop/start their API processes many times during their development workflow. Because of this, it makes sense to us to keep the cli-server running like a daemon. So, I believe we would like to keep the current behavior. If you would like to kill the cli-server after the command completes, you may run `api daemon:stop` - does that meet your needs?

bojan88:
Thanks for explaining. It completely makes sense.
My assumption was wrong, and I didn't know about `daemon:stop` command. It's not in the CLI help output, so that might be something worth adding.

PR #773, opened 2021-05-16
Diff page improvements

bojan88:
## Why
* Capture dropdown was not working (diffs below it were not being updated on change). That's fixed here
* Nervous dropdown button fix. Please check the video to see how it works without this change (CPU was throttled on the video to 6x to make it a bit slower and easier to spot)
* Rendering performance improvements. There were a lot of excess re-renders, and unmounts (that's why the dropdown button was so nervous). See the screenshots attached (Chrome exports are a bit big, but I can attach them too if you need them)

https://user-images.githubusercontent.com/1783133/118403815-3b0f0600-b63e-11eb-9eb5-7be18793cbae.mp4

--
--

#### Before
optic-capture-change-develop

#### After
optic-capture-change-fix

## What
Just some bug fixes for Diff page, and performance improvements

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/ui-v2/src/optic-components/pages/diffs/CapturePage.tsx
+++ b/workspaces/ui-v2/src/optic-components/pages/diffs/CapturePage.tsx
@@ -16,6 +16,7 @@ import {
 } from '@material-ui/core';
 import { AddedDarkGreen, OpticBlue, OpticBlueReadable } from '../../theme';
 import { CaptureSelectDropdown } from '../../diffs/contexts/CaptureSelectDropdown';
+import { LoadingDiffReview } from '../../diffs/LoadingDiffReview';
 import { useSharedDiffContext } from '../../hooks/diffs/SharedDiffContext';
 import {
   useDiffEnvironmentsRoot,
@@ -34,13 +35,11 @@ export function CapturePage(props: { showDiff?: boolean }) {
 
   const noCaptures =
     !capturesState.loading && capturesState.captures.length === 0;
+  const shouldRedirect =
+    !capturesState.loading && !props.showDiff && !!capturesState.captures[0];
 
   useEffect(() => {
-    if (
-      !capturesState.loading &&
-      !props.showDiff &&
-      capturesState.captures[0]
-    ) {
+    if (shouldRedirect) {
       history.push(
         diffEnvironmentsRoot.linkTo(
           'local',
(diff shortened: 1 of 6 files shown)


workspaces/ui-v2/src/optic-components/pages/diffs/ReviewDiffPages.tsx
```diff
@@ -46,6 +46,9 @@ export function DiffReviewPages(props: any) {
     allPaths.loading ||
     allRequestsAndResponsesOfBaseSpec.loading;
 
+  // TODO: we have a problem here because isLoading is false on mount,
+  // then true on next render and after that it becomes false upon loadin
+  // this results in a page content flash before it starts loading data
```
niclim, line 51: I'm not sure I follow this - are you talking about initial mount, or a capture change? [On initial mount](https://github.com/opticdev/optic/blob/develop/workspaces/ui-v2/src/optic-components/hooks/useDiffForCapture.ts#L30) `diff.loading` is `true`.

workspaces/ui-v2/src/optic-components/pages/diffs/CapturePage.tsx
```diff
@@ -129,6 +129,10 @@ export function CapturePage(props: { showDiff?: boolean }) {
   );
 }
 
+export const CapturePageWithDiff = (props: any) => (
```
niclim, line 132: This works for now to reduce rerenders - but I think there's some amount of work we can do with [NavigationRoute](https://github.com/opticdev/optic/blob/96010d79fc65e0b4d5af65a4feb35017384e9e2f/workspaces/ui-v2/src/optic-components/navigation/NavigationRoute.tsx#L14-L31) so we aren't creating new react elements (instead invoking these as functions). This is fine for now though

niclim (commented):
Thanks for the PR! This looks good, just a question about your TODO comment

workspaces/ui-v2/src/optic-components/pages/diffs/ReviewDiffPages.tsx
bojan88: I was wrong about this... my bad. You are right, `isLoading` is true on mount.
My assumption was that `isLoading` was responsible for page content flash before displaying the loading indicator. To reproduce this go to Docs page, and then go to Diffs page. But this minor problem is caused by rendering `'/apis/:specId/diffs'` and then redirecting to `'/apis/:specId/diffs/local/:captureId'`. I added another commit to fix this

workspaces/ui-v2/src/optic-components/pages/diffs/CapturePage.tsx
bojan88: Maybe something like `const MemoedComponent = React.useMemo(() => Component, [path]);` in `NavigationRoute`, but that might not be the best solution.

niclim: I think the most straightforward thing would be to change `<Component />` to `Component()` - i.e. fn call instead of react.createElement - although there's probably better ways to compose / design these components so we don't need to inject components.

workspaces/ui-v2/src/optic-components/pages/diffs/ReviewDiffPages.tsx
niclim: Thanks for digging into this!

niclim (approved):
Looks great - thanks for the PR and the changes!

PR #776, opened 2021-05-17
refactor: `ci` flag always implies coverage, exit code

jshearer:
I had wanted to allow explicitly disabling these features (`print-coverage`, `pass-exit-code`, etc). But since they have defaults specified, there's no way to know whether they were explicitly set to false, or defaulted to false as a result of not being set (and so they weren't getting the `true` default here).

Instead, let's just have `--ci` imply them. Is this a bad idea?

### DIFF
--- a/workspaces/local-cli/src/shared/local-cli-task-runner.ts
+++ b/workspaces/local-cli/src/shared/local-cli-task-runner.ts
@@ -107,10 +107,9 @@ export async function LocalTaskSessionWrapper(
   deprecationLogger.enabled = true;
 
   if (flags['ci']) {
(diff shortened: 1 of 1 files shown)


devdoshi:
makes sense to me but I'd defer to @acunniffe and @LouManglass

jshearer:
@acunniffe approved per call

LouManglass:
Sweet. Yeah, that sounds reasonable to me - They were originally individual items and we initially planned a different ci approach ( a separate agent). This is a more holistic design that takes into account all of the features at this point, and it's a solid choice IMO.

PR #778, opened 2021-05-17
shapehash: Handle '{}' without crashing

raybejjani:
## Why
toJson iterated over `item.fields` but this is empty when a JSON object is empty. The result was a crash while iterating on `undefined`.

## What
We now only iterate if the `.fields` is valid.

## Validation
`yarn ws:test` passes the new test without crashing

### DIFF
--- a/workspaces/shape-hash/src/json-to-shape-hash.ts
+++ b/workspaces/shape-hash/src/json-to-shape-hash.ts
@@ -75,11 +75,12 @@ export function toJsonExample(hash: string): any {
 function toJson(item: any) {
   switch (item.type) {
     case types.OBJECT:
-      const newObj = {};
-      item.fields.forEach(({ key, hash }: any) => {
-        //@ts-ignore
-        newObj[key] = toJson(hash);
-      });
+      const newObj: Record<string, any> = {};
+      if (item.hasOwnProperty('fields')) { 
+        item.fields.forEach(({ key, hash }: any) => {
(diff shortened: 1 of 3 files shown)


workspaces/shape-hash/src/json-to-shape-hash.ts
```diff
@@ -76,10 +76,12 @@ function toJson(item: any) {
   switch (item.type) {
     case types.OBJECT:
       const newObj = {};
-      item.fields.forEach(({ key, hash }: any) => {
-        //@ts-ignore
-        newObj[key] = toJson(hash);
-      });
+      if (item.hasOwnProperty('fields')) {
```
raybejjani, line 79: Is there a better way to do this? Perhaps a variant of the `forEach`?

devdoshi: you can use item.fields.reduce(accumulator, newObj), but if it's possible for types.OBJECT to have .fields be undefined then you need the hasOwnProperty check. I believe the //@ts-ignore could be removed if newObj were typed appropriately as well

devdoshi (approved):

workspaces/shape-hash/src/json-to-shape-hash.ts
raybejjani: 👍  done. I'll merge once the tests pass.

PR #781, opened 2021-05-17
docs: Fix typo in `README.md`

jshearer:
## Why
Fixes typo I made when fixing a typo in `README.md`

## What
The actual key is `useTask`, not `usesTask` :(


LouManglass (approved):
I guess I should increase my text zoom level, since I approved that last change and totally missed the extra `s`.

PR #782, opened 2021-05-17
refactor: Make `absolutePathPatternWithParameterNames` return `/` for the root path

jshearer:
## Why
In this particular case, it felt smelly to special-case `pathId === "root"` in the coverage renderer, so as per my conversation with @devdoshi, I did this instead. Also included shall be updates to the snapshot tests that were expecting empty string.

## What
This makes the field `absolutePathPatternWithParameterNames` return `/` as opposed to empty string when processing the root path. 

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/graph-lib/src/endpoints-graph/index.ts
+++ b/workspaces/graph-lib/src/endpoints-graph/index.ts
@@ -260,7 +260,13 @@ export class PathNodeWrapper implements NodeWrapper {
   get absolutePathPatternWithParameterNames(): string {
     let path = '';
 
-    for (const component of this.components()) {
+    const components = this.components();
+
+    if (components.length === 1 && components[0].value.pathId === 'root') {
+      return '/';
+    }
+
+    for (const component of components) {
       if (component.value.pathId === 'root') continue;
       if (component.value.isParameterized) {
         path = `${path}/{${component.value.name}}`;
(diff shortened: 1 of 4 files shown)


jshearer:
Hmmm. Everything passes here. 

It looks like there are snapshots that should be failing:

https://github.com/opticdev/optic/blob/a97b04ddd114fd5bdeb4ec2fb88d82d1a4f350d8/workspaces/ui-v2/src/lib/__tests/__snapshots__/stable-path-batch-generator.test.ts.snap#L98

https://github.com/opticdev/optic/blob/a97b04ddd114fd5bdeb4ec2fb88d82d1a4f350d8/workspaces/spectacle/tap-snapshots/test-apply-commands.ts-TAP.test.js#L16

Are the corresponding tests here not running in CI? How can I run them to update these snapshots? (@devdoshi/@acunniffe?)

devdoshi:
@jshearer only the rust tests currently run in ci. you can run `yarn ws:test` in each workspace to run the tests to update the snapshots for now

jshearer:
It appears that the snapshots weren't updated a while ago when `absolutePathPatternWithParameterNames` was added, so I included those here as well

devdoshi (approved):

PR #784, opened 2021-05-17
Add source to opticrc

niclim:
## Why

Adding a source to distinguish between different usages

## What

Added source to `.opticrc` - also consolidated the 3 usages of analytics into 1 place (previously, we called identify on the UI + local CLI, and sent events through the UI + cli-server)

I noticed we also are using the same analytics provider for all instances (localcli, public example and cloudviewer) - I'm going to follow up this PR with a updated analytics store to handle all three instances

In the future:
- Consolidate events (UI and cli)
- Update the event bus to merge contexts (instead of overwrite)
- Rethink the analytics interfaces (context should be automatic, data should be specifiable can be passed through from UI / cli invocations) 

## Validation

tested this out locally and saw events hitting the event bus - also saw my local opticrc get updated with the default source


workspaces/cli-client/src/client.ts
```diff
@@ -10,15 +10,13 @@ class Client {
...
-      return JsonHttpClient.putJson(url, { user });
-    }
+  postTrackingEvents(events: TrackingEventBase<any>[]) {
+    const url = `${this.baseUrl}/tracking/events/apiname`;
+    return JsonHttpClient.postJson(url, { events });
   }
 
-  postTrackingEvents(apiName: string, events: TrackingEventBase<any>[]) {
-    const url = `${this.baseUrl}/tracking/events`;
+  postTrackingEventsWithApi(apiName: string, events: TrackingEventBase<any>[]) {
```
niclim, line 18: This is a little odd, we might want to change the interfaces of these such that you can pass arbitraty data and that gets forwarded through, or it's attached to each event.

Right now this should be identical to what we are currently sending

workspaces/cli-server/src/analytics.ts
```diff
@@ -60,6 +61,12 @@ const inDevelopment = process.env.OPTIC_DEVELOPMENT === 'yes';
 // segment io sink
 const token = 'RvYGmY1bZqlbMukS8pP9DPEifG6CEBEs';
 const analytics = new Analytics(token);
+// Identify user
+getOrCreateAnonId().then((anonymousId) =>
```
niclim, line 65: This should run on file load, which I think should work identically to the current implementation. I didn't port over the jwt decoding since we aren't using login, but if we reimplement a login, we can readd it here

workspaces/analytics/src/eventbus.ts
```diff
@@ -14,6 +14,7 @@ export function newAnalyticsEventBus(
     emit: async (...events: TrackingEventBase<any>[]) => {
       const batchId = uuidv4();
       const context = await getContext(batchId);
+      // TODO - don't overwrite context
```
devdoshi, line 17: as in Object.freeze?

niclim: No, `events` contains a `context` key, but here we completely ignore it and overwrite it. We probably should merge these, but if we take a step back we should probably distinguish what we use context for (and what we use data for)

Data could be used as a "event specific" items, and context is something that should be sent with every event. Which would involve updating the typings (such that events, and UI events passed through event specific data, and context is applied always in the eventbus)

workspaces/ui-v2/src/analytics/index.tsx
```diff
@@ -86,24 +68,34 @@ export function AnalyticsStore({ children }: { children: ReactNode }) {
...
-          try {
-            FullStory.event(event.name, event.properties);
-          } catch (e) {}
-        }
+        // TODO consolidate UI and cli events types
+        cliClient.postTrackingEvents([
+          {
+            type: event.name,
+            data: event.properties,
+            // TODO update the context typing, this currently gets overriden in
```
devdoshi, line 81: lingering todo?

workspaces/analytics/src/eventbus.ts
devdoshi: Oh I see, yep that makes sense.

devdoshi (approved):
approved pending @acunniffe

acunniffe:
This is great @niclim  -- I'm glad we made time for this in the release. Thanks for owning it

PR #785, opened 2021-05-17
Fix misc contributions

niclim:
## Why

Two bugs:
- Response contributions weren't showing a default value
- If you try to do two saves on the same documentation page - the old state is still there and you can't resave

## What

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


workspaces/ui-v2/src/optic-components/pages/docs/components/EditContributionsButton.tsx
```diff
@@ -50,17 +50,18 @@ export function EditContributionsButton() {
       >
         {contents}
       </ToggleButton>
-      <CommitMessageModal
```
niclim, line 53: Changing this so it gets remounted such that state is reset each time we open + close

devdoshi (approved):

PR #786, opened 2021-05-18
Fix getrandom

devdoshi:
Please ignore for now :)


devdoshi:
we can revisit this next time we visit js tests

PR #788, opened 2021-05-18
Fix bug where undocumented body learner would miss certain optional fields

JaapRood:
## Why
In some of our testing, it has stood out how sometimes after adding a new undocumented endpoint, there would still be open diffs, all of which were meant to be optional. Adding a new endpoint should generate a spec that is fully compliant with whatever interactions it has seen, and should yield no further diffs.

## What
The merging of object field sets turned out not to be be symmetrical, which lead field sets to not be recorded. In addition, reducing all known field sets for an object also wasn't symmetrical, allowing the order in which field sets were seen to influence which were considered optional (where it should be unaffected). This PR adds a test that requires this symmetry, and a fix to ensure it.

In addition, it verifies that the commands generated for the test bodies, result in no further shape diffs being able to be generated.

## Validation
* [x] CI passes
* [x] adding new endpoints through the UI for our test sessions no longer yields additional diffs
* [x] all command generation tests verify no additional shape diffs

### DIFF
--- a/workspaces/diff-engine/src/learn_shape/result.rs
+++ b/workspaces/diff-engine/src/learn_shape/result.rs
@@ -527,6 +527,7 @@ mod test {
   use super::*;
   use crate::learn_shape::observe_body_trails;
   use crate::projections::SpecProjection;
+  use crate::shapes::diff as diff_shapes;
   use crate::state::body::BodyDescriptor;
   use cqrs_core::Aggregate;
   use insta::assert_debug_snapshot;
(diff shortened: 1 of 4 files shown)


JaapRood:
Got to add assertions for all the learner command generation tests to verify no additional shape diffs are yielded, then this should be ready.

JaapRood:
@acunniffe all covered now, feel free to merge upon an accepted review!

JaapRood:
@acunniffe seems like this is failing on the tests you've added to ui-v2, can you help resolve them? 

There's a panic in trying to apply root array commands, which I think you're addressing at #789? 

Apart from that it seems to be asserting whether the commands generated are correct. I'd suggest not doing that, as that's already covered by tests lower down, where we don't only have snapshots, but also more direct assertions about commands correctness (like whether they can be applied and don't yield any further diffs when done so). The consuming code should only care about that: that the commands apply properly and have the desired effect, not exactly how they are written, in what order, with what metadata, etc. Otherwise every subtle change will ripple through down the snapshots, even if there's no change in the API contract, seems silly.

JaapRood:
@acunniffe @devdoshi now with everything passing we're but a quick review away from a merge and another bug squashed 😅

devdoshi (approved):

PR #790, opened 2021-05-18
Split analytics implementations for different UI variants

niclim:
## Why

We've got a few different UI variants. The cloudviewer and some instances of the public viewer (demo, locally developed ui) do not have local-cli servers running - meaning we can't proxy through the cli server and have to go through segment directly. 

Note that this PR targets this PR https://github.com/opticdev/optic/pull/784

## What

- Split out into 3 different implementations - public + cloud viewer are similar, but they could diverge, especially if we want to track / not track
- I imagine at some point, we'll move all of this into a redux middleware layer, so we can access the same states / fns / getters, etc which will make this easier to juggle / access common state - this will also make it easier for us to control / send certain events on certain variants (cloud viewer only, or localcli only, etc)
- Added the `uiVariant` tag to be able to see different sources / usages of this.

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


workspaces/ui-v2/src/analytics/AnalyticsStore.tsx
```diff
@@ -0,0 +1,84 @@
...
+  clientAgent: string;
+  apiName: string;
+};
+const defaultMetadata = {
+  clientId,
+  clientAgent: 'anon_id',
+  apiName: '',
+};
+
+export type AnalyticsStoreProps = {
```
niclim, line 28: Exposed these fns which act as implementations. I think this would fit well as redux middleware.

Currently we need to access shared state through hooks, which isn't ideal since some hooks _may_ only be usable when the cli-server is running (meaning we either have to build the hooks to run in all contexts) or just access the data programatically here

I also considered building three implementations of AnalyticsStore, but we want to use the same `useAnalytics` hook - and to do that with rendered components is a little challenging / bad for render optimization

workspaces/ui-v2/src/spectacle-implementations/cloud-viewer.tsx
```diff
@@ -107,7 +112,13 @@ export default function CloudViewer() {
             capturesService={data.opticContext.capturesService}
           >
             <BaseUrlProvider value={{ url: match.url }}>
-              <AnalyticsStore>
+              <AnalyticsStore
+                getMetadata={getMetadata(() =>
```
niclim, line 116: There's something weird that I had to call `getMetadata(() => fn())` instead of `getMetadata(fn)` - where the first instance / render of this would return an invalid function...

Error I was getting `opticContext.configRepository.getApiName() -> `this.dependencies` is undefined` - which crashed the page

I didn't dig too much more since I think this is better replaced if we use redux + implement analytics middleware, but it is interesting.

workspaces/ui-v2/src/analytics/implementations/publicExampleAnalytics.ts
```diff
@@ -0,0 +1,98 @@
...
+
+export const initialize: AnalyticsStoreProps['initialize'] = async (
+  metadata,
+  appConfig
+) => {
+  if (!appConfig.analytics.enabled) {
+    return;
+  }
+
+  if (appConfig.analytics.segmentToken) {
```
niclim, line 53: In the future we may or may not want to exclude / not send events here - or at least distinguish between onboarding vs actual usage

workspaces/ui-v2/src/analytics/implementations/cloudViewerAnalytics.ts
```diff
@@ -0,0 +1,77 @@
...
+const segmentAnalytics = new Analytics();
+
+export const getMetadata: (
+  apiNameFetcher: () => Promise<string>
+) => AnalyticsStoreProps['getMetadata'] = (apiNameFetcher) => async () => {
+  const apiName = await apiNameFetcher();
+
+  return {
+    clientId,
+    clientAgent: 'cloud_viewer_anon_id',
```
devdoshi, line 21: I feel like it's useful to provide an actual uuid here so we can differentiate between different visitors more easily. But  I don't feel strongly.

workspaces/ui-v2/src/analytics/AnalyticsStore.tsx
devdoshi: I thought as long as the context provider was the same, the useAnalytics hook could be ignorant of the value provided by the different store impls?

workspaces/ui-v2/src/spectacle-implementations/cloud-viewer.tsx
devdoshi: i think that's normal behavior if you pass a class method without rebinding it, but seems fine to wrap it anyway

devdoshi: also passing a function to some hooks call it immediately, so you need to wrap, that might be it as well

workspaces/ui-v2/src/analytics/implementations/cloudViewerAnalytics.ts
niclim: Hmm, I initially added this as a fallback, but I think using localStorage to generate an id is probably _good enough_

There's obvious pitfalls with this with reporting, but we'll get a "good enough" idea using local storage

workspaces/ui-v2/src/analytics/AnalyticsStore.tsx
niclim: Correct - the key part is the context provider needs to be the same. If we need to call different hooks conditionally in different contexts, we'd probably need to mount different components in the `AnalyticsStore` which can get messy

In this, I'm just pulling out any hook related content and changing them work programatically

workspaces/ui-v2/src/spectacle-implementations/cloud-viewer.tsx
niclim: I was writing it as `getMetadata(data.opticContext.configRepository.getApiName)` which should have passed the function reference to the correct place. I think the first render is that it hasn't loaded yet (loading state) - and because i _think_ it can't tell the fn identity changes - it calls the old one. But wrapping this with an inline function means it's accessed on call?

It could also be that I'm using a useRef here. It's odd because the function is being called at the same time, it just had a reference to a stale instance and wrapping a fn means that we are able to access the correct function reference

Either way - this works 🤷

workspaces/ui-v2/src/analytics/implementations/cloudViewerAnalytics.ts
niclim: Ok I've added in a localStorage id - used always in cloudviewer and used as a fallback for the other two.

Note that localStorage isn't shared across domains - so cloudviewer and localcli/public example will never (?) point to the same user

devdoshi (approved):

JaapRood (commented):
Retroactively: looks good to me, dealt with sensibly!

PR #792, opened 2021-05-18
Add Spectacle documentation

smizell:
This PR focuses on Spectacle documentation and making GraphiQL more accessible to the users.

* Adds a CLI command `api spectacle` for opening up GraphiQL in the browser
* Adds documentation to the GraphQL schema
* Adds a page in the Optic documentation
* Adds a note in the Scripting documentation page about the `SPECTACLE_URL` environment variable

Note: there are a few places in the schema that I wasn't for sure about the description, so I've left them out for now.

### DIFF
--- a/website/docs/using/spectacle.mdx
+++ b/website/docs/using/spectacle.mdx
@@ -1,0 +1,84 @@
+---
+title: Using Spectacle
+---
+
+Optic provides you with the ability to build [scripts](/docs/apiops/scripts/) around your API Ops process. Sometimes you may want to access data that Optic gathers as it's documenting your APIs in your scripts. This might include the spec itself or changes to the spec. Optic provides a tool called Spectacle to make this possible.
+
+Spectacle is a GraphQL API that enables you to interact with Optic's data, either through exploring it in your browser with GraphiQL or querying in your code using Optic scripts.
+
+## Accessing Spectacle through GraphiQL
+
+To open GraphiQL for your locally-running instance of Spectacle, run the CLI command:
+
+```sh
+api spectacle
+```
+
+This will ensure Spectacle is running and open your browser to the Spectacle URL where you can explore the API and its documentation.
+
+## Building scripts around Spectacle
+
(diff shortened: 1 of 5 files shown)


website/docs/using/spectacle.mdx
```diff
@@ -0,0 +1,84 @@
...
+    body: JSON.stringify(body),
+    headers: { "Content-Type": "application/json" },
+  });
+  return await result.json();
+}
+```
+
+1. Install [Node.js](https://nodejs.org/en/) if you haven't already
+1. Save this code above as `spectacle.js`
+1. Install `node-fetch` with `npm install node-fetch`
```
devdoshi, line 67: i've never done this without doing npm init first, does it work?

devdoshi (approved):
Looks good to me. If we need something richer, we can expose graphiql on the frontend as well, so we can use it in our browser demos. https://github.com/graphql/graphiql/blob/main/packages/graphiql/README.md#usage-npm-module

website/docs/using/spectacle.mdx
smizell: Yeah, it creates the `package.json` with just the dependencies:

```js
{
  "dependencies": {
    "node-fetch": "^2.6.1"
  }
}
```

smizell:
Oh, I like that idea, too. That would make it even more visible.

website/docs/using/spectacle.mdx
```diff
@@ -0,0 +1,84 @@
...
+
+We'll look at an example using a Node.js script. The one below is using the `requests` query to get information about requests and responses in the spec. It's using `node-fetch` as the library to query Spectacle.
+
+```javascript
+const fetch = require("node-fetch");
+
+main();
+
+async function main() {
+  const result = await querySpectacle({
```
acunniffe, line 31: @smizell -- shouldn't we show people how to get the `querySpectacle` to work? Do they need to import the Spectacle package? something else? 

Might be good to expose a function in the package that looks for the ENV in script and just makes it dead simple to call it.l

PR #793, opened 2021-05-18
Fix ui snapshots

niclim:
## Why

Some UI snapshots are broken since the id generation strategy was [changed](https://github.com/opticdev/optic/pull/765/files#diff-d8200f71e4d7532465bb20e37eeae97b3730dfe4883caaf43488acc54c2aa99cR274). 

These are generated in rust - and I'm not sure what / how we'd generate deterministic ids when controlling this from the JS code? I threw this PR up to filter out any key with a generated ID - but we really want to be testing / generating the ids

If there's a better way to do this, we can close this - just wanted to fix our tests

Also there seems to be a non-related change in the snapshot (i'll comment in the file)

## What

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


niclim:
Closing in favor of intercepting test call and calling sequential

PR #794, opened 2021-05-18
feature: Implement spec hashing

jshearer:
## Why
Running with the analogy of Optic as git for APIs, one important feature of git is the fact that each atomic unit of change has a content-addressable ID that will always refer to that and only that exact point in "repo space" (exact diff, ontop of the exact set of commits, by exactly that author, etc). In order for us to start talking about Optic's concept of APIs, we need a similar content-addressable identifier for a specific point in "API space", which is what this PR is all about.

The most immediate use of the spec hash will be to support uploading coverage reports associated with a particular spec. Soon afterwards, those specs will all become part of "named APIs", akin to how your arbitrary git commits are all associated with a particular repo in github. All of that depends on having stable unique IDs for a particular point in the timeline of a particular spec. We _could_ just randomly generate them, but the added quality of being able to statelessly derive the ID from the spec simplifies a bunch of things, and feels quite elegant to boot.

## What
I originally thought (as @devdoshi suggested) that I could just serialize each `SpecEvent` to a JSON string, hash it, and be done with. Unfortunately, the ordering of object keys in JSON is explicitly undefined, so I needed to find some way of getting a stable JSON string out. I evaluated two different libraries for this:

* https://borsh.io/
* https://crates.io/crates/serde-hashkey

I would have liked to use Borsh, since it seems faster and more of a whole _project_ rather than "just" a library that somebody put together. But unfortunately since Borsh explicitly avoids working in Serde land (for performance, among other things), I would have had to add a `BorshSerialize` impl to _every_ variant of `SpecEvent`, of which there are quite a few.

Instead, I went with `serde-hashkey` which effectively gives you a stable equivalent to `serde_json::serialize()`.

The hash is propagated out through a spectacle query, so it can be used anywhere Spectacle may be found.

## Validation
* [x] Tests pass
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/diff-engine/src/projections/hash.rs
+++ b/workspaces/diff-engine/src/projections/hash.rs
@@ -1,0 +1,58 @@
+use crate::{RfcEvent, SpecEvent};
+use cqrs_core::{Aggregate, AggregateEvent, Event};
+use serde_hashkey::to_key;
+use sha2::{Digest, Sha256};
+use std::{collections::BTreeMap, hash::Hash};
+
+#[derive(Debug, Clone)]
+pub struct SpecHashProjection {
+  hasher: Sha256,
+}
+
+impl SpecHashProjection {
+  pub fn to_string(&self) -> String {
+    format!("{:x}", self.hasher.clone().finalize())
+  }
+
+  pub fn with_event(&mut self, evt: SpecEvent) {
+    let key = to_key(&evt).unwrap();
+    let key_str = serde_json::to_string(&key).expect("should be able to serialize");
+
+    self.hasher.update(key_str);
+  }
+}
+
+impl Default for SpecHashProjection {
+  fn default() -> Self {
+    SpecHashProjection {
+      hasher: Sha256::new(),
+    }
+  }
+}
+
+impl Aggregate for SpecHashProjection {
+  fn aggregate_type() -> &'static str {
+    "spec_hash_projection"
+  }
+}
+
+impl<I> From<I> for SpecHashProjection
+where
+  I: IntoIterator,
+  I::Item: AggregateEvent<Self>,
+{
+  fn from(events: I) -> Self {
+    let mut projection = SpecHashProjection::default();
+    for event in events.into_iter() {
+      projection.apply(event);
+    }
+    projection
+  }
+}
+
+// Events
+impl AggregateEvent<SpecHashProjection> for SpecEvent {
+  fn apply_to(self, projection: &mut SpecHashProjection) {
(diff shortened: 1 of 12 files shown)


workspaces/diff-engine/src/projections/hash.rs
```diff
@@ -0,0 +1,58 @@
...
+  hasher: Sha256,
+}
+
+impl SpecHashProjection {
+  pub fn to_string(&self) -> String {
+    format!("{:x}", self.hasher.clone().finalize())
+  }
+
+  pub fn with_event(&mut self, evt: SpecEvent) {
+    let key = to_key(&evt).unwrap();
```
devdoshi, line 18: It's not _too_ different but if we can get away with sticking with serde_json and the answer here: https://stackoverflow.com/questions/42723065/how-to-sort-hashmap-keys-when-serializing-with-serde
then we can also use that to make our snapshots deterministic everywhere we use hashmaps, which is valuable

devdoshi: also not having the extra dependency

workspaces/spectacle/src/graphql/schema.ts
```diff
@@ -23,6 +23,7 @@ type Query {
   endpointChanges(sinceBatchCommitId: String): EndpointChanges
   batchCommits: [BatchCommit]
   diff(diffId: ID): DiffState
+  hash: String
```
devdoshi, line 26: @niclim was thinking of exposing a generic metadata projection, I think this might fit better aligned with that

jshearer: Yep, that makes sense. `hash` feels a bit naked here as it is, it could certainly use some company.

JaapRood:
See the internal discussions, but just as a safety gap against merging this: I reckon going this route is the wrong trade-off and I was advice against this approach.

JaapRood (changes requested):
Your implementation looks clean and sane, nice job! However, I reckon going this route is the wrong trade-off and I would advice against this approach. See our internal discussion for details, but in short: yes, being able to derive hashes from contents can be nice, but it locks us down significantly in making changes to that content, implementing fault tolerance mechanisms, etc. Instead, we should go for an approach where the a version that identifies a version of the spec is generated at write, and extracted afterwards, like systems like Git and many DB internals do.

jshearer:
Putting this on hold per convo with @devdoshi / @JaapRood. tl;dr we can just use the spec ID from the backend rather than this hash for now.

PR #795, opened 2021-05-18
Fix snapshots

niclim:
## Why

Intercepts the test snapshot call to use sequential ID generation

There's snapshot changes in one of the tests `new request body has correct commands` - I'm not that familiar with the changes that's been happening - is this intended?

## What

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/ui-v2/src/lib/__tests/diff-helpers/__snapshots__/new-regions-interpretations.test.ts.snap
+++ b/workspaces/ui-v2/src/lib/__tests/diff-helpers/__snapshots__/new-regions-interpretations.test.ts.snap
@@ -7,116 +7,81 @@ Object {
       "AddShape": Object {
         "baseShapeId": "$string",
         "name": "",
-        "shapeId": "shape_6699",
+        "shapeId": "shape_6697",
       },
     },
     Object {
       "AddShape": Object {
         "baseShapeId": "$boolean",
         "name": "",
+        "shapeId": "shape_6698",
+      },
+    },
+    Object {
+      "AddShape": Object {
+        "baseShapeId": "$string",
+        "name": "",
+        "shapeId": "shape_6699",
+      },
+    },
+    Object {
+      "AddShape": Object {
+        "baseShapeId": "$string",
+        "name": "",
         "shapeId": "shape_6700",
-      },
-    },
-    Object {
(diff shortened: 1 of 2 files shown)


workspaces/ui-v2/src/lib/__tests/diff-helpers/fixture.ts
```diff
@@ -32,6 +32,14 @@ export const testCase = (basePath: string) => async (
     require(path.join(__dirname + '/universes', basePath, name + '.json'))
   );
 
+  // Overwrite the learn_undocumented_bodies to ensure deterministic id generation
```
devdoshi, line 35: cool, i'm fine with this, but ideally we'd "just" pass the id generation strategy from the entry point i.e. when creating an inmemory opticcontext

devdoshi: the entry point should be trusted to set the right environment for clocks, ids, etc.

workspaces/ui-v2/src/lib/__tests/diff-helpers/__snapshots__/new-regions-interpretations.test.ts.snap
```diff
@@ -7,116 +7,81 @@ Object {
...
       },
     },
     Object {
       "AddShape": Object {
         "baseShapeId": "$object",
         "name": "",
-        "shapeId": "shape_6708",
-      },
-    },
-    Object {
```
niclim, line 41: @devdoshi here's the shape diff from the snapshot - is this diff related to the diff engine changes that's been going on?

devdoshi (approved):

workspaces/ui-v2/src/lib/__tests/diff-helpers/__snapshots__/new-regions-interpretations.test.ts.snap
devdoshi: yeah I imagine it'll be in flux as we fix some bugs around these areas, we'll be responsible for verifying the final snapshots

JaapRood:
Rebased #788 on top of this, still fails the snapshots :(

PR #799, opened 2021-05-19
docs and sourceme changes

devdoshi:
## Why
@bojan88 brought up that our defaults can be confusing if you follow the docs. the aliases exposed in sourceme   should allow for developers to iterate on ui alone, cli alone, and both ui and cli together.

## What
if you were using `apidev` while iterating on the ui, you should switch to `uidev` (I don't have strong opinions about the names, and since we are developers we're free to make our own aliases)

## Validation
* [ ] CI passes
* [ ] consensus on naming

### DIFF
--- a/Developer-setup.md
+++ b/Developer-setup.md
@@ -47,12 +47,23 @@ Each workspace should have a package.json file with a name field like `@useoptic
 
 ## Validate Setup to contribute to Optic’s Local CLI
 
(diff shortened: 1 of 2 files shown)


LouManglass (approved):
Pending bojan88's thoughts, but for purposes of review, this looks reasonable to me.

acunniffe (approved):

bojan88 (approved):
Thanks for including me @devdoshi, and sorry for the late reply.
I think this is much clearer now, and straightforward for anyone setting up the project for the first time.

PR #800, opened 2021-05-19
Fix ui snapshots

niclim:
## Why

~~Snapshot tests are failing again - and I'm not sure how this happened. ~~The snapshots started failing after [this commit](https://github.com/opticdev/optic/commit/200736b092c73aa5e77760f661cae38754b3fbab) but the new-regions-interpretations looks identical to the fix here https://github.com/opticdev/optic/pull/795~~ - ACTUALLY the snapshots were added back here https://github.com/opticdev/optic/commit/9feba24afc15c66d33a0efdbf401f81e960e880f~~

Turns out i didn't rebuild the workspace 🤦 - here's the fix that actually should work

I'm fairly confused, why the snapshot tests were failing, and why the commit passed before.  I _think_ we can avoid this by turning on `Require PRs to be up to date before merging` though...?

## What

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/ui-v2/src/lib/__tests/diff-helpers/__snapshots__/shape-interpretations.test.ts.snap
+++ b/workspaces/ui-v2/src/lib/__tests/diff-helpers/__snapshots__/shape-interpretations.test.ts.snap
@@ -6376,11 +6376,9 @@ Object {
     Object {
       "SetResponseBodyShape": Object {
         "bodyDescriptor": Object {
-          "ShapedBodyDescriptor": Object {
-            "httpContentType": "application/json",
-            "isRemoved": false,
-            "shapeId": "shape_0",
-          },
+          "httpContentType": "application/json",
+          "isRemoved": false,
+          "shapeId": "shape_0",
         },
         "responseId": "response_1",
       },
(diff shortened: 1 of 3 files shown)


workspaces/ui-v2/src/lib/__tests/diff-helpers/__snapshots__/shape-interpretations.test.ts.snap
```diff
@@ -6376,9 +6376,11 @@ Object {
     Object {
       "SetResponseBodyShape": Object {
         "bodyDescriptor": Object {
-          "httpContentType": "application/json",
-          "isRemoved": false,
-          "shapeId": "shape_0",
+          "ShapedBodyDescriptor": Object {
```
devdoshi, line 6379: This should have been removed, so the previous version is correct. It smells like someone overwrote changes in a merge?

niclim: Oh i don't think I rebuilt the workspace

niclim:
@devdoshi turns out i needed to rebuild the workspace to actually get this to work and generate the correct snapshots

devdoshi (approved):

PR #803, opened 2021-05-20
Fix/field key from trails

acunniffe:
## Why
Adding extra information to the shape trails reduces the # of queries that have to run when interpreting diffs. 

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


workspaces/diff-engine/package.json
```diff
@@ -5,7 +5,7 @@
     "postinstall": "node scripts/install",
     "preuninstall": "node scripts/uninstall",
     "test": "cd tests && tap ./lib",
-    "ws:build-binaries": "scripts/build-binaries.sh",
+    "ws:build": "cargo build",
```
devdoshi, line 8: @JaapRood is this right?

workspaces/ui-v2/src/lib/shape-trail-parser.ts
```diff
@@ -73,10 +66,9 @@ export async function shapeTrailParserLastId(
       return {
         lastObject: lastObjectOption
           ? lastObjectOption[0]
-          : field.parentObjectId,
+          : fieldTrail.parentObjectShapeId,
```
devdoshi, line 69: Could this just always use the trail now?

workspaces/diff-engine/package.json
JaapRood: Yeah that should be fine.

workspaces/ui-v2/src/lib/shape-trail-parser.ts
acunniffe: this is needed for new / nested fields.

PR #804, opened 2021-05-20
Fix bug where no Response would be learned for interactions without response bodies

JaapRood:
## Why
In some debug captures there were diffs left over after learning all endpoints. These were caused by the undocumented bodies learner not writing the commands for their responses. In fact, all learnings for any interaction with an empty response body would fail and not be learned, silently skipped.

## What
The root cause of this bug was in the way `InteractionTrail` from diffs didn't map correctly onto `BodyAnalysisLocation`, expecting to but not finding a status code. A different variant of `InteractionTrailComponent` had to be matched for responses with empty bodies, but because a catch all `_` was used, Rust allowed this to go through without issue, until it panicked with downstream code expecting a response to always have a status code!

The fix handles the variant correctly and relies on exhaustive matching to know we're not missing any others. Also comes with a bonus test making sure the right commands are written, too.

## Validation
* [x] CI passes
* [x] Interactions with empty response bodies are learned correctly.


JaapRood:
Ready to be merged once reviewed!

devdoshi (approved):

PR #807, opened 2021-05-20
fix array polymorphism

niclim:
## Why

Changing polymorphism now works
Screen Shot 2021-05-20 at 1 24 21 PM

## What
- Added/updated typing to make this easier to keep updated - proper typing would have caught this

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/ui-v2/src/optic-components/hooks/useShapeDescriptor.ts
+++ b/workspaces/ui-v2/src/optic-components/hooks/useShapeDescriptor.ts
@@ -1,4 +1,9 @@
-import { IShapeRenderer, JsonLike } from '../shapes/ShapeRenderInterfaces';
+import {
+  IShapeRenderer,
+  JsonLike,
+  IArrayRender,
+  IObjectRender,
+} from '../shapes/ShapeRenderInterfaces';
 import { SpectacleContext } from '../../spectacle-implementations/spectacle-provider';
 import { useContext, useEffect, useState } from 'react';
 import sortBy from 'lodash.sortby';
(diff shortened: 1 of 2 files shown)


workspaces/ui-v2/src/optic-components/shapes/ShapeRenderInterfaces.ts
```diff
@@ -15,12 +15,13 @@ export interface IShapeRenderer {
   jsonType: JsonLike;
   asArray?: IArrayRender;
   asObject?: IObjectRender;
-  value: any;
+  value?: any;
```
niclim, line 18: We aren't actually querying this, and i believe `any` actually include nullable - but I added a `?` to make it more explicit that we can't rely on this value.

workspaces/ui-v2/src/optic-components/hooks/useShapeDescriptor.ts
```diff
@@ -49,6 +54,30 @@ const changesSinceShapeQuery = `
...
+    }
+  | {
+      id: string;
+      jsonType: JsonLike.ARRAY;
+      asArray: IArrayRender;
+      asObject?: undefined;
+    }
+  | {
+      id: string;
+      jsonType: Exclude<JsonLike, JsonLike.OBJECT | JsonLike.ARRAY>;
```
devdoshi, line 72: gonna have to take your typescript course @niclim :)

devdoshi (approved):

workspaces/ui-v2/src/optic-components/hooks/useShapeDescriptor.ts
```diff
@@ -49,6 +54,30 @@ const changesSinceShapeQuery = `
     }
 }`;
 
+type ShapeChoice =
```
niclim, line 57: I'd want to see if there's a way to generate typings off gql queries, but I'm not sure how practical that is - in the meantime, this needs to be kept in sync with the gql query above

workspaces/ui-v2/src/optic-components/hooks/useShapeDescriptor.ts
niclim: Haha it's a lot less glamorous than the code - more of a try stuff and see if the compiler accepts my typings

PR #808, opened 2021-05-20
Add typings to spectacle queries

niclim:
## Why

Slowly removing `any` typings - this time, I tackled all the spectacle queries and stuff. Given that I'm just updating typings, there shouldn't be any changes to the code (there's one place in the code that was changed, and that was related to the previous PR https://github.com/opticdev/optic/pull/807 around shapes which I forgot to update one part)

Adding typings should also make it easier to refactor + give us more confidence in refactoring

## What

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/ui-v2/src/optic-components/hooks/useEndpointBodyHook.ts
+++ b/workspaces/ui-v2/src/optic-components/hooks/useEndpointBodyHook.ts
@@ -1,44 +1,48 @@
 import { useSpectacleQuery } from '../../spectacle-implementations/spectacle-provider';
 import { IChanges } from '../changelog/IChanges';
 
-export function useEndpointBody(
-  pathId: string,
-  method: string,
-  renderChangesSince?: string
-): { requests: IRequestBody[]; responses: IResponseBody[] } {
-  const spectacleInput =
-    typeof renderChangesSince === 'undefined'
-      ? {
-          query: `{
-    requests {
+const EndpointBodyQueryWithoutChanges = `
+{
+  requests {
+    id
+    pathId
+    method
+    requestContributions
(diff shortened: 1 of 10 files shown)


workspaces/ui-v2/src/optic-components/hooks/useEndpointBodyHook.ts
```diff
@@ -47,27 +51,57 @@ export function useEndpointBody(
...
+      bodies: Body[];
+    }[];
+  }[];
+};
+
+export function useEndpointBody(
+  pathId: string,
+  method: string,
+  renderChangesSince?: string
+): { requests: IRequestBody[]; responses: IResponseBody[] } {
```
niclim, line 85: There's probably some "cool" things we could do here with generics where if we specify `renderChangesSince` as a string, it'll return us changelog information - but I think splitting out the doc page from the contributions page is probably a better idea (since the doc page could diverge in functionality, especially with manual edits) meaning that we wouldn't need this to be generic

workspaces/ui-v2/src/optic-components/hooks/useShapeDescriptor.ts
```diff
@@ -145,7 +145,7 @@ export function useShapeDescriptor(
               })
             );
             choice.asObject.fields = newFields;
-            return choice;
+            return { ...choice, shapeId: choice.id };
```
niclim, line 148: These are the only changes we need to do in code - which is related to the https://github.com/opticdev/optic/pull/807 PR (i.e. shim id -> shapeId)

devdoshi (approved):

PR #816, opened 2021-05-21
Add specification ID

niclim:
## Why

Adding a new specification identifier in a metadata key for specifications

## What

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


workspaces/analytics/src/consistentAnonymousId.ts
```diff
@@ -1,2 +0,0 @@
-import { v4 as uuidv4 } from 'uuid';
```
niclim, line 1: Cleaning up unused... This also would have generated a new id each app start / cli start

workspaces/analytics/src/events/diffs.ts
```diff
@@ -1,6 +1,6 @@
 import { DescribeEvent } from '../interfaces/RegisterEvent';
```
niclim, line 1: I started touching this file but realized I didn't need to be here. That being said, I'm not a huge fan of the JOI structure for a couple of reasons:
- We aren't using the typing correctly. The idea behind this seems to be to ensure consistent typing / event structures - we can achieve the same thing with typescript with a lot less indirection (since typescript will statically check)
- Our UI events are using a completely different thing

What I'm planning to do in a future PR (unless there's any pushback here) is to:
- Remove context from the `TrackEventsBase` (and pass in context automatically)
- Consolidate UI events + CLI events - these are effectively just - an event name and an object, we don't need two different ways to declaring events.
- Build a lighter weight layer with better typesafety - we can probably remove a bunch of this stuff by just specifying event creators (functions with arguments, or even just regular object types and typescript can statically check these)

The most important thing here is making sure our events are backwards compatible (for the ones we care about)

workspaces/spectacle/src/index.ts
```diff
@@ -311,6 +312,35 @@ export async function makeSpectacle(opticContext: IOpticContext) {
...
+                  AddContribution: {
+                    id: metadataKey,
+                    key: specIdKey,
+                    value: specId,
+                  },
+                },
+              ],
+              uuidv4(),
+              'Set spec metadata ID',
+              { clientId: '', clientSessionId: '' }
```
niclim, line 338: Is there an (easy way) to add a contribution without creating a batch commit? Ideally this is transparent and should not show up in history? Or would this be an antipattern (especially since we are using batch commits as _any_ change to a spec)?

workspaces/ui-v2/src/store/MetadataProvider.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+type SpectacleMetadataResponse = {
+  metadata: {
+    id: string;
+  };
+};
+
+const MetadataContext = createContext<MetadataContextValue | null>(null);
+
+export const SpecMetadataProvider: FC = ({ children }) => {
```
niclim, line 23: This is what we'd need to do to fix the many async calls we make + stop the `React memory leak with setState` - but I think this is what I envisage our FE state management / redux store to take the place of. The idea is we can fairly easily switch this store out.

There's also a couple of different options we have with handling loading state. In this case, we always fetch specMetadata so that the UI always has access to it and does not need to wait for loading. In other cases we could also pass down loading state and individual UI pieces could view loading state separately

workspaces/spectacle/src/index.ts
```diff
@@ -311,6 +312,35 @@ export async function makeSpectacle(opticContext: IOpticContext) {
...
+      metadata: async (parent: any, args: any, context: any, info: any) => {
+        const metadataKey = 'metadata';
+        const specIdKey = 'id';
+        const metadata =
+          context.spectacleContext().contributionsProjection[metadataKey] || {};
+        let specId = metadata[specIdKey];
+
+        if (!specId) {
+          specId = uuidv4();
+          await context
```
niclim, line 324: So this causes spectacle to be reset (which is intended), but this means on:
- Public examples
- Cloud viewer with no specId metadata
- LocalCLI with no specId

The spectacle instance will fetch twice on load (i.e. initial load will set spectacle, and then this will cause spectacle to reload)

This is part of the reason I blocked loading so only the spec metadata load is sent at the start

workspaces/analytics/src/interfaces/NoProperties.ts
```diff
@@ -1 +0,0 @@
-export interface NoProperties {}
```
niclim, line 1: Also unused

workspaces/spectacle/src/index.ts
devdoshi: I think it's fine to put it in a batch commit. but we can definitely expose a way that doesn't. We can put a link to docs on the metadata id in the commit message so it's not confusing / there's an external reference for what it is and why it exists. if we change contributions, I believe these will stay, and then we'd have ExternalContributionAdded(contentUrl or contentPath, contentChecksum) or something like that

devdoshi (approved):

workspaces/spectacle/src/index.ts
niclim: I think I'm going to change this to something like "initialize specification attributes" - on existing specs, this could be confusing. But for new specifications this would make sense

PR #820, opened 2021-05-25
remove extra cargo build

devdoshi:
## Why
The `cargo build` was already being run slightly differently above, so when it runs again here on mac it doesn't complete instantly, resulting in a timeout

## Validation
* [x] regression passes


devdoshi:
since it's triggered manually: https://github.com/opticdev/optic/actions/runs/875255108

notnmeyer (approved):

PR #822, opened 2021-05-25
spectacle should rebuild projections when the spec changes

devdoshi:
## Why
The .optic/api/specification.json file may change for various reasons, mainly when a user changes their git state. 

## What
Now when the .optic/api/specification.json file changes, spectacle will reload its projections. You can observe the console.count behavior in your console. You may also directly invalidate the projections through the graphql mutation

## Validation
* [ ] CI passes

### DIFF
--- a/workspaces/cli-server/src/spectacle/index.ts
+++ b/workspaces/cli-server/src/spectacle/index.ts
@@ -34,6 +34,7 @@ import { InteractionDiffWorkerRust } from '@useoptic/cli-shared/build/diffs/inte
 import { IPathMapping, readApiConfig } from '@useoptic/cli-config';
 import { CapturesHelpers } from '../routers/spec-router';
 import { IgnoreFileHelper } from '@useoptic/cli-config/build/helpers/ignore-file-interface';
+import fs from 'fs-extra';
 
 ////////////////////////////////////////////////////////////////////////////////
 export interface LocalCliSpecState {}
@@ -50,8 +51,22 @@ export class LocalCliSpecRepository implements IOpticSpecReadWriteRepository {
 
   constructor(private dependencies: LocalCliSpecRepositoryDependencies) {
     this.notifications = dependencies.notifications;
+    this.initialize();
   }
 
+  initialize() {
+    fs.watch(
+      this.dependencies.specStorePath,
+      {
+        encoding: 'utf8',
+        persistent: false,
+        recursive: false,
+      },
+      (event, filename) => {
+        this.notifications.emit('change');
+      }
+    );
+  }
   async applyCommands(
     commands: any[],
     batchCommitId: string,
--- a/workspaces/spectacle/src/graphql/schema.ts
+++ b/workspaces/spectacle/src/graphql/schema.ts
@@ -7,6 +7,10 @@ schema {
 type Mutation {
   applyCommands(commands: [JSON], batchCommitId: ID, commitMessage: String, clientId: ID, clientSessionId: ID): AppliedCommandsResult
   startDiff(diffId: ID, captureId: ID): StartDiffResult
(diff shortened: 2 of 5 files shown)


workspaces/spectacle/src/graphql/schema.ts
```diff
@@ -7,6 +7,10 @@ schema {
 type Mutation {
   applyCommands(commands: [JSON], batchCommitId: ID, commitMessage: String, clientId: ID, clientSessionId: ID): AppliedCommandsResult
   startDiff(diffId: ID, captureId: ID): StartDiffResult
+  invalidateCaches: InvalidateCachesResult
+}
+type InvalidateCachesResult {
+  batchCommitId: ID
```
niclim, line 13: `invalidateCaches` seems to return nothing - should this be removed?

niclim (approved):

acunniffe (approved):
Tested, it works!

workspaces/spectacle/src/index.ts
```diff
@@ -216,6 +216,11 @@ export async function makeSpectacle(opticContext: IOpticContext) {
     return projections;
   }
 
+  opticContext.specRepository.notifications.on('change', () => {
+    console.count('reloading because of specRepository change');
+    reload(opticContext);
```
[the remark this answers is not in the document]
JaapRood: This is a short coming of using EventEmitters for this, rather than a Stream: they don't allow for async handling. That, in turn, introduces a race here: multiple changes can cause multiple reloads in-flight, and a race over which one finishes. It's small, but once it's hit, it could make a real head scratcher of a bug (especially as mutations _are_ reading in the current spec).

workspaces/spectacle/src/index.ts
```diff
@@ -265,6 +270,9 @@ export async function makeSpectacle(opticContext: IOpticContext) {
           notificationsUrl: '',
         };
       },
+      invalidateCaches: async (parent: any, args: any, context: any) => {
+        await reload(context.spectacleContext().opticContext);
+      },
     },
```
JaapRood, line 276: This isn't getting called anywhere and leaks implementation details to its consumers. I'd propose we don't expose it  unless we find a really good reason for it.

workspaces/cli-server/src/spectacle/index.ts
```diff
@@ -50,8 +51,22 @@ export class LocalCliSpecRepository implements IOpticSpecReadWriteRepository {
 
   constructor(private dependencies: LocalCliSpecRepositoryDependencies) {
     this.notifications = dependencies.notifications;
+    this.initialize();
   }
 
+  initialize() {
+    fs.watch(
+      this.dependencies.specStorePath,
```
JaapRood, line 59: `fs.watch` (proxied by `fs-extra` directly from `fs`) [is pretty unreliable](https://github.com/paulmillr/chokidar#why), especially on macOS as ti doesn't use the OS's native FSEvents API :

> Node.js fs.watch:
> 
> - Doesn't report filenames on MacOS.
> - Doesn't report events at all when using editors like Sublime on MacOS.
> - Often reports events twice.
> - Emits most changes as rename.
> - Does not provide an easy way to recursively watch file trees.
> - Does not support recursive watching on Linux.
> 
> Node.js fs.watchFile:
> 
> - Almost as bad at event handling.
> - Also does not provide any recursive watching.
> - Results in high CPU utilization.

Fortunately, there's modules like [chokidar](https://www.npmjs.com/package/chokidar) or [fsevents](https://www.npmjs.com/package/fsevents) that do this properly.

JaapRood:
Happy this landed so Optic v10 could ship, I'll look at some of the identified issues separately!

PR #823, opened 2021-05-25
Handle redirects when visiting invalid URLs

niclim:
## Why

Fixes:
- Visiting an invalid url -> redirects to a base url (`/documentation` if not on a valid entry point)
  - e.g. visiting `/examples/todos-partial/documentation/123123123` redirects you back to `/examples/todos-partial/documentation` and other likewise cases
  - e.g. visiting a page on the diff page and refreshing on it will keep you on the same page (unfortunately you lose progress, and also you cannot refresh on the pending endpoint page... I'm not entirely sure why (think it's due to learning / state required to set up))

## What

- Removed `NavigationRoute` and replaced it's usage with a PageLayout component - some weird things going on with this, 
  - Caused excess rerenders / remounts due to using `React.createElement` over fns
  - Some really weird behavior that `exact` wasn't matching - I am not sure why but I think it's because `Routes` rendered in children components _might_ be treated differently? Basically some routes weren't resolving until I pulled it out of the NavigationRoute but I'm not sure why and haven't dug into it

If this is too large / too much to verify, this PR can probably wait until after we cut Optic 10

## Validation
Best way to test this is:
- run through each instance (cloud viewer, local cli, public examples) and go through page navigation
- places to navigate
  - endpoints list -> endpoint page
  - changelog endpoints list -> changelog endpoint page
  - diff flow (and submit)


workspaces/ui-v2/src/optic-components/pages/docs/DocumentationRootPage.tsx
```diff
@@ -87,7 +96,9 @@ export function DocumentationRootPage(props: {
                     disableGutters
                     style={{ display: 'flex' }}
                     onClick={() =>
-                      props.onEndpointClicked(endpoint.pathId, endpoint.method)
+                      history.push(
+                        `${match.url}/paths/${endpoint.pathId}/methods/${endpoint.method}`
```
niclim, line 100: This handles both the documentation + changelog pages since it uses the current `match.url` - which is supplied from either the changelog page or the documentation page

workspaces/ui-v2/src/spectacle-implementations/cloud-viewer.tsx
```diff
@@ -122,11 +128,19 @@ export default function CloudViewer() {
               >
                 <SpecMetadataProvider>
                   <Switch>
-                    <>
-                      <DiffReviewEnvironments />
-                      <DocumentationPages />
-                      <ChangelogPages />
-                    </>
+                    <Route
+                      path={`${match.path}/changes-since/:batchId`}
```
niclim, line 132: We can't use the `navigation/Routes` helpers since it relies on baseUrl - which is only rendered in this component. We could move this out into a separate component. 

Side note, I'm generally not a fan of the current route helpers - it provides a centralized way to specify routes but there's a fair amount of indirection we need to go through (hook -> function call / path param) - but we still have somewhat scattered route logic (i.e. useParams needs to know what context it's rendered in, and what the named path params are)

workspaces/ui-v2/src/optic-components/pages/diffs/CapturePage.tsx
```diff
@@ -24,19 +28,29 @@ import {
...
   const capturesState = useCaptures();
   const history = useHistory();
   const diffEnvironmentsRoot = useDiffEnvironmentsRoot();
 
-  const noCaptures =
-    !capturesState.loading && capturesState.captures.length === 0;
+  const noCaptures = capturesState.captures.length === 0;
+  // The only current environment supported is local
+  const isValidEnvironment = environment === 'local';
+  // TODO this currently does not get reached if we have an invalid boundary,
```
niclim, line 46: https://github.com/opticdev/optic/blob/04bcab1e719b100e79ce36e06509ee6664b6fdef/workspaces/ui-v2/src/optic-components/pages/diffs/ReviewDiffPages.tsx#L33 this part errors if in the url we specify an incorrect / invalid diff.

Only possible if the user enters the url manually - I left this for now, and I think this can kind of fit into better error handling

workspaces/ui-v2/src/optic-components/pages/diffs/CapturePage.tsx
```diff
@@ -24,19 +28,29 @@ import {
...
+  }>();
   const capturesState = useCaptures();
   const history = useHistory();
   const diffEnvironmentsRoot = useDiffEnvironmentsRoot();
 
-  const noCaptures =
-    !capturesState.loading && capturesState.captures.length === 0;
+  const noCaptures = capturesState.captures.length === 0;
+  // The only current environment supported is local
+  const isValidEnvironment = environment === 'local';
```
niclim, line 55: I'm not sure if this is correct - but this is what I've found everywhere in the UI at least? And I can't see where else this is set

acunniffe: yeah, it'll change when we move the other envs into the OSS project, but this is correct today

workspaces/ui-v2/src/optic-components/pages/diffs/CapturePage.tsx
acunniffe: agreed on the approach here.

acunniffe (approved):

PR #826, opened 2021-05-26
Eliminate potential racing and performance issues in reloading spec projections from spec changes

JaapRood:
## Why
In #822 a first version of a mechanism got introduced that lets Spectacle respond to the spec changes on disk, due to things like Git operations, manual edits, etc. While it unblocked us for Optic v10, it included a couple of data races and a relied on the not to be trusted `fs.watch`. This had the potential for some really funky bugs (as data races always tend to produce), for the system to get overwhelmed in the case of many concurrent file changes, high CPU usage during watching, etc.

## What
This re-implements the mechanism for detecting changes to the Spec from the file system by using an `AsyncGenerator` instead of event emitters, as well as `chokidar` instead of straight up `fs.watch`. 

By using an `AsyncGenerator` the reload mechanism can exert back-pressure and prevent being overwhelmed. It also makes sure there is strict control over the consumption of these change events, allowing strict control over timing, concurrency, etc.  With the use of `chokidar` we don't only get more performant and reliable file watching (especially cross platform), but also an initial layer of debouncing of change events (which can come in quick successions in certain scenarios, like Git rebases for example). Projections are not reloaded until the spec has been stable for 2 seconds, a reasonable amount of latency in an eventually consistent system.

## Validation
* [ ] CI passes
* [x] Manual changes to `specification.json` show up in the Docs UI without daemon restart.
* [x] Projections aren't rebuilt constantly during rapid sequences of changes like a Git rebase, but when stable again.

### DIFF
--- a/workspaces/spectacle/src/index.ts
+++ b/workspaces/spectacle/src/index.ts
@@ -74,6 +74,8 @@ export interface IOpticSpecReadWriteRepository extends IOpticSpecRepository {
     commandContext: IOpticCommandContext
   ): Promise<void>;
 
+  changes: AsyncGenerator<void>;
+
   notifications: EventEmitter;
 }
 
@@ -207,6 +209,7 @@ export async function makeSpectacle(opticContext: IOpticContext) {
     shapeViewerProjection: any,
     contributionsProjection: ContributionsProjection;
 
+  // TODO: consider debouncing reloads (head and tail?)
   async function reload(opticContext: IOpticContext) {
     const projections = await buildProjections(opticContext);
     endpointsQueries = projections.endpointsQueries;
@@ -216,12 +219,23 @@ export async function makeSpectacle(opticContext: IOpticContext) {
     return projections;
   }
 
-  opticContext.specRepository.notifications.on('change', () => {
-    console.count('reloading because of specRepository change');
-    reload(opticContext);
-  });
+  async function reloadFromSpecChange(
+    specChanges: AsyncGenerator<void>,
+    opticContext: IOpticContext
+  ) {
+    await reload(opticContext);
 
-  await reload(opticContext);
+    for await (let change of specChanges) {
+      console.count('reloading because of specRepository change');
+      await reload(opticContext);
+    }
+  }
+
+  // TODO: make sure this Promise is consumed somewhere so errors are handled. Return from makeSpectacle perhaps?
+  const reloadingSpecs = reloadFromSpecChange(
+    opticContext.specRepository.changes,
+    opticContext
+  );
 
   const resolvers = {
     JSON: GraphQLJSON,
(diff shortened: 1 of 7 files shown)


workspaces/spectacle/src/index.ts
```diff
@@ -207,6 +209,7 @@ export async function makeSpectacle(opticContext: IOpticContext) {
     shapeViewerProjection: any,
     contributionsProjection: ContributionsProjection;
 
+  // TODO: consider debouncing reloads (head and tail?)
```
JaapRood, line 212: Chokidar already comes with a 2 second debounce to deal with fast sequential operations like Git rebases, atomic writes from editors, etc. That should be enough as a starting point.

workspaces/spectacle/src/index.ts
```diff
@@ -216,13 +219,24 @@ export async function makeSpectacle(opticContext: IOpticContext) {
...
+  ) {
+    for await (let generation of specChanges) {
+      console.log('reloading because of specRepository change', generation);
+      await reload(opticContext);
+    }
+  }
 
   await reload(opticContext);
 
+  // TODO: make sure this Promise is consumed somewhere so errors are handled. Return from makeSpectacle perhaps?
```
JaapRood, line 234: @devdoshi right now, in browsers and modern versions of node, errors here should cause a `UnhandledPromiseRejectionWarning`, so it _shouldn't_ get swallowed silently. Good enough for now, or should we bubble this up somewhere? I can imagine combining all async operations in a `Promise.all` and returning in a `result` promise.

JaapRood:
@niclim @devdoshi the snapshot tests definitely still seem indeterministic in there results: definitely didn't change anything here, yet new ids are being written. Jest runs these tests concurrently as far as I'm aware, maybe the id generation isn't isolated properly somewhere?

devdoshi:
I didn't get a chance to look at this in depth today but I see 
```
console.error
      command could not be applied: Validation("Command failed validation: shape id must be assignable to add shape, \"AddShape(AddShape { shape_id: \\\"shape_10000\\\", base_shape_id: \\\"$boolean\\\", name: \\\"\\\" })\"")
```
in the output which suggests things are competing on the same spec instance, such that an id generated from `let mut sequential_id_generator = SequentialIdGenerator { next_id: 9999 };` has already been created

devdoshi (approved):
looks good aside from the tests failing

JaapRood:
> I didn't get a chance to look at this in depth today but I see
> 
> ```
> console.error
>       command could not be applied: Validation("Command failed validation: shape id must be assignable to add shape, \"AddShape(AddShape { shape_id: \\\"shape_10000\\\", base_shape_id: \\\"$boolean\\\", name: \\\"\\\" })\"")
> ```
> 
> in the output which suggests things are competing on the same spec instance, such that an id generated from `let mut sequential_id_generator = SequentialIdGenerator { next_id: 9999 };` has already been created

That doesn't seem to be a failing test, so perhaps intended?

PR #827, opened 2021-05-26
Rename Rust packages to better describe their intended purpose

JaapRood:
## Why
When we started on the Rust codebase, we were purely focused on the Diff Engine, making the name `diff-engine` appropriate. However, we're now doing a lot more than just diffing, covering basically all the most core Optic business and domain logic, and will continue to expand its capabilities. More appropriate naming should clarify what's cover and it's intended use.

There's also still some code in the other optic-core repository, which is the home of the Scala implementation that we're trying to let go. We should bring that over into the main repo as well.

## What

We're renaming the diff-engine packages, extracting some additional logic elsewhere and pulling in some other packages from elsewhere.

Workspaces / packages: 
- `diff-engine` -> `optic-engine`
- `diff-engine/cli` + `diff-engine/lib` -> `optic-engine-native`
- `diff-engine-wasm` -> `optic-engine-wasm`
- `diff-engine-wasm/streams` + `domain-types` -> `optic-domain`

~~Published binaries:~~
- ~~`optic_diff` -> `optic_engine`~~

Since every package receives a new name, only newly published versions of these packages will be affected, leaving the currently published packages untouched and available for use. 

In addition, a new [Cargo feature](https://doc.rust-lang.org/cargo/reference/features.html) `avro` feature has been added to `optic-engine`, disabling the non-functional Avro deserialising of `HttpInteraction`s by default. This reduces the amount of dependences that have to build with about 30, speeding it up considerably.

## Validation
* [x] CI passes
* [x] Builds locally
* [x] Tests pass locally
* [x] Release flow dry run passes


workspaces/optic-streams/tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "extends": "../config/tsconfig.json",
   "compilerOptions": {
-    "outDir": "lib",
+    "outDir": "build",
```
devdoshi, line 4: yay

devdoshi:
looks good so far :)

workspaces/optic-streams/tsconfig.json
JaapRood: "lib" used to be where node packages had their JS code (generated or not), but TS seems to have muddled the water, so might as well be consistent with the rest of our workspaces 😅

JaapRood:
Things are definitely getting a bit hairier now, with the Files tab on this PR not really that functional anymore. Biggest part of that is the test snapshots for the Rust code, all needing to be changed because of the changing crate names, some also requiring file renames.

I'm thinking of changing the binary name from `optic_diff` to `optic_engine` in a separate PR, as there's moving pieces outside of this Github repo the are required for that. Reviewing all the changes to make that work without our browsers crashing would be great.

There's a couple of things left over to do here:

- extract `optic-engine` from `optic-engine-native`, leaving the CLI target and Node integration with that as `optic-engine-native`, moving out all the common Rust logic. Hopefully that makes it clear that `optic-engine` is the core code and `optic-engine-native` and `optic-engine-wasm` specific interfaces to the core.
- verify the Github workflows still work, with the task files having been moved and renamed in the process of all this.

devdoshi:
Looks good to me and I agree with the strategy for the remaining changes

JaapRood:
I extracted the last bits and got the documentation in a semi decent state, explaining briefly the purpose of each package and how it's to be used.

That leaves dry running the release building and publishing to verify we haven't broken anything there.

JaapRood:
With lots of final fixes and tweaks, I think this is ready for review and merging (@devdoshi). 

There is one significant additional improvement I included: making the `avro-rs` dependency of `optic-engine` optional,  hiding it behind the new `avro` feature and disabling that new feature by default. The Avro deserialising of interactions is non-functional, yet `avro-rs` requires the building of about 30 additional dependences, reducing the normal build from ~140 to ~110. It's a significant increase of build times!

devdoshi:
@JaapRood I believe the types have been moved into cli-shared optic-types.ts but that might still not be the right place

JaapRood:
I ended up taking domain-types and `cli-shared/optic-types` and merging it with `optic-streams`, renaming the whole thing `optic-domain`. I added a README to explain that `optic-domain` is the TS / JS layer to work with the inputs and outputs of the Optic Engine, with the latter remaining the place the actual business logic lives. Hopefully that's a clean framing that's not too confusing, while allowing use to keep the amount of packages under control.

JaapRood:
@devdoshi if crashing browser windows don't keep you from it, would love a review of this before I merge. From there I'll look at some other consumers of the legacy packages and leave them notes on the changes / how to upgrade!

devdoshi (approved):
reviewed in sublime merge, looks good!

PR #829, opened 2021-05-26
Consolidate analytics helpers into typescript functions

niclim:
## Why

Previously we were using two types of analytics helpers (UI which used a class + generic objects, and cli events which used a lib helper). Given we are using typescript now, we have runtime type checks are generally less useful for analytics - as static typing will help us catch improperly typed events earlier in the development process.

As a result, I've created basic typescript functions to generate events for sending (which match `TrackingEventBase`). I also consolidated UIOpticEvents to use TrackingEventBase so we don't need to juggle two types for analytic events.

This should help with:
- Being easier / lighter weight to add new events
- Provide better type safety (before we weren't using the correct typings + overriding it with `any`)

## What
- Removed context from TrackingEventBase (context is automatically passed in on CLI event sends) - The UI doesn't have the same context, which means some UI events will have different context.
- We have a somewhat implicit convention that UI events are snake_case, and CLI events are PascalCase
- Changed CLI events to use typescript fn creates
- Use TrackingEventBase instead of OpticUIEvent type
- Lots of deleted code (removed unnecessary dependencies

## Validation
Typescript should be able to validate _most_ of these changes


workspaces/analytics/src/cliEvents.ts
```diff
@@ -0,0 +1,103 @@
+import { CheckAssertionsResult } from './interfaces/ApiCheck';
+import { TrackingEventBase } from './interfaces/TrackingEventBase';
+
+export enum EventsEnum {
```
niclim, line 4: I only ported over the events that were currently being used in code

workspaces/analytics/tests/__snapshots__/EventBus.test.ts.snap
```diff
@@ -3,18 +3,21 @@
...
-    "arch": "mips",
     "release": "1998",
-    "apiName": "testing",
-    "clientSessionInstanceId": "testing",
-    "clientTimestamp": "testing",
+    "source": "user",
   },
-  "data": Object {
-    "apiName": "Hello World API",
+  "event": Object {
```
niclim, line 16: This is the main change in the snapshot, event and context are separated (since TrackingEventBase no longer contains context)

workspaces/analytics/src/eventbus.ts
```diff
@@ -2,32 +2,36 @@ import { EventEmitter } from 'events';
...
       const context = await getContext(batchId);
-      // TODO - don't overwrite context
       events.forEach((event) => {
-        eventEmitter.emit('event', { ...event, context });
+        eventEmitter.emit('event', { event, context });
       });
     },
     eventEmitter,
-    listen: (callback: (e: TrackingEventBase<any>) => void): void => {
+    listen: (callback: ListenCallbackType): void => {
```
niclim, line 22: This is probably the part that requires the most attention, since I don't think typescript can infer through events in the event emitter (it just trusts the events we give it)

devdoshi (approved):
nice cleanup! fyi the original goal with the runtime type checking/sanitization/validation was for ingesting the events on the backend

PR #836, opened 2021-05-28
Enable workspace tests

devdoshi:
## Why
We should be running all of our workspace tests now

## What
Failing tests will fail the build now

## Validation
* [ ] CI passes


Taskfile.yml
```diff
@@ -69,12 +69,19 @@ tasks:
...
+    cmds:
+      - yarn wsrun --stages --report --fast-exit --exclude-missing ws:test
+
   workspaces:build:ci:
     desc: CI workflow
     cmds:
       - task: workspaces:setup
       - task: workspaces:clean
       - task: workspaces:build
+      - cargo build
```
devdoshi, line 83: @JaapRood seems like this is required for the cli-shared tests. I'm not clear on where we're supposed to be running cargo build for what. Can you clarify? Anyway, I think we can get your diff-engine reorg merged and then apply most of the changes from here on top

JaapRood: Not really sure, especially as `workspaces:build` should include the building of any native binaries 🤔. `cli-shared` does test the native <> Node.js interface, so there having to be a build seems right. But why the separate commands is necessary needs further investigation. 

As you said, probably better based on top of the re-organisation, as there's some chance this has solved itself already.

devdoshi:
closing in favor of what I accidentally pushed to develop :/

PR #837, opened 2021-05-30
demo should build and publish to docker registry

devdoshi:
## Why
We want to bring back the demo UI build so we can show Optic 10's UI without people needing to install Optic

## What
Previously the Optic 9 demo UI build had extra content and behaviors. For now, the new demo build only has the in-memory spec viewer linked to the public example-sessions

## Validation
* [ ] CI passes
* [ ] Verified in staging


workspaces/ui-v2/Dockerfile
```diff
@@ -1,5 +1,5 @@
 FROM library/nginx:alpine
-ARG APP
-COPY build-${APP}/ /usr/share/nginx/html/
```
devdoshi, line 3: based on the inconsistent naming of our repos, I thought it was better to remove that assumption

workspaces/ui-v2/Taskfile.yml
```diff
@@ -47,19 +48,13 @@ tasks:
       - >
         BUILD_PATH=build-cloud-spec-viewer
         yarn build:cloud
-    sources:
```
devdoshi, line 50: I wasn't sure what the utility or correctness of these were in this context

workspaces/ui-v2/src/entry-points/demo/demo-examples.tsx
```diff
@@ -0,0 +1,207 @@
...
+    analytics: {
+      enabled: false,
+    },
+    documentation: {
+      allowDescriptionEditing: true,
+    },
+  },
+};
+
+export default function DemoExamples(props: { lookupDir: string }) {
```
devdoshi, line 53: Aside from the name, this is identical to PublicExamples

devdoshi:
manually triggering here: https://github.com/opticdev/optic/actions/runs/891296683

not sure how to test on staging

devdoshi:
@acunniffe can set up the path changes for `demos/github` and `demos/github-with-diffs` and add in the examples either in this PR or a follow-up before we actually publish it

notnmeyer:
> not sure how to test on staging

look at the original projects/ui stuff in monorail for reference,
https://github.com/opticdev/monorail/blob/master/projects/ui/Taskfile.yml#L4-L12

but basically, to deploy for now, edit the "demo" deployment and update the image tag with the new build. 

i can clean up the deployment bits when im back if theyre not sorted. deployments were never done from this repo, since i didnt want it talking directly to k8s.

workspaces/ui-v2/Taskfile.yml
jshearer: @notnmeyer does this break things? I have a very vague memory of you telling me that this `sources`/`generates` fields were important, but I can't quite remember why.

jshearer: Ah you're on vacation. No worries then, it can't break too many things, and if it does then we'll fix 'em.

jshearer (approved):
The substance of the changes look good to me, and @devdoshi convinced me that the taskfile stuff is fine. 👍

PR #838, opened 2021-05-31
Fix links to `Monitor Live APIs with Optic` link

taras:
## Why
I'm working on [OpenAPI spec for Backstrage's Catalog plugin](https://github.com/backstage/backstage/issues/2566). I wanted to find a way to validate the spec that I created against the real API. In the process, I found Optic which I'm going to try today. While reading the docs, I realized that the link to "Monitor Live APIs with Optic" was broken.

## What

Since "Monitor Live APIs with Optic" page exists at the specified path, I'm assuming that `http://locahost:4000` was a mistake. I removed the host so the link will work.

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


acunniffe:
Thanks for fixing this @taras 👍  -- I'll go through the docs today and click through everything. 

Separately, I like backstage / know a lot of users so I'm happy to help out with this today, feel free to book some of my time https://calendly.com/optic-onboarding/aidan-chat

Lmk if there's anything else we can do to help

PR #839, opened 2021-05-31
CI should require FLAGS_FILE to be set

devdoshi:
## Why
Releases have been made without FLAGS_FILE set, resulting in misconfigured releases

## What
CI will fail when the FLAGS_FILE environment variable is not set

## Validation
* [ ] CI passes

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -42,14 +42,14 @@ tasks:
       - package.json
       - yarn.lock
 
-  workspaces:build:
-    desc: Build Yarn workspaces
+  workspaces:setenv:
+    desc: Ensure workspace projects have .env files
     vars:
       FLAGS_FILE: '{{.FLAGS_FILE}}'
-    deps: ["workspaces:setup"]
     cmds:
       - |
-        if [[ -n "$FLAGS_FILE" ]]; then
+        if [[ -n "$FLAGS_FILE" ]]
+        then
           echo "Feature flags found ($FLAGS_FILE), copying to packages for build and runtime."
           cp $FLAGS_FILE .env
           cp $FLAGS_FILE workspaces/local-cli/.env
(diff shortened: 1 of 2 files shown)


Taskfile.yml
```diff
@@ -42,38 +42,51 @@ tasks:
...
+          echo "Please specify the FLAGS_FILE environment variable"
+          exit 1
         fi
-        yarn wsrun --stages --report --fast-exit --exclude-missing ws:build
+
+  workspaces:build:
+    desc: Build Yarn workspaces
+    deps: ["workspaces:setup"]
+    cmds:
+      - yarn wsrun --stages --report --fast-exit --exclude-missing ws:build
```
notnmeyer, line 70: rogue newline

notnmeyer (approved):
👍

Taskfile.yml
```diff
@@ -42,38 +42,51 @@ tasks:
...
-        if [[ -n "$FLAGS_FILE" ]]; then
+        if [[ -n "$FLAGS_FILE" ]]
+        then
           echo "Feature flags found ($FLAGS_FILE), copying to packages for build and runtime."
           cp $FLAGS_FILE .env
           cp $FLAGS_FILE workspaces/local-cli/.env
           cp $FLAGS_FILE workspaces/cli-server/.env
           cp $FLAGS_FILE workspaces/cli-scripts/.env
           cp $FLAGS_FILE workspaces/diff-engine/.env
           cp $FLAGS_FILE workspaces/ui-v2/.env.production.local
```
notnmeyer, line 59: i know this isnt part of the PR directly, this should be using something like `$ENVIRONMENT` or the file should be renamed to not include the env

devdoshi: hmm there are other considerations like .gitignore etc.; not sure what your intent is

notnmeyer: i meant the hardcoded "production" in ".env.production.local"

devdoshi: Oh  I see - https://create-react-app.dev/docs/adding-custom-environment-variables/#what-other-env-files-can-be-used it needs to be production for production builds (from the React build's perspective)

notnmeyer: ohhhhhh 👍

PR #846, opened 2021-06-01
Fix safari path parameter editing

niclim:
## Why

On safari, the named path parameter we could only edit one character at a time (before needing to unblur and reselect). Fixing this uncovered another error where the delete icon in safari could not be clicked.

## What

- The issue was we had `userSelect: none` on the input component. The reason we could focus on it is because we programmatically focused on the input (using the `autoFocus` element)
- Removed the edit state (which is similar to isPathParameter). This is because `e.relatedTarget` for blur events is the target that _receives_ focus, which is different in safari. There's not a great way to sequence the onBlur to be after the onClick event. Instead, I removed the edit state such that we dont need the blur listener. This means the `x` shows up on all edited name fields (which I think might be better anyway, since it's easier to undo changes if you have named a bunch of parameters)

Screen Shot 2021-06-01 at 10 36 20 AM

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/ui-v2/src/pages/diffs/AddEndpointsPage/components/UndocumentedUrl.tsx
+++ b/workspaces/ui-v2/src/pages/diffs/AddEndpointsPage/components/UndocumentedUrl.tsx
@@ -189,44 +189,22 @@ function PathComponentRender({
   const classes = useStyles();
   const [name, setName] = useState(pathComponent.name);
 
-  const [isEditing, setIsEditing] = useState(false);
-
   const onStartEdit = () => {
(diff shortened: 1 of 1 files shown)


devdoshi (approved):
nice improvements across the board! aside: i think it would be cool if the name defaulted to singularize(the previous component) + 'Id' or something like that but that's a can of worms

niclim:
> nice improvements across the board! aside: i think it would be cool if the name defaulted to singularize(the previous component) + 'Id' or something like that but that's a can of worms

Ohh that's a great idea - I'll add it to my todo list, I dont think it'd be too hard to do

PR #847, opened 2021-06-01
Split changelog page from docs page

niclim:
## Why

In the future, we will be building out / iterating on the editing flows. Since editing flows are only going to be in the documentation view, I've split out these two pages to reduce the overhead (i.e. only docs will have editing functionality). There's more components to move into shared files / better abstractions - but this PR aims to remove the `ContributionEditingStore` in the changelog page (which might be removed with a more generic "editing store" for our future iterations). 

## What

- Renamed the `shapes` component to ShapeRenderer + only exported functions / components that were publicly used (so we can more easily mess / update internals in the ShapeRenderer component (it's a fairly complicated component that will likely see more and more edits with future iterations)
- Copy + pasted code from docs -> changelog and then specialized each component (changelog page does not need edit flow, docs page does not need changelog information)
- There are definitely better places for a lot of these components (in `src/pages/docs/components` and `src/pages/changelog/components`) to live or better abstractions. But I think these will become more obvious as we work on them.

## Validation
the changelog and documentation pages still work and show the appropriate changes + are editable


JaapRood (approved):
This looks like a useful clean up that separates some concerns in components we'll be working on soon, making the water less muddled there. I'm totally on board with not instantly trying to create the most generic possible components out of this, as `no abstraction > the wrong abstraction`. Letting time prove to us whether's there's something generic to extract and whether that's even worth extracting, is the right call made!

Just a small side note: I think we're missing a couple of design and iteration steps to say with certainty that editing will only happen on the documentation view. That has no further bearing on the changes made here as far as I can see, but good to note that none of that is set in stone.

niclim:
> Just a small side note: I think we're missing a couple of design and iteration steps to say with certainty that editing will only happen on the documentation view. That has no further bearing on the changes made here as far as I can see, but good to note that none of that is set in stone.

You are probably right here - there are instances we we _might_ want to add in contributions to changelog / older versions (or to attached changelogs to older versions). That being said, I think it's less likely we want to delete / manually modify fields (perhaps if you wanted to edit from old version... but that's weird unless we wanted forked specificiations). Either way, splitting these out means we can diverge the behaviors from these pages

Issue #849, opened 2021-06-02
Asterisk in the path of an endpoint


nikita2206:
An asterisk in the path of an endpoint currently breaks Optic's UI when, after it's seen me navigating to that endpoint I try to add it in the diff tool. 
The full path in my case looks like `/v1/foo/**`.

LouManglass:
Oof! I'll give that a closer look and have an update for you in a day. If you have any other details to add, please let me know and I'll be sure to include them in my testing.

niclim:
Thanks for raising this @nikita2206 - I've got a fix up here https://github.com/opticdev/optic/pull/850 where I also identified a couple of other issues that should also be fixed with that PR! We'll follow up with you when there's a new version that has this problem fixed!

nikita2206:
Thanks for such a quick response!

LouManglass:
Thanks for checking in on the PR as well, and the added notes. We'll keep you posted as we work through a proper fix and get a new version tested and released.

niclim:
Hi @nikita2206 - we've just released optic version `10.0.5` which should fix this problem (you should be able to have `*` in endpoints without crashing)

I did want to follow up on this comment you left here https://github.com/opticdev/optic/pull/850#issuecomment-853993249
> I wonder if this also tackles the issue with having complex regexps in the ignore rules? For example I was trying to use a negative lookahead pattern which is something like (?!do not match) and I was getting an error saying that my pattern is invalid.

Can you give some examples or use-cases where you'd want to have negative lookups (or other regexp patterns that the ignore rules currently don't support)? This would be helpful to understand where there's gaps and how we could address them. Thanks!

nikita2206:
Hey @niclim thanks! We have a haproxy in front of all web services, so that they all are accessible by the same domain but with different path prefixes. I just wanted to exclude one of the services by doing a negative match on its path prefix. I must say the use case is weird and I'm absolutely fine without being able to do that (I simply deploy Optic for the specific service only, not for everything that's behind haproxy). On the other hand it does look like the library that Optic uses for matching paths support using arbitrary regexps so maybe Optic could relatively easily add that support.

niclim:
[...]

PR #850, opened 2021-06-03
Handle special characters in api names

niclim:
## Why

As raised in https://github.com/opticdev/optic/issues/849 - the UI crashes when there are special characters. After looking into this I discovered a couple of other issues:
- Special characters (specifically `?`, `+`, `*`, `[`, and `{`) broke the usage of the pathToRegexp library
- Named path parameters with a `-` in the key or any special character (i.e. [any character not part of this](https://github.com/pillarjs/path-to-regexp/blob/77df63869075cfa5feda1988642080162c584427/src/index.ts#L55-L64)) would cause our match to be incomplete (see test cases for examples)
  - e.g. `/api/:show-me-everything/feed` would create a regex that looked something like `/api/(.*)-me-everything/feed` instead of `/api/(.*)/feed`

In general, I think we want to be more loose with matching / allowing names in apis (since we are capturing traffic, we can't dictate what sort of name an api _should_ be). I'm not sure if I missed any obvious cases here or if there's any reason we _wouldn't_ want to treat special characters (and other cases that are detailed in the [test cases](https://github.com/opticdev/optic/pull/850/files#diff-832bc26c2d2685a7f791886bdb5336fc60ee3f1acf12840edc5a0a659c1ad143R1)) - but happy to revise if I've missed anything

## What

- Built a custom function that wraps this library - since we want to be more loose with acceptable user input / paths.
- Note that there's another usage of pathToRegexp that I didn't use the wrapped function - since that is considered user input (rather than traffic identified endpoints)

## Validation

I'm able to learn the endpoints with special characters without crashing the page
Screen Shot 2021-06-02 at 6 36 16 PM
Screen Shot 2021-06-02 at 6 36 26 PM


workspaces/ui-v2/src/utils/pathToRegexpEscaped.ts
```diff
@@ -0,0 +1,41 @@
...
+  s.replace(specialCharactersRegex, '\\$&');
+
+// The pathToRegexp lib doesn't treat `-` as the same path parameter
+const replaceDashWithUnderscoreInNamedParameter = (s: string) =>
+  s
+    .split('/')
+    .map((part) => (part.startsWith(':') ? part.replace(/-/g, '_') : part))
+    .join('/');
+
+// @gotcha this swallows query parameters - it's more appropriate that this function doesn't use query parameters
```
niclim, line 16: This is a little odd, but the way that I imagine we have to handle query parameters is that it is treated separately from the path

workspaces/ui-v2/src/utils/pathToRegexpEscaped.ts
```diff
@@ -0,0 +1,41 @@
...
+// We want to be looser with our definition of path parameters so we'll remove special characters
+// the only thing the name after : is used for is group matching
+const replaceSpecialCharactersInNamedParameter = (s: string) =>
+  s
+    .split('/')
+
+    .map((part) =>
+      part.startsWith(':')
+        ? part.replace(specialCharactersRegex, '') + 'a' // The regexp lib expects at least 1 character after a :
```
niclim, line 25: Another quirk - `:` expects at least one character - so something like `/:\+` will throw an error. We can get around this by throwing a character at the end of this (since we don't actually use the named parameter from the regex at all this should be fine)

LouManglass:
Some special characters (the [gen-delims in RFC3986](https://datatracker.ietf.org/doc/html/rfc3986)) have special meaning - we probably should be careful around those. Of the characters specifically listed above, this includes `?` and `[` For example, I'm not sure where this process stops, but we'd definitely want `?` available to delimit URL params still when we get around to implementing that. I expect this change won't affect that (we'll build the path, and then it can be properly parsed into url params at the appropriate stage) but wanted to call it out in case I misunderstood the scope of this.

niclim:
> Some special characters (the [gen-delims in RFC3986](https://datatracker.ietf.org/doc/html/rfc3986)) have special meaning - we probably should be careful around those. Of the characters specifically listed above, this includes `?` and `[` For example, I'm not sure where this process stops, but we'd definitely want `?` available to delimit URL params still when we get around to implementing that. I expect this change won't affect that (we'll build the path, and then it can be properly parsed into url params at the appropriate stage) but wanted to call it out in case I misunderstood the scope of this.

This brings up some interesting points - do we want to be _less_ strict if users are _not_ using uri strings. While writing this, I figured the best way to handle this is that this should specifically handle just the path component (query strings, and other parts of the URI should be handled separately)

When we handle query parameters, we could split this out / handle this explicitly (i believe query parameters will show up in the UI and will crash when trying to learn)

nikita2206:
I wonder if this also tackles the issue with having complex regexps in the ignore rules? For example I was trying to use a negative lookahead pattern which is something like `(?!do not match)` and I was getting an error saying that my pattern is invalid.

niclim:
> I wonder if this also tackles the issue with having complex regexps in the ignore rules? For example I was trying to use a negative lookahead pattern which is something like `(?!do not match)` and I was getting an error saying that my pattern is invalid.

This PR does not fix that issue, although I think I know what we'd need to do to fix that. Let me take a look at that and think through some of the implications / some potential fixes (we are currently using the same path-to-regexp lib which means user input regex / some more custom things might not be valid)

LouManglass:
I did a visual check, ran against other existing patterns, this seems to work as expected without having a negative impact on anything obvious.

niclim:
@devdoshi - not sure if this fell through your notifications - but pinging you incase you it did!

devdoshi (approved):

PR #854, opened 2021-06-03
Implement redux + migrate useEndpointsHook

niclim:
## Why

This is our first part moving over to redux. As part of this, I'm planning to show / iterate on how we can handle the following concerns:
- Loading + reloading spectacle, and using different implementations of spectacle
- Handling a simulated command store

After thinking about this a bit, I think it's easiest to keep our current contexts, and use dependency injection as part of the action creators / thunks. This should make it easy to migrate to, and also maintain the load + reloading behavior we already have implemented. Additionally, I've dug into the redux code a little and I don't think we should have a problem by having nested redux stores (in place of the simulated command context) - [this POC PR](https://github.com/opticdev/optic/pull/855) should be all we have to do to handle the simulated command store 

As part of proving out this solution, I've migrated the `useEndpointsHook` (which is a fairly straightforward data fetch and doesn't have complex UI interactions / user inputs to join)

## What

- Implement redux store
- Migrate useEndpointsHook to redux fetching (now we are also not fetching on every hook mount - only at the entry points when explicitly requested)
- Split out changelog information from the useEndpointsHook (since it's only used on one page - the changelog page)

## Validation
Validate against:
- [x] public example
- [ ] demo
- [ ] local cli
- [ ] cloud viewer


workspaces/ui-v2/src/pages/changelog/ChangelogEndpointRootPage.tsx
```diff
@@ -49,20 +50,31 @@ const ChangelogRootComponent: FC<
...
+  const endpointsState = useAppSelector((state) => state.endpoints.data);
+
+  const dispatch = useAppDispatch();
+  const spectacle = useSpectacleContext();
+  useEffect(() => {
+    // the typing is correct, we just arent setting it correctly above
+    // TODO remove ts-ignore
+    // @ts-ignore
+    dispatch(endpointActions.fetchEndpoints({ spectacle }));
+  }, [dispatch, spectacle]);
```
niclim, line 62: This is what interacting with the store would look like - I'm probably going to pull these into separate functions such that we can reuse this easier but the core interactions with redux is here

workspaces/ui-v2/src/types/endpoints.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export interface IEndpoint {
+  pathId: string;
+  method: string;
+  purpose: string;
+  description: string;
+  fullPath: string;
+  pathParameters: IPathParameter[];
+  group: string;
+}
```
niclim, line 17: I pulled out `changelog` from this interface, because it is only used in one place and should be fetched separately

workspaces/ui-v2/src/store/root.ts
```diff
@@ -0,0 +1,16 @@
...
+  reducer: {
+    endpoints: endpointsReducer,
+  },
+});
+
+export type RootState = ReturnType<typeof store.getState>;
+export type AppDispatch = typeof store.dispatch;
+
+export const useAppDispatch = () => useDispatch<AppDispatch>();
+export const useAppSelector: TypedUseSelectorHook<RootState> = useSelector;
```
niclim, line 16: We use these custom dispatch / selectors such that our hooks have the appropriate redux types

workspaces/ui-v2/src/pages/changelog/ChangelogEndpointRootPage.tsx
```diff
@@ -49,20 +50,31 @@ const ChangelogRootComponent: FC<
...
   }>
 > = ({ match }) => {
-  const { endpoints, loading } = useEndpoints();
+  const endpointsState = useAppSelector((state) => state.endpoints.data);
+
+  const dispatch = useAppDispatch();
+  const spectacle = useSpectacleContext();
+  useEffect(() => {
+    // the typing is correct, we just arent setting it correctly above
+    // TODO remove ts-ignore
```
niclim, line 59: https://github.com/opticdev/optic/pull/853 will fix this

workspaces/ui-v2/src/store/endpoints.ts
```diff
@@ -0,0 +1,127 @@
...
+      endpointId: `${request.pathId}.${request.method}`,
+    })),
+    description: request.pathContributions.description || '',
+    purpose: request.pathContributions.purpose || '',
+  }));
+};
+
+const fetchEndpoints = createAsyncThunk(
+  'FETCH_ENDPOINTS',
+  async ({ spectacle }: { spectacle: IForkableSpectacle }) => {
```
niclim, line 112: I think this makes the most sense to do dependency injection here rather than trying to include this in middleware

workspaces/ui-v2/src/pages/changelog/ChangelogEndpointRootPage.tsx
niclim, line 62: Also note, this _should_ cause this action to be refetched if the spectacle instance changes (which is what we want). There are two options we have to handle the old data
- send a reload actions that all redux stores reset to loading
- just let all fetch.fulfilled actions overwrite the old store (we need to make sure we build the store with this requirement in mind)

workspaces/ui-v2/src/types/endpoints.ts
```diff
@@ -0,0 +1,17 @@
+export interface IPathParameter {
```
niclim, line 1: I couldn't think of a better name for this file - but it's kinda any _domain_ type (specifically for the FE)

workspaces/ui-v2/src/hooks/useFetchEndpoints.ts
```diff
@@ -0,0 +1,15 @@
...
+
+import { useSpectacleContext } from '<src>/contexts/spectacle-provider';
+import { endpointActions, useAppDispatch } from '<src>/store';
+
+export const useFetchEndpoints = () => {
+  const dispatch = useAppDispatch();
+  const spectacle = useSpectacleContext();
+  useEffect(() => {
+    // the typing is correct, we just arent setting it correctly above
+    // TODO remove ts-ignore
```
niclim, line 11: https://github.com/opticdev/optic/pull/853 will fix this

workspaces/ui-v2/src/hooks/useFetchEndpoints.ts
```diff
@@ -0,0 +1,15 @@
...
+    // @ts-ignore
+    dispatch(endpointActions.fetchEndpoints({ spectacle }));
+  }, [dispatch, spectacle]);
```
niclim, line 14: Also note, this _should_ cause this action to be refetched if the spectacle instance changes (which is what we want). There are two options we have to handle the old data
- send a reload actions that all redux stores reset to loading
- just let all fetch.fulfilled actions overwrite the old store (we need to make sure we build the store with this requirement in mind)

workspaces/ui-v2/src/pages/diffs/ReviewDiffPages.tsx
```diff
@@ -32,7 +33,8 @@ export function DiffReviewPages(props: any) {
   //dependencies
   const diff = useDiffsForCapture(boundaryId, diffId);
   const allRequestsAndResponsesOfBaseSpec = useAllRequestsAndResponses();
-  const allEndpointsOfBaseSpec = useEndpoints();
+  useFetchEndpoints();
```
niclim, line 36: This is used to initiate the fetch request. This request in particular could be globally triggered (i.e. anywhere since it's used in every page) but for now i think it's fine to leave it in the top level of each page's entry point.

workspaces/ui-v2/src/pages/diffs/EndpointDocumentationPane.tsx
```diff
@@ -40,14 +40,12 @@ export const EndpointDocumentationPane: FC<
...
   ...props
 }) => {
-  const { endpoints, loading } = useEndpoints();
-  const bodies = useEndpointBody(pathId, method, lastBatchCommit);
-  const thisEndpoint = endpoints.find(
-    (i) => i.pathId === pathId && i.method === method
+  const thisEndpoint = useAppSelector((state) =>
+    state.endpoints.results.data?.find(
+      (endpoint) => endpoint.pathId === pathId && endpoint.method === method
+    )
```
niclim, line 46: In this case (and a few other diff page cases) - the loading state is handled in a parent component. In the future we could have selectors that query the paths, and throw errors if there wasn't a loading state / error state to speed up development and avoid us having to build loading + error states everywhere we want to access `AsyncStatus` fields (with the assumption the loading / error states are handled somewhere above the tree)

workspaces/ui-v2/src/pages/diffs/ReviewDiffPages.tsx
niclim: Also, I think it's a little weird since it could be easy to "forget" to initiate a request (but fairly obvious since the loading state will never complete) - but we _want_ this level of decoupling and being explicit here increases the control we have

niclim:
@devdoshi @JaapRood I've built out what I think is a good foundation for our FE state management flow. There's some upcoming work where i think having a state management solution is helpful (not a blocker).

Ideally I'd like to be able to merge this, but completely understand if there's any sticking points and we can use this PR as a kick off point to tackle them (hopefully having code implement is easier to reason about changes)

I've tried to address the following:
- Spectacle loading and reloading (left this the same as existing) - reloading handled by https://github.com/opticdev/optic/pull/854#discussion_r645217797)
- Forking + simulated context - https://github.com/opticdev/optic/pull/855
- Type safety
- Granular control over loading + accessing data + rerenders

JaapRood:
Sorry @niclim, I totally meant to review this today but caught up in #856 and before I had to leave for home for other life stuff. Marking it as the first thing next time I work!

workspaces/ui-v2/src/store/endpoints.ts
```diff
@@ -0,0 +1,127 @@
+import {
+  createSlice,
+  createAsyncThunk,
+  SerializedError,
+} from '@reduxjs/toolkit';
+import { IForkableSpectacle } from '@useoptic/spectacle';
+import { AsyncStatus, IEndpoint } from '<src>/types';
+
+function sharedStart(array: string[]): string {
```
devdoshi, line 9: is this new functionality? it's not obvious what it's for / if it's the same as what was there before. How should I verify?

devdoshi (approved):
looks fine, but please address my comment

workspaces/ui-v2/src/store/endpoints.ts
niclim: This was copied over from https://github.com/opticdev/optic/pull/854/files#diff-0e32d0f84390e78d11a2a1ca6d12fbb8d2acbc039ad7dce8aea4224ffbebd6ffL157 - let me see if I can simplify / refactor it since I agree it's not super clear to follow

niclim: Pulled this fn out into a util file (called `findLongestStartingSubstring`) - changed some variable names and added a few simple tests

devdoshi: thanks! it's still a slightly weird algorithm for what it's doing but it works!

niclim: Agreed - it's "clever" to use the javascript sort and I'm fairly convinced it works in all cases, but it's definitely not the clearest lol - I also did find the [stack overflow post](https://stackoverflow.com/a/1917041) where the original solution must have come from while trying to think of other ways to do this

workspaces/ui-v2/src/pages/changelog/ChangelogEndpointRootPage.tsx
```diff
@@ -19,6 +18,7 @@ import {
 import { useEndpointBody } from '<src>/hooks/useEndpointBodyHook';
 import { SubtleBlueBackground } from '<src>/constants/theme';
 import { CodeBlock, EndpointTOC, TwoColumn } from '<src>/pages/docs/components';
+import { useAppSelector } from '<src>/store';
```
JaapRood, line 21: Just to verify for my own sanity: this to use a selector on the app-level-state (Redux store), right? Read a bit as if a different app can be selected 😅.

workspaces/ui-v2/src/pages/changelog/ChangelogListPage.tsx
```diff
@@ -41,16 +46,26 @@ export const ChangelogListPage: FC<
...
-  const { endpoints, loading } = useEndpoints(props.changelogBatchId);
+  const endpointsState = useAppSelector((state) => state.endpoints.results);
+  const changelog = useEndpointsChangelog(props.changelogBatchId);
+  const changelogByEndpointId = changelog.reduce(
+    (acc: Record<string, EndpointChangelog>, endpointChange) => {
+      acc[`${endpointChange.method}${endpointChange.pathId}`] = endpointChange;
+      return acc;
+    },
+    {}
+  );
```
JaapRood, line 57: Nit: shall we have our own version of the "to reduce or not to reduce" debate that happened last week on Twitter 😂🙈? I don't really care either way, but try (and often forget)  to add a comment to my reduces to indicate what kind of operation it's implementing (a `group-by` in this case). Not that this is hard to deduce from reading this, but it seems to address many people's issues over reduce readability.

workspaces/ui-v2/src/store/endpoints.ts
JaapRood: Especially true if the instance can change! I've made the mistake of writing things as middleware more than once before, and have regretted it. There's definite reasons to going the middleware route, but I've learnt the hard way that defaulting to simple dependency injection is most likely the best path.

workspaces/ui-v2/src/store/root.ts
JaapRood: Ahh gotcha, makes sense!

JaapRood (approved):
Great! Easy to read and follow. As someone who's used to reading Redux code, I found the action dispatches and selectors making it a lot clearer what the data flow is and how I can work with it.

workspaces/ui-v2/src/pages/changelog/ChangelogEndpointRootPage.tsx
niclim: Correct! It's a reexport of useSelector but with the typing of the current app's store typing (otherwise we cannot use typescript safety checks). Alternative names are welcome

workspaces/ui-v2/src/pages/changelog/ChangelogListPage.tsx
niclim: Hmmm in this case, a group-by might be more intuitive, and I believe we are already using groupby - I'll update this part

niclim: Ok I just tried implementing this, it's actually doing a `keyBy` instead of `groupBy` - in which we don't have the dependency installed (so I'm going to keep it as is)

In this case, I think the name is probably sufficient enough - but I agree, if there's anything more complicated with a reduce we should either:
- use a common library / util function
- annotate it with a comment

JaapRood: It is a `indexBy` / `keyBy`, you're right! Kind of proving the point that a comment could instantly take away all guessing.  Apart from that, it's just a few lines of code without any surprising behaviours, so not sure that warrants extracting it / pulling in a util library.

PR #856, opened 2021-06-04
Implement Commands Generator for deleting of Endpoint

JaapRood:
## Why
To let the user delete endpoints from a spec, the Optic Engine needs to be able to generate the commands that achieve this. The UI should be able to query these commands, so it can be in control of when the deletion is actually applied and in what context (simulated for previews, or to the persisted spec).

## What
The WASM interface now exposes a `spec_endpoint_delete_commands`, which uses the new `EndpointQueries.delete_endpoint_commands` to generate commands that result in the removal of all requests and responses for a given (path, http method) pair. Once the result events are applied to the internal `SpecProjection` (used for the diffing, but _not_ the Spectacle queries), the affected requests or responses are no longer considered part of the spec.

A couple of things will be tackled separately, which is possible with the feature still being hidden behind an opt-in feature flag:

- **Updating the Spectacle's spec projection**. The Optic Engine exposes a graph representation to Spectacle which it uses to resolve the queries it receives. Excellent existing snapshot coverage will help us handle the new events.
- **Pruning of path components**. To allow a deleted endpoint to be rediscovered as an "undocumented url" from the Optic UI, any path components without requests + responses should be pruned from the spec. This should either happen through the generation of additional "garbage collection" commands after applying a match, or through performing the pruning on a projection / query level.

## Validation
* [x] CI passes
* [x] Deletion commands can be generated
* [x] Deletion commands are valid and converted into events.
* [x] Events, when applied, removed all responses and requests from the spec.

### DIFF
--- a/workspaces/optic-engine-wasm/src/lib.rs
+++ b/workspaces/optic-engine-wasm/src/lib.rs
@@ -11,7 +11,7 @@ use optic_engine::{
 };
 use std::collections::HashMap;
 use uuid::Uuid;
-use wasm_bindgen::prelude::*;
+use wasm_bindgen::{__rt::WasmRefCell, prelude::*};
 
 #[wasm_bindgen(start)]
 pub fn init() {
@@ -378,6 +378,28 @@ pub fn spec_resolve_response(
 
   serde_json::to_string(&response)
     .map_err(|err| JsValue::from(format!("responses could not be serialized: {:?}", err)))
+}
+
+#[wasm_bindgen]
+pub fn spec_endpoint_delete_commands(
+  spec: &WasmSpecProjection,
+  path_id: String,
+  method: String,
+) -> Result<String, JsValue> {
+  let endpoint_queries = spec.endpoint_queries();
+
+  let commands = endpoint_queries
+    .delete_endpoint_commands(&path_id, &method)
+    .ok_or_else(|| {
+      JsValue::from("delete endpoint commands could not be generated for unexisting endpoint")
+    })?;
+
+  serde_json::to_string(&commands).map_err(|err| {
+    JsValue::from(format!(
+      "delete enpoints commands could not be serialized: {:?}",
+      err
+    ))
+  })
 }
 
 ////////////////////////////////////////////////////////////////////////////////////////////////
--- a/workspaces/optic-engine/src/commands/endpoint.rs
+++ b/workspaces/optic-engine/src/commands/endpoint.rs
@@ -225,7 +225,7 @@ pub struct UnsetRequestBodyShape {
 #[derive(Deserialize, Debug, Clone, Serialize)]
 #[serde(rename_all = "camelCase")]
 pub struct RemoveRequest {
-  request_id: RequestId,
+  pub request_id: RequestId,
 }
 
 // Responses
@@ -282,7 +282,7 @@ pub struct UnsetResponseBodyShape {
 #[derive(Deserialize, Debug, Clone, Serialize)]
 #[serde(rename_all = "camelCase")]
 pub struct RemoveResponse {
-  response_id: ResponseId,
+  pub response_id: ResponseId,
 }
 
 // Headers
@@ -473,6 +473,17 @@ impl AggregateCommand<EndpointProjection> for EndpointCommand {
         ))]
       }
 
+      EndpointCommand::RemoveRequest(command) => {
+        validation.require(
+          validation.request_exists(&command.request_id),
+          "request id must be in use to be removed",
+        )?;
+
+        vec![EndpointEvent::from(endpoint_events::RequestRemoved::from(
+          command,
+        ))]
+      }
+
       // Responses
       // ---------
       EndpointCommand::AddResponseByPathAndMethod(command) => {
@@ -497,6 +508,17 @@ impl AggregateCommand<EndpointProjection> for EndpointCommand {
         )?;
 
         vec![EndpointEvent::from(endpoint_events::ResponseBodySet::from(
+          command,
+        ))]
+      }
+
+      EndpointCommand::RemoveResponse(command) => {
+        validation.require(
+          validation.response_exists(&command.response_id),
+          "response id must be in use to be removed",
+        )?;
+
+        vec![EndpointEvent::from(endpoint_events::ResponseRemoved::from(
           command,
         ))]
       }
@@ -1008,6 +1030,46 @@ mod test {
   }
 
   #[test]
+  pub fn can_handle_remove_request_command() {
+    let initial_events: Vec<EndpointEvent> = serde_json::from_value(json!([
+      {"PathComponentAdded": {"pathId": "path_1","parentPathId": "root","name": "todos"}},
+      {"RequestAdded": {"requestId": "request_1", "pathId": "path_1", "httpMethod": "GET"}},
+      {"RequestAdded": {"requestId": "request_2", "pathId": "path_1", "httpMethod": "POST"}},
+      {"RequestRemoved": {"requestId": "request_2" }},
+    ]))
+    .expect("initial events should be valid endpoint events");
+
+    let mut projection = EndpointProjection::from(initial_events);
+
+    let valid_command: EndpointCommand = serde_json::from_value(json!(
+      {"RemoveRequest": {"requestId": "request_1" }}
+    ))
+    .expect("example command should be a valid command");
+
+    let new_events = projection
+      .execute(valid_command)
+      .expect("valid command should yield new events");
+    assert_eq!(new_events.len(), 1);
+    assert_debug_snapshot!("can_handle_remove_request_command__new_events", new_events);
+
+    // TODO: enable this usecase once EndpointProjection handles new removing of requests
+    // let unexisting_request: EndpointCommand = serde_json::from_value(json!(
+    //   {"RemoveRequest": {"requestId": "request_2" }}
+    // ))
+    // .unwrap();
+    // let unexisting_request_result = projection.execute(unexisting_request);
+    // assert!(unexisting_request_result.is_err());
+    // assert_debug_snapshot!(
+    //   "can_handle_remove_request_command__unexisting_request_result",
+    //   unexisting_request_result.unwrap_err()
+    // );
+
+    for event in new_events {
+      projection.apply(event); // verify this doesn't panic goes a long way to verifying the events
+    }
+  }
+
+  #[test]
   pub fn can_handle_add_response_by_path_and_method_command() {
     let initial_events: Vec<EndpointEvent> = serde_json::from_value(json!([
       {"PathComponentAdded": {"pathId": "path_1","parentPathId": "root","name": "todos"}},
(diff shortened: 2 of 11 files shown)


workspaces/optic-engine-wasm/src/lib.rs
```diff
@@ -11,7 +11,7 @@ use optic_engine::{
 };
 use std::collections::HashMap;
 use uuid::Uuid;
-use wasm_bindgen::prelude::*;
+use wasm_bindgen::{__rt::WasmRefCell, prelude::*};
```
JaapRood, line 14: Not sure how this got here, might be an auto import that needs to be cleaned up 🤔

workspaces/optic-engine-wasm/src/lib.rs
```diff
@@ -380,6 +380,28 @@ pub fn spec_resolve_response(
     .map_err(|err| JsValue::from(format!("responses could not be serialized: {:?}", err)))
 }
 
+#[wasm_bindgen]
+pub fn spec_endpoint_delete_commands(
+  spec: &WasmSpecProjection,
+  path_id: String,
+  method: String,
+) -> Result<String, JsValue> {
```
JaapRood, line 388: This is the main interface for Spectacle to integrate with. Thanks to the `wasm_bindgen` annotation this results in `import { spec_endpoint_delete_commands } from '@useoptic/optic-engine-wasm` to become functional.

workspaces/optic-engine-wasm/src/lib.rs
```diff
@@ -380,6 +380,28 @@ pub fn spec_resolve_response(
...
+  let endpoint_queries = spec.endpoint_queries();
+
+  let commands = endpoint_queries
+    .delete_endpoint_commands(&path_id, &method)
+    .ok_or_else(|| {
+      JsValue::from("delete endpoint commands could not be generated for unexisting endpoint")
+    })?;
```
JaapRood, line 395: Since we accept arbitrary strings for both the `path_id` and `method`, we have to deal with the case where commands are requested for an endpoint that doesn't exist. Here we map the `Option<T>` to a `Result<T, JsValue>`, which is so far the best way we've found to forward errors.

workspaces/optic-engine-wasm/src/lib.rs
```diff
@@ -380,6 +380,28 @@ pub fn spec_resolve_response(
...
+
+  serde_json::to_string(&commands).map_err(|err| {
+    JsValue::from(format!(
+      "delete enpoints commands could not be serialized: {:?}",
+      err
+    ))
+  })
```
JaapRood, line 402: For simplicity's sake, we serialise the results to a string and require the consuming code to deserialise again. Note that this doesn't mean necessarily Spectacle: it could choose to pipe the results straight through to the UI, where serialisation has to happen anyway.  There's alternatives for creating Javascript objects, or returning constructs  in the shared WASM memory (which are basically opaque JS objects containing only a pointer into memory) , however:

- creating Javascript objects [isn't more performant or efficient](https://rustwasm.github.io/wasm-bindgen/reference/arbitrary-data-with-serde.html#an-alternative-approach-serde-wasm-bindgen)
- working with opaque data structures is a pain, especially once you try to debug or inspect them. After going that route with the ScalaJS stuff, we reckoned it's more straightforward for the JS side to always work with JSON / Javascript objects.

workspaces/optic-engine-wasm/src/lib.rs
```diff
@@ -380,6 +380,28 @@ pub fn spec_resolve_response(
...
 }
 
+#[wasm_bindgen]
+pub fn spec_endpoint_delete_commands(
+  spec: &WasmSpecProjection,
+  path_id: String,
+  method: String,
+) -> Result<String, JsValue> {
+  let endpoint_queries = spec.endpoint_queries();
+
```
JaapRood, line 390: Why are we using endpoint queries here and aren't putting this logic anywhere else? Lets look at the typical CQRS cycle:

`... -> Events -> Projections -> [Queries] -> Commands -> Events -> ...`.

Since we're generating commands here, that makes the obvious place to implement these the queries (drawn as optional above, as sometimes projections are simple enough that we implement a query straight on it, but conceptually it's pretty much the same thing).

So why _Endpoint_Queries, specifically?

So far, the Events, Projections and Queries have all been broken up in `Endpoint*`, `Shape*` and `Rfc*`, describing request level, shape / body level and more metadata like parts of the chain, respectively. They're all variant of grouping `Spec*` variants. There are _other_ projections that cross some of these concerns, but generally it's a partitioning that has held up really well.

JaapRood:
@niclim let me know if this level of review I'm adding is either too little, too much, if you have any questions or comments :)

workspaces/optic-engine/src/projections/endpoint.rs
```diff
@@ -287,6 +287,43 @@ impl EndpointProjection {
       None
     }
   }
+
+  pub fn get_response_node_indexes<'a>(
+    &'a self,
+    path_id: &'a PathComponentId,
+    method: &'a HttpMethod,
+  ) -> Option<impl Iterator<Item = NodeIndex> + 'a> {
```
JaapRood, line 295: Thanks to @ejpbruel for helping me figure out how to get this function signature right. Has been something I've been scratching my head over for a while now, even avoiding iterators all together.

JaapRood:
The basic commands that remove the requests and response are now being generated. There's a test to try and verify they are indeed correct and have the expected results, however these new commands aren't handled by the Spec command handlers yet. Manual inspection has verified they are, though!

I've intentionally left out the generating of PathComponent pruning commands for now. I've come to believe this process is perhaps best implemented at the command application stage, as a round of "garbage collection". That's because the commands required for this depend on all the aforementioned having been applied and producing a new version of the projection / queries. Perhaps it's _even_ that can be, or is better done a projection level, rather than with additional commands. Enough questions to have a look at the a little bit later.

I'm keen to get this merged sooner rather than later, so @niclim and I have something common to work with, code we can start to integrate with the front end. However, perhaps it's best I at least add the spec command handlers (the part that verifies the commands and persists the resulting events to the `specification.json`). That way, nothing should crash, even if the projections don't reflect that things have been removed. As long as the new UI is hidden behind a feature flag, that should allow us to move forward without breaking anyone's spec files.

niclim:
> @niclim let me know if this level of review I'm adding is either too little, too much, if you have any questions or comments :)

Thanks - this was interesting to read through - I'm sure when I come to start implementing things I'll have more questions + understand smaller things in terms of the domain model + rust code but it makes sense from a high level here

workspaces/optic-engine/src/commands/endpoint.rs
```diff
@@ -465,6 +473,17 @@ impl AggregateCommand<EndpointProjection> for EndpointCommand {
         ))]
       }
 
+      EndpointCommand::RemoveRequest(command) => {
+        validation.require(
+          validation.request_exists(&command.request_id),
+          "request id must be in use to be removed",
+        )?;
```
JaapRood, line 480: I chose to be strict here, where removing a request that doesn't exist fails, but I'm not sure it's necessary. Generally, the outcome the user wanted has been achieved, which is why you see a lot of APIs implemented to be lenient on this. However, it in this case, I imagine we want the user (read: developer issuing this command) to be aware they might be trying to delete something that doesn't exist, and that this might be a bug on their end.

@devdoshi what do you reckon?

workspaces/optic-engine/src/commands/endpoint.rs
```diff
@@ -493,6 +512,17 @@ impl AggregateCommand<EndpointProjection> for EndpointCommand {
         ))]
       }
 
+      EndpointCommand::RemoveResponse(command) => {
+        validation.require(
+          validation.response_exists(&command.response_id),
+          "response id must be in use to be removed",
+        )?;
```
JaapRood, line 519: See discussion about removing request validation on my thoughts about being strict here.

JaapRood:
The command handlers are now implemented, meaning the issued remove request / response commands will yield events that are subsequently stored in the `specification.json`. While I originally thought that would be enough for this PR, the implementation of these handlers + their tests have highlighted how the `EndpointProjection` must be updated from these events for their validation logic to be right (and tested). With that in mind, I'll have a look at whether we can at the very least naively support updating the `EndpointProjection` as well. If it turns out to be too involved, I'll do that in another PR.

workspaces/optic-engine/src/projections/endpoint.rs
```diff
@@ -129,6 +130,33 @@ impl EndpointProjection {
...
+    if let Some(method_parent_edge_index) = method_parent_edge_index {
+      self.graph.remove_edge(method_parent_edge_index); // prevents request to be resolved from path node
+    }
+    self.node_id_to_index.remove(&request_id); // prevents request node to be looked up by request id
+
+    // GOTCHA: we're not deleting the request node itself, as that would invalidate self.node_id_to_index
+    // as the graph indexes shift.
+    // TODO: figure out the implications of the above, like correctness of queries and
+    // eventual garbage collection.
+  }
```
JaapRood, line 158: @devdoshi would value your thoughts on this approach and the resulting gotchas. What this does is:

- removes the edge between the `HttpMethod` node and `Request` node. This should prevent the queries to resolve this request, but allows Response to still exist (to be removed separately through its own event, even though it's technically possible to not issue those, leaving responses without a request).
- removes the request id from the node id to node index map. Should prevent queries that try to resolve the node per index. Does mean that technically these ids could be re-used, which I'm not 100% sure would have any negative implications. We could keep a separate list of deleted ids to prevent this.
- leaves all the actual graph nodes in place, as removing them will require the `node_id_to_index` to be rebuilt, which seems non-trivial. That does mean we have orphaned nodes hanging around. Probably not immediately an issue, but something we'd want to keep tabs on.

devdoshi:
@JaapRood I would expect to see inputs into the snapshot tests that include these removals so we can see and verify the output impact on the projections. Since we're not removing paths, how would you contrast this with just removing bodies, which we have domain commands for already? What's the overlap / gap in that approach vs this one?

JaapRood:
> Since we're not removing paths, how would you contrast this with just removing bodies, which we have domain commands for already


We will be removing paths, which is important for the UX around relearning endpoints to work correctly, but just not as part of these changes. I [wrote](https://github.com/opticdev/optic/pull/856#issuecomment-854829504) earlier: 

> I've intentionally left out the generating of PathComponent pruning commands for now. I've come to believe this process is perhaps best implemented at the command application stage, as a round of "garbage collection". That's because the commands required for this depend on all the aforementioned having been applied and producing a new version of the projection / queries. Perhaps it's even that can be, or is better done a projection level, rather than with additional commands. Enough questions to have a look at the a little bit later.

> I would expect to see inputs into the snapshot tests that include these removals so we can see and verify the output impact on the projections. 

Sure, of course, but I'm still wondering about the reasons for the separate Spectacle projection, whether it's something vestigial from early implementations, something that could perhaps be unified (because even with separate test coverage, having to implement and cover something should probably have a reason for it!).

JaapRood:
The internal `EndpointProjection` now deals with responses being removed as well. The tests for `endpoint_queries.delete_endpoint_commands` can now verify that after applying it's generated commands, the responses and requests for a given endpoint no longer show up in query results.

---

> Sure, of course, but I'm still wondering about the reasons for the separate Spectacle projection, whether it's something vestigial from early implementations, something that could perhaps be unified (because even with separate test coverage, having to implement and cover something should probably have a reason for it!).

It's used deep in Spectacle, giving the Node.js side a generic graph querying interface, so regardless something we'll have to update. It looks like if we mutate the graph correctly, all the queries implemented in the Typescript _should_ continue to work, hopefully with the snapshot tests helping us out there.

It _does_ show that this approach of exposing a graph directly is a trade off: it prevents a wild growth of more specific Queries interfaces from being exposed, which is probably useful at the time everything is still changing. However, it does tightly couple Rust Optic Engine code directly to it's consumer, with the graph representation pretty much locked unless you're keen to start changing things on the consumer side as well.

Not that this suggests it would have to change. It just shows, that no matter the representation, where a contract between two sides exists, there's effort involved with changing things down stream due to upstream changes. Given Spectacle exposes a GraphQL interface itself, the generic graph is probably the right call.

---

I'll go find out where the `SpectacleEndpointProjection` tests live and get those updated as well. As long as it's covered by snapshots properly, we should be able to make changes with relative confidence.

JaapRood:
A change of plans: lets get this PR merged without the support in Spectacle yet. This way all the basics of it land, and perhaps @niclim or somebody else can work on the further needed enhancements separately.

devdoshi (approved):
looks good for now, there should definitely be additional changes before this gets released (related to adding snapshot scenarios and handling spectacle changes) but this looks good to start.

PR #857, opened 2021-06-04
Add UI elements for deleting endpoint on documentation page

niclim:
## Why

Adding in the UI components for deleting an endpoint - this is all behind a feature flag as it's not fully functional

note: this is branched off the [implement-redux branch](https://github.com/opticdev/optic/pull/854) - I'll land the implement-redux PR first (after @JaapRood gets a chance to review it), but opening this up for review first

## What

In this PR, everything is behind a feature flag (except for `Edit Descriptions` -> `Edit`). This just adds in the UI components - there are follow up PRs for the UI side to be ready to go.

Separate PRs
- Migrate contributions to redux + hook up deletion state into redux
  - As part of this, I'll also update to block edits to contributions (and other states) if the endpoint is deleted (it's much easier to do this if we add it into the redux store)
- connect spectactle queries to the command generation + make mutations through spectacle command
- design consistency + move towards a design system

Screenshots
Endpoint page delete + undo deletion
Screen Shot 2021-06-04 at 1 56 25 PM
Screen Shot 2021-06-04 at 1 56 11 PM

Endpoint list page delete + undo deletion
Screen Shot 2021-06-04 at 12 51 45 PM

Confirmation modal for deleting an endpoint
Screen Shot 2021-06-04 at 2 01 34 PM

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


workspaces/ui-v2/src/pages/docs/DocumentationRootPage/DocumentationRootPage.tsx
```diff
@@ -0,0 +1,89 @@
+import React, { FC, useMemo } from 'react';
```
niclim, line 1: This is mainly the same page - just moved over (git cant tell because there's a fair amount of changes in this file)

workspaces/ui-v2/src/pages/docs/DocumentationRootPage/EndpointRow.tsx
```diff
@@ -0,0 +1,140 @@
...
+  const classes = useStyles();
+  const endpointId = getEndpointId({
+    method: endpoint.method,
+    pathId: endpoint.pathId,
+  });
+  const { isEditing } = useContributionEditing();
+  const { showDeleteEndpointUi } = useFeatureFlags();
+  const [deleteModalOpen, setDeleteModalOpen] = useState(false);
+
+  // TODO redux-delete-implement replace this with redux selector query
```
niclim, line 40: Will address this in a follow up PR (which will involve moving the contributions editing flow into redux and then attaching delete state into that)

workspaces/ui-v2/src/pages/docs/components/DeleteEndpointConfirmationModal.tsx
```diff
@@ -0,0 +1,62 @@
...
+            Are you sure you want to delete endpoint the following endpoint?
+            <EndpointName
+              method={endpoint.method}
+              fullPath={endpoint.fullPath}
+            />
+          </div>
+          <div className={classes.confirmationTextContainer}>
+            This will permanently delete this endpoint. You will lose all
+            contributions associated with this endpoint and will no longer see
+            this endpoint in your documentation page.
```
niclim, line 42: Is there anything else we should raise to the user (with how we plan to implement delete in the domain model)?

workspaces/ui-v2/src/contexts/config/AppConfiguration.tsx
```diff
@@ -1,6 +1,8 @@
 import React, { ReactNode, useContext } from 'react';
 import invariant from 'invariant';
-interface IFeatureFlags {}
+interface IFeatureFlags {
+  showDeleteEndpointUi: boolean;
+}
```
JaapRood, line 6: I know this interface was already here, but as we discussed after #681, we don't want to codify temporary feature flags (e.g. flags used for hiding features that will at some point be enabled and then no longer be configurable) as Typescript types. Instead, we'll want to use the `REACT_APP__` feature flags as described in [CreateReactApp](https://create-react-app.dev/docs/adding-custom-environment-variables). Yes, those are ugly, and get unwieldy over time, but all of that should be a driving force in getting rid of these feature flags as soon as possible, or getting rid features we'll never enable.

workspaces/ui-v2/src/pages/docs/components/DeleteEndpointConfirmationModal.tsx
JaapRood: Once we've also implemented revert, this wouldn't have as much of a disastrous impact, as the delete could be reverted. Also, we do actually still store the contributions (indefinitely until we implement implement compaction of the specification.json). 

That said, better warn sternly here and have it turn out not that bad.

JaapRood (commented):
Thanks for the detailed PR description, screenshots really make it easier to see what's going on here. Implementation looks like a decent place to start as well, should allow us to get something working end-to-end sooner rather than later. With a bit of luck we'll have some appetite left over to tweak the design of it (and the Docs edit page more generally). 

I'd highly recommend we instantly move away from the `IFeatureFlags` as I reckon that should never have landed in the first place. That said, if need really be, that can be a separate pull request.

JaapRood (approved):

PR #858, opened 2021-06-05
Switch out contributions store with redux and connect delete endpoint

niclim:
## Why

Part of adding in the new delete endpoint flow touches the contributions store. Figured this was a good time to be able to rip this out and implement redux. There are some parts that might be a little more verbose (there are ways we can make this simpler) - however this _does_ fix the performance issues from editing (see PR here https://github.com/opticdev/optic/pull/740) - meaning we don't need to debounce every state since redux allows us to have more control over updates.

## What

In future PRs
- [ ] Block editing when an endpoint is (staged to be) deleted
- [ ] Hook up spectacle

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


workspaces/ui-v2/src/pages/docs/components/EditContributionsButton.tsx
```diff
@@ -3,20 +3,60 @@ import { ToggleButton } from '@material-ui/lab';
...
+  useAppDispatch,
+  documentationEditActions,
+  selectors,
+} from '<src>/store';
 
 export function EditContributionsButton() {
   const classes = useStyles();
+  const spectacle = useSpectacleContext();
+  const clientId = useClientAgent();
+  const clientSessionId = useSessionId();
```
niclim, line 22: So these things could sit in a `metadataSlice` - which means we could expose them in selectors, but we wouldn't need to manually pass these in for the `save` action (which instead we could derive from the current state)

workspaces/ui-v2/src/store/documentationEdit.ts
```diff
@@ -0,0 +1,187 @@
...
+  },
+  {
+    dispatch: AppDispatch;
+    state: RootState;
+  }
+>(
+  'SAVE_DOCUMENTATION_CHANGES',
+  async ({ spectacle, commitMessage, clientId, clientSessionId }, thunkApi) => {
+    const state = thunkApi.getState();
+    // TODO fetch from spectacle
```
niclim, line 34: Will follow this up in a subsequent PR - a spectacle query needs to be exposed which hooks up to https://github.com/opticdev/optic/pull/856

workspaces/ui-v2/src/store/selectors/documentationEditSelectors.ts
```diff
@@ -0,0 +1,35 @@
+import { RootState } from '../root';
+import { IContribution } from '<src>/types';
+
+// Valid changes dedupes:
+// - same contribution value as already set
```
niclim, line 5: We aren't currently doing this since all the state currently isn't in redux - this is _possible_ but kind of convoluted to get the relevant state / initial values for each contribution.

This is identical behavior to how it's currently implemented, however once we move all state into redux this is easier (not trivial, since contribution values will be stored in lots of different locations based on how they are fetched - e.g. fields, bodies, shapes, etc).

workspaces/spectacle-shared/src/local-cli/client.ts
```diff
@@ -22,7 +22,10 @@ import {
 import { IApiCliConfig } from '@useoptic/cli-config';
 
 export class LocalCliSpectacle implements IForkableSpectacle {
-  constructor(private baseUrl: string, private opticEngine: IOpticEngine) {}
+  private spectacleUpdateEvents: ((...args: any) => any)[];
+  constructor(private baseUrl: string, private opticEngine: IOpticEngine) {
+    this.spectacleUpdateEvents = [];
```
niclim, line 27: We need to implement this because:
- The local cli version doesn't have a spectacle watcher on the UI side
- only the in memory spectacle instances have a spec change notification watcher

We force an identity change in spectacle (which is the same as how the in memory instances work) whenever we call a mutation. Since we aren't actually making a web socket connection to our cli-server, we listen to when we send a mutate event. 

We can tackle this in a separate PR, but do we want to create a web-socket connection between the cli-server and the UI? Would that make sense, or would we expect users to refresh their page if they switched branches, or something changed on the file system outside of optic? (i.e. connecting this https://github.com/opticdev/optic/pull/826 to the UI)

workspaces/ui-v2/src/pages/docs/components/RenderBody.tsx
```diff
@@ -53,7 +60,12 @@ export function TwoColumnBodyEditable(props: TwoColumnBodyProps) {
       }
     />
   );
-}
+};
+
+// Memoize this as there could be a large number of contribution groups
```
niclim, line 65: It's probably also worth virtualizating this since huge specs cause a lot of rerenders outside of the viewport, leading to slow responsiveness. 

The good thing about redux here though is that we have more fine grained control over renders making performance improvements (where we need it) much simpler

workspaces/spectacle-shared/src/local-cli/client.ts
```diff
@@ -45,6 +55,10 @@ export class LocalCliSpectacle implements IForkableSpectacle {
     // send query to local cli-server
     return JsonHttpClient.postJson(`${this.baseUrl}/spectacle`, options);
   }
+
+  registerSpectacleUpdateEvent<T extends (...args: any) => any>(fn: T) {
+    this.spectacleUpdateEvents.push(fn);
+  }
```
JaapRood, line 61: Probably best to pull in node's `EventEmitter` here if you're going the events + listeners way (with AsyncGenerators as the alternative) rather than building our own. It's already being imported and bundled in other places, so no reason to stay away from it for that. Also will allow your hook to deregister itself and not leak any handles.

workspaces/ui-v2/src/entry-points/local/local-cli.tsx
```diff
@@ -115,11 +115,17 @@ export function useLocalCliServices(
...
+  const [spectacle, setSpectacle] = useState(
+    new LocalCliSpectacle(apiBaseUrl, opticEngine)
   );
-  const capturesService = React.useMemo(
+  const refEngine = useRef(opticEngine);
+  const refApiUrl = useRef(apiBaseUrl);
+  useEffect(() => {
+    spectacle.registerSpectacleUpdateEvent(() => {
+      setSpectacle(new LocalCliSpectacle(refApiUrl.current, refEngine.current));
+    });
```
JaapRood, line 126: We'll want to return [an effect clean up function here](https://reactjs.org/docs/hooks-effect.html#example-using-hooks-1), so we can tidy up the event listener. Might look like a nit pick, but bugs caused due to unregistered event emitters tend to show themselves in super obscure ways and can be a massive PITA.

workspaces/ui-v2/src/pages/docs/EndpointRootPage.tsx
```diff
@@ -218,10 +230,11 @@ export const EndpointRootPage: FC<
           }
         />
 
-        {bodies.requests.map((i, index) => {
+        {bodies.requests.map((i) => {
           return (
             <TwoColumnBodyEditable
-              key={index}
+              key={i.rootShapeId}
```
JaapRood, line 236: Yes, nice one! Using array index as component keys is a mistake that leads to some nasty rendering bugs / incorrect react component instancing!

workspaces/ui-v2/src/pages/docs/components/RenderBody.tsx
JaapRood: Possible premature optimisation, or did we already have concrete cases of this being a performance bottleneck? Memoization has a cost and can actually lead to degraded performance, especially when applied in multiple places in the hierarchy, which can be tricky to reason about. Lets not apply any of it until there's concrete evidence that's where lots unnecessary renders happen, and then memoisation reduces those numbers.

JaapRood (approved):
Apart from a couple of small meta things I'd recommend tackling for our long term sanity, this looks good!

workspaces/ui-v2/src/pages/docs/components/RenderBody.tsx
niclim: This does actually cause an issue on the `debug-github-capture` example - there's ~200 fields to render, and on first render (starting edit flow) - there's 200 fields to update at once (which needs to render the input field).

workspaces/ui-v2/src/entry-points/local/local-cli.tsx
niclim: Bleh - I was like I need to handle this and totally forgot about useEffect's teardown yesterday (was frustrated with the internet being out!) Thanks for the reminder!

workspaces/spectacle-shared/src/local-cli/client.ts
niclim: Interesting - I didn't realize that we could use the nodejs implementation - but apparently webpack includes this for us https://www.npmjs.com/package/events

workspaces/ui-v2/src/pages/docs/components/RenderBody.tsx
niclim: This fixes the case where we _initially_ type into a new field (which changes pendingCount -> causes a rerender int he `EndpointRootPage` component -> causing this to rerender when it doesn't need to) - it does not fix the case where we actually need to rerender these fields (where if we have ~200 fields, we have to render a large number of UI elements which causes UI jank - in which case virtualizing this could help)

PR #861, opened 2021-06-07
Fix pending endpoint documentation display

niclim:
## Why

Turns out I did need to implement the simulated store with redux (the endpoint documentation pane) - The pending endpoint page should now load - instead of saying there is no endpoint here

Screen Shot 2021-06-07 at 11 08 06 AM

## What

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/ui-v2/src/pages/diffs/contexts/SimulatedCommandContext.tsx
+++ b/workspaces/ui-v2/src/pages/diffs/contexts/SimulatedCommandContext.tsx
@@ -4,12 +4,10 @@ import { v4 as uuidv4 } from 'uuid';
 import { IForkableSpectacle } from '@useoptic/spectacle';
 
 import { Loading } from '<src>/components';
-import {
-  SpectacleStore,
-  useSpectacleContext,
-} from '<src>/contexts/spectacle-provider';
+import { SpectacleStore } from '<src>/contexts/spectacle-provider';
 import { useSessionId } from '<src>/hooks/useSessionId';
-import { createReduxStore, useAppDispatch, endpointActions } from '<src>/store';
+import { createReduxStore } from '<src>/store';
+import { useFetchEndpoints } from '<src>/hooks/useFetchEndpoints';
 
 type SimulatedCommandStoreProps = {
   spectacle: IForkableSpectacle;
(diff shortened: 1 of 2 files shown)


workspaces/ui-v2/src/pages/diffs/contexts/SimulatedCommandContext.tsx
```diff
@@ -22,6 +26,7 @@ export const SimulatedCommandContext = React.createContext<SimulatedCommandConte
 export function SimulatedCommandStore(props: SimulatedCommandStoreProps) {
   const value = { previewCommands: props.previewCommands };
   const [isProcessing, setIsProcessing] = useState(true);
+  const store = useMemo(() => createReduxStore(), []);
```
JaapRood, line 29: Wait, so we create an entire new app state for simulated commands? I'm guessing the assumption here is that all the state in the main store is invalidated by a spec change? It _seems_ like that's a bit much and we'd want something more surgical, but at the same time I can't tell you what that would be. Could also happen later, of course.

JaapRood (approved):
LGTM, just a bit confused on something that probably just shows I haven't touched this part of the codebase yet.

workspaces/ui-v2/src/pages/diffs/contexts/SimulatedCommandContext.tsx
niclim: So the simulated command store is used for us to "preview" the state of say an endpoint after a diff is applied and learnt. We need a spectacle instance to run in memory, and likewise need a completely different state to represent the "simulated commands" - since the source of truth is not the main redux store (since we haven't applied the commands). Now I agree this is fairly heavy weight to load the redux store - and the things we need to load up are a subset of this (endpoints, shapes, anything in the EndpointDocumentationPanel)

The reasons I decided to pull in the new redux store is:
- It allows us to use the same appSelectors regardless of where it is (we can reuse our existing components that already implicitly rely on this behavior)

- A new redux store is effectively a large object with no initial data (we control the initial fetch state) - which _shouldn't_ be too much overhead to load in

The alternative is to either make any component that uses the simulated command store "render only" and accept props for their data, build a context wrapper that handles the simulated command store or to figure out how not load the entire redux store

Creating a new store and keeping the same interface is a tradeoff of iteration speed and simplicity against performance (i'm unclear on what / how much performance gains there would be here). It feels a little odd, but I've come to think this makes the most sense in the simulated command context - i.e. each simulated command context has it's own "state" (think of it as an iframe)

PR #862, opened 2021-06-08
Create redux metadata slice

niclim:
## Why

Create a metadata slice in redux (follow up from this comment https://github.com/opticdev/optic/pull/858#discussion_r646813551). This also means we aren't making fetches on every mount of a component that uses the hook (i.e. it's fetched up front once).

## What

New / added
- MetadataSlice which includes
  - apiName
  - clientAgent / clientId
  - clientSessionId
  - specificationId
- MetadataFetcher component (to trigger the fetches) - this could be renamed to global fetcher (since there will probably be more things that need to be globally fetched for each UI instance - local cli, public example, etc)

Deleted
- useApiNameHook
- useClientSessionId
- useClientId
- MetadataProvider

Misc fixes:
- Renamed `findLongestStartingSubstring.ts` to `findLongestStartingSubstring.test.ts`

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


workspaces/ui-v2/src/components/navigation/TopNavigation.tsx
```diff
@@ -11,13 +11,13 @@ import { Container } from '@material-ui/core';
...
 import ChangeHistoryIcon from '@material-ui/icons/ChangeHistory';
-import { useApiName } from '<src>/hooks/useApiNameHook';
 import { useDiffReviewPageLink, useDocumentationPageLink } from './Routes';
 import { useAppConfig } from '<src>/contexts/config/AppConfiguration';
+import { useAppSelector } from '<src>/store';
 
 export function TopNavigation(props: { AccessoryNavigation: any }) {
   const classes = useStyles();
-  const apiName = useApiName();
+  const apiName = useAppSelector((state) => state.metadata.data?.apiName || '');
```
niclim, line 20: When we come to render this component, the appSelector should always return the apiName (given we are rendering loading + error states higher in the tree) - typescript cannot infer this - and there are cases where we could forget to handle loading + error states (if we refactor, etc)

My thoughts here are to:
- Create a selector that checks and throws an error if `data` is undefined - this way we have type safety and also have quicker developer feedback if we haven't explicitly handled error states

Thoughts? Any alternative (rather than having to handle these with casting to some default value)?

JaapRood: I think it's fine for this not to be safe, as failure doesn't have any major impact on the rest of the app's flow, it's only being used in a single spot so it's not too hard to get right. If in QA or through user testing this pops up as being missing sometimes, we can always come back and address this (properly with the separate selector as you've mentioned).

JaapRood (approved):
Looks good, although I can't help but escape the thought that called this "session" instead of "metadata" would more concisely describe the function of this, with "metadata" being pretty.... meta 😂. Either way, this does what it should do!

workspaces/ui-v2/src/components/navigation/TopNavigation.tsx
niclim: I think the selectors are the best long term solution - but for now this should be fine. There are cases where we have the same pattern where we'll handle loading states higher in the tree, and will want the assumption we've already loaded the data / handled errors (loading endpoint data, etc).

niclim:
> Looks good, although I can't help but escape the thought that called this "session" instead of "metadata" would more concisely describe the function of this, with "metadata" being pretty.... meta 😂. Either way, this does what it should do!

Haha yeah, metadata is a fairly loose term (kind of lazy naming here) - but I don't think session is correct, since specificationId, clientAgent and apiName are all different (and more "associated" data with the current spec)

PR #864, opened 2021-06-09
refactor: Change paths to include `personId`

jshearer:
Part of https://github.com/opticdev/monorail/pull/307

## Why
Individual specs (as opposed to specs as part of a larger project, which will come as a later feature) live on the `Person` aggregate in `backend-web`. Since we aren't using snapshots yet, and the cost is minimal to include the extra piece of data, we now also include the `personId` that a particular spec lives on when fetching it.

## What
Paths change to `/public-specs/{personId}/{specId}`.
GitBot learns about this here: https://github.com/opticdev/optic-changelog/pull/45

### DIFF
--- a/workspaces/ui-v2/src/entry-points/cloud/TopLevelRoutes.tsx
+++ b/workspaces/ui-v2/src/entry-points/cloud/TopLevelRoutes.tsx
@@ -11,7 +11,7 @@ export default function TopLevelRoutes() {
         <Route
           strict
           //@TODO: centralize this path pattern
-          path="/public-specs/:personId/:specId"
+          path="/person/:personId/public-specs/:specId"
           component={CloudViewer}
         />
 
--- a/workspaces/ui-v2/src/entry-points/cloud/TopLevelRoutes.tsx
+++ b/workspaces/ui-v2/src/entry-points/cloud/TopLevelRoutes.tsx
@@ -11,7 +11,7 @@ export default function TopLevelRoutes() {
         <Route
           strict
           //@TODO: centralize this path pattern
-          path="/public-specs/:specId"
+          path="/public-specs/:personId/:specId"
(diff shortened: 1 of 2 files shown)


workspaces/ui-v2/src/entry-points/cloud/TopLevelRoutes.tsx
```diff
@@ -11,7 +11,7 @@ export default function TopLevelRoutes() {
         <Route
           strict
           //@TODO: centralize this path pattern
-          path="/public-specs/:specId"
+          path="/public-specs/:personId/:specId"
```
devdoshi, line 14: Would it make sense to change the resource hierarchy so it's a person who has public-specs, as opposed to the other way around? How would you accommodate an org having public specs as well?

workspaces/ui-v2/src/entry-points/cloud/cloud-viewer.tsx
```diff
@@ -73,11 +73,14 @@ export default function CloudViewer() {
           return process.env.REACT_APP_STAGING_API_BASE;
         }
       })();
-      const response = await fetch(`${apiBase}/api/specs/${specId}`, {
-        headers: { accept: 'application/json' },
-      });
+      const response = await fetch(
+        `${apiBase}/api/specs/${personId}/${specId}`,
```
devdoshi, line 77: similar comment here; this resource hierarchy is inconsistent and doesn't accommodate orgs

jshearer:
@devdoshi changed paths to be `/person/:personId/public-specs/:specId`

devdoshi:
> @devdoshi changed paths to be `/person/:personId/public-specs/:specId`

darn english language - should we do people/:personId?

devdoshi (approved):

PR #866, opened 2021-06-09
Connect endpoint deletion commands to UI via spectacle

niclim:
## Why

Combines the latest code on `develop` with the code in https://github.com/opticdev/optic/pull/856 (soon to be merged). The github diff is hard to read because it requires code from develop _and_ PR #856 - once the PR is merged this should be easier to read / reason about.

This works all the way from selecting things on the UI (`REACT_APP_FF_SHOW_DELETE_ENDPOINT=true yarn start`) to deleting + generating the deletion commands to applying the delete commands.

This currently does _not_ work end to end - as the deletion commands are not applied to the spectacle_endpoint projection (which this PR https://github.com/opticdev/optic/pull/865 should handle). I believe (or hope) when all these PRs are merged we can delete an endpoint end to end (with some minor follows to be handled - described in jaap's pr)

## What

- Hooks up the wasm code to spectacle. Merging https://github.com/opticdev/optic/pull/856, https://github.com/opticdev/optic/pull/865 and this PR should have a "working" end to end deletion flow
- Redirect to different root page if deleting endpoint

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


workspaces/ui-v2/src/pages/docs/DocumentationRootPage/EndpointRow.tsx
```diff
@@ -94,7 +108,10 @@ export const EndpointRow: FC<EndpointRowProps> = ({ endpoint }) => {
         >
           <EndpointNameMiniContribution
             id={endpointId}
-            endpointId={endpointId}
+            endpoint={{
```
niclim, line 111: I had to change this interface since we want to specifically split out method and pathId when storing these in contributions and in the deleted endpoint state.

workspaces/ui-v2/src/store/documentationEdit/slice.ts
```diff
@@ -4,82 +4,8 @@
...
-          batchCommitId: uuidv4(),
-          clientId,
-          clientSessionId,
-        },
-      });
-    }
-  }
-);
+import { createSlice, PayloadAction } from '@reduxjs/toolkit';
+import { saveDocumentationChanges } from './thunks';
```
niclim, line 8: Restructured to:
```
- documentationEdit (folder)
   - index.ts (reexports slice.ts)
   - slice.ts (redux slice logic0
   - thunks.ts (async)
```

niclim:
@JaapRood - I'm opening up this PR for review, but please merge in your PR https://github.com/opticdev/optic/pull/856 or just view this commit diff https://github.com/opticdev/optic/pull/866/commits/ce457e4c547e898502ec5c79b29ee0d779b23ed0 as this PR required changes from develop and your branch

workspaces/spectacle/src/index.ts
```diff
@@ -307,6 +334,10 @@ export async function makeSpectacle(opticContext: IOpticContext) {
       shapeChoices: async (parent: any, args: any, context: any, info: any) => {
         return context.spectacleContext().shapeViewerProjection[args.shapeId];
       },
+      endpoint: async (parent: any, args: any, context: any, info: any) => {
+        const { pathId, method } = args;
+        return context.spectacleContext().getEndpointProjection(pathId, method);
+      },
```
JaapRood, line 340: Since this is the only field in this schema so far, it's fine to generate the whole thing at once. Eventually, we might want to move each `command` field to its own resolver so the various commands are only generated if the field is requested by a client. In this case the commands are pretty cheap to generate, but that's definitely not always the case.

JaapRood (approved):
Solid stuff!

PR #867, opened 2021-06-10
make path nullable

niclim:
## Why

Testing out what happens with nullable paths - turns out the behavior isn't too good - deleting an endpoint looks liek this
Screen Shot 2021-06-09 at 5 45 18 PM

We actually want this filtered...

## What

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


niclim:
This doesn't work well - closing POC

PR #868, opened 2021-06-10
filter out unreachable nodes in spectacle endpoints graph

niclim:
## Why

This is probably the ideal fix, but I'm not sure why the changelog is weird - I need to spend some time to dig and figure out why. The changelog for the github example things everything is _updated_ when in reality, these are new endpoints (the oranges should be green).

Screen Shot 2021-06-09 at 5 45 53 PM

I'm also not sure if we need to check for reachable shapes in the shapesGraph - I don't think so though? 

I'm also not sure what else we need to handle here

## What

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


niclim:
This probably works (the batch commit nodes aren't being attached back to requests / response nodes which is why the changelog is wonky) - but this is probably better suited in the rust layer (in serializing the graph) - will add into this PR https://github.com/opticdev/optic/pull/865

PR #869, opened 2021-06-10
Prune unused paths left after deleting endpoints

JaapRood:
## Why
Now that endpoints can be deleted, we have to start pruning the paths these endpoints were registered with. Without this, a deleted endpoint request / response bodies will just be re-learned again automatically next time traffic is observed, as a URL is matched. By deleting path components that have become unused after the deletion of an endpoint, they will once again show up as "undocumented urls".

One of the use cases this resolves is where an incorrect path parameter got documented:

> - I accidentally leave `aidanc` as a constant , instead of `{username}`
> - `/users/friends/aidanc` *— this is endpoint I want to delete*
> - `/users/friends`
> - `/users/profile`
>    - ^ when I delete this endpoint, it should also delete `aidanc` component since no other API endpoint depends on it. But since `/friends` `/users` are used elsewhere, those should be retained.
>    - **If we don't prune path parameters the endpoint will just be relearned next time traffic is observed**

## What
A new query is introduced to detect all unused paths in the current spec, meaning they don't have any endpoints. This includes any path components that are unused due to all their children paths being unused as well in addition to not having any endpoints.

The new `PrunePathComponents` has been added for the client to be able to trigger the pruning process. This is a higher order level command which performs the above query and issues and path component + path parameter removal commands. 

The `EndpointsProjection` has been updated to handle the newly issued `RemovedPathComponent` and `RemovedPathParameter` events. To keep this do-able and not get into conditions where handling these events also means deleting nested requests or responses, the constraint was added to the `RemovePathComponent` and `RemovePathParameter` commands to only be valid when they don't have any children paths or request / responses. The unused paths query takes this into account, returning unused paths descendants first, making sure paths are removed in the correct order.

## Validation
* [x] CI passes
* [x] Unused paths are removed from a spec upon issuing of the command.


JaapRood:
With the latest changes I implemented the finding of unused paths. It uses a petgraph's generic depth-first-search, starting from root, allowing the logic to be implemented without recursion on our end. The logic is pretty generic and should hold up well, with a basic test asserting it produces the right results.

While implementing and developing a more subtle understanding of this subject as we go, quite a few alternative approaches came to light for me:

- Rather than _always_ performing the path pruning step, it could be driven by a command from the consumer side. There could be a single `PruneUnusedPaths` that the client could include after all the endpoint deletion commands it's trying to apply. This would yield more control over when paths are pruned, rather than just always.
- For controlling also _which_ paths might get pruned instead of only _when_, every generated set of "Delete endpoint commands" could include a final `PruneEndpointPaths { pathId, method }`. As an added bonus, this removes the need for a complete visiting of the paths graph. With the generic finding of unused paths we have now, that should still work, even with successive and interdepend endpoint removals. I'm leaning towards this one.

Fortunately the logic just written still applies and can easily be repurposed for the above changes in mechanism, it could perhaps just be run from a path component further down the graph, rather than the root path component node.

While we think the above mechanisms over, I'll focus on the generating of the right path removal commands, and their conversion into events.

JaapRood:
The command generation and handling is now implemented.

I ended up going for a `PrunePathComponents` command, that for now we'll rely on the client to issue rather than automatically for each batch. The implementation was pretty smooth and straightforward, which I take as an indicator the logic is in the right spot. Meanwhile, we still allow the client control over when paths pruning happens, while not requiring it to have knowledge about all the constraints a projection must be built under in order to make the pruning effective. As long as it happens _after_ any removals or adding (which makes intuitive sense), it'll work out well. 

Anything more specific, like a prune command for only specific ids is something to look at in the future, if a global pruning turns out to be too blunt of a tool.

Next up, I'll get the newly generated path removal events handled in the internal `EndpointsProjection`.

JaapRood:
While I implemented the internal `EndpointProjection` applying of removing path events, it became obvious that for this to remain sane, the paths should only be allowed to be removed if they are unused. Without that, removing a path or parameter wouldn't just be a matter of deleting an edge on the graph and path id, but walking it's entire subtree and removing all children entities (like requests and responses as well).

To give ourselves a chance of maintaining that constraint, I added it to the command handler: removing path components or path parameters requires them to:

- not have any child paths
- not have any requests
- not have and responses 

While that was straightforward enough to implement once I had the walking of the graphs working, I realised that the current query for discovering unused paths **only returns unused paths for leaf path components**. Those that don't have any children paths, but also no requests and responses. That misses the possibility where a path component is just as well unused if it's does have children, but all of it's children are unused as well. The entire sub tree of the path graph should be pruned.

However, when that's implemented, the order of these unused path ids becomes important as well: children have to be removed before their parents are.

--- 

For the pruning to be complete and have the effect that we want, we should implement the behaviour described above. I'll look at that now.

JaapRood:
In the version I just pushed I've adapted the unused path discovery to include paths that are unused due to only having unused children. They're also returned in an order that would lead to valid commands when the paths are chosen to deleted, listing the children before their parents.

Screenshot 2021-06-11 at 18 38 35

In the above example (dump from the updated snapshot test), `path_3` is correctly identified to also be unused, but not after `path_4` is listed as such first.

--- 

I'm pretty confident that this covers the main logic. It should be ready for a review and a merge.

JaapRood:
@niclim and @devdoshi, I'll leave this up for you to finish. However, if you run into anything highly disagreeable, feel free to leave it for when I'm working next.

niclim:
Ok I'm going to take over the rest of the PR. Here's a list of things that will need to be done for this to work:
- [ ] Send `PrunePathComponents` when deleting an endpoint [here](https://github.com/opticdev/optic/blob/develop/workspaces/ui-v2/src/store/documentationEdit/thunks.ts#L70-L76)
- [ ] Handle `without_path` and `without_path_parameter` in the `spectacle/endpoints` projection
- [ ] Handle `isDeleted` paths and path parameters in the UI? (TBD if this is required - my guess is that we need to do this in the diff flow)

niclim:
Ok @devdoshi I think this is ready to review. I've connected the PrunePathComponents command to the UI, and surfaced the relevant information to the UI (such that relearning is possible) - I've also tested relearning an endpoint and it's working as I expect.

This screenshot is of me using the `examples/todos-partial` example - deleted the `/api/lists/:listId/completed` endpoint, relearnt the captures as two separate endpoints
Screen Shot 2021-06-14 at 1 58 52 PM

devdoshi (approved):
Looks good! I would prefer a design where PrunePathComponents is not part of the Command API surface area, and instead following the pattern we've established where we can query to get commands to achieve a certain effect. This would enable us to have the client choose to only get commands to prune unused paths in the context of a particular endpoint, for example. However, it's about 90% of the same code, so I'm happy with this approach.

niclim:
Hmm I think what you suggest is makes more sense and is more correct, however I don't think it's a huge amount of work to migrate later on if / when we find out this is important - so i'm inclined to merge this in as is and test it + update this later on (keeping it as is doesn't seem likely to cause us any issues in the short term)

PR #872, opened 2021-06-10
optic-engine-native workspace needs .env

devdoshi:
## Why
When we reorganized the optic-engine folder structure we did not preserve the invariant where the workspace that is responsible for installing the prebuilt binaries needs the .env to tell it where to get them from. 

## What
optic-engine-native workspace needs .env  in the published npm artifact so it knows where to install binaries from. However, this should not affect production, which is set by default when the .env is not present

## Validation
* [ ] CI passes
* [ ] Verified in side-channel build

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -56,6 +56,7 @@ tasks:
           cp $FLAGS_FILE workspaces/cli-server/.env
           cp $FLAGS_FILE workspaces/cli-scripts/.env
           cp $FLAGS_FILE workspaces/optic-engine/.env
+          cp $FLAGS_FILE workspaces/optic-engine-native/.env
           cp $FLAGS_FILE workspaces/ui-v2/.env.production.local
         else
           echo "Please specify the FLAGS_FILE environment variable"


Taskfile.yml
```diff
@@ -56,6 +56,7 @@ tasks:
           cp $FLAGS_FILE workspaces/cli-server/.env
           cp $FLAGS_FILE workspaces/cli-scripts/.env
           cp $FLAGS_FILE workspaces/optic-engine/.env
+          cp $FLAGS_FILE workspaces/optic-engine-native/.env
```
devdoshi, line 59: note for posterity: we can probably remove optic-engine and cli-scripts now, since it doesn't appear they use it anymore. optic-engine's usage moved to optic-engine-native, and cli-scripts' usage moved to cli-server/optic-engine-native as part of getting rid of our scala code

JaapRood:
Thanks for tackling this one, totally missed it during the split up!

Taskfile.yml
notnmeyer: maybe this should get logged somewhere? no one is ever going to look back at comments in a merged PR for backlog items.

PR #875, opened 2021-06-10
Releases/10 0 5 release

LouManglass:
## Why
Release 10.0.5. Manual usability and extended tests run


niclim (approved):

niclim:
I think that we may need to include this PR (https://github.com/opticdev/optic/pull/872) as the build is failing on:
```
 |                 \      `optic-engine-native binary could not be found (pre-built download
 |                 for ${OS.arch()} ${OS.type()} or local build ${debugBinPath}). Consider
 |                 installing a pre-built version or building locally.`\r
```

LouManglass:
> I think that we may need to include this PR (#872) as the build is failing on:
> 
> ```
>  |                 \      `optic-engine-native binary could not be found (pre-built download
>  |                 for ${OS.arch()} ${OS.type()} or local build ${debugBinPath}). Consider
>  |                 installing a pre-built version or building locally.`\r
> ```

Yeah, seems right. This isn't urgent, I'll check in with Dev tomorrow so we're all up to speed. For now we have the tested changesets broken out in this branch so we have a good landmark.

devdoshi:
This is a different issue, regarding where the code is looking for the locally built binaries, as opposed to the prebuilt binaries. @JaapRood can you take a look since it seems to be flaky? I wouldn't block the release on it if it's been tested.

niclim:
Hmm if this is flaky, we should be able to rerun the tests... I kicked off a rerun and we'll see how it goes!

niclim:
This does in fact look flaky, this time mac and linux passed, but windows failed.

JaapRood:
As far as I understand, that error should ultimately have to be a blocker: this seems to be a problem of trying to use a locally built version of the Optic Engine CLI, where as a published version will always install the prebuilt binary that will have been published by that time.

@devdoshi if this is non blocking, I'll have to get back to this after my trip, as the flakiness and sometimes being a windows only issue makes it a real head scratcher that risk derailing the other active bets (got some pruning of commands to finish up so that stuff can ship).

My best guess, for anyone who wants to dive into this before I get the chance to, is that it's related to us [wiring `cargo build` straight into the `ws:build`](https://github.com/opticdev/optic/commit/8f8c033cb84d436f6540f8f895607489a21f213b#diff-b76767237a8edfde503f8deb18bf9fcc7fb997825597667edca1b94288f0787c). I couldn't remember then why it had been separated into it's own script and only called for _some_ of the tasks (back then still in sourceme.sh I believe), which I had traced as vestigial (preventing people from having the Rust toolchain installed while trying to make a build when the diff engine was still new). However, it does seem like _the_ place where `cargo build` might be getting called from not the root of the workspace. Not sure how it all connects, but where I'd start looking into further.

niclim:
@JaapRood I'm not sure I follow why this should be a blocker - it's something we should get fixed, but my understanding is that majority of the time, it should be pulling from prebuilt binaries (i.e. install downloads the prebuilt binary and then is subsequently used)? Is there anything that I'm missing, if not @LouManglass and I will release this and we'll look into why this is erroring as a follow up

devdoshi:
@JaapRood at least on my current copy of the code, `workspaces/optic-engine-native` `ws:build` runs the `build-binaries.sh` but I don't believe it ever ran `cargo build` from the root; 
```
node scripts/uninstall
cargo build
```
would suggest the cwd is/was in the `workspaces/optic-engine-native` dir. 

It looks kind of like a race condition, where perhaps the yarn workspace dependencies are not reflecting reality.

Since this only affects CI/local, it's not a blocker to me.

niclim:
Going to kick off this release, will do some quick sanity checks of pulling down the new optic-cli to verify we can download / install the prebuilt binaries

PR #876, opened 2021-06-10
Update spectacle projections to handle endpoint deletion

niclim:
## Why

Handle the remove commands in the spectacle projection. 

## What

I've taking the approach to add in deleted state (instead of the other failed attempts to remove these nodes from the graph https://github.com/opticdev/optic/pull/865). This is more correct since we still want spectacle to have the information that the nodes are there (changelog), however don't want these to show up in the UI.

To follow this PR:
- [ ] Handle changelog view for deleting an endpoint (currently this completely stops returning the deleted endpoint in all history) - all the information should be here to render the changelog information, we might just need to change our queries...

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


niclim:
Ok I think this looks good and works as we expect. @devdoshi what's the current state of snapshot testing? I think a spectacle snapshot (or even spectacle query snapshot tests) would be really good here to test the deletion functionality. 

Where would the best place to put these sorts of tests? I could implement the snapshots i added in here https://github.com/opticdev/optic/pull/865#discussion_r648603751 but they don't cover the spectacle part

workspaces/spectacle/src/index.ts
```diff
@@ -320,14 +320,15 @@ export async function makeSpectacle(opticContext: IOpticContext) {
...
+            .endpointsQueries.listNodesByType(endpoints.NodeType.Path, false)
+            .results
         );
       },
       requests: (parent: any, args: any, context: any, info: any) => {
         return Promise.resolve(
           context
             .spectacleContext()
-            .endpointsQueries.listNodesByType(endpoints.NodeType.Request)
+            .endpointsQueries.listNodesByType(endpoints.NodeType.Request, false)
```
devdoshi, line 331: I prefer to expose constants with meaningful names instead of stray booleans

workspaces/graph-lib/src/endpoints-graph/index.ts
```diff
@@ -310,36 +319,57 @@ export class GraphQueries {
...
+      : nodesByType.filter((node) => !isNodeDeleted(node));
+    return this.wrapList(type, filteredNodesByType);
   }
 
   //@TODO add singular find* variant
-  listIncomingNeighborsByType(id: NodeId, incomingNeighborType: NodeType) {
+  listIncomingNeighborsByType(
+    id: NodeId,
+    incomingNeighborType: NodeType,
+    includeDeleted: boolean = true
```
devdoshi, line 337: let's be consistent and use Removed instead of Deleted everywhere

devdoshi (approved):
looks good, just a few small changes. Regarding the additional snapshots, I'll have to do some work to make it easier to add cases, so don't worry about that for now.

workspaces/graph-lib/src/endpoints-graph/index.ts
JaapRood: Was here to mention the same thing ☺️

PR #877, opened 2021-06-10
Fix undelete from unstaging all endpoints

niclim:
## Why

When a user undelete one endpoint, they expect only one endpoint to be undeleted.

## What

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


JaapRood (approved):
Makes sense, doesn't it? 😅

PR #878, opened 2021-06-11
Handle endpoint removal in changelog

niclim:
## Why

When removing an endpoint, we want to show these in the changelog history

Also, wherever I am saying delete, I really mean remove (old habits die hard)

## What

Had to change a couple of things to get this to work
- Added in a RemovedIn edge when removing a request / response
- Moved the remove endpoint filtering logic (implemented in https://github.com/opticdev/optic/pull/876) to the UI (since we want to conditionally show deletions)
- Update the buildEndpointChanges function to handle a RemovedIn edge
- Update the changelog endpoint root page to handle a removed view
  - Show a banner if the endpoint was removed
  - If the endpoint should _not_ be visible in the changelog history view (i.e. would have been filtered) - redirect to the documentation root page (on the current changelog version)

I was also going to add in `getAssertedEndpoints` - but we need to consolidate our loading states further up the tree, and then we can use getAssertedEndpoints more effectively (this function would throw an error if state hasn't loaded and returns a non-nullable endpoint array) - there was only 2 places this would have been used so it wouldn't have been that useful.

To follow this PR:
- I am not sure if the gitbot will be able to handle deletion events. I'll have to follow up with this and see how the gitbot works (given that some of the required logic is on the UI side now) - although my guess is, yes it works

## Screenshots

Documentation page (current API spec)
Screen Shot 2021-06-11 at 3 12 09 PM

Changelog page with a deleted endpoint
Screen Shot 2021-06-11 at 3 12 13 PM

Changelog endpoint page
Screen Shot 2021-06-11 at 4 02 13 PM

## Validation
Before merging, we'll want to verify that the changelog still handles creation + update events correctly
- Creating a new endpoint
- Updating an endpoint (new response added, fields added, field changed)

Scenarios tested:
- Endpoints added (request added , resulting in an added endpoint, with requests, responses, bodies added)
- Responses added (results in an updated)
- Endpoint deleted (request deleted, resulting in a deleted endpoint)
-  Endpoint was deleted in batchCommit 3, but we are only viewing batchCommit 4 onwards - this should not show the deleted endpoint

### DIFF
--- a/workspaces/optic-engine/src/projections/spectacle/endpoints.rs
+++ b/workspaces/optic-engine/src/projections/spectacle/endpoints.rs
@@ -123,7 +123,10 @@ impl AggregateEvent<EndpointsProjection> for EndpointEvent {
         }
       }
       EndpointEvent::RequestRemoved(e) => {
-        projection.without_request(e.request_id);
+        projection.without_request(&e.request_id);
+        if let Some(c) = e.event_context {
+          projection.with_remove_history(&c.client_command_batch_id, &e.request_id);
+        }
       }
       EndpointEvent::ResponseAddedByPathAndMethod(e) => {
         projection.with_response(
@@ -138,7 +141,10 @@ impl AggregateEvent<EndpointsProjection> for EndpointEvent {
         }
       }
       EndpointEvent::ResponseRemoved(e) => {
-        projection.without_response(e.response_id);
+        projection.without_response(&e.response_id);
+        if let Some(c) = e.event_context {
+          projection.with_remove_history(&c.client_command_batch_id, &e.response_id);
+        }
       }
       EndpointEvent::RequestBodySet(e) => {
         //@GOTCHA: this doesn't invalidate previous RequestBodySet events for the same (request_id, http_content_type)
@@ -324,10 +330,10 @@ impl EndpointsProjection {
     self.domain_id_to_index.insert(request_id, node_index);
   }
   ////////////////////////////////////////////////////////////////////////////////////////////////////
-  pub fn without_request(&mut self, request_id: RequestId) {
+  pub fn without_request(&mut self, request_id: &RequestId) {
     let request_node_index = *self
       .domain_id_to_index
-      .get(&request_id)
+      .get(request_id)
       .expect("expected request_id to have a corresponding node");
 
     if let Some(Node::Request(request_node)) = self.graph.node_weight_mut(request_node_index) {
@@ -362,10 +368,10 @@ impl EndpointsProjection {
     self.domain_id_to_index.insert(response_id, node_index);
   }
   ////////////////////////////////////////////////////////////////////////////////////////////////////
-  pub fn without_response(&mut self, response_id: ResponseId) {
+  pub fn without_response(&mut self, response_id: &ResponseId) {
     let response_node_index = *self
       .domain_id_to_index
-      .get(&response_id)
+      .get(response_id)
       .expect("expected response_id to have a corresponding node");
 
     if let Some(Node::Response(response_node)) = self.graph.node_weight_mut(response_node_index) {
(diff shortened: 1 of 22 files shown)


workspaces/ui-v2/src/pages/diffs/ReviewDiffPages.tsx
```diff
@@ -35,6 +35,10 @@ export function DiffReviewPages(props: any) {
   const allRequestsAndResponsesOfBaseSpec = useAllRequestsAndResponses();
   useFetchEndpoints();
   const endpointsState = useAppSelector((state) => state.endpoints.results);
+  const filteredEndpoints = useMemo(
```
niclim, line 38: We need to useMemo this, since changes in identity here cause something in the SharedDiffStore to reset state. This should be less brittle when we move it to redux, just wanted to call it out since it's easy to make a mistake here and break the diff flow.

workspaces/spectacle/src/helpers.ts
```diff
@@ -84,13 +83,14 @@ export function buildEndpointChanges(
   shapeQueries: shapes.GraphQueries,
   sinceBatchCommitId?: string
 ): EndpointChanges {
+  // Sorted in reverse order - the latest change takes precedence
   const sortedBatchCommits = endpointQueries
     .listNodesByType(endpoints.NodeType.BatchCommit)
     .results.sort((a: any, b: any) => {
       return a.result.data.createdAt < b.result.data.createdAt ? 1 : -1;
-    });
+    }) as endpoints.BatchCommitNodeWrapper[];
```
niclim, line 91: I think there might be some "cool" things we can do with typescript generics such that `listNodesByType` returns the actual node type instead of the generic `NodeWrapper`. Haven't gotten to dig into this though

workspaces/optic-engine/src/projections/spectacle/endpoints.rs
```diff
@@ -123,7 +123,10 @@ impl AggregateEvent<EndpointsProjection> for EndpointEvent {
         }
       }
       EndpointEvent::RequestRemoved(e) => {
-        projection.without_request(e.request_id);
+        projection.without_request(&e.request_id);
```
niclim, line 126: We could also change the other consumers here to pass in references instead of `clone()` - where there's a with_creation_history, etc

workspaces/ui-v2/src/store/selectors/endpointSelectors.ts
```diff
@@ -0,0 +1,49 @@
...
+} from '<src>/hooks/useEndpointsChangelog';
+import { getEndpointId } from '<src>/utils';
+
+export const filterRemovedEndpoints = (endpoints: IEndpoint[]): IEndpoint[] =>
+  endpoints.filter((endpoint) => !endpoint.isRemoved);
+
+/**
+ * Filters removed endpoints not removed in changes and joins changelog information
+ */
+export const filterRemovedEndpointsForChangelogAndMapChanges = (
```
niclim, line 14: This is the main piece of work on the FE (the name is a bit of a mouthful, could split up the functions or rename - but I think it's easier to keep it in one function for it's intended use in changelog):
- An endpoint should show up if it is not removed
- An endpoint that _is_ removed should not show up with no changes (changelog pages should use this page)
- An endpoint that is removed and that deletion is in the range of batch commits _should_ show up with a `changes: 'removed'`
- An endpoint that is removed and that deletion is outside the range of batch commits _should not_ show up

workspaces/spectacle/src/helpers.ts
devdoshi: yep most of the graph queries should be able to produce more meaningful types, except the arbitrary neighbors/descendants ones, but I never got around to it

devdoshi (approved):

PR #879, opened 2021-06-14
usability-tests: updates

LouManglass:
## Why
Updated usability tests per release walkthrough, chats

### DIFF
--- a/Usability-tests.md
+++ b/Usability-tests.md
@@ -175,11 +175,11 @@ At this point, all observed traffic should be documented. The status of the proj
 - [ ] Run `api start` to bring up the project.
 - [ ] Send new traffic to the project: either modify an existing endpoint, or send traffic to a new endpoint. _Note_ this traffic will show up in the Optic review dashboard. Do not ignore/document/approve it.
 - [ ] Stop the process with `ctrl+c`.
-- [ ] Run `api status --pre-commit || [[ $? = 0 ]] && echo "API Status is clean" || echo "API Status is dirty"`: Expect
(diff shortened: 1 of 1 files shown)


Usability-tests.md
```diff
@@ -163,10 +175,13 @@ At this point, all observed traffic should be documented. The status of the proj
...
 - [ ] Send new traffic to the project: either modify an existing endpoint, or send traffic to a new endpoint. _Note_ this traffic will show up in the Optic review dashboard. Do not ignore/document/approve it.
 - [ ] Stop the process with `ctrl+c`.
-- [ ] Run `api status --pre-commit && echo $?`: Expect
+- [ ] Run `api status --pre-commit || echo $?`: Expect
 	- Diffs and undocumented URLs are reported as observed.
 	- `1` is echoed to the terminal (API status is "dirty" and requires review)
 - [ ] Run `api status --review` and document/accept observed differences.
+- [ ] Run `api status --pre-commit || echo $?`: Expect
+	- No diffs or undocumented URLs are reported.
+	- `0` is echoed to the terminal (API status is "clean")
```
notnmeyer, line 184: i understand why this is run a second time and we expect the output to change, but it seems awfully confusing as a manual testing step to run the same command but needing to interpret the result differently. rather than requiring interpretation of the exit status, why not just echo a meaningful message that says what the status actually represents? that both makes the command more meaningful and simplifies the documentation.

LouManglass: Sure, commit landing now.

niclim (approved):

PR #880, opened 2021-06-14
Add in upload sentry sourcemaps to production builds

niclim:
## Why

New sourcemaps will be uploaded on release - this will help us triage and fix any bugs affecting end users.

I was also trying to get sourcemaps to upload for older versions, but the bundle hashes are different so they don't work. It'll only work for newer releases of optic.

## What

Todo:
- [x] Inject the relevant auth token

## Validation
I've tested the upload flow and have seen the outputs of source maps getting uploaded (all the files we expect).

We can also create sidechannel releases (will require a version bump or something) to verify this, but I believe all the pieces should fit together properly.

### DIFF
--- a/workspaces/ui-v2/config/webpack.config.js
+++ b/workspaces/ui-v2/config/webpack.config.js
@@ -6,6 +6,7 @@ const webpack = require('webpack');
 const resolve = require('resolve');
 const PnpWebpackPlugin = require('pnp-webpack-plugin');
 const HtmlWebpackPlugin = require('html-webpack-plugin');
+const SentryWebpackPlugin = require('@sentry/webpack-plugin');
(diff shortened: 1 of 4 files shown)


workspaces/ui-v2/config/webpack.config.js
```diff
@@ -591,6 +595,24 @@ module.exports = function (webpackEnv) {
         )
       ),
       new OpticWasmWebpackPlugin(),
+      shouldUploadSentrySourcemaps &&
```
niclim, line 598: This should only run when we are actually distributing the published versions (since bundles are generated with hashes that can change)

notnmeyer (approved):
seems cromulent

PR #887, opened 2021-06-15
Revive debugOptic and update error page

niclim:
## Why

Revived debugOptic so it dumps relevant debugging information. Also updated the error page to be simpler and more streamlined (sending debug state is the most effective way for us to track errors if we've missed handling cases anywhere in the app).

## What

The debugOptic function now:
- Returns the spec events
- Dumps the SharedDiffContext state - this is effectively everything that has passed through the diffing process. Any errors in the rust process perhaps should be dumped out into a debug dump log? Although we have sentry errors sent (it would be nice to get captures / dump logs of interactions that caused the rust engine to die.

The updated error UI looks like
Screen Shot 2021-06-15 at 10 34 33 AM

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/ui-v2/src/components/DebugOpticComponent.tsx
+++ b/workspaces/ui-v2/src/components/DebugOpticComponent.tsx
@@ -1,0 +1,58 @@
+import { FC, useEffect } from 'react';
+
+import { IOpticSpecRepository } from '@useoptic/spectacle/build';
+
+declare global {
+  interface Window {
+    __GLOBAL_DIFF_DEBUG_FUNCTION__?: () => any;
+    debugOptic?: () => void;
+  }
+}
+
+const GLOBAL_DIFF_DEBUG_FUNCTION = '__GLOBAL_DIFF_DEBUG_FUNCTION__';
+
+// In the future, we could directly dump the redux store
+export const useGlobalDiffDebug = (fn: () => any) => {
+  useEffect(() => {
+    window[GLOBAL_DIFF_DEBUG_FUNCTION] = fn;
+    return () => {
+      delete window[GLOBAL_DIFF_DEBUG_FUNCTION];
+    };
+  }, [fn]);
+};
+
+const debugDump = (specService: IOpticSpecRepository) => {
+  return async function () {
+    const events = await specService.listEvents();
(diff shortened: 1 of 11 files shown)


workspaces/ui-v2/src/contexts/MetadataLoader.tsx
```diff
@@ -24,7 +24,7 @@ export const MetadataLoader: FC = ({ children }) => {
   }
   if (result.error) {
     console.error(result.error);
-    return <>error</>;
+    return <>error loading app metadata</>;
```
niclim, line 27: We'll want to start handling errors explicitly everywhere we expect them, as well as tighten our typescript / gql schemas so we aren't casting or using `any`. Giving a more specific error message here with better typings will help us more quickly identify where the errors are coming from (rather than the catch all)

Also, there's separate work to make the error states look nicer (i.e. not a single line of text) - but I think that'll come later

workspaces/ui-v2/src/components/DebugOpticComponent.tsx
```diff
@@ -0,0 +1,58 @@
...
+      delete window[GLOBAL_DIFF_DEBUG_FUNCTION];
+    };
+  }, [fn]);
+};
+
+const debugDump = (specService: IOpticSpecRepository) => {
+  return async function () {
+    const events = await specService.listEvents();
+    const diffDebugFunction = window[GLOBAL_DIFF_DEBUG_FUNCTION];
+    const diffState = diffDebugFunction ? diffDebugFunction() : false;
```
niclim, line 28: Thoughts on just dumping the diffState? I think it's probably the most useful since those are inputs to the UI - such that if the UI crashes we know why it crashed + what interaction. Is this enough information for us to debug the diff flow (i.e. everything in https://github.com/opticdev/optic/blob/develop/workspaces/ui-v2/src/pages/diffs/contexts/SharedDiffContext.tsx)?

devdoshi: I think it's a good starting point

devdoshi: I don't think it would give us the interaction

devdoshi (approved):
I'm looking at getting better logging on the initial-bodies and trail-values side

PR #891, opened 2021-06-16
The missing pieces of a Design First approach

smizell:
This adds the blog post on looking at the missing pieces to a design first approach.


website/blog/missing-pieces-of-design-first.md
```diff
@@ -0,0 +1,91 @@
...
+
+Unfortunately, the API consumer is left out of workflows today. The burden is on them to keep up with API changes, and many times they find out about change when it's too late.
+
+## Putting these ideas to practice
+
+If you're trying to garner adoption in your company around a Design First process, consider these other areas that are an important part of the approach. It's easy to pick Design First as a practice without considering the full experience around how people build APIs.
+
+You can find ways to help people get more visibility into their API changes, collaborate around the changes, and ensure each change meets the quality standards of the company—all without requiring teams to abandon their existing workflows. You can help teams transition to a better way to build APIs in their own context.
+
+In a future post, we'll look at the ways Optic tries to help in all these areas to encourage a collaborative design and development experience for the entire team.
```
devdoshi, line 91: It might fit in a few different places, but perhaps now is not the right time to talk about it: there's also the additional layer of understanding how each consumer uses the API, aka Optic's API analytics, to go deeper than just conformance.

devdoshi (approved):

website/blog/missing-pieces-of-design-first.md
smizell: Would you see this as another section in the post? I've tried to fit this into the existing ones, and while it's related, it doesn't feel like it fits directly. It's almost like there's an "observability" component that lets teams analyze and understand the way people use the API to help inform the design.

I think this is a good idea, and it's something that isn't in the Design First discussion—I even left it out here myself :)

Do you think it's worth adding? If so, how do you feel about the length of the post since this will add a few paragraphs?

devdoshi: I think it can come later :) definitely worth adding but can be part of the optic or another article

PR #892, opened 2021-06-16
Fix endpoint removal when endpoint is the last endpoint in the spec

niclim:
## Why

When removing an endpoint and that endpoint is the last one in the node graph - PrunePathsCommand removes all nodes, and also marks the root node for removal - which cannot be done and rust panics!

## What
We need to preserve the root node (since we can only have 1 unique id in the graph, and root is always a deterministic id and we are orphaning nodes). We also "can't" remove the root node since remove == orphan

Also updated the PrunePathsCommand to be appended to after every endpoint deletion (this is more "correct" and also does not add a PrunePathsCommand every time we create a contribution.

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


devdoshi (approved):
oopsy :)

PR #893, opened 2021-06-16
Fix/common start issues

niclim:
## Why

This fixes two bugs:
- we have two endpoints `/api/list` and `/api/ladders` - the longest common substring function returns `/api/l`which is incorrect
- When we compute the common start for endpoints, we include all removed endpoints - which doesn't match what we show on the UI (depending on the changelog diff, we may or may not show removed endpoints)

## What
- Updated function to longestCommonPath (uses `split by /` instead of a string match)
- Moved group endpoints to the render methods of changelog and documentation - this is so we can pass in and derive the groups (after the filter logic relevant to the UI has been applied)

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


workspaces/ui-v2/src/pages/docs/DocumentationRootPage/DocumentationRootPage.tsx
```diff
@@ -42,10 +44,21 @@ export function DocumentationRootPage() {
...
-  const tocKeys = Object.keys(grouped).sort();
+  const groupedEndpoints = useMemo(() => {
+    const commonStart = findLongestCommonPath(
+      filteredEndpoints.map((endpoint) => endpoint.fullPath)
+    );
+    const endpointsWithGroups = filteredEndpoints.map((endpoint) => ({
+      ...endpoint,
+      // If there is only one endpoint, split['/'][1] returns undefined since
+      // commonStart.length === endpoint.fullPath.length
+      group: endpoint.fullPath.slice(commonStart.length).split('/')[1] || '',
```
devdoshi, line 55: spectacle exposes the pathComponents as an array as well as a string. The array is for use cases like this. What is a human-readable way to describe the desired sorting and grouping of this ui?

niclim: `group` is a string which is where the endpoint should be grouped in the documentation page.

E.g.
`/api/list`
`/api/todos`
`/api/todos/:todoId`

The common start is `/api` and each would be grouped under the next path parameter (in this case, there would be two groupings - todos and list)

devdoshi: Cool we can revisit later but this is definitely less broken than before 😓 please use the spectacle-provided path components if possible

devdoshi (approved):

workspaces/ui-v2/src/pages/docs/DocumentationRootPage/DocumentationRootPage.tsx
niclim: Oh I just saw this - I'll follow this PR up with a separate fix (changing the function signature)

PR #894, opened 2021-06-16
Add support for merging base OpenAPI documents into the generated one

smizell:
## Why

This allow people to add a default OpenAPI document that will get merged into the generated one so they can automate the entire process. The generated OpenAPI document looks like:

```yaml
openapi: 3.0.3
info:
 title: Optic Generated OpenAPI
 version: 1.0.0
# ...
```

We want to allow the user to override the title and add any other field.

## What

This changes the `local-cli` command for `generate:oas` to include a `baseDocument` flag.

```sh
api generate:oas --baseDocument=base-openapi.yml
```

where `base-openapi.yml` is:

```yaml
info:
  title: My custom title
```

This adds two dependencies: js-yaml and deepmerge.


workspaces/local-cli/src/commands/generate/oas.ts
```diff
@@ -75,7 +89,7 @@ export async function emit(
   if (shouldOutputYaml) {
     const outputFile = path.join(outputPath, 'openapi.yaml');
     //@ts-ignore
-    await fs.writeFile(outputFile, yaml.safeDump(parsedOas, { indent: 1 }));
+    await fs.writeFile(outputFile, yaml.dump(parsedOas, { indent: 1 }));
```
smizell, line 92: The version of `js-yaml` I installed makes `dump` safe by default now.

acunniffe:
Love this — I have a user in product board who built something similar. I’ll reach out and see if this works for them!

devdoshi (approved):

PR #896, opened 2021-06-17
Fix rust snapshots post rust release

niclim:
## Why

It looks like the latest rust version that got released today changed snapshot behavior https://blog.rust-lang.org/2021/06/17/Rust-1.53.0.html - see failing run here https://github.com/opticdev/optic/runs/2851483321

Specifically escaping `'` - I don't see this in the patch notes, but I imagine that this would not be a user facing change?

## What

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -72,6 +72,12 @@ jobs:
       uses: actions/setup-node@56899e050abffc08c2b3b61f3ec6a79a9dc3223d # v1.4.4
       with:
         node-version: "14.x"
+    - name: 'Rust toolchain'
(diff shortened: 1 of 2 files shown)


niclim:
It looks like the `Run Tests / node_packages` is pulling an older version of rust, which is defaulting to the older behavior...?

.github/workflows/run_tests.yml
```diff
@@ -72,6 +72,12 @@ jobs:
       uses: actions/setup-node@56899e050abffc08c2b3b61f3ec6a79a9dc3223d # v1.4.4
       with:
         node-version: "14.x"
+    - name: 'Rust toolchain'
```
niclim, line 75: The `rust_binaries` test use this toolchain, which pulls the latest stable version of rust. Otherwise we rely on the assumption that the `ubuntu-latest` container includes the same rust version as the toolchain (which is not always the case)

PR #897, opened 2021-06-17
Feature: Spec sharing

jshearer:
## Why
Users want to be able to share their specs. Now they can!

<details>
  <summary>UI Flow</summary>
  
Screen Shot 2021-06-17 at 4 05 23 PM

Screen Shot 2021-06-17 at 3 56 43 PM
Screen Shot 2021-06-17 at 3 56 50 PM
Screen Shot 2021-06-17 at 3 57 38 PM
Screen Shot 2021-06-17 at 3 57 50 PM

  ### 

</details>

## What
@niclim introduced the ability to fetch a list of all events from the local cli spec service, which I brought in and exposed through the `SpecRepositoryContext`. I changed the `DebugOpticComponent` to use this context, instead of requiring it be passed in as a prop.

Sharing is gated by the `config.sharing.enabled` flag in the `AppConfiguration` context, and only the local spec viewer is enabled, since that's only place we want sharing right now, and the only place it makes sense.

The majority of the rest of the PR should be UI stuff.

### TODO:
- [ ] Add `intent` to `AddSpecCommand`/`SpecAddedEvent` to allow tracking the intent of these specs that users are going to upload. I'm thinking something like 
```rust
pub enum PublicSpecUploadIntent {
    #[serde(default)]
    Unknown,
    GitBot,
    ShareWithCustomer,
    ShareWithTeam
}
```


workspaces/ui-v2/src/entry-points/local/local-cli.tsx
```diff
@@ -54,9 +56,27 @@ const appConfig: OpticAppConfig = {
...
+    api: {
+      domain:
+        (window.location.hostname.indexOf('useoptic.com') >= 0
+          ? process.env.REACT_APP_PROD_API_BASE
+          : process.env.REACT_APP_STAGING_API_BASE) || 'https://api.o3c.info',
+    },
+    sharing: {
+      enabled: true,
+      specViewerDomain:
+        (window.location.hostname.indexOf('useoptic.com') >= 0
```
niclim, line 68: Wouldn't this always return false? Since the local cli will be run against `localhost` which will always run in staging?

In what cases would we want to upload to prod and what cases would we want to upload to staging?

workspaces/ui-v2/src/entry-points/cloud/cloud-viewer.tsx
```diff
@@ -52,6 +53,13 @@ const appConfig: OpticAppConfig = {
     documentation: {
       allowDescriptionEditing: false,
     },
+    api: {
```
niclim, line 56: What api are we referring to here? Is this the cli api, oauth api or some other api?

workspaces/ui-v2/src/contexts/SpecRepositoryContext.tsx
```diff
@@ -0,0 +1,28 @@
+import React, { FC } from 'react';
+import { useContext } from 'react';
+import { IOpticSpecRepository } from '@useoptic/spectacle';
+
+export function useSpecRepository(): {
+  specRepo: IOpticSpecRepository;
+} {
+  const specRepo = useContext(SpecRepositoryContext)!;
```
niclim, line 8: Given we are creating the context with a default of null - null will be returned if there isn't a parent provider. Can we do a `if !specRepo -> throw error` - similar to https://github.com/opticdev/optic/blob/develop/workspaces/ui-v2/src/contexts/spectacle-provider.tsx#L23-L25

Also we could just return `specRepo` straight instead of wrapping it in an object (unlikely we'd want to return anything other than the context when the hook only reads from a single context value

workspaces/ui-v2/src/components/sharing/ShareButton.tsx
```diff
@@ -0,0 +1,139 @@
...
+    if (isAuthenticated) {
+      const token = await getAccessTokenSilently();
+
+      const response = await fetch(`${baseDomain}/api/person`, {
+        headers: { Authorization: `Bearer ${token}` },
+      });
+      const resp_data = await response.json();
+      return resp_data.id;
+    }
+  }, [getAccessTokenSilently, isAuthenticated, baseDomain]);
```
niclim, line 49: Can we put this state into redux? It might be good to put this into the `metadata` slice (we are slowly moving everything into redux - excluding non-serializable things like SpecRepo / spectacle, since it's likely we'll want to use the person's id / are they logged in, in other places in the app)

When / how does a user log in? Can they log in through the UI - if so, the relevant state will need to change if a user can go from not-logged in -> logged in

workspaces/ui-v2/src/components/sharing/ShareModal.tsx
```diff
@@ -0,0 +1,223 @@
...
+function generateBadgeUrl({ endpointCount }: { endpointCount: number }) {
+  const params = {
+    label: 'API Docs',
+    message: `${endpointCount} Endpoints`,
+    style: 'flat',
+    color: 'rgb(43,123,209)',
+    // Took https://github.com/opticdev/optic/blob/develop/website/static/img/logo-bare.svg
+    // Ran it through https://jakearchibald.github.io/svgomg/
+    // base64 encoded it (on macos, you can do: `pbpaste | base64 | pbcopy` to quickly base64 encode whatever is in your paste buffer)
+    // prepend with `data:image/svg+xml;base64,`
```
niclim, line 45: Perhaps this could go in `src/constants`?

workspaces/ui-v2/src/components/sharing/ShareModal.tsx
```diff
@@ -0,0 +1,223 @@
...
+  };
+
+  const queryParams = Object.entries(params)
+    .map(([k, v]) => `${k}=${encodeURIComponent(v)}`)
+    .join('&');
+
```
niclim, line 53: Perhaps this is a good `src/utils` function?

workspaces/ui-v2/src/components/sharing/ShareButton.tsx
niclim: Oh I see there's a login flow, we could load the initial user's state (in the inital metadata fetch) - and then as a callback for loading a user, it could hydrate the redux store)

workspaces/ui-v2/src/components/sharing/ShareModal.tsx
```diff
@@ -0,0 +1,223 @@
...
+export const ShareModal: React.FC<{
+  share: (intent: ShareTarget) => Promise<string>;
+}> = (props) => {
+  const styles = useStyles(props);
+
+  const { isAuthenticated, isLoading, loginWithPopup } = useAuth0();
+
+  const endpointCount = useAppSelector(
+    (state) => state.endpoints.results?.data!.length || 0
```
niclim, line 77: If we are casting here anyway - it's probably more correct to write `data?.`. It also looks like from the type signature, results will always be non-null - it's only data that is nullable

workspaces/ui-v2/src/components/sharing/ShareModal.tsx
```diff
@@ -0,0 +1,223 @@
...
+  );
+
+  const [shareType, setShareType] = useState<ShareTarget | null>(null);
+
+  let inputRef = useRef<HTMLInputElement>();
```
niclim, line 82: We probably don't need to use an input tag here - we already have the shareUrl as state, so we can use that instead of overriding the user's focus + select state (probably not a big deal in this specific example) - we could build a copyToClipboard util function here in place of this

workspaces/ui-v2/src/components/sharing/ShareModal.tsx
```diff
@@ -0,0 +1,223 @@
...
+
+  const copy = useCallback(() => {
+    const input = inputRef.current;
+    if (input) {
+      input.focus();
+      input.select();
+      document.execCommand('copy');
+      setCopied(true);
+    }
+  }, [inputRef, setCopied]);
```
niclim, line 100: FYI you don't need to specify `refs` and `setState` fns in react dependency arrays

workspaces/ui-v2/src/components/sharing/ShareModal.tsx
```diff
@@ -0,0 +1,223 @@
...
+
+  let inputRef = useRef<HTMLInputElement>();
+  const [copied, setCopied] = useState(false);
+
+  useEffect(() => {
+    if (copied) {
+      let timeout = setTimeout(() => setCopied(false), 7500);
+      return () => clearTimeout(timeout);
+    }
+  }, [copied, setCopied]);
```
niclim, line 90: So this works, but it's probably not ideal - what might be better here is to add this setTimeout to the end of the `copy` function (otherwise you have to handle things like `if copied` and `return () => clearTimeout` which can be easy to miss things and cause infinite loops!

workspaces/ui-v2/src/components/sharing/ShareModal.tsx
```diff
@@ -0,0 +1,223 @@
...
+    if (input) {
+      input.focus();
+      input.select();
+      document.execCommand('copy');
+      setCopied(true);
+    }
+  }, [inputRef, setCopied]);
+
+  // Call share when authenticated and you set a shareType
+  const shareUrl = useAsyncMemo(async () => {
```
niclim, line 103: Sort of minor, but since this is an async state - we should start trying to add in loading and error states when we get into these flows - since it's possible the share callback is loading or errors (and our UI won't surface that inforamtion)

workspaces/ui-v2/src/components/sharing/ShareModal.tsx
```diff
@@ -0,0 +1,223 @@
...
+}: {
+  endpointCount: number;
+  shareUrl: string;
+}) {
+  return `[![API Docs: ${endpointCount} Endpoints](${generateBadgeUrl({
+    endpointCount,
+  })})](${shareUrl})`;
+}
+
+export const ShareModal: React.FC<{
```
niclim, line 69: Sort of a high level comment - this flow seems fine, but what do you think about:
- Having a log in state up front (i.e.  you need to be logged in)
- Changing the buttons for share type to a radio button + condensing that to a single flow (i.e. showing the radio buttons and urls at the bottom up front) - allows users to change and experiment rather than having to make that decision up front. This probably means we'll want to have some async loading + error state when they change the share type (we can also cache props.share calls) though

niclim (commented):
Great stuff! Couple of comments

niclim:
Also, the rust_binaries test is failing because of https://github.com/opticdev/optic/pull/896

The node_packages looks like it's failing because of linting errors

workspaces/ui-v2/src/entry-points/local/local-cli.tsx
jshearer: Oh wait derp thank you, you're totally right. This should probably actually select whether the frontend is running in dev mode or in a release build

niclim: Do you think it's better to just have a single env variable that gets injected per build? or two separate ones that look at release status

workspaces/ui-v2/src/entry-points/cloud/cloud-viewer.tsx
jshearer: Good point. The backend (api.useoptic.com). Will clarify. `backendApi`?

niclim: Perhaps opticApi? We can always change it later fairly easily as since it's only referenced in code

workspaces/ui-v2/src/contexts/SpecRepositoryContext.tsx
jshearer: Done, and done

workspaces/ui-v2/src/components/sharing/ShareModal.tsx
jshearer: done

workspaces/ui-v2/src/components/sharing/ShareModal.tsx
jshearer: done

workspaces/ui-v2/src/components/sharing/ShareModal.tsx
jshearer: done

workspaces/ui-v2/src/components/sharing/ShareModal.tsx
jshearer: huh, TIL

workspaces/ui-v2/src/components/sharing/ShareModal.tsx
jshearer: Happy to move it there, like this?

```typescript
  const copy = useCallback(() => {
    const input = inputRef.current;
    if (input) {
      input.focus();
      input.select();
      document.execCommand('copy');
      setCopied(true);
      setTimeout(() => setCopied(false), 7500);
    }
  }, []);
```

I do wonder though, don't I still need to `return () => clearTimeout` here?

workspaces/ui-v2/src/components/sharing/ShareModal.tsx
jshearer: I'd say I want input from @acunniffe here -- the flow in this PR came from a discussion with him

workspaces/ui-v2/src/components/sharing/ShareModal.tsx
jshearer: Yeaaaah.. the annoying thing stopping me from adding all of this to redux is that `useAuth0()` hook, which maintains its own state in a pretty convenient little package. I also think the desire here is to not force everyone to log in before using the UI -- logging in is only required for sharing.

Eventually (soon) I'm going to start focusing on the OpticHub UI, which likely _will_ require logging in to use (except for explicitly public parts, like the spec viewer that'll move over there)

workspaces/ui-v2/src/components/sharing/ShareModal.tsx
jshearer: We talked in a DM about this, I think tentatively came to the conclusion that changing focus to select the text that you clicked on a button to copy seems okay/expected behavior. Do we still want to do this "invisibly"? Happy to do either, I don't really have a strong opinion -- my weak opinion is that this works fine, and doesn't seem to be that jarring

jshearer:
So I think there are two points of discussion left:

### The actual UI flow
@niclim suggested having login first, and simplifying the 3/4 step flow to 2/3 steps: 
* Click "Share", UI presents radio/toggle buttons for team/customer
* Select one 
* Optionally log in
* Share links appear below, so you can change/play with the toggle to see what the difference is. 

I kinda like this. @acunniffe thoughts?


### State management
Optimally, all of the application state would be stored in/managed by redux. This PR introduces a bunch of stateful logic: 
* Login status (`isAuthenticated`, `isLoading`, `async getAccessTokenSilently`)
* User metadata like ID that can only be loaded once authenticated successfully
* Share URL fetched/uploaded to asynchronously

Usually I'd say absolutely, let's redux all of it. In this case, there are two things stopping me:
* Auth0 provides a convenient `useAuth0()` hook to manage all of the auth0 state in a nice package. Logging in, logging out, fetching/caching/refreshing auth tokens, etc are all taken care of. Integrating this with redux doesn't seem like a trivial amount of work, though it is certainly not impossible or even all that complicated.
* Fairly soon I'm going to start working on the account (/org/shared spec/SaaS in general) dashboard: OpticHub. In that context, I'd say putting in the work to handle auth as part of the state layer (i.e redux) makes way more sense: auto-fetching user metadata on page load, exposing prepared ways to call APIs such that you don't need to explicitly get the user's auth token, manage loading states in redux with e.g `createAsyncThunk`, authenticated routes, etc etc. 

  Given that OpticHub is going to pull in a lot of components from `ui-v2`, I'd say that once we write that infrastructure for OpticHub, we should teach `ui-v2` how to use it/interact with it in a shared way.

workspaces/ui-v2/src/components/sharing/ShareModal.tsx
niclim: Bleh, ok I think what you have is fine - I was trying to avoid changed useEffects - but in this case if we have a setTimeout with setState, if this unmounts we have a memoryleak (it's likely not a big impact, apart from having a react warning.

This is probably _fine_ to keep as is

jshearer: What's the memory leak? `return () => clearTimeout(timeout);` cancells the `setTimeout` on unmount, righT?

acunniffe:
> Click "Share", UI presents radio/toggle buttons for team/customer

Only reason to do this first, is that the login button will be a big churner? How big, I'm not sure, we should measure that. 

Also we should measure which radio people click re: Team or Consumer. That's really the only reason to do it first IMO. 

I think sharing should require login. It's a reasonable line to draw and it also makes the service safer

niclim:
> Only reason to do this first, is that the login button will be a big churner? How big, I'm not sure, we should measure that.

You could also show this on a single UI flow. Team + consumer radio buttons, with a "you must be logged on to share" block early - rather than having a multi stage click. 

Re: state management
Keeping it now as is is fine since it's just used in one place. We probably don't even need to move it into redux, just using it as a selector + if we need to derive states from multiple places since it's using a provider and the relevant data is cached. It just "feels nicer" if everything is in redux, which isn't really a good reason to do things 🙃

workspaces/ui-v2/src/components/sharing/ShareModal.tsx
niclim: Yeah I think that's fine, it's not a huge deal. These interactions are minor - at some point we'll have some UI / UX patterns + consistency where we'll want to have the same behavior across the app

workspaces/ui-v2/src/components/sharing/ShareButton.tsx
```diff
@@ -0,0 +1,139 @@
+import React, { useCallback, useMemo, useState } from 'react';
```
niclim, line 1: From lint:
```
 | src/components/sharing/ShareButton.tsx
 |   Line 1:30:   'useMemo' is defined but never used          @typescript-eslint/no-unused-vars
 |   Line 32:52:  'logout' is assigned a value but never used  @typescript-eslint/no-unused-vars
```

niclim (approved):

workspaces/ui-v2/src/components/sharing/ShareButton.tsx
```diff
@@ -0,0 +1,139 @@
...
+import ShareIcon from '@material-ui/icons/Share';
+import { ShareModal } from './ShareModal';
+import { useAuth0 } from '@auth0/auth0-react';
+import { useSpecRepository } from '<src>/contexts/SpecRepositoryContext';
+import { useAppConfig } from '<src>/contexts/config/AppConfiguration';
+import { useAsyncMemo } from 'use-async-memo';
+
+export enum ShareTarget {
+  TEAM,
+  CUSTOMER,
```
devdoshi, line 19: I believe this is meant to be Consumer? As in, if I am sharing my API I either want to discuss/share it with my team or with external consumers? But there's overlap in the venn diagram between teammates and consumers. So perhaps this can be revisited.

devdoshi:
as long as we track the share intent starting (so we can measure against completions), their selections, and completions, then we should have the data we need. I'm not sure about the Team vs Customer dichotomy, so best to have Other too? I think it makes sense to require login. What are the related features to this? OAS download?

workspaces/ui-v2/src/components/sharing/ShareButton.tsx
jshearer: Fixed

acunniffe:
I’m not sure if there was a misspeak or a mistype but I expected the categories to be 

Team (my team) or External Consumer

Issue #881, opened 2021-06-14
Combining targetUrl and useTask makes the process hang indefinitely


aledalgrande:
Hi, we run our test server inside the test runner, so it can share the same connection pool and see the data created during test transactions (they are rolled back after each test so never committed).

The issue is that, while trying to setup Optic, I realized that this setup makes the process hang:
```
name: "my project"
tasks:
  start:
    command: "rails s --port $PORT"
    inboundUrl: http://localhost:3000
  docs:
    targetUrl: http://localhost:7368
    inboundUrl: http://localhost:5000
  specs:
    useTask: docs
    command: bundle exec rspec spec/acceptance_optic
```

The tests hit `http://localhost:5000` while `http://localhost:7368` is the Rails server started by the test runner. When the test passes and the RSpec process exits, I have to stop the command with Ctrl-C, which obviously isn't ideal if I want to run on CI.

Do you have any workaround for this?

For reference, if I set `useTask: start` it works normally (but then my tests don't pass).

LouManglass:
Howdy @aledalgrande ,

Normally when running in what we would call [dependent task mode](https://www.useoptic.com/docs/using/advanced-configuration#dependent-tasks) like your `specs` task, we'd expect that command process to exit gracefully which would bring down the Optic tasks. I understand that's not happening here, and your specific flow isn't one I've tried before. 

Is there any chance that this is an open-source project I can look at locally, or failing that, would you have time to [have a quick chat](https://calendly.com/d/pgv6-9tp7/30-minute-meeting) to show off your setup? I'd like to learn more. I'm sure we can find a workaround with a little digging.

aledalgrande:
Thanks will try to setup a small repo to test out. If I run `api run docs` and then `bundle exec rspec spec/acceptance_optic` in a separate terminal it works without any issue.

aledalgrande:
Here's a minimal repro: https://github.com/Latermedia/optic_task_issue

If you run `rspec` + `api run docs` in two separate terminals the test suite will pass and exit normally. But with `api run specs` it will get stuck.

LouManglass:
Great, thanks! I'll dig into this a bit and see what our next steps are.

notnmeyer:
@aledalgrande I think this is a bug related to `useTask`. Using the your example repo, try this optic.yml as a workaround,

```
name: "my project"
tasks:
  start:
    command: "rails s --port $PORT"
    inboundUrl: http://localhost:3000
  specs:
[...]

PR #898, opened 2021-06-17
Ingest capture method

acunniffe:
## Why
We expect Optic middlewares and alternative capture methods to continue to grow in scope, esp as more teams push to run tests with Optic. 

## What

This PR introduces a new command `api exec "<command>` 

You could run `api exec "bundle exec specs"` or `api exec "yarn run test"` 

This will spin up a capture-saver on a local server. The server is guaranteed to be up and running before your command is run.  In the environment will be a variable called "OPTIC_LOGGING_URL". This is the base url of the capture service and will be a random port on `localhost`. Could one get fancy and start making it a hosted URL 🙉  shh?

Each`IngestTrafficService` will host routes that differ depending on the format of data that accept. Right now we only implemented `/ecs` to support the SDK work. You could imagine it being `/morgan` or `/har`, `/aws-lambda2` .... etc. It's up to each middleware to append the proper route based on the style of traffic they want to dump. 

If you `POST` arrays of traffic to that endpoint, you'll end up with samples getting saved. Run with `DEBUG=optic*` to see for yourself:

```
  optic-debug saw sample {
  http: {
    response: { body: [Object], status_code: 302, headers: [Object] },
    request: { body: [Object], method: 'PUT', headers: [Object] },
    version: '1.1'
  },
  url: {
    full: 'http://localhost:3000/status/302',
    path: '/status/302',
    domain: 'localhost'
  },
  client: { address: '::1', ip: '::1', port: 60876 },
  user_agent: { original: 'PostmanRuntime/7.28.0' }
} +2s
```

The current implementation dumps each batch to the file system of the cwd -- this will help us collect a bunch of snapshot test inputs. This is WIP work, we need @ukmadlz to go run an SDK this way, and copy over all the `x.json` files that get littered in his repo so we can finish writing proper tests for the ECS to interaction code pathway. 

---

You can also add tasks that only have a command field -- these will get rewritten to `api exec <>` under the hood. Took 3 mins to make it do this, maybe we don't want to. Curious to hear @LouManglass and @devdoshi thoughts on this. 

```
  start-tests:
    command: yarn run tests
```

Thanks to the awesome work of @jshearer we're able to just use coverage with `api exec <>`...no headaches. Pretty awesome!

## Validation
* [x] Team buy in
* [x] CI passes
* [x] Verified in staging
* etc...


devdoshi:
Looks good - do we have any stories/reqs for the scenario where we aren't managing a command but just receiving data until told to stop?

LouManglass:
Phwoa! Good stuff.

I'm in favor of a more descriptive use of parameters. Our current model of command-driven task vs. proxy-driven is a bit leaky, and being able to mix and match the parameters you need to achieve a goal makes sense. If you just need a command (because the project is already plumbed to send data to the logging service), awesome. If you still need the proxy but your command addresses everything else, give [proxy and command parameters](https://github.com/opticdev/optic/issues/881#issuecomment-862478470). 

I think this will also come in to play in a broader discussion around how we organize documentation, but that's definitely not a blocker, and we can continue that conversation in the proper place/time. 

Nice work! IMO ship with command only field and we can document the capability properly for users on launch.

acunniffe:
@devdoshi - I did not spend as much time with the manual start/stop flows. The `IngestTrafficService` could definitely spin up on its own. All it needs is a capture saver. 

Managing the commands still feels useful, so we can dump out the set of diffs / traffic for users. 

The story where you might want to run that out-of-band definitely could exist, maybe running multiple test suites?

Do you have some I'm missing? I may not see what you see. Wouldn't take long to add, but don't see it as blocking either. 

--

I was imagining in the case replacing the current proxy with running scoop or slurp locally, or in CI -- those would just run as normal processes, and point at the ingest service. Gives us a lot of flexibility.

devdoshi (approved):
My comments aren't blocking this, this is a good step forward

ukmadlz:
Here are the captures from the Express middleware https://drive.google.com/file/d/1Cr8wIS_uVpCRGPAGYgiam96V0IbiSziU/view?usp=sharing

PR #899, opened 2021-06-17
Implement changelog history view 

niclim:
## Why

As part of the revert bet - we want to add a revert to commit button. This could go in two places:
- On each changelog diff page (i.e. revert to this commit)
- On a new changelog history page

@JaapRood and I discussed and thought that having a changelog history page as a first class entry point was worth it - and given that a large part of Optic's value is in viewing temporal changes - it'd make sense to have a history endpoint view.

Right now, the changelog history page doesn't really provide much extra information over the dropdown (whcih is why I went back and forwards on it a lot whether it's worth while) - but I think in the future there's a lot richer information we can add to this page:
- Summary of changes on this page - e.g. # endpoints created, updated, changed or commit author, links out to discussions / proposals around these changes, etc
- Ways to manage changes (i.e. merging, conflicts, etc)
- Approvals, etc
- Changes that had breaking changes, etc

## What

Everything is behind a feature flag `REACT_APP_FF_SHOW_REVERT_COMMIT`. 

Given this is a new page, I attempted to add in some more design structure to how we want to build pages in the future - with some CSS style helpers, a Page component and other things.

I want to get @JaapRood's feedback when he's back but since this is all behind a feature flag, we could merge this and then make changes to things. My questions are below:
- Thoughts on the useSpacingStyles (the more I look at it, the less I like it) - I think I actually prefer predefined global css classes for atomic spacing (mt-l - marginTop-large, p-m - padding-medium)
- Thoughts on having css layout styles in line (and other css styling like cursor: pointer, or textAlign: right) - we _could_ run everything through material ui's makeStyles function, but I'm not a huge fan of it - we need to write more, to do the same thing (we aren't saving any time, we effectively write the same thing - with extra hooks + the same object shape) - unless we are using `theme` variables - in certain cases this is useful* , in other cases like spacing, it's an extra layer of indirection. we also need to think of names (minor - but what do you name something wrapping a button? what happens if you need two divs to wrap a button with styling)
  - *this is useful when we have things like the `Page` components with stylable props - e.g. `padded: true | false`

Screenshots:
The styling could use some work - we'll definitely want style this better before we release this
Screen Shot 2021-06-17 at 3 20 18 PM

Updated the dropdown to only show the last 5 commits. The assumption is that users won't want to often change back to something a long time ago - in which case they would want a more comprehensive view with more information (the changelog history page, currently lacking extra information)
Screen Shot 2021-06-17 at 3 20 26 PM

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


workspaces/ui-v2/src/App.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import CssBaseline from '@material-ui/core/CssBaseline';
-import { appTheme } from './constants/theme';
+import { appTheme } from './styles';
```
niclim, line 3: Moved into `<src>/styles`

workspaces/ui-v2/src/components/Page/Page.tsx
```diff
@@ -0,0 +1,86 @@
+import React, { FC, useEffect } from 'react';
```
niclim, line 1: @JaapRood Revived your page component - we currently have a lot of usage with PageLayout, but I think this should take over those adhoc pagelayout components, etc

workspaces/ui-v2/src/components/loaders/index.ts
```diff
@@ -1,3 +1,3 @@
-export * from './FullPageLoader';
-export * from './Loading';
```
niclim, line 2: Removed some loading components where they were just reexports - There's more work to do to consolidate and standardize our loading + error states

workspaces/ui-v2/src/components/navigation/TopNavigation.tsx
```diff
@@ -3,61 +3,79 @@ import { makeStyles } from '@material-ui/core/styles';
...
+                  />
+                )}
+              </div>
+              <div className={classes.spacer} />
+              <div>{AccessoryNavigation && <AccessoryNavigation />}</div>
+            </Toolbar>
+          </AppBar>
+        </Container>
+      </div>
+      <div className={classes.spacing}></div>
```
niclim, line 78: Added a spacing component such that the TopNavigation automatically spaces everything below the `position:fixed` top bar

workspaces/ui-v2/src/styles/stylingHooks.ts
```diff
@@ -0,0 +1,49 @@
+import { darken, makeStyles, Theme } from '@material-ui/core';
+
+type SpacingInputs = {
+  size?: number;
+};
+
+export const useSpacingStyles = makeStyles<Theme, SpacingInputs>((theme) => ({
```
niclim, line 7: I added these useSpacingStyles but I don't think I like this. I would think the general usecases here are that we'd want margin + padding to be _different_ (i.e. margin top and bottom with size 2 and left and right with size 1, with other padding stuff).

I actually don't like the material ui makeStyles, especially for ad-hoc usages with styling. My preference is to use predefined css classes - e.g. `mt-l mb-l` which correspond to `marginTop: large -> 24px` etc where we have different preset spacing units.

workspaces/ui-v2/src/styles/stylingHooks.ts
```diff
@@ -0,0 +1,49 @@
...
+  }),
+  marginRight: ({ size = 1 }) => ({
+    marginRight: theme.spacing(size),
+  }),
+  marginBottom: ({ size = 1 }) => ({
+    marginBottom: theme.spacing(size),
+  }),
+}));
+
+export const useUtilityStyles = makeStyles<Theme, { baseColor?: string }>(
```
niclim, line 40: I think this is actually something I think would be useful - it might be _harder_ to get this right though (do we want a consistent darken, across all shades? do we need to handle specific pseudo classes when we want to overwrite material ui classes, etc)

workspaces/ui-v2/src/utils/dateFormat.ts
```diff
@@ -0,0 +1,11 @@
+// @ts-ignore
```
niclim, line 1: Pulled from usages across the UI app

workspaces/ui-v2/src/pages/changelogHistory/ChangelogHistory.tsx
```diff
@@ -0,0 +1,75 @@
...
+} from '<src>/components';
+import { useBatchCommits } from '<src>/hooks/useBatchCommits';
+import { useSpacingStyles, useUtilityStyles } from '<src>/styles';
+import { formatTimeAgo } from '<src>/utils';
+
+export const ChangelogHistory: FC = () => {
+  const containerSpacing = useSpacingStyles({
+    size: 3,
+  });
+  const batchCommitSpacing = useSpacingStyles({});
```
niclim, line 18: Looking at this, this feels like it's too rigid which will end up requiring _more_ code to get the css applied as we want it

workspaces/ui-v2/src/pages/changelogHistory/ChangelogHistory.tsx
```diff
@@ -0,0 +1,75 @@
...
+        padded
+        className={containerSpacing.padding}
+        style={{
+          display: 'flex',
+          alignItems: 'center',
+        }}
+        loading={loading}
+      >
+        <h1>Changelog</h1>
+        <div style={{ width: 800, display: 'flex', flexDirection: 'column' }}>
```
niclim, line 38: For things like layout, and width - I think inline styling actually works well here - usually you want them granular enough that you can be specific (e.g. display flex, with all the other required css styles)

niclim:
@devdoshi - feel free to review or not review this tomorrow, I think there's a fair amount of things jaap and I might want to change / update with how we handle the styling here.

I can spend tomorrow working on the actual implementation of reverting so I'm not blocked by this merging or anything

devdoshi (commented):
Thanks @niclim - I'll leave it up to you and @JaapRood to refine later!

workspaces/ui-v2/src/components/navigation/TopNavigation.tsx
```diff
@@ -72,6 +90,10 @@ const useStyles = makeStyles((theme) => ({
     position: 'fixed',
     width: '100%',
   },
+  spacing: {
+    height: 41, //40px height + 1px border
```
JaapRood, line 94: Having a class named "spacing" smells of css hackery. Haven't actually dug into how this is all structured, but 98/100 times it's an indicator of layout needing to be structured differently.

workspaces/ui-v2/src/pages/changelogHistory/ChangelogHistory.tsx
```diff
@@ -0,0 +1,94 @@
...
+                    {batchCommit.commitMessage}
+                  </h4>
+                  <span className={classes.commitTime}>
+                    {formatTimeAgo(new Date(batchCommit.createdAt))}
+                  </span>
+                </Link>
+              </li>
+            ))}
+          </ol>
+        </section>
```
JaapRood, line 53: @niclim here's some of the changes I've made to the new page, representing fairly well how I think we should approach the new pages. See some of our private conversation about why I think going with semantic classnames is the better trade-off to make, especially from the perspective of working in tandem with a designer (and their tools). Happy to go into more detail if you'd like.

niclim:
@JaapRood and I discussed this offline, going to merge this in as is + branch off and tackle things individually. If there's any other comments we want to address this in follow up PRs

JaapRood:
Should've given it an approve after discussing, sorry!

PR #902, opened 2021-06-18
New fields, within objects, inside of lists [BREAK] -> Do not break

acunniffe:
## Why
We tracked down a bunch of the Optic 10 diff issues folks have been having and they seem to have a small/common cause. The new shape trail parser did not include the object ID when the last shape trail was `ListItemTrail`, and the list shape could be an object. 

This meant that the diff would be interpreted as a new field, but when it tried to lookup the object it had to attach itself to, it would hit an invariant we coded. 

## What
The shape trail parser has been updated. Interestingly, we do have snapshot tests for this scenerio, but we didn't make a strong assertion that the lastObject should be defined in these cases. Lots of JSON to approve, it was simply missed. I've updated the cases and manually reviewed them. 

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/shape-hash/src/json-to-shape-hash.ts
+++ b/workspaces/shape-hash/src/json-to-shape-hash.ts
@@ -76,14 +76,18 @@ function toJson(item: any) {
   switch (item.type) {
     case types.OBJECT:
       const newObj: Record<string, any> = {};
-      if (item.hasOwnProperty('fields')) { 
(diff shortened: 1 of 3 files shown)


devdoshi (approved):
Nice

LouManglass (approved):
This works as intended for the invariant violation discovered.

PR #903, opened 2021-06-18
Prevent root node from being learnt multiple times

niclim:
## Why
Fixes:
- The root path node from being learnt multiple times
- Grouped endpoints showing up as `undefined`

The root path node can be learnt multiple times as when the interaction diff traverser gets a root path node, the `endpoint_queries.resolve_interaction_path(&interaction);` always returns `None` - https://github.com/opticdev/optic/blob/develop/workspaces/optic-engine/src/interactions/traverser.rs#L25

## What

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/optic-engine/src/queries/endpoint.rs
+++ b/workspaces/optic-engine/src/queries/endpoint.rs
@@ -41,9 +41,6 @@ impl<'a> EndpointQueries<'a> {
   }
 
   pub fn resolve_path(&self, path: &str) -> Option<PathComponentIdRef> {
-    // If the path is a root path, we need to check it separately
-    // There is always going to be a node for the root path, so we need to check
-    // If it has been learnt, there will be an associated HTTP method
     if path.eq("/") {
       return Some(ROOT_PATH_ID);
     }
@@ -541,6 +538,19 @@ mod test {
       {"ResponseAddedByPathAndMethod": {"responseId": "response_1", "pathId": "root", "httpMethod": "GET", "httpStatusCode": 200 }},
     ]))
     .expect("should be able to deserialize test events");
+
+    let spec_projection = SpecProjection::from(events);
+    // dbg!(Dot::with_config(&spec_projection.endpoint().graph, &[]));
+
+    let endpoint_queries = EndpointQueries::new(spec_projection.endpoint());
+
+    assert_eq!(endpoint_queries.resolve_path("/").unwrap(), "root");
+  }
+
+  #[test]
+  pub fn resolve_path_should_resolve_against_an_empty_spec() {
+    let events: Vec<SpecEvent> =
+      serde_json::from_value(json!([])).expect("should be able to deserialize test events");
 
     let spec_projection = SpecProjection::from(events);
     // dbg!(Dot::with_config(&spec_projection.endpoint().graph, &[]));
--- a/workspaces/optic-engine/src/queries/endpoint.rs
+++ b/workspaces/optic-engine/src/queries/endpoint.rs
@@ -550,6 +550,44 @@ mod test {
   }
 
   #[test]
+  pub fn resolve_path_can_resolves_root_path() {
+    let events: Vec<SpecEvent> = serde_json::from_value(json!([
+      {"PathComponentAdded": { "pathId": "path_1", "parentPathId": "root", "name": "posts" }},
+      {"PathComponentAdded": { "pathId": "path_2", "parentPathId": "path_1", "name": "favourites" }},
+      {"PathComponentAdded": { "pathId": "path_3", "parentPathId": "root", "name": "authors" }},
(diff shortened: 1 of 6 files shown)


workspaces/optic-engine/src/queries/endpoint.rs
```diff
@@ -41,6 +41,32 @@ impl<'a> EndpointQueries<'a> {
   }
 
   pub fn resolve_path(&self, path: &str) -> Option<PathComponentIdRef> {
+    // If the path is a root path, we need to check it separately
+    // There is always going to be a node for the root path, so we need to check
+    // If it has been learnt, there will be an associated HTTP method
+    if path.eq("/") {
```
niclim, line 47: I think this is correct, we have to treat this as a special case since we always have a root node, so resolving the path isn't enough to check if we have "seen" this path. I think we need to treat this if it has _any_ http method associated with this.

I'm not sure if I've missed anything here though

devdoshi: Hmm, I think resolve_path should just be concerned with paths. perhaps the logic might be more clear if we used petgraph's DFS/BFS visitors. However, I believe the problem is:
```
let mut path_components = path.split('/');
    // skip leading empty
    path_components.next();
    let mut last_resolved_path_id = Some(ROOT_PATH_ID);
    while let Some(s) = path_components.next() {
      // eprintln!("trying to match segment {}", s);
      let node_index = self
        .graph_get_index(last_resolved_path_id.unwrap())
        .expect("expected a node with node_id to exist");
```

So when the path is `/`, we split by `'` skip the first `''` and try to search for the second `''` relative to the root path component. IIRC at the time this was written I didn't want to make assumptions about component names being empty. So we should be able to say if path.eq("/") return Some(ROOT_PATH_ID). 

With any change in behavior we should see a change in tests, so whatever the solution ends up being, I'd like to see a test that would fail with the previous version of the code that would pass with this one. Sorry for the bug :)

niclim: I added two tests here to illustrate (also, not sure if we should move these into modules + have separate tests):
- `resolve_path_can_resolves_root_path` - before this PR, the code always returned `None` where it should have returned root
- `resolve_path_does_not_include_root_when_no_methods_attached` - implementing `if path.eq("/") return Some(ROOT_PATH_ID)` _always_ returns the root node - even if we haven't learnt anything. Given that we always have the root node, we treat the root node as being seen if we've seen any HTTPmethods learnt from this. Otherwise what happens is we always see the root node as learnt - even if we haven't learnt any endpoints
Screen Shot 2021-06-18 at 3 38 45 PM

This seemed incorrect to me, which is why there's extra logic in there - but perhaps the logic isn't in the right place, but I'm not sure where a better place is to put this. Thoughts?

devdoshi: Well definitely resolve_path should only care about paths. I think it's ok that / shows up as a known path, but i might be missing what else that actually affects?

niclim: I think that is the only difference, they both show up in the undocumentedUrls component and go through the same flows. I've updated this to remove the extra logic such that it's just `if path.eq("/") return Some(ROOT_PATH_ID)`

devdoshi: Great. I think that makes the most sense; there kind of has to be an established root path. Curious the bug made it this long haha.

workspaces/optic-engine/src/queries/endpoint.rs
```diff
@@ -41,6 +41,13 @@ impl<'a> EndpointQueries<'a> {
   }
 
   pub fn resolve_path(&self, path: &str) -> Option<PathComponentIdRef> {
+    // If the path is a root path, we need to check it separately
```
devdoshi, line 44: we can remove these comments now

workspaces/optic-engine/src/queries/endpoint.rs
```diff
@@ -523,6 +530,26 @@ mod test {
...
+      {"ResponseAddedByPathAndMethod": {"responseId": "response_1", "pathId": "root", "httpMethod": "GET", "httpStatusCode": 200 }},
+    ]))
+    .expect("should be able to deserialize test events");
+
+    let spec_projection = SpecProjection::from(events);
+    // dbg!(Dot::with_config(&spec_projection.endpoint().graph, &[]));
+
+    let endpoint_queries = EndpointQueries::new(spec_projection.endpoint());
+
+    assert_eq!(endpoint_queries.resolve_path("/").unwrap(), "root");
```
devdoshi, line 550: should be able to test against an empty spec

devdoshi (approved):

PR #904, opened 2021-06-18
Fix capture for context type text/html with no text body

niclim:
## Why

When a non-json response is returned with a content length 0 - the proxy captures this as null - when it should be an empty string. 

TODO
- [ ] test this proxy out and send 0 length non-json responses

## What

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/ui-v2/src/pages/diffs/ReviewEndpointDiffPage/ReviewEndpointDiffPageContainer.tsx
+++ b/workspaces/ui-v2/src/pages/diffs/ReviewEndpointDiffPage/ReviewEndpointDiffPageContainer.tsx
@@ -44,7 +44,7 @@ export const ReviewEndpointDiffContainer: FC<
 
   return (
     <PageLayout AccessoryNavigation={DiffAccessoryNavigation}>
-      {!endpoint || shapeDiffs.loading ? (
+      {!endpoint || shapeDiffs.loading || newRegionDiffs.loading ? (
         <Loading />
       ) : (
         <ReviewEndpointDiffPage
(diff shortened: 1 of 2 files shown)


workspaces/ui-v2/src/pages/diffs/ReviewEndpointDiffPage/ReviewEndpointDiffPageContainer.tsx
```diff
@@ -44,7 +44,7 @@ export const ReviewEndpointDiffContainer: FC<
 
   return (
     <PageLayout AccessoryNavigation={DiffAccessoryNavigation}>
-      {!endpoint || shapeDiffs.loading ? (
+      {!endpoint || shapeDiffs.loading || newRegionDiffs.loading ? (
```
niclim, line 47: Ok this was another bug that will allow a user to recover from this state - there's a race condition where diffDescriptions can be undefined here...

JaapRood:
Seems like a great find of a bug!

> When a non-json response is returned with a content length 0 - the proxy captures this as null - when it should be an empty string.

Can you maybe explain why this is the case? And how the difference between the two makes a difference down stream / to the ui / ux?

LouManglass (approved):
Per discussion - tested against some generated request/response data and the behavior has changed as expected. We'll need to get it in front of more user traffic to be sure, though it looks good at the general level.

JaapRood (approved):
Assuming this is the right assumption (seems to from an internal thread), this looks good!

PR #910, opened 2021-06-21
Install rust toolchain for npm package release and side channel

niclim:
## Why

https://github.com/opticdev/optic/pull/896#discussion_r653877736
> my guess is that the base container comes with a 1.52.0 version of rust, and when you run the rust-toolchain action, it upgrades it to 1.53.0

As a result, we should apply these to the release branches where we run the tests otherwise the tests fail and run a different verison of the build (I also noticed the side channel and the release workflow specify different versions of node - sidechannel == 14, release ==12) is there a better way to standardize this so we have the same env for our builds? @notnmeyer ?

## What

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/.github/workflows/release-side-channel.yml
+++ b/.github/workflows/release-side-channel.yml
@@ -21,6 +21,12 @@ jobs:
         uses: actions/setup-node@56899e050abffc08c2b3b61f3ec6a79a9dc3223d # https://github.com/actions/setup-node/releases/tag/v1.4.4
         with:
           node-version: 14
+      - name: 'Rust toolchain'
+        uses: actions-rs/toolchain@b2417cde72dcf67f306c0ae8e0828a81bf0b189f # https://github.com/actions-rs/toolchain/commits/v1
+        with:
+          toolchain: stable
+          profile: minimal
+          override: true
       - name: Install Task
         uses: Arduino/actions/setup-taskfile@9d04a51fc17daddb0eb127933aaa950af1e3ff97 # they dont give us any tags :\
         with:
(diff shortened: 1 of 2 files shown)


niclim:
Examples of where this is happening:

In this PR - https://github.com/opticdev/optic/pull/896 the rust_packages + npm_packages runs both run the same rust tests, but one fails and one passes (because of rust version differences)

In this sidechannel build https://github.com/opticdev/optic/runs/2878606119?check_suite_focus=true this also fails because it's pulling a different version of rust

notnmeyer:
https://github.com/actions-rs/toolchain#inputs says `toolchain` can be set to a version if we want to control the verison. we intentionally opted for floating on "stable" and fixing stuff as it came up, but i dont really have an opinion about it.

related, https://doc.rust-lang.org/edition-guide/rust-2018/rustup-for-managing-rust-versions.html#managing-versions

> sidechannel == 14, release ==12) 
 
the sidechannel releases are newer as a concept and were put on 14 to see if it worked. i'd also support those being the same version.

niclim:
This is slightly different - it's that the base container for ubuntu comes with a version of rust, but it's out of date. Calling the rust toolchain command for node / npm package builds updates the rust version to the latest. The issue is that we weren't calling the rust toolchain action in node builds (which still call rust code)

I'll create a separate PR for sidechannel + release versions so we can decide if we want to include that in the upcoming release

notnmeyer:
oh, right. got it 👍

notnmeyer (approved):

PR #911, opened 2021-06-21
Use node-version 14 for releases

niclim:
## Why

Re: https://github.com/opticdev/optic/pull/910#issuecomment-865312849 - specifying node version 14 - which is the latest LTS node version - keeping this separate and we can decide if we want to include this in the next release, or include this into the next release (the sidechannel build is on node version 14, everything else is on version 12)

## What

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


notnmeyer (approved):

LouManglass (approved):
14 is great, it's LTS.

PR #912, opened 2021-06-21
Spectacle typings

niclim:
## Why

As part of my ongoing challenge to remove all references to `any` in our codebase - I added typing to spectacle

## What

- Split out spectacle/index.ts -> spectacle.ts and types.ts
- To view the diff better of what has changed in this PR, see: https://github.com/opticdev/optic/pull/912/files/86c535f5921988342e62b4664317f1fad6220838..c937b3cf92eb5e6f7cf7f32dff2254182cc90f0b

Ongoing
- Remove more any typings from spectacle
- Start making our graphql queries return non-nullable instances (none of our schemas are non-nullable)
  - we'll want to be cautious here, because graphql will error the entire query if it receives a nullable value when it should be non-null. Our UI currently handles all these cases, but that's a band-aid for the underlying problem (receiving a nullable value in the first place)

## Validation

I've gone through a pass of the UI - since we've slightly changed the graphql schema to be _slightly_ more strict now (i.e. adding `!` - which can cause errors in certain parts of the code.


yarn.lock
```diff
@@ -3904,6 +3904,11 @@ argparse@^1.0.7:
   dependencies:
     sprintf-js "~1.0.2"
 
+argparse@^2.0.1:
```
niclim, line 3907: Some changes picked up in `task postpull`

workspaces/spectacle/src/graphql/schema.ts
```diff
@@ -258,10 +258,7 @@ type ChangesResult {
 """
 Endpoint Changes
 """
-type EndpointChanges {
-  # URL for Optic change documentation
-  opticUrl: String
```
niclim, line 263: This isn't called anywhere, and isn't returned in the buildEndpointChanges function

JaapRood:
I'm not sure how I feel about losing all the Git history for `spectacle/index.ts` that results from splitting up to two separate files for types and implementation. There's now two whole new files, all blamed to @niclim instead of the original author, with all line by line info and why what changed how gone. Not sure if cost is worth something that could have happened in that same original file.

If there are reasons for this beyond a bit of clean up, I'd suggest attempting this in a way that keeps most of that original history in tact. That would mean extracting types to a new file, but leaving the original `spectacle/index.ts`, and only renaming it at the last moment (so Git registers it as a rename and continues history).

niclim:
@JaapRood I did some git commit juggling and I think i've updated this to preserve git history of `spectacle.ts` - see blame history here https://github.com/opticdev/optic/blame/a419d832d0a3b1ff26a5d852dc31e112a6208b9b/workspaces/spectacle/src/spectacle.ts#L181 - the magic commit is the rename commit here https://github.com/opticdev/optic/pull/912/commits/03a9646bdf7664bbdc0bfec41350442a5eb55df1

JaapRood (approved):
Looks like a good clean up to me. Going to tag @devdoshi explicitly here for a retroactive review, as this is his recent and original design and might have a more informed opinion.

niclim:
Sounds good - I'm going to merge this in and I'll address any comments retroactively in a follow up PR

PR #913, opened 2021-06-21
Releases/10.0.6

LouManglass:
## Why
Latest release, including some issue fixes:

- Issue 895 (will require clearing out existing captures first)
- API root requests will not show up continually as diffs
- Intercept mode will show diffs reliably

## What

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


LouManglass:
Release in flight.

PR #914, opened 2021-06-21
Implement reset to commit spectacle commands

niclim:
## Why

Set up the resetToCommit spectacle command handler as part of the reset to commit bet. This includes the inMemory and local cli implementation

## What

- Implemented the in memory resetToCommit (a simple in memory filter)
- Implemented the local cli resetToCommit (uses streams + async iterators)
- Added resetToCommit to the spectacle mutation gql schema

## Validation
Tested via implementing https://github.com/opticdev/optic/pull/919 and testing with my own test project (local cli + in memory examples)


workspaces/cli-server/src/spectacle/index.ts
```diff
@@ -97,9 +132,10 @@ export class LocalCliSpecRepository implements IOpticSpecReadWriteRepository {
 
   async listEvents(): Promise<any[]> {
     if (isEnvTrue(process.env.OPTIC_ASSEMBLED_SPEC_EVENTS)) {
-      const events = OpticEngine.readSpec({
+      const streamArray = StreamArray.withParser();
```
niclim, line 135: I'm not sure how this is intended to be used, because if it's going to use a stream to avoid loading this into memory - all instances will need to be updated to consume a stream

workspaces/cli-server/src/spectacle/index.ts
```diff
@@ -109,6 +145,38 @@ export class LocalCliSpecRepository implements IOpticSpecReadWriteRepository {
...
+      'specification.json'
+    );
+    const tempOutputFile = path.join(
+      this.dependencies.specDirPath,
+      `temp-spec-file${uuidv4()}.json`
+    );
+    const tempFileDeletion = path.join(
+      this.dependencies.specDirPath,
+      `spec-to-delete${uuidv4()}.json`
+    );
```
niclim, line 162: I'm not sure if this is necessary? But if node crashes while the stream is in flight, this could (maybe) corrupt the specification.json file?

Is there a better way to handle this?

workspaces/cli-server/src/spectacle/index.ts
```diff
@@ -12,10 +17,9 @@ import {
   IOpticSpecRepository,
   StartDiffResult,
 } from '@useoptic/spectacle';
-import { EventEmitter } from 'events';
 import { AsyncTools, AsyncTools as AT, Streams } from '@useoptic/optic-domain';
-import * as OpticEngine from '@useoptic/optic-engine-native';
-import * as opticEngine from '@useoptic/optic-engine-wasm';
+import * as OpticEngineNative from '@useoptic/optic-engine-native';
+import * as OpticEngineWasm from '@useoptic/optic-engine-wasm';
```
niclim, line 22: Renamed to be more explicit about usage

workspaces/ui-v2/src/entry-points/local/local-cli.tsx
```diff
@@ -151,7 +151,7 @@ export function useLocalCliServices(
     baseUrl: apiBaseUrl,
     spectacle,
   });
-  const specRepository = new LocalCliSpecRepository({ baseUrl: apiBaseUrl });
+  const specRepository = new UILocalCliSpecRepository({ baseUrl: apiBaseUrl });
```
niclim, line 154: There's two localcli spec repositories (one that runs in the cli, one that runs in memory in the UI)

niclim:
@JaapRood - this still has a little bit of work, but this could use some tests, but I'm not sure where the best place to put these are. I know dev was working on our snapshot tests, but in the meantime this could use some basic snapshot tests (for both the local cli implementation and the in memory implementation) - where do you think the best place to put tests are for this

workspaces/cli-server/src/spectacle/index.ts
JaapRood: Ensuring some safety of the data is a good thing and the process going away (killed, crashing, etc) an hopefully infrequent but plausible situation we must consider, nice that you've thought about this. 

We're in luck, though: in this case, `specification.json` that has any historical value is tracked in their Git repo, so it's hard to really destroy something. Git is likely to hold at least _some_ version for us to mitigate the damage caused by crashes. I'd say that's probably enough of a safety net for now, given the appetite for this bet and it's fairly experimental nature.

What if we _do_ want to increase safety some more?

Renaming files seems like a reasonable way to do so: the chance of data getting lost is definitely minimised this way, as data isn't destroyed until the previous operation has written. However, it does introduce a whole bunch more failure cases as well that could lead to interesting situations. What happens, for instance, when the process crashes _after_ writing the tmp output file, but _before_ it can rename it? Or before the temp file can be deleted? How should the process recover, or are we stuck with old, potentially partly written files?

Generally, the way to deal with these kinds of problems is to implement a crash-recovery mechanism, or a processing mechanism that can continue where it left off. Doing that efficiently gets you into the world or Write-Ahead-Logs or similar concepts: basically we're implementing a database at that point.

There are some other routes to look into, like using `fs.truncate`, but to make that work you have to be pretty precise as well, so there's no guarantees that actually makes Optic more instead of less fault-tolerant.

For now I'd say: lets rely on the fact that these specs are tracked files in Git repositories as our safety net: the rest of the app definitely does so. If we _do_ want to start improving fault tolerance beyond that, I'd say we should look at implementing that sort of stuff on the Rust side.

JaapRood (commented):
The approach here is the one I'd go for: using streams to filter the current specification.json and writing a new one, until the given batch commit is detected. The implementation has a couple of bits of refactoring the codebase would benefit from.

Note: I'm not sure how much this actually relies on #912, but since there has been some concerns raised there, it might be worth rebasing this onto `develop` if at all possible, to make sure merging of functionality doesn't get blocked by clean up / refactors.

JaapRood:
> where do you think the best place to put tests are for this

@niclim I'd say if there isn't any particular end to end test where this get covered, keep the tests close to where they are implemented. Spectacle itself has a good example of how to do simple snapshot tests in Node itself, using `Tap`. It might be worth trying to add tests there, or if it's very specific to a particular spectacle implementation, add a test suite _like_ that to where that implementation lives (like `cli-server`). 

Perhaps that also shows that the `IOpticSpecReadWriteRepository` and similar interfaces should only be concerned with reading and writing, not this additional reset logic, dealing only in providing or accepting streams of spec events. That way, the logic for what constitutes a reset can live in the main spectacle package, as well as its tests. Food for thought, I hope 🤔.

workspaces/cli-server/src/spectacle/index.ts
niclim: Ah yeah, it completely slipped my mind that anything of value would most likely be backed by version control, so it's good enough to rely on that rather than build our own crash recovery mechanisms

workspaces/cli-server/src/spectacle/index.ts
```diff
@@ -109,6 +112,28 @@ export class LocalCliSpecRepository implements IOpticSpecReadWriteRepository {
...
+    const specEvents = Streams.SpecEvents.fromJSONStream()(rawEventStream);
+    const filteredEvents = pipe(
+      Streams.SpecEvents.takeBatchesUntil(batchCommitId),
+      AsyncTools.intoJSONArray
+    )(specEvents);
+
+    const writeableStream = fs.createWriteStream(specJsFile);
+
+    for await (const chunk of filteredEvents) {
+      writeableStream.write(chunk);
```
niclim, line 134: I'm not sure if this is correct, but this is how I managed to pipe the stream into the writeable stream.

I am a little concerned that this function will exit / the promise will resolve before the last write can occur. I also don't know if there's any race conditions since we aren't waiting for each write to complete (but i presume this _should_ work because its a stream)?

Any other ways to do this? Or is this fine as it is?

workspaces/optic-domain/src/async-tools.ts
```diff
@@ -37,6 +38,22 @@ export function lastBy<T>(
   };
 }
 
+export async function* intoJSONArray<T>(
```
niclim, line 41: I tried looking for an existing function that did this for me, but I couldn't find anything

workspaces/optic-domain/src/async-tools.ts
```diff
@@ -54,17 +71,22 @@ export function fromReadableJSONL<T>(): (stream: Readable) => AsyncIterable<T> {
...
-  key: string;
-  value: T;
-}> {
+export function fromJSONMap<T>(): (stream: Readable) => AsyncIterable<T> {
   return async function* (source) {
     let parseResults = source.pipe(StreamObject.withParser());
 
     for await (let parseResult of parseResults) {
-      yield parseResult;
+      yield parseResult.value;
```
niclim, line 79: This function (fromJSONMap) isn't currently used, and `parseResult.value` is used in other fromXYZ functions (e.g. fromReadableJSON)

workspaces/cli-server/src/spectacle/index.ts
```diff
@@ -28,11 +32,42 @@ import {
...
+      this.foundBatchCommit = true;
+    }
+
+    if (!this.foundBatchCommit) {
+      return next(null, chunk);
+    } else {
+      next();
+    }
+  }
+}
```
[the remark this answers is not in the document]
niclim: This makes sense, didn't know we had existing patterns for this - this makes a lot of sense!

niclim:
@JaapRood I've updated this to use async iterables (very cool stuff, took me a while to wrap my head around and figure out how they worked together though).

Responses to your comments above:
- I added a test to the `takeUntilBatches` function
- I agree, I think logic could be moved from `IOpticSpecReadWriteRepository` into spectacle, and leave it solely in charge of the read/write operations (to memory, or to disk). This seems like a slightly larger task, but would help us write better overall tests.
- I updated this to retarget `develop` - I removed the spectacle updates, but those will be pretty quick to make (i.e. updating the gql schema, adding in resolvers for the new mutation) - so I'll leave that until I'm happy with the rest of this code

As for now, let me know if these changes make sense - I'll either create a new PR with the relevant spectacle commands (and the reset button) or include it in this PR

workspaces/optic-domain/src/async-tools.ts
JaapRood: I _think_ that in this case parseResult is right as we also want to return the `parseResult.key`, because an iterator of a `Map` is generally it's entries (key => value pairs). However, since it's not being used, there's no proof of that, so 🤷‍♂️.

workspaces/optic-domain/src/async-tools.ts
JaapRood: `stream-json` doesn't only have parsers, but also what they call [stringers](https://github.com/uhop/stream-json/wiki/Stringer) (for both json and jsonl, see docs), which are generally used in combination with a [Disassembler](https://github.com/uhop/stream-json/wiki/Disassembler). The dissambler accepts entities and returns a token stream, which the stringer uses to emit JSON. Even the biggest entities can be smoothly rendered as JSON that way!

JaapRood (commented):
Great progress, thanks for digging into the world of Async Iterators / Generators! There's definitely a bit of a learning curve with them, so I really appreciate the effort. Hopefully I've given enough details to dot the i's and cross the t's, let me know otherwise!

workspaces/optic-domain/src/async-tools.ts
niclim: Ah, yeah that'd make sense - I'll revert this change and it can be updated later if required

workspaces/cli-server/src/spectacle/index.ts
niclim: Thanks for the detailed explanation, I updated with your suggestions above

workspaces/optic-domain/src/async-tools.ts
niclim: I updated this function to return a `Readable` stream - since this is likely it be used to serialize back into a IO using a stream. I could update this to return an async iterable but I'm not sure what usecase this would be necessary

workspaces/spectacle/src/graphql/schema.ts
```diff
@@ -8,6 +8,7 @@ type Mutation {
   applyCommands(commands: [JSON!]!, batchCommitId: ID!, commitMessage: String!, clientId: ID!, clientSessionId: ID!): AppliedCommandsResult
   startDiff(diffId: ID!, captureId: ID!): StartDiffResult
   invalidateCaches: InvalidateCachesResult
+  resetToCommit(batchCommitId: ID!): Boolean
```
niclim, line 11: We aren't returning anything here, GQL doesn't have a way to define explicit null types. The omission of a `!` means we can return null. 

There isn't much too useful to return here (apart from whether there was a batchCommit removed - but that's not likely to be super useful given you are likely querying batch commits -> calling this once you have a list of valid commits which is why I just returned null.

niclim:
Ok I've updated the code with your suggestions and added the spectacle schema, so it should be callable from the FE - @JaapRood

workspaces/optic-domain/src/streams/spec-events.ts
```diff
@@ -0,0 +1,25 @@
...
+  return takeWhile((event) => {
+    if (seenBatchCommitId) {
+      return false;
+    }
+    if (event.BatchCommitEnded?.batchId === batchCommitId) {
+      seenBatchCommitId = true;
+    }
+    return true;
+  });
+}
```
niclim, line 25: I realized my initial implementation was incorrect - we want to _include_ the batchCommit specified here, everything else should be ignored. This code now is correct (updated the test as well)

workspaces/cli-server/src/spectacle/index.ts
```diff
@@ -109,6 +114,40 @@ export class LocalCliSpecRepository implements IOpticSpecReadWriteRepository {
...
+      `optic-temp-spec-file-${uuidv4()}.json`
+    );
+    const tempFileDeletion = path.join(
+      this.dependencies.specDirPath,
+      `optic-temp-spec-to-delete-${uuidv4()}.json`
+    );
+    const specJsFile = path.join(
+      this.dependencies.specDirPath,
+      'specification.json'
+    );
```
niclim, line 130: It turns out we did need these temp files. If you open a read file stream and try write to that same file, it turns out that doesn't work and the parser just crashes. 

I prefixed the temp files as `optic-temp` and added this to the `.gitignore` file. The problem is older optic installs won't have gitignore. This brings up the question how do we update the local states of users with outdated versions (perhaps we could add an optic version to optic.yml and then run relevant "migration" steps) - this means we could also extract out spec id from the spec.json + add in missing gitignores. Something we'll probably want to look into later.

JaapRood (approved):
Looking good! There's some small stuff that could be tidied up a bit further if anyone feels keen, but given our appetite for this feature, it's dealer's choice whether to spend the effort.

PR #916, opened 2021-06-22
Merge release -> develop

niclim:
## Why

There's probably a better release process we can set, so that we can do FF merges or not have to juggle develop -> release + merge backs. But currently we are missing the release version bump commit in the develop branch. Merging this back will pull back the relevant merge commits (so we _could_ FF merge develop -> release) and also include the 10.0.6 release bump

## What

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


LouManglass (approved):

LouManglass:
> Describe the motivation for the changes.
> 
> There's probably a better release process we can set, so that we can do FF merges or not have to juggle develop -> release + merge backs. But currently we are missing the release version bump commit in the develop branch. Merging this back will pull back the relevant merge commits (so we _could_ FF merge develop -> release) and also include the 10.0.6 release bump

Following up on a recent conversation we've had - I threw some time on your calendar today to have a chat about release workflow, get the conversation started. I'll be happy to bring it to the rest of the team and make a shift. We can defer this conversation if need be, as this merge should give us some quick wins already, I just wanted to get this process moving.

PR #917, opened 2021-06-23
docs: updated install instructions

LouManglass:
## Why

Just cleaning up the install instructions, is all.

## What

We no longer need the `--legacy-peer-deps` flag. Yay!


LouManglass:
Also added a fix for a broken link

JaapRood (approved):

PR #918, opened 2021-06-23
Implement a basic History page UI

JaapRood:
## Why
We had a very basic page rendering a list of batch commits, but it didn't look too much like a history view yet.

## What
A short pass to:

- establish the History page name and route (instead of changelog)
- design the History page to look somewhat like a timeline

Screenshot 2021-06-23 at 18 09 01

## Validation
* [ ] CI passes


JaapRood:
@niclim feel free to merge this if you want to work of it! There's maybe a bit more to do to style the buttons, but that might as well be a follow-up PR.

niclim (approved):
Looks great - what a difference some styling makes!

PR #919, opened 2021-06-23
Add in confirm reset modal and hook up UI to spectacle

niclim:
## Why

Connects the UI to the resetToCommit spectacle handlers implemented in https://github.com/opticdev/optic/pull/914 - this is currently targeting https://github.com/opticdev/optic/pull/914

## What

Main changes:
- Added a confirm reset commit modal
Screen Shot 2021-06-23 at 1 56 17 PM
- Hooked up the UI to call spectacle (see deleted commit after confirming above)
Screen Shot 2021-06-23 at 1 56 28 PM

Misc changes:
- Changes the latest commit (compare -> view)
- Does not allow resetting the initialize specification metadata (there are edge cases that if you had an existing spec pre-optic10, you can reset _past_ this commit. It also feels more and more that the specification id should not be in the specification.json (maybe more in the .opticyml or in a separate file)
- Added a Button component with a loading state included (idea here is we can add / compose more functionality on top of material UI + have more consistency in our app)
- added analytic event for reset

## Validation

Tested this in the local cli + the in memory implementations and it worked as expected (specification.json updated and able to relearn)


workspaces/ui-v2/src/components/navigation/NavButton.tsx
```diff
@@ -13,9 +14,15 @@ export type NavButtonProps = {
 export function NavButton(props: NavButtonProps) {
   const classes = useStyles();
   const { Icon, title, to } = props;
+  const history = useHistory();
 
   return (
-    <Link className={classes.root} href={to}>
+    <Link
```
niclim, line 20: @JaapRood I know you changed this in your previous PR - realized as part of building this, we need to actually use onClick so we use the react-router redirect rather than the full page browser redirect.

I _think_ the button / link should maintain the appropriate accessibility / keyboard utility (I think material UI should handle this correctly)

workspaces/ui-v2/src/pages/changelogHistory/ChangelogHistory.tsx
```diff
@@ -1,23 +1,32 @@
...
   useChangelogPages,
   useDocumentationPageLink,
 } from '<src>/components';
-import { useBatchCommits } from '<src>/hooks/useBatchCommits';
+import { BatchCommit, useBatchCommits } from '<src>/hooks/useBatchCommits';
 import { formatTimeAgo } from '<src>/utils';
 
+import { ConfirmResetModal } from './components';
+
+const SPECIFICATION_METADATA_COMMIT_MESSAGE =
```
niclim, line 16: This needs to match up with the spectacle commit message (this is probably the "best" way to identify this commit). There's downsides (obviously) with using the commit message - we could also make the assumption the first commit is unrevertable; which for most cases first commit === specification metadata commit. 

The longer term thing here is the metadata probably should live outside of the temporal spec file

workspaces/ui-v2/src/components/navigation/NavButton.tsx
```diff
@@ -15,7 +15,7 @@ export function NavButton(props: NavButtonProps) {
   const { Icon, title, to } = props;
 
   return (
-    <Link className={classes.root} href={to}>
+    <Link className={classes.root} to={to}>
```
niclim, line 18: Used the react-router link so we can use a react-router browser redirecet rather than a full page browser reload

workspaces/ui-v2/src/pages/changelogHistory/ChangelogHistory.tsx
```diff
@@ -43,20 +54,22 @@ export const ChangelogHistory: FC = () => {
                 <div className={classes.commitControls}>
                   <Button
                     className={classes.commitCompareButton}
-                    href={
-                      i === 0
-                        ? documentationPage.linkTo()
-                        : changelogPage.linkTo(batchCommit.batchId)
+                    onClick={() =>
```
niclim, line 57: @JaapRood I know you changed this in your previous PR - realized as part of building this, we need to actually use onClick so we use the react-router redirect rather than the full page browser redirect.

From my testing, we still maintain the relevant keyboard navigation behavior which is provided by material ui (i.e. we can press enter when focused on the button)

niclim:
@JaapRood if you have any design critique I'm all ears, design definitely isn't one of my strengths

workspaces/ui-v2/src/components/navigation/NavButton.tsx
JaapRood: Wow, what a miss on my part 😅😭😬

workspaces/ui-v2/src/pages/changelogHistory/ChangelogHistory.tsx
JaapRood: Lets rewrite this with a `Link` from react-router then, as no href and only an on click handler causes accessibility and ux issues, keyboard navigation aside.

JaapRood (commented):
Looking good apart from an accessibility issue with the Button. Anything one can click that navigates to a new page (or part of a page) should be an anchor tag, to be able to benefit from all the browser's accessibility features around links. Should be fairly straightforward to resolve, though!

JaapRood:
I pushed a couple of changes that resolves the accessibility issues with the docs / compare buttons. While I was there, I couldn't help but refine the UI a bit more, mostly because I felt friction and clutter in how the first item in the list (the current version) had different buttons. By more clearly (although hopefully subtly at the same time) distinguishing it as a "special" commit, it should be more obvious to the user why they can't compare or reset it.

Screenshot 2021-06-24 at 12 13 56

workspaces/ui-v2/src/pages/changelogHistory/ChangelogHistory.tsx
JaapRood: Uhm, isn't this a reset _to_ the selected commit, making the selected commit the new "current" commit? That should make the initial commit a fine candidate as a reset target. Once it becomes the current, it can no longer be reset back any further.

workspaces/ui-v2/src/pages/changelogHistory/ChangelogHistory.tsx
```diff
@@ -143,7 +214,45 @@ const useStyles = makeStyles((theme) => ({
...
+
+    ...(variant === 'outlined'
+      ? {
+          border: `1px solid ${LightBlueBackground}`,
+        }
+      : {
+          background: LightBlueBackground,
+        }),
+  };
+}
```
JaapRood, line 258: This is a very quick and dirty version of a mixin that could eventually live on the theme object. Over time, those should be the source of truth of what buttons look like in Optic, with generic a `Button` component applying the default styles. That way, defaults can be used, but customisation remains straightforward as well.

However, that's way beyond the scope of what we can justify to do here, so this will just have to be the start.

workspaces/ui-v2/src/pages/changelogHistory/ChangelogHistory.tsx
JaapRood: Fixed in https://github.com/opticdev/optic/pull/919/commits/448903fb3a776eaf66145ea9319d440b8c5d29aa: turns out you can pass a `component` prop to a `Button`, a feature that seems to have been added for [exactly this purpose](https://material-ui.com/components/buttons/#third-party-routing-library).

workspaces/ui-v2/src/pages/changelogHistory/ChangelogHistory.tsx
```diff
@@ -29,45 +51,81 @@ export const ChangelogHistory: FC = () => {
...
-                        console.log(`resetting to ${batchCommit.batchId}`);
-                      }}
-                    >
-                      Reset
-                    </Button>
-                  )}
-                </div>
-              </li>
-            ))}
+            {batchCommits.map((batchCommit, i) => {
```
niclim, line 54: At some point we could also move this into it's own component (such that changes in render don't rerender the entire list) - unlikely to happen in this case (editing commit messages, etc)

niclim:
Going to merge this in so it can be tested - @JaapRood and I discussed offline and are happy for this to land (it's also still behind a FF)

PR #920, opened 2021-06-23
Fix UI gql queries

niclim:
## Why

When I added this - https://github.com/opticdev/optic/pull/912/files#diff-9d0ba504d2645e362388f01514b44ae2c69158d09f52ca3aab7c722ec5cd26cfR8-R46 I missed a couple of usage points here. Graphql queries + mutations need to match the underlying schemas (which is somewhat annoying, and we need to be careful about changing these queries in the future to be stronger typed (for any arguments)) - the biggest one will be making sure when we update the query, we need to update gitbot (and if / when we expose spectacle as a first class thing, treat these as breaking changes).

## What

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/spectacle-shared/src/local-cli/client.ts
+++ b/workspaces/spectacle-shared/src/local-cli/client.ts
@@ -150,7 +150,7 @@ export class LocalCliDiffService implements IOpticDiffService {
 
   async listDiffs(): Promise<IListDiffsResponse> {
     const result = await this.dependencies.spectacle.query<any, any>({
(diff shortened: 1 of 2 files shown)


niclim:
@JaapRood I'm going to merge this in first since it fixes the current develop code, but if you have concerns, i'll address them in a follow up PR

JaapRood:
@niclim I'm all here for retroactive review on things like this!

PR #921, opened 2021-06-23
Remove endpoint delete debug command

niclim:
## Why

Given we now have a way to delete an endpoint - we should remove the WIP way. The endpoint deletion functionality through the UI also provides changelog history - where this does not provide the changelog history

## What

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


acunniffe:
Given this was undocumented and hidden from view, I say "not a breaking change" remove it!

JaapRood:
I wasn't aware this even existed, to be honest, but that also means I can't really speak for what its practical use has been. I could imagine someone still uses this for debugging or QA? Since this is just a bit of clean up, I'll abstain from approving it and let people who actually used this do the talking :)

LouManglass:
I'll mark this deprecated on the docs site reference as well, as we add a "here's how to delete" article for the UI. I won't pull it down entirely yet from the command reference, as folks may not have upgraded quite yet, but we'll make the strong preference to upgrade and proper delete and then get rid of the command reference "soon".

PR #923, opened 2021-06-24
Enable Deleting of Endpoint for all release environments

JaapRood:
## Why
The new deleting endpoint feature should be ready for an initial release.

## What
The feature flag that governs the feature is enabled for all release environments.

## Validation
* [x] CI passes
* [x] Final QA


LouManglass (approved):
Testing has gone really well. I'm still digging in, but I've deleted endpoints that are part of longer paths in specs, deleted endpoints that should've been parameterized and mistakenly weren't, working on delete and re-add...in the general cases I've tried, this has been solid. The underlying feature is ready.

niclim (approved):

PR #924, opened 2021-06-24
Only show share button when on docs page

jshearer:
@niclim pointed out that the share button only really makes sense on the docs page, so I moved it into the `DocsPageAccessoryNavigation` component.

### DIFF
--- a/workspaces/ui-v2/src/pages/docs/components/DocPageNavigation.tsx
+++ b/workspaces/ui-v2/src/pages/docs/components/DocPageNavigation.tsx
@@ -5,6 +5,7 @@ import { useAppConfig } from '<src>/contexts/config/AppConfiguration';
 import { useAppSelector, selectors } from '<src>/store';
 
 import { EditContributionsButton } from './EditContributionsButton';
+import { ShareButton } from '<src>/components/sharing/ShareButton';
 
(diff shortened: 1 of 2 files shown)


workspaces/ui-v2/src/pages/docs/components/DocPageNavigation.tsx
```diff
@@ -22,6 +23,7 @@ export const DocsPageAccessoryNavigation: FC = () => {
       {appConfig.documentation.allowDescriptionEditing && (
         <EditContributionsButton />
       )}
+      {appConfig.sharing.enabled && <ShareButton />}
```
niclim, line 26: I think this might be better placed to on on the left, rather than on the right? i.e. the order of these buttons are - share button, changelog dropdown, edit contributions?

I don't really have a good reason other than it's my preference so feel free to disagree with me here

niclim (approved):

workspaces/ui-v2/src/pages/docs/components/DocPageNavigation.tsx
jshearer: I have even less of an opinion, so I shall defer to your slightly stronger opinion. changed!

PR #926, opened 2021-06-24
Remove reset button on the last commit

niclim:
## Why

I missed this message when reviewing https://github.com/opticdev/optic/pull/919#discussion_r657821610 - we can rely on the assumption the first commit is the specification initialization commit. Also with the changes to how the reset works, we are resetting + including the commit, rather than removing the commit

Screen Shot 2021-06-24 at 10 58 48 AM

Going to merge this in as for testing + retroactive review

## What

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


niclim:
@LouManglass I'm not sure if you've created a release branch yet - but this should also be included in it for the reset to commit feature

LouManglass (approved):
This seems right for the description.

PR #931, opened 2021-06-24
Move discard endpoint on diff flow to recently added rows

niclim:
## Why

Discard Endpoint is kind of misleading. The discard only does something if you've staged the endpoint. It also makes more sense to discard endpoints from the recently added view

## What

### Before

Screen Shot 2021-06-24 at 2 03 29 PM

### After
Screen Shot 2021-06-24 at 2 00 26 PM
Screen Shot 2021-06-24 at 2 00 33 PM

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


LouManglass:
No review, but I like this change in UX. Makes a lot more sense this way.

I could see a case for discarding an unlearned endpoint (say, I see it only has a 5xx series status and I know it's because there was a bug, and I want to throw this traffic out this time). That's different though, and we can revisit that idea and its tradeoffs later.

acunniffe:
Fast response to users award goes to 🔥

acunniffe (approved):

PR #936, opened 2021-06-24
Feature/query param UI mock

niclim:
## Why

Mocking out the spectacle implementation - the spectacle queries will be used for the documentation page and changelog page (and the diff flow endpoint preview containers). What I have in mind for the UI is something like this (very similar to path parameters - maybe a little bit of different styling)

Screen Shot 2021-06-24 at 4 51 37 PM

This will be sufficient for our first pass, but let's briefly consider more complicated shapes (so we aren't completely shutting ourselves out of future iterations):

### Arrays
Arrays are simple to display, but we'd need to differentiate parsing strategy (i.e. same key or csv or other)

### Objects
There's a few ways to describe these - `field[key][nestedKey]` or `field: stringifiedJSON`.

For `field[key]` variants, we could probably rely on primitive behaviors above since we mostly capture everything. The biggest thing we miss here is around optional objects (i.e. being able to infer whether the entire object is optional vs partial keys and variants) - but I think this is _good enough_ for most cases. If a user wanted more specific we could parse this (or allow custom parsing to treat this as an `opticShape.asObject` variant) - but the bigger issue is how do you 

For `field: stringifiedJSON` - this (if we parse this) we can render this as an `opticShape.asObject` variant. For now this would be a string.

We'd probably need to use the shape renderer to handle this and figure out a different UI flow to above. I would say for now, this is likely a not a common case and not support rendering object types (even though our data model could support it)

### Displaying parsing strategy
This - depending on the level of granularity (per document, per endpoint, per field) could be queried from spectacle, and this information could be displayed inline (since the use case of viewing the documentation is likely to learn _how_ to use the endpoint).

## What

I also annotated some parts around the `change` query since endpoint change and change result are effectively the same thing, but have two different interfaces (and also added in a removed key).

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


workspaces/spectacle/src/graphql/schema.ts
```diff
@@ -106,6 +106,24 @@ type HttpBody {
...
+  # The query parameter key
+  key: String!
+
+  # Reference to the shape for this query parameter
+  rootShapeId: String!
+
+  # Is the query removed
+  isRemoved: Boolean!
+
+  changes(sinceBatchCommitId: String): ChangesResult!
```
niclim, line 124: I _think_ this is necessary for the HTTP query since we also have `key`, but I'm not 100% sure here - we'll be able to find out as we add in more

niclim:
@JaapRood did some mocking out of the spectacle schema for querying query parameters - will spend some time tomorrow thinking through some of the diff workflow points and that should cover the integration points.

Let me know what you think here and if it makes sense (this should be safe to merge as well since nothing should be calling this, other than the changes included to changelog)

workspaces/spectacle/src/spectacle.ts
```diff
@@ -312,6 +355,23 @@ export async function makeSpectacle(opticContext: IOpticContext) {
...
+            rootShapeId: 'shape-1',
+            isRemoved: false,
+          },
+          {
+            id: 'queryid-2',
+            key: 'limit',
+            rootShapeId: 'shape-2',
+            isRemoved: false,
+          },
+        ];
```
JaapRood, line 373: This isn't how the Optic Engine will be able to represent what the query parameters for a Request are, especially considering how we'll need to implement diffing. Instead of returning an array of query nodes, each with a `key` and it's own `rootShapeId`, each `Request` will have a single `Body` that represents the entire set of query parameters. This `Body` will point to a root shape which is always an object, with fields for each query parameter.

This approach is important because this allows us to leverage the existing Shape differ, since that just works on Bodies.

JaapRood (commented):
> Arrays
> Arrays are simple to display, but we'd need to differentiate parsing strategy (i.e. same key or csv or other)
> Objects
> There's a few ways to describe these - field[key][nestedKey] or field: stringifiedJSON.

It's important to realise that we're only supporting `key: string => value: string` initially, with no further interpretation / parsing of these keys or values to deduce implied structure (as that's where all the complexity lies). So the only way document that, is by assuming the exact same thing: we don't care whether there are delimiters in a value that might indicate an array, or square brackets in keys that might indicate an array or objects.

The _only_ non-string structure we can safely infer is an array through _reoccurring_ keys, which will be powered by simple string matching.

niclim:
> It's important to realise that we're only supporting key: string => value: string initially
> 
> The only non-string structure we can safely infer is an array through reoccurring keys, which will be powered by simple string matching.

Agreed! I just wanted to keep this in mind such that we have the ability to improve on this in the future for more complicated parsing strategies when we go down those routes (or provide customization / plug in points) for representing objects, etc

workspaces/spectacle/src/spectacle.ts
niclim: Would this be an HTTPBody? If so, what would the contentType be (would this even be applicable)? Would we want to create our own type that doesn't include content type

I've updated this so it query returns a single HTTPBody

JaapRood: Working on #937 has shown it's not _quite_ an `HTTPBody`, as there's no content type to consider. It's own type with only a `rootShapeId` and `isRemoved` seems appropriate.

niclim:
@JaapRood Updated to match your rust implementation https://github.com/opticdev/optic/pull/937/files - This should be good to merge, but will await your approval incase anything else changes (I can branch off this and merge in your branch to continue work)

JaapRood (approved):
Sometimes there's nothing like the actual code to get aligned :) Looking good!

PR #937, opened 2021-06-25
Implement learning of query parameters for undocumented endpoints

JaapRood:
## Why
This is our first step in supporting query parameters with the Optic Engine. Implementing the learner will allow us to model the new events and commands around query parameters, as well as getting our first usable path to generating specs with query parameters end to end, but without having to implement the diffing as well.

## What
Governed by the `REACT_APP_FF_LEARN_UNDOCUMENTED_QUERY_PARAMETERS` feature flag (disabled for all targets by default), the undocumented bodies learner now visits parsed query parameters of the interaction it's learning about. For requests with query parameters, the new `SetRequestQueryParametersShape` command is generated, linking an object described through `Shape*` commands to a `Request`.

The `SetRequestQueryParametersShape` is able to be handled and turned into a `RequestQueryParametersShapeSet` event, which is handled by the internal `EndpointProjection` to track the query parameters set for each request. 

Two basic end-to-end tests verify this entire process to be working on the happy path.

This first implementation also generated a lot of follow up that we'll look at separately:

- exact command + events to be used to represent how the query parameter shape is linked to the request
- where to implement / execute the query string -> query parameters parsing, how this might fit in with future requirements of custom parsers
- wider (snapshot) test coverage
- diffing
- the UI

## Validation
* [ ] CI passes
* [x] happy path end to end verified
* [x] no query parameters are parsed, visited or documented without the feature flag set


JaapRood:
So far, I've implemented the logic that makes the undocumented bodies learner visit the query parameters of interactions that don't match any specced endpoints. The observations made are aggregated per endpoint (in the same way as happens for request and response bodies), to eventually have shape commands generated for them and included as part of the commands of the matching request in the `EndpointBodies` output type.

By including the commands for generating the query parameter spec in with the request, the client shouldn't have to make additional changes. Per request it can apply the learned commands in a simulated context to preview the endpoint, and apply them to the spec to persist.

Now this core bit has been implemented, there's two sides where more work has to be done to make this work end-to-end:

1. the parsing of the query string into an actual `BodyDescriptor`, so the learner will a body to learn from.
2. the generating of `QueryParameter*` commands that will tie the generated shape commands to request. 

Since the latter involves the modelling of the new commands (and with that the events) I'll work on that next, as there's more surface area for discussions and overall has a larger impact on the rest of the decisions we make. The parsing logic is more isolated and easier to mock in the meantime!

workspaces/optic-engine/src/projections/learners/undocumented_bodies.rs
```diff
@@ -10,29 +10,69 @@ use crate::JsonTrail;
...
       let (root_shape_id, body_commands) =
         observations.into_commands(id_generator, &JsonTrail::empty());
-      let endpoint_body =
-        EndpointBody::new(&body_location, root_shape_id, body_commands, id_generator);
+      let mut endpoint_body = EndpointBody::new(&body_location, root_shape_id, body_commands);
+
+      if let BodyAnalysisLocation::UnmatchedRequest {
+        ref path_id,
+        ref method,
+        ref content_type,
```
niclim, line 51: What's does the usage of the `ref` keyword do here? And why is it necessary?

workspaces/optic-engine/src/projections/learners/undocumented_bodies.rs
```diff
@@ -10,29 +10,69 @@ use crate::JsonTrail;
...
+      } = body_location
+      {
+        let expected_query_params_location =
+          BodyAnalysisLocation::UnmatchedRequestQueryParameters {
+            path_id: path_id.clone(),
+            method: method.clone(),
+            content_type: content_type.clone(),
```
niclim, line 58: I've seen usages of `clone()` for `&str` through our code. Wouldn't it be better to pass the references through (and make sure the lifetimes we are using are correct)? Or are we doing this to avoid specifying lifetimes + so we can get something functional out faster, or do we expect to mutate these in each usage (or something else that I'm missing)

niclim (commented):
Attempting to follow along + ask comments about rust to attempt to understand this part of the code base a little better

workspaces/optic-engine/src/projections/learners/undocumented_bodies.rs
JaapRood: In this case we need to clone, as we like the output type to own its data. That's the  most straightforward way to know it can be safely adapted, rewritten and serialised later on (even if JSON serde support serialising by reference). 

That said, it's also definitely because it's naive and less complicated way to do things at first, and nothing in the performance we've measured suggests it's worth the cost of optimising this (making it more complicated to write initially, but also change and maintain generally).

workspaces/optic-engine/src/projections/learners/undocumented_bodies.rs
JaapRood: It's to make sure `path_id`, `method` and `content_type` are captured as references to whatever is inside `body_location`, rather than moving them out (which is what would happen without, but I think you can also write without `ref` and instead match on `&body_location`). We want to do that, as `body_location` is used again later.

JaapRood:
I've pushed a version where a new `SetRequestQueryParametersShape` command is generated as part of the undocumented bodies learner for interactions that were observed to use query parameters. The command is able to be handled and generates an `RequestQueryParametersShape` event, which is handled by the internal `EndpointProjection` to track whether and what query parameters were set for each request.

I chose this approach for a single command and event after reviewing how the same was done for Request body shapes. Like the request body shape, there is a 1:1 relationship between the request and it's query parameters. The alternative of `QueryParametersAdded` and an additional commands to add this to the request, seems most appropriate when there isn't that 1:1 relationship. This does prevent re-use of query parameters across endpoints, but I reckon we would want to implement such a feature differently. @devdoshi relying on you to let me know if there's something I've missed here, you're the original author of the Optic spec after all and dealt with this / thought about this before.

Assuming I've made the right call here, I'm going to continue on the other side of this and tackle the parsing of query strings to a `BodyDescriptor`, so the learner can actually visit them.

workspaces/optic-engine/src/interactions/result.rs
```diff
@@ -229,6 +229,11 @@ pub enum BodyAnalysisLocation {
     method: String,
     content_type: Option<String>,
   },
+  UnmatchedRequestQueryParameters {
+    path_id: PathComponentId,
+    method: String,
+    content_type: Option<String>,
```
devdoshi, line 235: Seems like content_type is unnecessary

workspaces/optic-engine/src/events/http_interaction.rs
```diff
@@ -32,7 +32,7 @@ pub struct Request {
   pub path: String,
   pub headers: ArbitraryData,
   // #[serde(skip)]
-  pub query: ArbitraryData,
+  pub query: QueryParametersData,
```
devdoshi, line 35: What's the intent of this change? In the original design, the query string should be saved raw in asText, and either at capture time or pre-diffing should be parsed by a pluggable parser into a JSON object. Currently we have 2 capture sources; the local cli proxy and the rust proxy, with more coming. I would probably not use the existing query string parser implementation the proxy is using, and instead use the node url module by default but I would leave a pathway for it to be changed. The contract is that the query string should parse into a diffable object, so it can use the existing diffing logic and commands (aside from the new command to replace the root shape ID)

workspaces/optic-engine/src/events/http_interaction.rs
```diff
@@ -103,6 +109,20 @@ impl From<&ArbitraryData> for Option<BodyDescriptor> {
   }
 }
 
+impl From<&QueryParametersData> for Option<BodyDescriptor> {
+  fn from(query_param_data: &QueryParametersData) -> Self {
+    let data = &query_param_data.data;
+
+    if let Some(query_string) = &data.as_text {
```
devdoshi, line 116: What's the intent here?

devdoshi:
@JaapRood In the original design, adding endpoints should have also emitted RequestParameterAddedByPathAndMethod, ShapeAdded, and RequestParameterShapeSet events, enforcing the shape for a given (path * method)'s query string. This seems like it didn't make it out of the scala implementation: (https://github.com/opticdev/optic-core/blob/1c3a041987374b7005a0a3824c2d82be55138e5a/core/optic/shared/src/main/scala/com/useoptic/contexts/requests/RequestsAggregate.scala#L93)

The query string object would be empty by default, so the docs UI might elect to not show it when it is empty

Updating a shape should be done through the existing shapes commands when editing a shape, and something like RequestParameterShapeSet when changing the root shape (if that would ever be necessary) (https://github.com/opticdev/optic-core/blob/1c3a041987374b7005a0a3824c2d82be55138e5a/core/optic/shared/src/main/scala/com/useoptic/contexts/requests/Commands.scala#L57)

The touch points I see are:
- [ ] ensuring capturing saves at least the raw query string
- [ ] ensuring diffing has access to a parsed query string asJsonString or asShapeHashV1Base64
- [ ] diff visitors can resolve the expected RequestParameter shapeId to visit the query string RequestParameter and emit diffs with trails indicating the location is the query string
- [ ] learning endpoints should be able to set the initial query string shape - you'll probably have to use the shapeId you allocate during AddRequest to make changes, or perhaps have commands for manipulating the root shape, or another option.
- [ ] changelog projections and queries should reflect changes in query string shapes
- [ ] spectacle queries should expose query string resolvers
- [ ] ui docs page should show query string shapes
- [ ] ui diffs screen should show diffs for query strings and suggestions for changing the shape
- [ ] more?

Since we did implement most of this once before, it might help to dig up those commits; let me know.

workspaces/optic-engine/src/events/http_interaction.rs
JaapRood: It's a simple wrapper around `ArbitraryData` (flattened for serde, so invisible to the outside), so we can implement a separate way of parsing the `asText` of a query parameter to a `BodyDescriptor`. So why even bother with that, if the query string is also available as shape hash, json?

During the exploration and forming of the bet, we ran into the giant tyre fire that is the world of parsing query params, with sometimes not even respecting the RFC's for safe urls, using different key-value delimiters, etc. It's obvious that at some point, the user should be able to provide their own parsing strategy implementation, while the Optic one should be the most base one that makes sure all params make it into the documentation, but all as flat strings.

For the current bet, we're strictly on taking that most base way of parsing query parameters, with no other kind of structure supported. Not only to limit the complexity on the domain side, but also all the UI and UX work needed to establish base support for query parameters.

Now this was only a drat, but thinking ahead of how we would eventually move to a world where query parameters are parsed outside from the domain (like what is currently happening in the capture proxy with the `DefaultQueryParser`), I reckoned it might make sense to:

- For now only read query parameters from `asText` and perform the most base way of parsing query parameters (building a `Vec<(String, String)>` to represent all the entries) on the Rust side. That way, we know no matter the source of the capture, our base way of parsing query parameters is the same and known to the Engine. We'll know we won't get any more complicated shapes, we'll know what to expect building our first UI, etc.
- Later, start attempt to read query parameters from the ShapeHash or Json first, and treating that as custom parsed query strings (even if those parsers are default Optic ones). By that point we should expect pretty much any shape, with the additional UI and UX work on that to follow as well.

Writing this down now, I do wonder if that simple mechanism for choosing default parser vs. custom parser will work for non-local captures, where only the shape hash might be the acceptable way of transmitting that data. Perhaps not. If that's the case, we might have to come up with something more elaborate. Regardless, I reckon it's a good idea for the Rust codebase to be the place where the original query string -> parameters parser lives, just to make sure it's consistent everywhere (Node qs parsers have weird idiosyncrasies as well, as do the ones in pretty much any other language, and probably the Rust one too, but the more compelling the idea to make sure to use just one implementation).

workspaces/optic-engine/src/events/http_interaction.rs
JaapRood: See this other https://github.com/opticdev/optic/pull/937#discussion_r660377516.

JaapRood:
>  In the original design, adding endpoints should have also emitted RequestParameterAddedByPathAndMethod, ShapeAdded, and RequestParameterShapeSet events, enforcing the shape for a given (path * method)'s query string. This seems like it didn't make it out of the scala implementation:

Apart from the names and shape being different `RequestParameterShapeSet ` and `RequestQueryParametersShapeSet` are conceptually the same and do the same thing (waives hands). You'll find that a `ShapeAdded` command is also generated to represent the query shape as an object (also perhaps not by default, only when query parameters were learned, something we can tweak). The `RequestParameterAddedByPathAndMethod` is like the `QueryParameterAdded` event I suggested in [this comment](https://github.com/opticdev/optic/pull/937#issuecomment-869581006):

> I chose this approach for a single command and event after reviewing how the same was done for Request body shapes. Like the request body shape, there is a 1:1 relationship between the request and it's query parameters. The alternative of QueryParametersAdded and an additional commands to add this to the request, seems most appropriate when there isn't that 1:1 relationship. This does prevent re-use of query parameters across endpoints, but I reckon we would want to implement such a feature differently. @devdoshi relying on you to let me know if there's something I've missed here, you're the original author of the Optic spec after all and dealt with this / thought about this before.

I of course looked at the old Scala implementation as well, hence the similarities conceptually ☺️. The changes I've made came from the impression that some of these events used were from an older design, and perhaps made more general than we needed them to be, plus my own understanding of how to put the domain together. For example, `RequestParameterAddedByPathAndMethod` didn't seem to be used by anything else than query params, and query params are such a key part of most HTTP API's, it seemed to me they deserved their own event for assigning them to requests. They are subtly different in implementation, so overloading the one event seemed like maybe something we didn't want to continue with. 

That said, I obviously miss a whole year of context and deeper thought on that there, so I'm more than happy to take your guidance on that. All these names and specific shapes should be easy enough to change (thank you Rust 🦀). Just wanted to get something to get us going and form a base of discussion!

Finally, thanks for all the touch points, that aligns with our understanding of what needs to be done and the approach we're taking. This PR specifically is about the base Optic Engine support for the required events and allowing the undocumented bodies learner to create them, we're taking the rest on separately.

JaapRood:
The version I just pushed features the basic query string parser and a working learner tested with two end-to-end tests. @niclim this includes a fix for the crash you saw with #943, which was caused by the learner not dealing properly with requests without observed query parameters.

Unfortunately, the two e2e tests don't actually snapshot anything yet: the unstable ordering in the learning of shapes still makes it so that generated id's change on just about every run. That's an issue we really need to sort out, but probably as a separate PR from this one.

---

With the basic implementation of the learner working from end to end, and work on the UI starting to become blocked by it,  I want to focus on getting this PR merged, dealing with any further discussions (parsing strategies, command and events design, testing strategies, etc.) in separately. As long the learning of query parameters is hidden behind a feature flag, that should be totally fine. 

Given that, hiding the learning of query params behind a feature flag will be my next focus.

JaapRood:
I pushed a version with a temporary mechanism to disable the learning of query parameters, governed by a feature flag that's disabled by default. I've verified manually that with the feature flag not present or set to anything else but `true`, no query params are parsed, visited or have commands / events generated for them. That should prevent any of these new commands and events making into the wild, which should let us merge this even while the discussion on that is pending.

@niclim especially with you work on the UI being potentially blocked by this, I'm happy to have this merged will I'm off, given a sanity check of whether the feature flag is in place correctly. We'll have to keep in mind that some of the details might still shift from revisiting some of these topics, but should be preferable than just waiting for it all to land (especially as building the rest of what we need might inform some of these decisions!).

niclim:
I checked the FF is in place by:
- pulling the repo + building
- Learning an endpoint with ```          "query": {
            "shapeHashV1Base64": null,
            "asJsonString": null,
            "asText": null
          },
``` and with a shapwHasV1Base64 value
- These were learnt successfully

niclim (approved):

PR #939, opened 2021-06-25
Add cloud viewer tag on release instead of master

niclim:
## Why

It doesn't look like cloudviewer is getting deployed. I'm not 100% sure i follow the deployment pipeline - but I think that this fix should deploy the latest version of cloud viewer

This targets release (and I'll merge this back into develop)

## What

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/.github/workflows/build-ui-v2.yml
+++ b/.github/workflows/build-ui-v2.yml
@@ -51,7 +51,7 @@ jobs:
           echo "USER=${GITHUB_ACTOR}" >> $GITHUB_ENV
       - run: |
           task ui-v2:ci
-          if [ "$BRANCH" = "develop" ] || [ "$BRANCH" = "master" ]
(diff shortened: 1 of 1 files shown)


niclim:
As an aside, we should fix the regression build errors in the test regression scenarios

notnmeyer:
this workflow doesn’t deploy anything. it just pushes a docker image to our registry. historically, i have been against allowing this repo access to k8s

 i’m afk at the moment so cant refresh my memory on deploying it, but let’s plan on following up either later today or monday and we’ll sort it out.

niclim:
Ah ok - I'll close this Pr since this is not the correct fix - and lets follow up later today

LouManglass (approved):
This makes sense, as does the merge flow you suggested. I'd wait for Nate's eyes, but it seems sensible based on what I read.

Issue #929, opened 2021-06-24
Null content type crashes the review diff flow


jshearer:
I'm getting a crash in [new-regions-interpreter.ts](https://github.com/opticdev/optic/blob/develop/workspaces/ui-v2/src/lib/new-regions-interpreter.ts#L68) when I try to learn a new shape for an existing endpoint. It looks like `location.inRequest.contentType` is undefined.

<img width="1036" alt="Screen Shot 2021-06-24 at 4 02 53 PM" src="https://user-images.githubusercontent.com/4368270/123325128-a681a880-d505-11eb-986b-77996d03858f.png">

Here's the debug file with the interaction in question:

[debug-optic-1624564000735-tame-wolverine-62.json.zip](https://github.com/opticdev/optic/files/6711805/debug-optic-1624564000735-tame-wolverine-62.json.zip)

niclim:
A fix for the diff page crashing is in this PR https://github.com/opticdev/optic/pull/940 - however there's more work to do to handle the diff found in this capture - i.e.
- A GET request was made with a request body {} + learnt
- A subsequent GET request with no request body was sent (with null) - learning this causes two endpoints to show up (with the same id, i.e. same pathId and method

LouManglass:
The fix for the crash has shipped, minimum version 10.1.1. We'll let this sit open to resolve the rest of the issue.

niclim:
This is now fixed with the latest version of optic - `10.2.0`

PR #940, opened 2021-06-25
Handle content type null crashing diff flow

niclim:
## Why

@jshearer and I paired on this. It's possible to have content type null captured here - updating the UI to allow for this.

While working on this - the specific capture is interesting (from here https://github.com/opticdev/optic/issues/929) because:
- A GET request was made with a request body `{}` + learnt
- A subsequent GET request with no request body was sent (with null) - learning this causes two endpoints to show up (with the same id, i.e. same pathId and method
Screen Shot 2021-06-25 at 11 18 31 AM

Technically this should result in the request body being removed as a diff instead of creating a new AddRequest command. This could also be part of the manual edit work (depending on how we want to specify edits, if we go through the diff flow + with edits, or use similar components). Something like...
Screen Shot 2021-06-25 at 11 23 46 AM

## What

Don't crash the page if the diff has a null content type

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/ui-v2/src/lib/new-regions-interpreter.ts
+++ b/workspaces/ui-v2/src/lib/new-regions-interpreter.ts
@@ -65,7 +65,7 @@ function newContentType(
         shapes: [],
         copy: [
           plain('Document'),
-          code(location.inRequest!.contentType!.toString()),
+          code(location.inRequest!.contentType || 'null'),
           plain('Request'),
         ],
       },


jshearer (approved):
This change should fix the one UI crash we saw. I'm still a little concerned about that `cannot destructure ...` error, but since we couldn't reproduce that again.. it's probably fine?

niclim:
> This change should fix the one UI crash we saw. I'm still a little concerned about that `cannot destructure ...` error, but since we couldn't reproduce that again.. it's probably fine?

I think that was because we were screwing around with the capture (passing in an invalid shapeHash) which caused rust to crash and return undefined (which is why it couldn't destructure anything)

PR #941, opened 2021-06-28
Add the SDK implementations to the documentation

ukmadlz:
## Why
New SDKs require documenting as alternative capture implementations

## What
The express and hapi mdx files contain the additional references materials


website/docs/integrations/frameworks/express.mdx
```diff
@@ -4,11 +4,15 @@ sidebar_label: Express
 slug: /express
 ---
 
-Optic is easy to use with Express JS, no special library or large code changes required.
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+import GitHubButton from 'react-github-btn';
+
+Optic is really easy to use with Express JS, you can either use the [`api start` command](#api-start-command) without changing your application, or you can use [`api exec` and use middleware](#middleware-and-api-exec-command) without changing how you interact with your application:
```
acunniffe, line 11: I like how you frame this re: changing how your interact or changing code. Good way of stating tradeoffs. I'd maybe have users run `api start` for middlewares too but use one of the aliases that's supported 

```
start:
	command: npm start
```

website/docs/integrations/frameworks/hapi.mdx
```diff
@@ -51,6 +55,68 @@ Should you need any more help, or want to chat about the process, please reach o
...
+```sh
+npm install @useoptic/hapi-middleware
+```
+
+  </TabItem>
+</Tabs>
+
+### Configure the middleware
+
+The middleware accepts a configuration object with the following options:
```
acunniffe, line 88: Want might to pull this out and do a separate MDX file for it

acunniffe: Since it's reused

acunniffe (approved):
Looks good, suggested a few small things. I like the wording you use to contrast the middlewares from the proxy

website/docs/integrations/frameworks/express.mdx
ukmadlz: I thought the `OPTIC_LOGGING_URL` was only set and the capture endpoint started as part of the `api exec` flow?

acunniffe: if you do `api run <task>` where that task only has `command` specified, it's equivalent, and routes to `api exec <command>` under the hood.

ukmadlz: Added a note section https://github.com/opticdev/optic/pull/941/files#diff-9574856e1c965304e22bafdf49d8ace52de923bf88e79088b05c6dab34f23b03R9 to point this out, so the concepts appear separate till the user reads further without getting confused

PR #942, opened 2021-06-28
Change path parameter matcher to use a fn instead of pathToRegexp lib

niclim:
## Why

There's a number of cases where we need to shim the path and modify it for use in the pathToRegexp lib. There's also cases where this doesn't work and the pathToRegexp lib will incorrectly match on paths where a user has not specified a path parameter - specifically in the case where a user sends traffic of `/api/todos/:some-hard-coded-name` - even if the user does not specify `:some-hard-coded-name` as a parameter, pathToRegexp cannot differentiate this. A colon also crashes the diff page when trying to create a parameterized pathToRegexp expression

Previously, this was crashing the page when trying to parameterize (also creating an incorrect regexp if `:feed` was not specified as a path param)
Screen Shot 2021-06-28 at 9 59 46 AM

As a result, it's easier for us to write our own custom path matcher, based on user input (using `PathComponentAuthoring` and doing an array match)

## What

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


acunniffe (approved):

devdoshi (commented):
looks good and simpler too!

PR #943, opened 2021-06-28
Query connect spectacle to rust

niclim:
## Why

Connects the Rust intepretation of query parameter events to spectacle - uses the `REACT_APP_FF_LEARN_UNDOCUMENTED_QUERY_PARAMETERS` FF to gate showing these events

## What

- Handle with_query_parameters for the spectacle endpoint projection
- Exposes QueryParameterNode through graph-lib, spectacle
     
## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/spectacle/src/spectacle.ts
+++ b/workspaces/spectacle/src/spectacle.ts
@@ -312,12 +312,8 @@ export async function makeSpectacle(opticContext: IOpticContext) {
       method: (parent: endpoints.RequestNodeWrapper) => {
         return Promise.resolve(parent.value.httpMethod);
       },
-      query: async () => {
-        // TODO connect up to rust code
-        return {
-          rootShapeId: 'shape-1',
-          isRemoved: false,
-        };
+      query: async (parent: endpoints.RequestNodeWrapper) => {
+        return parent.query();
       },
       bodies: (parent: endpoints.RequestNodeWrapper) => {
         return Promise.resolve(parent.bodies().results);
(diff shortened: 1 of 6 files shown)


workspaces/spectacle/src/spectacle.ts
```diff
@@ -312,12 +312,11 @@ export async function makeSpectacle(opticContext: IOpticContext) {
...
         return Promise.resolve(parent.value.httpMethod);
       },
-      query: async () => {
-        // TODO connect up to rust code
-        return {
-          rootShapeId: 'shape-1',
-          isRemoved: false,
-        };
+      query: async (parent: endpoints.RequestNodeWrapper) => {
+        return process.env.REACT_APP_FF_LEARN_UNDOCUMENTED_QUERY_PARAMETERS ===
```
niclim, line 316: @JaapRood put this behind the same FF for simplicity of toggling - but if we decide we want to release learning separate from the UI, we can change this FF

niclim:
Ok this is ready to be reviewed + merged, going to branch off this PR which should include everything we need to build out the UI components (with mocked out events).

The UI work to learn query parameters on the diff flow looks fairly minimal (barring complications) - so the majority of the UI work looks like it'll be in building out the UI components for query parameters

workspaces/spectacle/src/spectacle.ts
JaapRood: Good call. Hopefully that won't be necessary at all (doesn't look like it) :)

JaapRood (approved):
👍

PR #945, opened 2021-06-29
verbose flag to replace api check

acunniffe:
## Why
As part of our rethinking of on-boarding and docs, @LouManglass and I identified `api check` as a major point of friction. 

We wanted to explore a possible `--verbose` flag that printed the same lifecycle information as your task ran normally.

This is A) useful during on-boarding and B) might be generally useful to leave on. It's a lot less noisy than `DEBUG=optic*`

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


devdoshi:
It would be nice to see a video illustrating the differences. but looks good!

PR #946, opened 2021-06-29
Add test command as alias

smizell:
## Why

Running tests is one of the most important steps while using Optic, especially in the build pipeline. We want to make it easy to set up testing for people.

## What

This adds a special `test` command that will allow people to start up Optic, their API, run their tests, and handle any non-0 exit codes in CI to stop the build. The specifics are:

* Adds a test command `api test` which is an alias for `api run test --ci`
* Modifies the `--ci` flag to include `--exit-on-diff`
* Fixes some of the spacing in the `optic.yml` init output
* Adds a test task to the `optic.yml` on `api init`

Once the docs are updated, I think a link from `opitc.yml` to the docs about testing would be great!

### DIFF
--- a/workspaces/local-cli/src/shared/local-cli-task-runner.ts
+++ b/workspaces/local-cli/src/shared/local-cli-task-runner.ts
@@ -102,6 +102,7 @@ export async function LocalTaskSessionWrapper(
     flags['print-coverage'] = true;
     flags['pass-exit-code'] = true;
     flags['collect-diffs'] = true;
+    flags['exit-on-diff'] = true;
   }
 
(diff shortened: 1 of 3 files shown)


workspaces/local-cli/src/shared/local-cli-task-runner.ts
```diff
@@ -102,6 +102,7 @@ export async function LocalTaskSessionWrapper(
     flags['print-coverage'] = true;
     flags['pass-exit-code'] = true;
     flags['collect-diffs'] = true;
+    flags['exit-on-diff'] = true;
```
smizell, line 105: This adds `--exit-on-diff` to the `--ci` flag.

acunniffe (approved):
Looks good to me

PR #948, opened 2021-06-29
Migrate endpoint bodies to redux

niclim:
## Why

This targets https://github.com/opticdev/optic/pull/943

Working through the query param stuff, the data fetching flows we are currently doing are a little odd. Figured it was a good time we could move this part of code into redux. There's heavy overlap between the data fetched in the endpoints data and the requests / response bodies (and later queries). This technically pulls in more data, but oddly enough we were doing that anyway (since spectacle does not currently have pagination, or id based queries for requests) - when spectacle does support this, we can be more targeted with our data fetching.

## What

- Moved useEndpointBodies data into the `IEndpoint` type + fetched this data in the `fetchEndpoints` thunk
- Removed useEndpointBodies and useAllRequestsAndResponses
- ~~Changed the type of `responses` -> `response` such that it's a single (nullable) object. It does not make sense to have multiple request bodies (polymorphic should be handled in shapes).~~

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


JaapRood (commented):
These changes are very welcome and they seem well implemented! One bit sticks out to me, though:

> Changed the type of responses -> response such that it's a single (nullable) object. It does not make sense to have multiple request bodies (polymorphic should be handled in shapes).

I'm not read up on all the details of _where_ you're applying this, but in general, we can't assume an endpoint as a single request definition. That's not to do with the polymorphism in the body (which is indeed handled by the shapes), but the fact that we document a separate Request for every content-type that's observed.

In practice, that often just means the single one: `application/json`, but so far that's not something we have wanted to assume. In some test projects I've used in the past, I have already seen multiple, where both XML or JSON could be sent. In another case, it was `application/json; charset=utf8` vs `application/json`. 

Does that have any impact on this?

niclim:
Hmmm you are right, and I totally forgot about this point. I'll reupdate the code to handle multiple request bodies

niclim:
Ok @JaapRood  - I've updated this to use requests so that multiple requests can be displayed

workspaces/ui-v2/src/hooks/useEndpointsBodyChanges.ts
```diff
@@ -0,0 +1,65 @@
...
+      added
+      changed
+    }
+    responses {
+      id
+      changes(sinceBatchCommitId: $sinceBatchCommitId) {
+        added
+        changed
+      }
+    }
```
JaapRood, line 19: I'm wondering whether `responses` being nested in `requests` is the right way to go here. There are very obviously no responses without requests (unless we allow for time travel 😅), but as with requests, there are multiple per status code and per content type. However, posting a request body with one content type, doesn't mean receiving the response in that content type (`Content-Type` describing the request body, `Accept` what it accepts in return). You do wonder whether such a disconnect actually is out there in the wild, so perhaps if it's massively inconvenient to have them separate, it's not worth the bother.

workspaces/ui-v2/src/store/endpoints/thunks.ts
```diff
@@ -63,10 +95,34 @@ export const endpointQueryResultsToJson = ({
...
+      rootShapeId: body.rootShapeId,
+      pathId: request.pathId,
+      method: request.method,
+      description: request.requestContributions.description || '',
+    })),
+    responseBodies: request.responses
+      .flatMap((response) => {
+        return response.bodies.map((body) => {
+          return {
+            statusCode: response.statusCode,
```
JaapRood, line 111: This `flatMap` -> `map` combo might indicate the same thing: that we should think of requests and responses separately. Haven't found the place where we actually use the fact that they're nested.

workspaces/ui-v2/src/hooks/useEndpointsBodyChanges.ts
JaapRood: Sorry for not spotting that in earlier review :(

JaapRood (approved):
Apart from the issue raised, this looks good. Whichever way that falls, should be ready to go.

workspaces/ui-v2/src/hooks/useEndpointsBodyChanges.ts
niclim: So this is how spectacle is currently defined - I agree it's confusing but here's how it maps
```
requests - refer to an endpoint (http method + path)
request.bodies - refer to the actual request bodies
request.responses - refer to the response tied to the endpoint
request.responses.body - refer to the response body
```
If anything, the top level item should probably be renamed `endpoint`

JaapRood: Makes sense!

PR #950, opened 2021-06-30
Learn a query parameters per Endpoint instead by each supported Request content type

JaapRood:
## Why
In the initial implementation of learning #937 we implemented the learning of query parameters per Request. While there seems a 1:1 mapping here, the Requests as modelled in the Optic Engine are per content type. That means that if an endpoint would support multiple request body content types, we'd spec a separate instance of query parameters as well. Instead, we want those to be singular per Endpoint.

## What

- [x] Introduction of command and event to register query parameters per Endpoint (path id + method)
- [x] Conversion of command and event to set query shape to be per newly added query parameters
- [x] Visiting of query parameter shapes per unmatched query parameters diffs, instead of unmatched request content type
- [x] Hiding of new query diff results from actual diff output (partially behind `REACT_APP_FF_DIFF_QUERY_PARAMETERS`, or fully disabled)
- [x] Including separate commands to learner result to document query parameters
- [x] Apply learned query parameters commands from the client when documenting a new endpoint (no longer included in the commands for requests).

## Validation
* [x] CI passes
* [x] Query parameters are learned for newly documented endpoints
* [x] No query parameter diffs are generated from diffing (yet!)


JaapRood:
The learner for undocumented bodies is now powered by new endpoint-level diffing of query parameters, so that they get learned for all interactions, instead of just those with request bodies. Since that's part of the overall diffing pipeline, I had to take care that those types of diffs don't leak beyond learning yet. For now, that's controlled through the `REACT_APP_FF_DIFF_QUERY_PARAMETERS` feature flag (although for calls from the WASM engine it's just fully disabled).

It highlighted how with `optic-engine-wasm` we have pretty little control over the JS interface that we're exposing, the way that it's currently setup. Basically, because Rust doesn't allow optional arguments, nor does the Typescript and JS side, with a newly introduced `options` argument for the `diff_interaction` becoming a breaking change (you can pass empty options, but you _have_ to pass options, because Typescript).

A better approach might be wrapping the generated Typescript / JS code by WASM bindgen, so we actually control the whole lot. Probably not something we have the appetite for now, but something that could prevent breaking a whole bunch of downstream consumers unnecessarily.

@niclim this should now be in a stable enough state where you could work off it, either contributing to or forking off into your own pull requests.

workspaces/optic-engine/src/interactions/result.rs
```diff
@@ -427,72 +488,74 @@ impl InteractionTrail {
...
+  //   #[derive(Default, Debug)]
+  //   struct InteractionIdentifiers<'a> {
+  //     path: Option<&'a String>,
+  //     method: Option<&'a String>,
+  //     request_content_type: Option<&'a String>,
+  //     response_content_type: Option<&'a String>,
+  //     response_status_code: Option<u16>,
+  //   }
+
+  //   impl<'a> From<&'a InteractionTrail> for InteractionIdentifiers<'a> {
```
JaapRood, line 507: I commented this out, as the new query parameters support to the `InteractionTrail` makes this more complicated, yet, I don't think this is being used at all anymore. I'll have to confirm, but pretty sure this was added as part of the shape diff affordances learner, but then abandoned. Rather than updating it, if it's dead code, we should get rid of it.

workspaces/optic-engine/src/interactions/result.rs
```diff
@@ -1,13 +1,14 @@
...
 use crate::shapes::{JsonTrail, ShapeDiffResult};
-use crate::state::endpoint::{PathComponentId, RequestId, ResponseId, ShapeId};
+use crate::state::endpoint::{PathComponentId, QueryParametersId, RequestId, ResponseId, ShapeId};
 use serde::{Deserialize, Serialize};
 use std::collections::hash_map::{DefaultHasher, HashMap};
 use std::hash::{Hash, Hasher};
 
 #[derive(Debug, Deserialize, Serialize, Hash)]
 pub enum InteractionDiffResult {
+  UnmatchedQueryParameters(UnmatchedQueryParameters),
```
JaapRood, line 11: This new variant of `InteractionDiffResult`, together with the interaction trail embedded in it, will let the client discern this is a diff in the query parameters. That will become important once we add support for that in the UI diff flow, so useful to be aware of :)

workspaces/optic-engine/src/interactions/result.rs
```diff
@@ -17,6 +18,8 @@ pub enum InteractionDiffResult {
   // Matches
   // -------
   #[serde(skip)]
+  MatchedQueryParameters(MatchedQueryParameters),
+  #[serde(skip)]
```
JaapRood, line 22: These `Matched*` types are only for internal use, so we can move on to shape diffing once there is an endpoint -> interaction match. In this case, that would mean that:

- the interaction's method and path were matched to a documented path.
- query parameters were documented for this path.
- query parameters were present on the interaction.

The latter is probably something that we'll want to revise once we get into the shape diffing logic: an interaction without query parameters could be considered as one with empty query parameters, which could indicate the documented query parameters are optional. We'll leave that until we're working on the diff flow.

JaapRood:
The most important in review would be making sure the new feature flag `REACT_APP_FF_DIFF_QUERY_PARAMETERS` is in place properly, which makes sure no query parameters related diffs make it to the diff flow for now. The learning of query parameters is already behind a feature flag. When hidden properly, we can get away with small issues and revisit those later.

JaapRood:
This should be ready to be merged, awaiting a passing review!

workspaces/cli-shared/src/diffs/initial-types.ts
```diff
@@ -1,18 +1,9 @@
...
-
-export interface ILearnedBody {
-  contentType: string;
-  statusCode?: number;
-  commands: any[];
-  rootShapeId: string;
-}
+export type ILearnedBodies = LearningResults.UndocumentedEndpointBodies.LearnedBodies;
+export type ILearnedBody = LearningResults.UndocumentedEndpointBodies.LearnedBody;
+export type ILearnedQueryParameters = LearningResults.UndocumentedEndpointBodies.LearnedQueryParameters;
```
niclim, line 6: 🎉

workspaces/optic-domain/src/streams/learning-results/undocumented-endpoint-bodies.ts
```diff
@@ -15,6 +16,11 @@ export interface LearnedBody {
   rootShapeId: string;
 }
 
+export interface LearnedQueryParameters {
+  commands: any[];
```
niclim, line 20: There's a CQRSCommands type in spectacle i believe (although I'm not sure if this will cause a circular dependency with this package)

workspaces/optic-engine/src/interactions/mod.rs
```diff
@@ -72,6 +81,27 @@ pub fn diff(
...
+  include_query_params: bool,
+}
+
+impl Default for DiffConfig {
+  fn default() -> Self {
+    Self {
+      include_query_params: false,
+    }
+  }
+}
```
niclim, line 95: Is there a reason we don't want to pass in `include_query_params` as an argument for `default` and instead mutate this via `with_query_params`?

workspaces/optic-engine/src/projections/learners/undocumented_bodies.rs
```diff
@@ -213,61 +202,84 @@ impl EndpointBody {
...
+        })
+      }
+      _ => panic!("EndpointBody should only be created for unmatched responses and requests"),
     }
   }
 
   fn append_endpoint_commands(&mut self, ids: &mut impl SpecIdGenerator) {
     match self {
+      EndpointBody::QueryParameters(query_parameters) => {
+        let query_params_id = ids.request(); // TODO: give query params their own prefix
```
niclim, line 263: Extra todo? Or is this done somewhere else?

niclim (approved):
I think this looks good to me - I'm not too familiar with the diff + learn bodies parts of the optic engine so I think it might be worth @devdoshi to take a second look over those parts

workspaces/optic-engine/src/interactions/result.rs
devdoshi: Yeah seems like a few things will need to change for the diff flow but things look good so far @niclim @JaapRood

workspaces/optic-engine/src/interactions/mod.rs
JaapRood: Yeah, that's because this is implementing the `Default` trait, which requires `default()` to be callable without any arguments. Hence the builder pattern: let those who don't care go with the default config, while allowing others to configure.

workspaces/optic-engine/src/projections/learners/undocumented_bodies.rs
JaapRood: Nice find, extra todo indeed!

workspaces/optic-domain/src/streams/learning-results/undocumented-endpoint-bodies.ts
JaapRood: There is, which should probably live in `optic-domain` so it can be used everywhere else. Lets document for a follow up PR for clean up.

workspaces/optic-engine/src/projections/learners/undocumented_bodies.rs
JaapRood: Resolved in https://github.com/opticdev/optic/pull/950/commits/1041cddb7e56381665ff6ce0e0a1d320edc68a0d

PR #951, opened 2021-06-30
Implement query parameter panel component

niclim:
## Why

Targets https://github.com/opticdev/optic/pull/948

This creates the UI component for a query parameter panel. This is used in 3 locations - changelog page, documentation page and diff review flow. There are other components (request / response body renderers, path parameters) that are used in these contexts and also have their own specialized version. This PR also attempts to build a more composable solution to these.

## What

Creates a QueryParameterPanel component with data fetcher / adaptor components (that can be removed if we move the data fetching into redux). 

This should all be gated by the `REACT_APP_FF_LEARN_UNDOCUMENTED_QUERY_PARAMETERS` FF - as no query parameters will be learnt or returned by spectacle.

Note the styling is currently inconsistent with the other request / response body panels (moved out header + description out of the two columns and different font families). What we'll want to do is migrate the other components to use these panels (rather than trying to emulate hard coded styles here) - ignore the `age1` field, was just testing the styling for multi row. We'll also want to adjust the styling when we render shapes (i.e. arrays, etc)

Screen Shot 2021-06-30 at 2 05 36 PM

There's a lot to follow this up. Here's an attempt to capture everything that should follow this (annotations in code related to this bet will be label `// TODO QPB` - i.e. query parameter bet

TODOS (stack ranked based on importance / sequence - might not get to everything as part of this bet):
- [ ] Update QueryParameterPanel to handle changelog renderering
- [ ] Implement QueryParameterPanel on the diff and changelog parts and handle relevant UI diff components
- [ ] Move BodyRender to components and make it use the `Panel` component
- [ ] Update PathParameters to be a PathParameterPanel and use the `Panel` component
- [ ] Move contributions into redux (split from main data models, i.e. endpoints / requests / shapes)
- [ ] Provide way to query changes per field / shape / body, etc
- [ ] Move shape accumulation into redux (requires changing the way we query for changes)
- [ ] Migrate + reduce number of usages of docs field params specializations
- [ ] Migrate current UI components to use the panel components created above (e.g. BodyPanel, PathParametersPanel, QueryParameterPanel)
- [ ] Consolidate color scheme (right now these are copy pasted from current usage)
- [ ] Reorganize `<src>/components` - we probably want a distinction between things like a `Button` and more specialized components like `QueryParameterPanel` - similar to the [atomic design](https://bradfrost.com/blog/post/atomic-web-design/) (naming tbd if we want to follow atoms / molecudes)

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/ui-v2/src/components/Panel.tsx
+++ b/workspaces/ui-v2/src/components/Panel.tsx
@@ -1,5 +1,6 @@
 import React, { FC } from 'react';
 import { makeStyles } from '@material-ui/core';
+import { FontFamilyMono } from '<src>/styles';
 
 type PanelProps = {
   header: React.ReactNode;
@@ -23,26 +24,22 @@ const useStyles = makeStyles((theme) => ({
     width: '100%',
   },
   header: {
+    display: 'flex',
     backgroundColor: '#e4e8ed',
     color: '#4f566b',
-    flex: 1,
-    fontSize: 13,
-    height: 35,
-    display: 'flex',
-    fontWeight: 400,
-    paddingLeft: 13,
-    fontFamily: 'Roboto',
+    fontSize: theme.typography.fontSize - 1,
+    height: theme.spacing(4),
+    paddingLeft: theme.spacing(2),
+    fontFamily: FontFamilyMono,
     alignItems: 'center',
+    overflow: 'hidden',
     borderTopLeftRadius: 8,
     borderTopRightRadius: 8,
-    overflow: 'hidden',
     borderBottom: '1px solid #e2e2e2',
   },
   content: {
     backgroundColor: '#f8fafc',
-    paddingTop: 8,
-    paddingBottom: 8,
-    borderTop: 'none',
+    padding: theme.spacing(1),
     maxHeight: '80vh',
     overflowY: 'auto',
     borderLeft: '1px solid #e4e8ed',
--- a/workspaces/ui-v2/src/components/Panel.tsx
+++ b/workspaces/ui-v2/src/components/Panel.tsx
@@ -1,0 +1,54 @@
+import React, { FC } from 'react';
+import { makeStyles } from '@material-ui/core';
+
+type PanelProps = {
+  header: React.ReactNode;
+};
+
+export const Panel: FC<PanelProps> = ({ children, header }) => {
+  const classes = useStyles();
+  return (
+    <div className={classes.wrapper}>
+      <div className={classes.header}>
+        <div>{header}</div>
+      </div>
+      <div className={classes.content}>{children}</div>
+    </div>
+  );
+};
+
+const useStyles = makeStyles((theme) => ({
+  wrapper: {
+    overflow: 'hidden',
+    width: '100%',
+  },
+  header: {
+    backgroundColor: '#e4e8ed',
+    color: '#4f566b',
+    flex: 1,
+    fontSize: 13,
+    height: 35,
+    display: 'flex',
+    fontWeight: 400,
+    paddingLeft: 13,
+    fontFamily: 'Roboto',
+    alignItems: 'center',
+    borderTopLeftRadius: 8,
+    borderTopRightRadius: 8,
+    overflow: 'hidden',
+    borderBottom: '1px solid #e2e2e2',
+  },
+  content: {
+    backgroundColor: '#f8fafc',
+    paddingTop: 8,
+    paddingBottom: 8,
+    borderTop: 'none',
+    maxHeight: '80vh',
+    overflowY: 'auto',
+    borderLeft: '1px solid #e4e8ed',
+    borderRight: '1px solid #e4e8ed',
+    borderBottom: '1px solid #e4e8ed',
+    borderBottomRightRadius: 4,
+    borderBottomLeftRadius: 4,
+  },
+}));
--- a/workspaces/ui-v2/src/components/Panel.tsx
+++ b/workspaces/ui-v2/src/components/Panel.tsx
@@ -1,4 +1,5 @@
 import React, { FC } from 'react';
+import classNames from 'classnames';
 import { makeStyles } from '@material-ui/core';
 import { FontFamilyMono } from '<src>/styles';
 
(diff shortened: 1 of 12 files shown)


workspaces/ui-v2/src/components/QueryParametersPanel.tsx
```diff
@@ -0,0 +1,63 @@
...
+  return queryParameters;
+};
+
+export const QueryParametersPanel: FC<QueryParametersPanelProps> = ({
+  parameters,
+}) => {
+  const classes = useStyles();
+  return (
+    // TODO QPB add in query parsing strategy here?
+    <Panel header={''}>
```
niclim, line 36: Here we can probably have something informational about:
- How this query parameter was parsed
- How you can structure an example request (or maybe this fits into the showExample feature)

workspaces/ui-v2/src/pages/docs/EndpointRootPage.tsx
```diff
@@ -229,13 +234,64 @@ export const EndpointRootPage: FC<
...
               </div>
             </CodeBlock>
           }
         />
+        {thisEndpoint.query && (
+          <div className={classes.bodyContainer} id="query-parameters">
+            <h6 className={classes.bodyHeader}>Query Parameters</h6>
+            <div className={classes.bodyDetails}>
+              <div>
+                {/* TODO QPB - change id from this to query id from spectacle */}
```
niclim: @JaapRood - here we'll need an id for a query parameter body or equivalent.

workspaces/ui-v2/src/pages/docs/components/DataFetchers.tsx
```diff
@@ -0,0 +1,56 @@
+import React, { FC } from 'react';
+import { createFlatList } from '<src>/components/FieldOrParameter';
+import { useShapeDescriptor } from '<src>/hooks/useShapeDescriptor';
+import { IShapeRenderer } from '<src>/components';
+
+type ContributionFetcherProps = {
+  rootShapeId: string;
+  endpointId: string;
+  // TODO QPB change this typing - currently this holds contributions, and the data for rendering the shape
```
niclim, line 9: This is confusing because we have two typings for contributions - `IContribution` - which is raw contribution information and `IContributions` which is a mash of contribution, and field data. The `FieldOrParameter` component needs to be changed (i.e. a more scoped interface) in order for us to remove this adhoc typing that is difficult to reason about

workspaces/ui-v2/src/pages/docs/components/DataFetchers.tsx
```diff
@@ -0,0 +1,56 @@
...
+      value: string;
+      endpointId: string;
+      depth: number;
+      name: string;
+      shapes: IShapeRenderer[];
+    }[]
+  ) => React.ReactNode;
+};
+
+// TODO QPB replace this by fetching the contributions in redux
```
niclim, line 23: A progression towards having more generic, composable components. What needs to happen next is:
- Build these fetchers for changelog  + diff pages + other components
- Migrate shapes / etc into redux
- rip out these data adaptors and pass in data directly from redux

workspaces/ui-v2/src/components/Panel.tsx
```diff
@@ -0,0 +1,51 @@
...
+      <div className={classes.content}>{children}</div>
+    </div>
+  );
+};
+
+const useStyles = makeStyles((theme) => ({
+  wrapper: {
+    overflow: 'hidden',
+    width: '100%',
+  },
```
JaapRood, line 25: These kinds of wrappers with full widths set and overflows to hidden generally worry me a bit, but I'm happy that more work needs to happen on the styling of these anyway :)

workspaces/ui-v2/src/components/Panel.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { FontFamilyMono } from '<src>/styles';
+
+type PanelProps = {
+  header: React.ReactNode;
+};
+
+export const Panel: FC<PanelProps> = ({ children, header }) => {
+  const classes = useStyles();
+  return (
+    <div className={classes.wrapper}>
```
JaapRood, line 12: Not something to instantly take action on, but more generally: given the way we're styling things, one really useful little pattern I've found is to accept `className` as a prop for general components like this, and add that to the wrapper div (in addition to the internal one). It allows the user of the component to control things like it's layout, for example child specific flex-box properties. `classnames` is an easy way to join internal css names with one optional one provided through the props.

workspaces/ui-v2/src/components/QueryParametersPanel.tsx
```diff
@@ -0,0 +1,63 @@
...
+        </div>
+      ))}
+    </Panel>
+  );
+};
+
+const useStyles = makeStyles((theme) => ({
+  queryComponentContainer: {
+    marginBottom: theme.spacing(1),
+    padding: `${theme.spacing(1)}px 0`,
```
JaapRood, line 52: Hot little nit tip: `theme.spacing` accepts anywhere between 1 to 4 arguments, allowing you to write `theme.spacing(1, 0)` here :)

workspaces/ui-v2/src/pages/docs/EndpointRootPage.tsx
```diff
@@ -286,4 +344,27 @@ const useStyles = makeStyles((theme) => ({
...
+    lineHeight: 1.6,
+    letterSpacing: '0.0075em',
+    margin: `${theme.spacing(3)}px 0`,
+  },
+  bodyDetails: {
+    display: 'flex',
+    width: '100%',
+    '& > div': {
+      width: '50%',
+      padding: `0 ${theme.spacing(1)}px`,
```
JaapRood, line 367: Another candidate for `theme.spacing(0, 1)`

JaapRood (approved):
Looking good! Happy to see incremental steps here, so we can adapt how far we get into this refactoring to the appetite we have left.

workspaces/ui-v2/src/components/Panel.tsx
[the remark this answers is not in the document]
niclim: Yeah there's a lot of copy pasted styles from previous implementations, and the style in general needs work

workspaces/ui-v2/src/components/Panel.tsx
```diff
@@ -0,0 +1,51 @@
...
+  content: {
+    backgroundColor: '#f8fafc',
+    padding: theme.spacing(1),
+    maxHeight: '80vh',
+    overflowY: 'auto',
+    borderLeft: '1px solid #e4e8ed',
+    borderRight: '1px solid #e4e8ed',
+    borderBottom: '1px solid #e4e8ed',
+    borderBottomRightRadius: 4,
+    borderBottomLeftRadius: 4,
```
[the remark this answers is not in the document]
niclim: Ah this helps a lot, probably shows I should read the documentation!

workspaces/ui-v2/src/components/Panel.tsx
niclim: Ah yeah, I'll update this - I added this with some other components but I think I got lazy here

PR #952, opened 2021-06-30
Add query param panel to changelog page

niclim:
## Why

This targets https://github.com/opticdev/optic/pull/951

Adding in the query param component for the changelog flow. I'm not sure why the changelog doesn't work (the todo-partial example has weird events that don't generate a changelog and I haven't had a chance to dig into this) - we'll definitely want to verify this when we have a more e2e pass for generating events.

Here's what it _would_ look like with correct events - there's better styling we can add here (and a background for an overall field added might not make sense if the shape renderer doesnt have a green background as well)
Screen Shot 2021-06-30 at 4 07 29 PM

## What

- Moved datafetchers (which should be temporary) from `<src>/pages/docs/components` -> `<src>/components`
- Fetch changelog information in useShapeDescriptor

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/ui-v2/src/components/QueryParametersPanel.tsx
+++ b/workspaces/ui-v2/src/components/QueryParametersPanel.tsx
@@ -1,8 +1,15 @@
 import React, { FC } from 'react';
 import { makeStyles } from '@material-ui/core';
+import classnames from 'classnames';
+
+import {
+  FontFamily,
+  AddedGreenBackground,
+  ChangedYellowBackground,
+  RemovedRedBackground,
+} from '<src>/styles';
 import { IFieldRenderer, IShapeRenderer, ShapeRenderer } from './ShapeRenderer';
 import { Panel } from './Panel';
-import { FontFamily } from '<src>/styles';
 
 type QueryParameters = Record<string, IFieldRenderer>;
 
@@ -16,7 +23,10 @@ export const convertShapeToQueryParameters = (
 ): QueryParameters => {
   const queryParameters: QueryParameters = {};
   if (shapes.length !== 1 || !shapes[0].asObject) {
-    console.error('unexpected format for query parameters');
+    if (shapes.length > 1) {
+      console.error('unexpected format for query parameters');
+    }
+    // otherwise loading
     return {};
   }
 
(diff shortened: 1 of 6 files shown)


workspaces/ui-v2/src/components/QueryParametersPanel.tsx
```diff
@@ -60,4 +76,13 @@ const useStyles = makeStyles((theme) => ({
...
   },
+  added: {
+    backgroundColor: `${AddedGreenBackground}`,
+  },
+  changed: {
+    backgroundColor: `${ChangedYellowBackground}`,
+  },
+  removed: {
+    backgroundColor: `${RemovedRedBackground}`,
+  },
```
JaapRood, line 87: 👏

workspaces/ui-v2/src/pages/changelog/ChangelogEndpointRootPage.tsx
```diff
@@ -51,7 +56,7 @@ const ChangelogRootComponent: FC<
   }>
 > = ({ match }) => {
   const { pathId, method, batchId } = match.params;
-  const styles = useStyles();
+  const classes = useStyles();
```
JaapRood, line 59: 🎉 I appreciate people that appreciate consistency :)

workspaces/ui-v2/src/pages/changelog/ChangelogEndpointRootPage.tsx
```diff
@@ -207,4 +257,27 @@ const useStyles = makeStyles((theme) => ({
...
+  bodyHeaderContainer: {
+    marginBottom: theme.spacing(2),
+  },
+  bodyHeader: {
+    fontSize: '1.25rem',
+    fontFamily: FontFamily,
+    fontWeight: 500,
+    lineHeight: 1.6,
+    letterSpacing: '0.0075em',
+    margin: `${theme.spacing(3)}px 0`,
```
JaapRood, line 273: Could be written as `theme.spacing(3, 0)` ☺️

JaapRood (approved):

PR #953, opened 2021-06-30
Add in query parameter renderer in diff preview

niclim:
## Why

Targets https://github.com/opticdev/optic/pull/952

Add in query param panel to diff preview flow.

Here's where it is - the styling needs work (there's a large amount of white space which is because the other components already include margins in different spots), but I am going to see if i can convert some code / components to use the shared components I just built. Again this is behind a FF so we shouldn't get any query params coming through right now
Screen Shot 2021-06-30 at 4 38 02 PM

## What

- Minor refactor of the `HighlightLocation` component - now supports highlighting query parameter changes
- Add to review diff endpoint preview

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/ui-v2/src/lib/Interfaces.ts
+++ b/workspaces/ui-v2/src/lib/Interfaces.ts
@@ -186,6 +186,7 @@ export interface IResponseBodyLocation {
 export interface IParsedLocation {
   pathId: string;
   method: string;
+  inQuery?: boolean;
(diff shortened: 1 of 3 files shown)


JaapRood (approved):
Another great incremental step!

> (there's a large amount of white space which is because the other components already include margins in different spots)

Good example of how I mean that layout is the "glue" in between re-usable components, specific to the context that it's used :)

PR #954, opened 2021-07-01
Refactor documentation page to use HttpBodyPanel

niclim:
## Why

Targets: https://github.com/opticdev/optic/pull/953

Now that we've built a way to compose these, we can get start ripping out the 3 specialized components that do the exact same thing. Also took the liberty to make the styling slightly more consistent (using our defined font family, and removing hard coded spaces - replacing with css, etc)

## What

- Refactor documentation page to use composable components (doesn't use the render body and looks more consistent)
Screen Shot 2021-06-30 at 5 24 45 PM

- Refactor changelog page to use the same components as above
- Refactor endpoint diff review page (as an aside, this looks odd on large screens - i checked out develop and it looks like this is has been the case)
Screen Shot 2021-06-30 at 6 18 45 PM

- Got rid of the very confusingly named RenderBody component (BodyRender will have to wait until the next PR!)

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


workspaces/ui-v2/src/pages/changelog/ChangelogEndpointRootPage.tsx
```diff
@@ -213,28 +213,111 @@ const ChangelogRootComponent: FC<
...
+            <div className={classes.bodyHeaderContainer}>
+              <h6 className={classes.bodyHeader}>Request Body</h6>
+              <ReactMarkdown
+                className={classes.contents}
+                source={requestBody.description}
+              />
+            </div>
+            <div className={classes.bodyDetails}>
+              <div>
+                <ContributionFetcher
```
niclim, line 230: The biggest callout here is that we use `ContributionFetcher` and `ShapeFetcher` which _both_ call `useShapeDescriptor` on the same root shape - meaning there's double the spectacle calls. This _shouldn't_ be an issue since spectacle is either in memory or hitting a local cli network. And this should be temporary since I'm planning to move this into redux (the idea here is that if these are split, it's much easier to incrementally split these out)

workspaces/ui-v2/src/pages/changelog/ChangelogEndpointRootPage.tsx
```diff
@@ -260,6 +343,7 @@ const useStyles = makeStyles((theme) => ({
   bodyContainer: {
     marginTop: theme.spacing(6),
     width: '100%',
+    height: '100%',
   },
```
JaapRood, line 347: These kinds of props on containers worry me a bit (most likely a `min-height: 100vh` is what's actually wanted), but happy if it's not doing anything particularly funky in this case, a bit of code that can survive another day!

JaapRood, line 347: These kinds of props on containers worry me a bit (most likely a min-height: 100vh is what's actually wanted), but happy if it's not doing anything particularly funky in this case, a bit of code that can survive another day!

workspaces/ui-v2/src/pages/diffs/EndpointDocumentationPane.tsx
```diff
@@ -108,15 +110,15 @@ export const EndpointDocumentationPane: FC<
...
 
       {thisEndpoint.query && (
         <HighlightedLocation
+          className={classes.bodyContainer}
           targetLocation={highlightedLocation}
           expectedLocation={Location.Query}
         >
           {/* TODO QPB - change id from this to query id from spectacle */}
-          <div className={classes.bodyContainer} id="query-parameters">
+          <div id="query-parameters">
```
JaapRood, line 121: I assume this `id` is here to allow deep / fragment links to it?

workspaces/ui-v2/src/pages/docs/EndpointRootPage.tsx
```diff
@@ -347,6 +433,7 @@ const useStyles = makeStyles((theme) => ({
   bodyContainer: {
     marginTop: theme.spacing(6),
     width: '100%',
+    height: '100%',
```
JaapRood, line 436: Some more slightly suss container css properties here. If a `min-height: 100vh` achieves the same thing, that would be preferable!

JaapRood (approved):

workspaces/ui-v2/src/pages/changelog/ChangelogEndpointRootPage.tsx
niclim: The reason we need this, is so the divs inherit the parent height such that the `position: sticky` panel can get a height

workspaces/ui-v2/src/pages/diffs/EndpointDocumentationPane.tsx
niclim: That is correct - there's potentially a better way to link these but I haven't thought about this that much

JaapRood: Perfectly fine way to link to something deeper on a page, I often hunt people's DOMs for them so my links can be more accurate :)

PR #955, opened 2021-07-01
Update PathParameters panel to use the Panel component

niclim:
## Why

More refactoring - since we have the Panel component, it means we can get rid of the more custom implementations (goodbye `BodyRender` and `TwoColumn`)

## What

Screen Shot 2021-07-01 at 1 23 36 PM

Screen Shot 2021-07-01 at 1 23 41 PM

Screen Shot 2021-07-01 at 1 24 01 PM

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


JaapRood (approved):
Bye `BodyRender` and `TwoColumn`, byeeeee.

PR #957, opened 2021-07-01
Move endpointChangelog into endpoint redux slice

niclim:
## Why

Part of the migration to redux - I tossed up between making changes / changelog as a separate store, but I think that would be a lot of work, for a very subjective gain - so I didn't. Instead I kept the current query pattern and split these out 

## What

- Endpoint slice now has `endpoints` and `changes` which is an object keyed by id (endpoint id, request or response id) - abstracting away the idea that there's two sources we query endpoint changes from.
- Removed useEndpointsBodyChanges and useEndpointsChangelog 

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


workspaces/ui-v2/src/pages/diffs/EndpointDocumentationPane.tsx
```diff
@@ -119,6 +120,7 @@ export const EndpointDocumentationPane: FC<
         >
           {/* TODO QPB - change id from this to query id from spectacle */}
           <div id="query-parameters">
+            {/* TODO QPB add changelogBG to this panel */}
```
niclim, line 123: This will likely involve creating a changes resolver for the query param body

JaapRood (approved):
Great stuff!

PR #958, opened 2021-07-02
Update spectacle projection to handle query parameter events

niclim:
## Why

Updated to handle the new query parameter events. This feels a little odd to me and it feels like we should have an endpoint node (I don't think we need a new event, but we can add this into the spectacle projection). This should work though

## What

Feel free to merge this if it makes sense @JaapRood 

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/graph-lib/src/endpoints-graph/index.ts
+++ b/workspaces/graph-lib/src/endpoints-graph/index.ts
@@ -247,15 +247,7 @@ export class RequestNodeWrapper implements NodeWrapper {
   }
 
   query(): QueryParametersNodeWrapper | null {
-    const neighbors = this.queries.listIncomingNeighborsByType(
-      this.result.id,
-      NodeType.QueryParameters
-    );
-    if (neighbors.results.length === 0) {
-      return null;
-    }
-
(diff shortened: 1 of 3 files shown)


niclim:
The only thing that needs to be updated is:
https://github.com/opticdev/optic/blob/698df4654356296002663d7137b1a4a5ec4706e5/workspaces/spectacle/src/spectacle.ts#L355-L376

id can be queried by `const id = parent.value.queryParameterId`

JaapRood:
@niclim the more I'm working on these projections, the more I think an an endpoint node would make sense as well, even if just to make it conceptually easier to understand what an endpoint is. Initially, we thought of the endpoint as a user level concept, not inherent to the business logic, but it does keep creeping in everywhere: as a logical unit, as a concurrency / partitioning unit, etc. Perhaps that's some tech debt we'll have appetite left to resolve at the end of bet.

JaapRood (approved):

JaapRood:
Makes total sense @niclim, thanks! Went ahead and merged it with your permission, but feel free to revert or re-open it if that wasn't right!

PR #960, opened 2021-07-02
Fix undocumented body not showing up in diff list

niclim:
## Why

Previously, undocumented bodies did not show up in the diff list. Now they do

Screen Shot 2021-07-02 at 2 52 50 PM

## What

DiffLinks was doing some odd things, where it was also sorting and grouping the requests / responses. This is confusing since the order is not the same as the sequence as the diffs are shown. If we want to do any sorting, we should do it on the allDiffs component such that it's consistent to a user

This also handles query parameters diffs

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/ui-v2/src/pages/diffs/ReviewEndpointDiffPage/DiffLinks.tsx
+++ b/workspaces/ui-v2/src/pages/diffs/ReviewEndpointDiffPage/DiffLinks.tsx
@@ -1,134 +1,41 @@
-import React, { useMemo } from 'react';
-import sortBy from 'lodash.sortby';
+import React from 'react';
 
-import {
-  IInterpretation,
-  IRequestBodyLocation,
-  IResponseBodyLocation,
-} from '<src>/lib/Interfaces';
+import { IInterpretation } from '<src>/lib/Interfaces';
 import { ICopyRender } from '../components/ICopyRender';
 import { makeStyles } from '@material-ui/styles';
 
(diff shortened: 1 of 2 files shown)


workspaces/ui-v2/src/pages/diffs/ReviewEndpointDiffPage/DiffLinks.tsx
```diff
@@ -1,134 +1,41 @@
...
-    return sections;
-  }, [allDiffs]);
-
   return (
     <List>
-      {sections.map((section) => {
-        if (section.requestId) {
-          return (
-            <div key={section.requestId}>
+      {/* TODO figure out why diffDescription can be nullable */}
```
JaapRood, line 19: Those are generated on the Typescript side, and probably _can_ be nullable because new types of diffs can be introduced for which it doesn't map a description yet? Like we're about the add new diff types for query params!

JaapRood (approved):

workspaces/ui-v2/src/pages/diffs/ReviewEndpointDiffPage/DiffLinks.tsx
niclim: Hmmm, that could be valid but I think diffDescription is nullable because of loading state - as for other diff types we might want a different way to model this

PR #963, opened 2021-07-06
Learn query add checkbox

niclim:
## Why

Based off of https://github.com/opticdev/optic/pull/950 

Now adds a checkbox to toggle on and off 
Screen Shot 2021-07-06 at 11 51 42 AM
Screen Shot 2021-07-06 at 11 51 47 AM

## What

Diff of changes can be viewed here https://github.com/opticdev/optic/commit/296dcdc47cf78f14807db34ff86bc330a67a6615 (since this contains develop + engine changes)

- There's some fairly confusing conditionals + weird typings here - this code could use with a refactor + migration to redux (but it might be better to migrate top down for redux - the shared diff store first before migrating this to redux)

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


devdoshi (approved):

JaapRood (approved):
The conditionals could definitely do with some refactoring, but as far as I can tell you got it all right :)

PR #964, opened 2021-07-06
Add stricter typing to diff flow

niclim:
## Why

I was digging around the diff flow for implementing query parameters and I believe that there are typing we can make more strict. These were not possible before as there were paths that it was possible to get `diffDescription` and `updateSpecChoices` as undefined (but it appears these code paths have been removed)

I checked the pathways that we generate IInterpretations (new region intepretation and shape diff intepretation) and it looks like we are always guaranteed to have an interpretation with non-nullable diffDescriptions and updateSpecChoices

## What

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/ui-v2/src/lib/parse-diff.ts
+++ b/workspaces/ui-v2/src/lib/parse-diff.ts
@@ -163,15 +163,6 @@ export class ParsedDiff {
   }
 }
 
-export function parseDiffsArray(
-  array: [IDiff, string[], string][]
-): ParsedDiff[] {
-  return array.map(
-    ([rawDiff, pointers, fingerprint]) =>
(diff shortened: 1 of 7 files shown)


workspaces/ui-v2/src/lib/parse-diff.ts
```diff
@@ -163,15 +163,6 @@ export class ParsedDiff {
   }
 }
 
-export function parseDiffsArray(
```
niclim, line 166: This is an unused function

JaapRood (approved):

niclim:
Going to hold off on merging this until friday since we are releasing tomorrow, and these changes are inherently more risky since we don't have type guarantees (where previously we could have been swallowing errors)

PR #965, opened 2021-07-06
Update UI to handle query parameters in diff

niclim:
## Why

The diff flow needs to be updated to handle a diff in the query parameter (undocumented or shape diff). 

## What

- Added support in the UI diff intepretation to handle UnmatchedQueryParameters and UnmatchedQueryParameterShapes
- Removed keymirror, it makes it hard for typescript to infer typings, and it doesn't even save us much keystrokes. and it's an extra dependency
- 

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


workspaces/cli-shared/src/diffs/diffs.ts
```diff
@@ -3,42 +3,45 @@
...
+  UnmatchedResponseStatusCode: 'UnmatchedResponseStatusCode',
+};
 
 export type IDiffWithShapeDiff =
+  | IUnmatchedQueryParametersShape
   | IUnmatchedResponseBodyShape
   | IUnmatchedRequestBodyShape;
 
+export interface IUnmatchedQueryParametersShape {
+  UnmatchedQueryParametersShape: {
```
niclim, line 38: I think this is missing from your PR in #950 - but we can modify either (this is similar to the request/response body shape diffs

workspaces/ui-v2/src/lib/diff-description-interpreter.ts
```diff
@@ -87,6 +92,7 @@ export async function descriptionForShapeDiff(
 
   const getJsonBodyToPreview = (interaction: any) => {
     const body =
+      (location.inQuery && interaction.query.body) ||
```
niclim, line 95: I'm not sure if this is correct, or how this will line up. This will also need to be checked when we integrate the two parts

JaapRood: I'm not sure either, the integration will tell us for sure.

JaapRood (commented):
Looking good so far, addressing the parts that I would think of needing changing.

niclim:
Ok so this mostly works:
new region
Screen Shot 2021-07-08 at 3 46 45 PM
shape diff
Screen Shot 2021-07-08 at 3 46 51 PM

The preview not showing up has something to do with `previewTabs[values].jsonTrailsByInteractions` being `{}` - which I need to do more digging to find out why. Digging into this code has me throwing in many "we need to refactor this because of _very very_ lose / poorly typed things". There's a lot of potential bugs _and_ difficulties we'll have with adding in new events in the future - currently it's really hard to iterate on this piece of code

That being said - @JaapRood my quick tests show that it's "working" end to end (i.e. the rust code seems to work correctly)

workspaces/cli-shared/src/diffs/trail-parsers.ts
```diff
@@ -131,34 +152,50 @@ export function locationForTrails(
...
 ): string | undefined {
-  //@ts-ignore
-  const Method: IMethod | undefined = interactionTrail.path.find((i) => {
-    return (i as any)['Method'];
+  const pathComponent = interactionTrail.path.find((pathComponent) => {
+    return 'Method' in pathComponent;
   });
 
-  if (Method) {
-    return Method!.Method.method;
```
niclim, line 145: i'm looking for 4 repeated words in a row now

workspaces/local-cli/src/commands/status.ts
```diff
@@ -138,49 +151,62 @@ export default class Status extends Command {
       variables: {},
     });
 
-    const endpoints: Endpoint[] = [];
-    const requestBodies: IRequestBodyForTrailParser[] = [];
-    const responseBodies: IResponseBodyForTrailParser[] = [];
-
-    for (const request of requests.data.requests) {
-      endpoints.push({
+    const endpoints = requests.map((request: any) => {
```
niclim, line 154: This is copy pasted from the ui redux store. there's probably a place we can put these in the same area - but it feels like this would go into spectacle. cli-shared is used by the UI and local-cli which is odd but maybe that's the intended location? I'm a little unsure where the best place for these shared data processing / items should live... Perhaps cli-shared is the best place

niclim:
Ok I figured out the preview... Kind of. The `InteractionBodyViewerAllJS` does a lot of things, and is hard to work on - it's going to be hard to work on if we need to fix / add any new features into it - but this should work for now.

See below:
Screen Shot 2021-07-08 at 5 20 16 PM
Screen Shot 2021-07-08 at 5 20 30 PM

The magic line that fixes this is https://github.com/opticdev/optic/pull/965/commits/d4ef6030e171730157e99c5a1a4f9f9c7d5441be (because our default values are an object of undefined, but I don't want to remove it since it's very loosely typed and I don't know what else this affects / shapes we should handle)

workspaces/ui-v2/src/lib/Interfaces.ts
```diff
@@ -30,12 +30,10 @@ export interface IPatchChoices {
 
 export interface IInteractionPreviewTab {
```
niclim, line 31: allowsExpand and ignoreRule aren't used at all in the code. It might be missed during porting or it's never used.

Either way, I think removing this simplifies it and if we need to reimplement it, these could have used rewrites anyway (it's hard to trace what they were intended to do - probably copy paste from old optic iterations)

workspaces/ui-v2/src/lib/diff-description-interpreter.ts
```diff
@@ -8,14 +9,53 @@ import {
 import { getExpectationsForShapeTrail } from './shape-diff-dsl-rust';
 import { code, ICopy, plain } from '<src>/pages/diffs/components/ICopyRender';
 import { IJsonObjectKey } from '@useoptic/cli-shared/build/diffs/json-trail';
-//@ts-ignore
-const { toJsonExample } = require('@useoptic/shape-hash');
+import { IHttpInteraction } from '@useoptic/optic-domain';
+import { toJsonExample } from '@useoptic/shape-hash';
+
+const getJsonBodyToPreview = (
```
niclim, line 15: Copied from below - the only difference it was pulling `location` in via a closure, now we pass it in explicitly so we don't need to duplicate

niclim:
@JaapRood I think this is ready to review - it's kinda messy because there's a lot of old code that I had to dig through to get this to work and modify places / types.

I've annotated code that I want to clean up as part of this query param bet (and there's probably a much simpler way we can parse / handle diffs) - but these will be follow up PRs

workspaces/local-cli/src/commands/status.ts
```diff
@@ -114,20 +110,37 @@ export default class Status extends Command {
     const { diffs } = await diffService.listDiffs();
     const { urls } = await diffService.listUnrecognizedUrls();
 
-    const requests: any = await spectacle.query({
+    const requestQuery = await spectacle.query<any>({
```
niclim, line 113: This is used in both local-cli and the ui, there's probably a better place to put this, so we don't need to duplicate, but I haven't found a great place for this (maybe the `cli-shared` workspace)?

niclim: I've also tested `api start` in the sidechannel build https://github.com/opticdev/optic/actions/runs/1016146932 and this works (less certain because of loose typings, but that's more to copy paste)

There's probably a better strategy since it's likely we'll want almost the same functionality in the UI as the local cli

workspaces/cli-shared/src/diffs/trail-parsers.ts
```diff
@@ -1,60 +1,100 @@
...
-  IRequestBody,
-  IResponseBody,
-  IResponseStatusCode,
-} from './interaction-trail';
-
-export interface IRequestBodyForTrailParser {
+import { IRequestSpecTrail } from './request-spec-trail';
+import { IInteractionTrail } from './interaction-trail';
+
+// TODO QPB - remove these types - these are the same as the types in ui-v2 - these should be moved to optic-domain
```
JaapRood, line 4: 👍

workspaces/local-cli/src/commands/status.ts
JaapRood: While _some_ of the logic might be shared, isn't the entire point of graph ql queries is that they can be and should be client specific? So that any can optimise exactly what it queries for, without worrying about affecting others? I'd be happy for code to be duplicated, especially if it's not particularly error prone. Especially for "glue" code, that's what I'd expect, as even if clients are context are similar, they hardly are ever the same, nor will they stay that way.

JaapRood (approved):

workspaces/local-cli/src/commands/status.ts
niclim: There's more transformation logic - I think as an ideal, there should be more that moves into spectacle, but I think the more complex our usecases become, we'll have specialized logic on top of that (that shouldn't live in spectacle) that need more complex transformations, etc. Lets say we implement a diff flow in the UI or other FEs, we'd probably want a shared layer between the UI and the cli, but spectacle probably shouldn't take all of these.

I think the biggest pain is that spectacle.query + defining our gql query produces an `any` result and we have to manually type the return, which causes the most pain

PR #966, opened 2021-07-06
Fix cannot read commands of undefined

niclim:
## Why

If there is an unmatched request or response body, and the learner does not find anything to learn, the UI crashes here. Adding this to be resilient to that (it's still not correct, and we'll still get sentry errors). The root cause is that there's a mismatch between the diff intepretation and the body learner (something I'm not too sure how that'd be an issue and what information would be useful to debug).

This will cause the UI to render _something_ but no commands will be generated (which I believe is better than crashing)

## What

This also sends learnedbody information and the diff information to sentry - do we want to sanitize any of this information? 

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/ui-v2/src/lib/new-regions-interpreter.ts
+++ b/workspaces/ui-v2/src/lib/new-regions-interpreter.ts
@@ -1,3 +1,5 @@
+import * as Sentry from '@sentry/react';
+
 import {
   CurrentSpecContext,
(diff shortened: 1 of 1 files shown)


devdoshi (approved):
I'll defer to @acunniffe and @LouManglass  on sanitizing data

PR #967, opened 2021-07-06
Add query parameter parsing strategy info tooltip

niclim:
## Why

Screen Shot 2021-07-06 at 3 59 17 PM

## What

In the future, a tooltip might not be the best component for complex query pattern parsing, but for now I think this is sufficient.

Also I am not a huge fan of the current copy, so any suggestions are appreciated

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/ui-v2/src/components/QueryParametersPanel.tsx
+++ b/workspaces/ui-v2/src/components/QueryParametersPanel.tsx
@@ -1,5 +1,6 @@
 import React, { FC } from 'react';
-import { makeStyles } from '@material-ui/core';
+import { makeStyles, Tooltip } from '@material-ui/core';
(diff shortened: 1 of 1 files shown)


JaapRood (approved):
Hmmm 🤔. We know we want to communicate with the user about the parsing strategy, and this more than what the version before did (zero communication). That said, a tooltip is fairly down the list of ways I would attempt that (they come with a bunch of accessibility and usability issues, especially on touch devices). What I would do differently exactly is something I'd have to spend some time on. I'm happy to have this in as a first pass at this, to be revisited as appetite allows.

niclim:
Hmmm good points, I agree, and its something we can spend more time later in the bet

PR #968, opened 2021-07-06
Remove unused spectacle queries and background components

niclim:
## Why

I was going to implement the changelog background for the endpoint documentation page - but the code path is actually not hit at all.

There is no changes resolver on the `request` or `responses` - which means they always resolve to null (i thought this would have crashed but apparently not)

## What

@devdoshi do you know if this is "safe" to remove from our schema (i.e. are there external users using spectacle)? Is there a better way to deprecate / remove fields. I checked optic-changelog and it does not use these unused fields.

I also want to update all the uses of ChangeResult so we return a consistent "change result" (i.e. a string enum of `'added' | 'updated' | 'removed'` rather than `{added: boolean, changed: boolean}`) over returning both

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


JaapRood:
I'll leave this one to you @devdoshi, @niclim is more informed on this than I am.

devdoshi (approved):
I believe the github actions are the only consumers, so if you checked those then it should be fine. I can't think of anything that's using this query. It's the same general API deprecation problem, but with graphql. Perhaps https://www.apollographql.com/docs/apollo-server/schema/directives/ ? Anyway unless we had an analytics server monitoring query usage we probably wouldn't really know whether it was safe.

niclim:
Given that this doesn't actually return anything, I think it's safe for us to remove it here. However as we expose spectacle to more and more people, we'll want to be more careful about breaking changes (perhaps our deprecations trategy should be marking fields as deprecated, and removing them in the next major version)

Github actions (i.e. optic changelog) uses `@useoptic/changelog` which calls `endpointChanges` - this was only used in the UI

PR #970, opened 2021-07-07
Diff and learn shapes of Query Parameters

JaapRood:
## Why
Now that we can learn query parameters for newly added endpoints, we'll want to support:

- Documenting query parameters for existing endpoints.
- Allowing query parameters to change through handing diffs.

## What

- [x] Generate a new `UnmatchedQueryParametersShape` diff variant, wrapping a shape diff of the query parameters.
- [x] Generate shape diff affordances for query parameter shape diffs, so the UI can reason about how to handle the diff.
- [x] Enable the returning of new diff result variants through `REACT_APP_FF_DIFF_QUERY_PARAMETERS`.
- [x] Provide test coverage for the various use cases of diffing query parameters.

## Validation
* [ ] CI passes
* [x] New diff variants are detected correctly
* [x] Shape diff affordance learner generates the correct affordances


JaapRood:
@niclim the new diff variants should be produced with the feature flag set, and the shape diff affordances should be generated. Not all use cases will be handled correctly yet (looking at the test coverage for that next), but hopefully it's good enough for you to start integrating the UI against (#965).

JaapRood:
I've added some basic end to end diff use cases around query parameters. These highlight how, while we have basic diffs of query parameters being generated correctly, there are some use cases that aren't working quite right yet. The most significant so far is that diffing of required query parameters yields no results when no query parameters were captured for an interaction, while they should be reported as missing.

To have a place for these new tests, I took this opportunity to also move the existing e2e tests for diffing to the approach we've been using for e2e tests as of late. Rather than embedding (and having to compile) events and interactions straight in the file, we read them from debug captures at run time. I know there are further plans to refine this strategy even further, but this is the best I'm aware of at the moment. The tests still have cryptic `scenario_x` names, but at least it makes a lot easier to add more tests with slightly more descriptive names and keeping the methodology consistent.

---

I'll look at fixing the failing use cases next.

workspaces/optic-engine/src/interactions/visitors/diff.rs
```diff
@@ -125,7 +125,7 @@ impl QueryParametersVisitor<InteractionDiffResult> for DiffQueryParametersVisito
       query_shape_id,
     ) {
       (None, None, None) => {}
```
JaapRood, line 127: It's this line that's preventing us from always representing the query parameters as an empty object when parsing it from the interaction: if no query params are specced and non are sent, we don't want to yield any diffs. With the default fallback, that would generate empty query parameters diffs for every existing endpoint for every spec out there, even if no query parameters are used.

So instead, we only consider the query parameters an object empty when they are not sent in the contexts where this is helpful. So far, that's only in shape diffing them. In that context, query parameters have definitely been specced, not sending any doesn't mean we just get to ignore the spec!

JaapRood:
We're now generating the right shape diffs, verified through tests, for required query parameters. Next up is verifying that the shape diff affordances are generated correctly, so that the UI can present all the options to patch the spec based on the diffs (or ignore them!).

JaapRood:
@niclim I've attempted to verify with an end to end test that the shape diff affordance learner (which tells the UI all the kinds of shapes it has observed) does this correctly and completely. It _looks_ like it does, but there's some logic on the UI side that mutates these learning results before it sends it back for generating commands for, especially in regards to missing fields in objects. I've yet to look into that, but perhaps most useful and pragmatic is to test whether the learner works correctly by piping it through the UI and verifying manually. That should also be a good step in figuring out how the automated tests should work.

niclim:
I can merge this branch into my UI branch, but perhaps the best way to review + merge this in and tackle any bugs from there

niclim:
I think your diff code looks good - my quick tests on the UI side looks solid. I think it might be worth merging and working on this since there's likely minor / small fixes as we test and work on this (there's a lot of incremental improvements to be had on the UI side)

JaapRood:
@niclim awesome that it's working end to end, it immediately brought some wrong behaviour to light, like handled diffs through the UI still yielding additional diffs (will go into further detail elsewhere). I added a test just to verify the diffing is correct, by not yielding diffs when there shouldn't be any. The rest we can address separately.

JaapRood:
Going ahead and merge this, with @niclim's confirmation that it's behind a FF properly (and having confirmed the same myself). Follow ups can be dealt with separately and more specifically.

PR #973, opened 2021-07-07
Hide reset to commit button for cloud viewer

niclim:
## Why

Cloud viewer should not be able to edit specs - as cloud viewer loads an inmemory instance. Updating the config to be more generally "edit" since that's more how we use these features (and we can make them more granular later if required).

Screen Shot 2021-07-07 at 8 19 03 AM

## What

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/ui-v2/src/components/navigation/TopNavigation.tsx
+++ b/workspaces/ui-v2/src/components/navigation/TopNavigation.tsx
@@ -51,8 +51,7 @@ export function TopNavigation(props: { AccessoryNavigation?: any }) {
                   Icon={SubjectIcon}
                 />
 
-                {/*@aidan: this needs to change*/}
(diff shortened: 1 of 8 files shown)


devdoshi (approved):
I feel like @JaapRood might take issue with renaming to 'features' which might be conflated with feature flagging, and I anticipate we'll revisit the differences between each viewer soon

niclim:
> I feel like @JaapRood might take issue with renaming to 'features' which might be conflated with feature flagging, and I anticipate we'll revisit the differences between each viewer soon

Hmm that's a good point, we shouldn't use the word features here  - removed that object so it's just `appConfig.allowEditing`. If we need better grouping we'll figure out a better name then

PR #976, opened 2021-07-07
Add events for landing on pages + deleting an endpoint

niclim:
## Why
Describe the motivation for the changes

decided to throw these through segment since i'm not sure if FS respects the analytic on / off triggers / if there's error handling. It's probably fine to just send everything / more through segment anyway

## What

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/analytics/src/uiEvents.ts
+++ b/workspaces/analytics/src/uiEvents.ts
@@ -6,6 +6,61 @@ export class OpticUIEvents {
   constructor(private dispatch: AnalyticsDispatcher) {}
 
   // in-use
+  // documentation page
+  documentationPageLoaded() {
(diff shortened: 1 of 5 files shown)


acunniffe (approved):

acunniffe:
Looks right to me 👍  -- I think those useEffects will work provided the analytics object is never re-created, which I don't believe it ever is.

niclim:
> Looks right to me 👍 -- I think those useEffects will work provided the analytics object is never re-created, which I don't believe it ever is.

It's not, it uses the context api which provides the same identity (analytics only changes if the actual function changes, which is wrapped by a useRef)

PR #977, opened 2021-07-07
Move diffs for capture state into redux

niclim:
## Why

As part of moving things into redux, the best place to start is from the top down. There'll be more to follow this - but we'll want to also move in intepretations (i.e. shape and new region intepretations), pending endpoints)

## What

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


workspaces/ui-v2/src/pages/diffs/ReviewDiffPages.tsx
```diff
@@ -12,28 +14,62 @@ import { SharedDiffStore } from '<src>/pages/diffs/contexts/SharedDiffContext';
...
+      const startTime = Date.now();
+      const actionResult = await dispatch(
+        diffActions.fetchDiffsForCapture({
+          capturesService,
+          captureId: boundaryId,
+        })
+      );
+
+      try {
+        const results = await unwrapResult(actionResult);
```
niclim, line 60: unwrapResult throws an error if the promise is rejected - we do this because we get better type safety through unwrapping. we also handle the error in the redux layer. This is just to connect analytics information

workspaces/ui-v2/src/pages/diffs/ReviewDiffPages.tsx
```diff
@@ -12,28 +14,62 @@ import { SharedDiffStore } from '<src>/pages/diffs/contexts/SharedDiffContext';
...
+        const endTime = Date.now();
+        setDiffService(results.diffService);
+        analytics.reviewPageLoaded(
+          results.data.diffs.length,
+          results.data.urls.length,
+          endTime - startTime,
+          results.numberOfEndpoints
```
niclim, line 67: This is odd, but it's because we are sending this information on a reviewPageLoaded. It probably should go in as a child and on a first mount. This is also more correct what's currently implemented, where we potentially send multiple reviewPageLoaded events

workspaces/ui-v2/src/pages/diffs/ReviewDiffPages.tsx
JaapRood: Nice! And maybe in addition we could also do with an `expectResult(actionResult, "assumption that makes us expect this")`, those have been invaluable in the Rust codebase.

workspaces/ui-v2/src/store/diff/slice.ts
```diff
@@ -0,0 +1,56 @@
...
+import { AsyncStatus } from '<src>/types';
+import { fetchDiffsForCapture } from './thunks';
+
+// This needs to be exported for typescript to be able to infer typings
+export type DiffState = {
+  state: AsyncStatus<
+    {
+      diffs: ParsedDiff[];
+      urls: IUnrecognizedUrl[];
+      trails: IValueAffordanceSerializationWithCounterGroupedByDiffHash;
```
JaapRood, line 16: Bonus points are available for renaming this horrible type. Or using an anonymous type that composes from it's components. Anything 😂😭

JaapRood (approved):

workspaces/ui-v2/src/store/diff/slice.ts
niclim: Ok I attempted to rename these - I think `IAffordanceTrails` and `IAffordanceTrailsDiffHashMap` are slightly better - but I'm not sure what these are _actually_ representing so there's probably a better name

PR #978, opened 2021-07-07
Remove navigation from diffs -> docs on pending endpoint flow

niclim:
## Why
Screen Shot 2021-07-07 at 4 32 45 PM

On this flow - removed the navigation to the documentation page. It doesn't really make sense for a user to go to the diff flow, add pending endpoints and then navigate to the documentation flow (where they would lose their current progress).

If a user wanted to view their current spec, they would go to the documentation view. 

## What

Removes navigation from diff -> docs from the list of existing endpoints 

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


JaapRood (approved):
Yes, thank you! If anything, the result of saving changes to history, would be to a place that then shows how you've made those changes, e.g. changelog with the previous version or history. But most appropriately, if there's something left to do in the same vain (like documenting more endpoints), we should just remain in that context!

PR #984, opened 2021-07-09
Fix bug where initial learning of query parameters could not result into optional bodies

JaapRood:
## Why
We uncovered a bug where query parameters for newly added endpoints would never be learned as optional, despite there being a mix of interactions for endpoints with and without query parameters.

## What
The query parameters diff visitor is updated to allow results to be generated in the case where an interaction doesn't have any interactions captured, and there isn't an existing spec yet. Before, that case was completely ignored, which is fine for the main diffing flow, but not enough for the learning of new query parameters, where we need to track whether query parameters were omitted in requests. With these changes, an internal variant of `UnmatchedQueryParameters` is added, used for the undocumented body learner and filtered out of output diff results.

## Validation
* [ ] CI passes
* [x] Query parameters can be learned as optional
* [x] No "undocumented query parameters" diffs for existing documented endpoints without any query parameters captured.

### DIFF
--- a/workspaces/optic-engine/src/interactions/result.rs
+++ b/workspaces/optic-engine/src/interactions/result.rs
@@ -35,7 +35,7 @@ impl InteractionDiffResult {
 
   pub fn interaction_trail(&self) -> &InteractionTrail {
     match self {
-      InteractionDiffResult::UnmatchedQueryParameters(diff) => &diff.interaction_trail,
+      InteractionDiffResult::UnmatchedQueryParameters(diff) => diff.interaction_trail(),
       InteractionDiffResult::UnmatchedQueryParametersShape(diff) => &diff.interaction_trail,
       InteractionDiffResult::UnmatchedRequestUrl(diff) => &diff.interaction_trail,
       InteractionDiffResult::UnmatchedRequestBodyContentType(diff) => &diff.interaction_trail,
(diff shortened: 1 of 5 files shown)


JaapRood:
@niclim I have to end my week, but thought I'd get this first bug fix out there for you, so you could of it if you wanted to!

niclim:
Ah I see my rebase has affected your merge - sorry didn't realized you branched off my PR - I'll fix it on this PR

workspaces/optic-engine/src/interactions/result.rs
```diff
@@ -97,17 +97,50 @@ impl UnmatchedRequestUrl {
...
 
+#[derive(Clone, Debug, Deserialize, Serialize, Hash)]
+#[serde(untagged)]
+pub enum UnmatchedQueryParameters {
+  Observed(UnmatchedQueryParametersDescriptor),
+
+  #[serde(skip)]
+  Unobserved(UnmatchedQueryParametersDescriptor),
+}
+
```
JaapRood, line 113: It was hard to find an exact way to express this new diff result variant. In a way there being no query parameters documented and no query parameters in an interaction, you can say that _matches_. However, all `Match*` variants so far all describe cases where we match something seen with something that exists in the spec, not both sides not existing. In the end, what we're willing to capture in both diffs and the learner is that we don't have a documented query parameters yet.

Going with "observed" and "unobserved" variants of the same main variant was the best I could come up with so far, keeping the close relationship between the two results in place.

JaapRood:
@niclim if this were to get a passing review, feel free to merge it so we can land the semi-stable version of query parameters into `develop`. Otherwise, I'll spend time on that myself tomorrow :)

niclim (approved):
I think this makes sense given your explanation

PR #985, opened 2021-07-09
fix: Use prod base domain when in prod, not the other way around

jshearer:
Looks like the condition to determine which backend base domain to use got reversed.. thanks @ukmadlz for finding this!


ukmadlz (approved):
Can confirm this is working on the side channel build when previously I was being sent to the staging ENVs.

PR #986, opened 2021-07-09
Refactor IParsedLocation to use typescript type narrowing

niclim:
## Why

Use type narrowing / inferences so we have better type safety - there's a bunch of things here that should really be included in the ParsedDiff class (it's a pretty leaky abstraction right now since it attempts to encapsulate diff logic, but you still need to pipe things together)

## What

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/ui-v2/src/lib/Interfaces.ts
+++ b/workspaces/ui-v2/src/lib/Interfaces.ts
@@ -210,33 +210,25 @@ export interface IResponseBodyLocation {
   responseId?: string;
 }
 
-// TODO QPB - refactor this type such that
-/**
- * 
 export interface IParsedLocation {
   pathId: string;
   method: string;
-  location: {
-    type: 'query'
-  } | {
-    type: 'request'
-    requestId: string
-    contentType: string
-  } | {
-    type: 'response'
-    statusCode: number
-    contentType: string
-    responseId: string
-  }}
- */
-export interface IParsedLocation {
-  pathId: string;
-  method: string;
-  inQuery?: {
-    queryParametersId: string;
-  };
-  inRequest?: IRequestBodyLocation;
-  inResponse?: IResponseBodyLocation;
+  data:
+    | {
+        type: 'query';
+        queryParametersId: string;
+      }
+    | {
+        type: 'request';
+        requestId: string;
+        contentType: string;
+      }
+    | {
+        type: 'response';
+        statusCode: number;
+        contentType: string;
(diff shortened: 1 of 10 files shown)


workspaces/ui-v2/src/lib/new-regions-interpreter.ts
```diff
@@ -36,19 +36,22 @@ export async function newRegionInterpreters(
...
+// TODO QPB - move IInterpretation generation into ParsedDiff
 function newContentType(
   udiff: ParsedDiff,
   location: IParsedLocation,
   learnedBodies: ILearnedBodies
 ): IInterpretation {
   if (udiff.isA(DiffTypes.UnmatchedRequestBodyContentType)) {
+    const contentType =
+      location.data.type === 'request' ? location.data.contentType : '';
     const learnedRequestBody = learnedBodies.requests.find(
```
niclim, line 48: This is a good example of why the ParsedDiff class is a bad abstraction - location is _from_ ParsedDiff - but to make this function correct, we need to check both the location and the diff type

workspaces/ui-v2/src/lib/Interfaces.ts
```diff
@@ -198,45 +198,25 @@ export const DiffInResponse = (key: string): boolean =>
...
+        type: 'request';
+        requestId: string;
+        contentType: string;
+      }
+    | {
+        type: 'response';
+        statusCode: number;
+        contentType: string;
+        responseId: string;
+      };
```
JaapRood, line 219: 👍 we should definitely not be too afraid to use anonymous types where appropriate!

workspaces/ui-v2/src/lib/Interfaces.ts
```diff
@@ -198,45 +198,25 @@ export const DiffInResponse = (key: string): boolean =>
...
- */
-export interface IParsedLocation {
-  pathId: string;
-  method: string;
-  inQuery?: {
-    queryParametersId: string;
-  };
-  inRequest?: IRequestBodyLocation;
-  inResponse?: IResponseBodyLocation;
+  data:
```
JaapRood, line 204: Nitty bit: I would have probably gone for `descriptor` over `data`, but that's only because "data" is such a generic and overloaded term. That said, "descriptor" isn't particularly concrete either, so 🤷‍♂️

JaapRood (approved):

PR #988, opened 2021-07-12
Add diff cases specfic to query parameters to diff-use-cases-with-events example session

JaapRood:
## Why
While query parameters produce shape diffs, which covers them under most shape diff use cases, there are some unique ways to how they are parsed, learnt and presented in the UI. We want to make sure we have an easy way of reproducing these use cases and verify they are handled correctly. 

For now this targets #984, but should target `develop` once that merges.

## What
This adds a partially documented `GET /query-parameters` endpoint and additional interaction that produces various diffs.

## Validation
* [x] CI passes
* [x] New diffs are generated


JaapRood:
@niclim another one that I'm happy to be merged once we have a passing review :)

niclim (approved):

PR #990, opened 2021-07-12
Fix/query param rendering

niclim:
## Why

Fixes misc rendering issues with query parameters

## What

- Label for undocumented query parameters is now undocumented query parameters
Screen Shot 2021-07-12 at 11 03 58 AM

- Optional query parameter keys are now rendered
- Polymorphic query parameter keys are now rendered

Screen Shot 2021-07-12 at 11 03 49 AM

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/ui-v2/src/components/ShapeRenderer/ShapeRenderer.tsx
+++ b/workspaces/ui-v2/src/components/ShapeRenderer/ShapeRenderer.tsx
@@ -1,6 +1,6 @@
 import React, { FC } from 'react';
 import { ShapeRenderStore } from './ShapeRenderContext';
-import { RenderRootShape } from './ShapeRowBase';
+import { RenderRootShape, OneOfRender } from './ShapeRowBase';
 import { IShapeRenderer } from './ShapeRenderInterfaces';
 
 type ShapeRendererProps = {
@@ -14,7 +14,11 @@ export const ShapeRenderer: FC<ShapeRendererProps> = ({
 }) => {
   return (
     <ShapeRenderStore showExamples={showExamples}>
-      <RenderRootShape shape={shape} />
+      {shape.length > 1 ? (
+        <OneOfRender shapes={shape} parentShapeId={'root'} />
+      ) : (
+        <RenderRootShape shape={shape} />
+      )}
     </ShapeRenderStore>
   );
 };
--- a/workspaces/ui-v2/src/components/ShapeRenderer/ShapeRowBase.tsx
+++ b/workspaces/ui-v2/src/components/ShapeRenderer/ShapeRowBase.tsx
@@ -65,18 +65,12 @@ export const RenderField = ({
         <ShapeRowBase depth={depth} changes={changes}>
           <span className={sharedClasses.shapeFont}>"{name}"</span>
           <span className={sharedClasses.symbolFont}>: </span>
-          <RenderFieldLeadingValue
-            parentId={parentId}
-            shapeRenderers={shapeChoices}
-          />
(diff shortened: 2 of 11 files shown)


workspaces/ui-v2/src/components/ShapeRenderer/ShapeRowBase.tsx
```diff
@@ -65,18 +65,12 @@ export const RenderField = ({
         <ShapeRowBase depth={depth} changes={changes}>
           <span className={sharedClasses.shapeFont}>"{name}"</span>
           <span className={sharedClasses.symbolFont}>: </span>
-          <RenderFieldLeadingValue
-            parentId={parentId}
```
niclim, line 69: parentId is unused in these components - removing since it's confusing

workspaces/ui-v2/src/components/ShapeRenderer/ShapeRenderer.tsx
```diff
@@ -14,7 +14,11 @@ export const ShapeRenderer: FC<ShapeRendererProps> = ({
 }) => {
   return (
     <ShapeRenderStore showExamples={showExamples}>
-      <RenderRootShape shape={shape} />
+      {shape.length > 1 ? (
+        <OneOfRender shapes={shape} parentShapeId={'root'} />
```
niclim, line 18: This isn't necessarily true, but the usage of this component means that we don't _actually_ care what the real parentShapeId is, it's the root of the rendered component (which is used in shapeRendererStore).

workspaces/ui-v2/src/lib/diff-description-interpreter.ts
```diff
@@ -135,17 +135,20 @@ export async function descriptionForShapeDiff(
...
+        plain(undocumentedLocation),
         code(lastIsField.key),
         plain('observed'),
       ],
       location,
       changeType: IChangeType.Added,
       diffHash: asShapeDiff.diffHash(),
-
-      assertion: [code('undocumented field')],
+      assertion: [code(undocumentedLocation)],
```
niclim, line 151: I'm not sure what this actually powers in the UI - can't seem to find anything that affects this behavior

niclim:
Somewhat related - the array render looks kinda odd (especially in the query parameter space)
Screen Shot 2021-07-12 at 11 07 59 AM

This is _consistent_ with how we render arrays, and this pattern works for more complex shape renders, but for simple arrays, and empty arrays/objects, there's follow up UI work we should do to render them inline (i.e. `[]` and `{}` or `["string"]` renders in a single line, and everything else goes to mulitline

workspaces/ui-v2/src/lib/diff-description-interpreter.ts
JaapRood: I'm not too sure either right of the bat. @acunniffe, could you perhaps enlighten us?

JaapRood:
@niclim I took the liberty to push some commits that moves us towards the user's model of query parameters and away from our implementation details a bit more. Figured that was the best way to explain what I had in mind 😅.

There's basically two enhancements:

- "query parameter is optional" instead of "this field is optional" in the list of choices.
- Framing a "list" of a query parameters as a "multiple" property. Given how we parse query parameters, that's actually what this diffing does. We don't really know (or _need_ to know!) what the server does with it, whether it turns it into a list (most likely) or something else. All we know is whether it's a valid thing in the first place. 

Screenshot 2021-07-13 at 11 48 57

Especially the reframing of a "list" to "multiple" is probably something we want to represent in the preview and documentation as well, but I've left that for now.

PR #991, opened 2021-07-12
Handle query parameters deletion on endpoint deletion

niclim:
## Why

When we delete an endpoint, we also need to delete query parameters (otherwise the rust engine cannot delete the paths since the paths are still in use)

## What

- Added `EndpointCommand::RemoveQueryParameters` and `EndpointEvent::QueryParametersRemoved`
- Updated delete endpoint command generation to include query parameters
- Updated internal endpoint and spectacle endpoint projections to handle these events.

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/optic-engine/src/projections/spectacle/endpoints.rs
+++ b/workspaces/optic-engine/src/projections/spectacle/endpoints.rs
@@ -172,10 +172,20 @@ impl AggregateEvent<EndpointsProjection> for EndpointEvent {
         }
       }
       EndpointEvent::QueryParametersAdded(e) => {
-        projection.with_query_parameters(e.path_id, e.http_method, e.query_parameters_id)
+        projection.with_query_parameters(e.path_id, e.http_method, e.query_parameters_id);
       }
       EndpointEvent::QueryParametersShapeSet(e) => {
-        projection.with_query_parameters_shape(e.query_parameters_id, e.shape_descriptor);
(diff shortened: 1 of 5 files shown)


workspaces/optic-engine/src/projections/spectacle/endpoints.rs
```diff
@@ -172,10 +172,20 @@ impl AggregateEvent<EndpointsProjection> for EndpointEvent {
...
+        projection.with_query_parameters_shape(&e.query_parameters_id, &e.shape_descriptor);
+        // We treat the query parameter only as created when there is a shape attached to it
+        if let Some(c) = e.event_context {
+          projection.with_creation_history(&c.client_command_batch_id, &e.query_parameters_id);
+        }
+      }
+      EndpointEvent::QueryParametersRemoved(e) => {
+        projection.without_query_parameters(&e.query_parameters_id);
+        if let Some(c) = e.event_context {
+          projection.with_remove_history(&c.client_command_batch_id, &e.query_parameters_id);
```
JaapRood, line 187: `with_` in the Rust world implies a transfer of ownership from something into the thing that it will now live with. Lets stick with that pattern and clone the event rather than giving a reference to it if it's used more than once. 

`without_` can remain to be by reference.

JaapRood (approved):
Great one @niclim! Apart from a nit pick, it fits right into the codebase as if it always belonged there :)

PR #992, opened 2021-07-12
Move types into optic domain

niclim:
## Why

Interactions, events and commands are optic domain specific - moving to the optic-domain package

## What
- Optic domain now has interactions, events and command typings

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


JaapRood (approved):
👏 YAS 🎉

PR #994, opened 2021-07-12
Refactor locations for trails function

niclim:
## Why

Cleaning up some of the TODOs I left while implementing the query parameters bet

## What

Refactor locations for trails to have better type safety. By doing this, I uncovered some odd odd variants that I don't know how they are hit :/ 

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


workspaces/cli-shared/src/diffs/trail-parsers.ts
```diff
@@ -74,84 +62,105 @@ export function locationForTrails(
...
-      contentType,
-      inResponse: true,
+      descriptor: {
+        type: 'response',
+        statusCode: statusCode,
+        responseId,
+        contentType,
+      },
     };
   } else if ('SpecPath' in trail) {
```
niclim, line 131: Do you know when this trail is generated? It doesn't have a request / response id, but mimics the shape of them and later on we coerce these to non-nullable (so it appears to not crash the UI currently) - but I'm really confused what this is supposed to represent

workspaces/cli-shared/src/diffs/trail-parsers.ts
```diff
@@ -1,70 +1,58 @@
...
-export interface IPathParameter {
-  id: string;
-  name: string;
-  isParameterized: boolean;
-  description: string;
-  endpointId: string;
-}
-
-interface IEndpoint {
+// This is a subset of IEndpoint in ui-v2
```
niclim, line 4: Creating the minimal amount of data we need that should match IEndpoint so we dont need to copy paste types, and instead, this function specifies the minimum data is needs (if it gets more, it ignores it, the consumers can pass in more)

workspaces/cli-shared/src/diffs/trail-parsers.ts
niclim: Ok it's generated from somewhere, found out from the snapshot tests - we should make these explicit, but I think this would do with a good refactor to split up diff types such we can define an abstract class, and depending on the diff type we have specific classes that'll generate spec trails + descriptions, etc).

workspaces/ui-v2/src/lib/parse-diff.ts
```diff
@@ -97,35 +91,14 @@ export class ParsedDiff {
       currentSpecContext.currentSpecEndpoints
     );
 
-    invariant(
-      Boolean(location!.pathId),
```
niclim, line 101: pathId is non-nullable via types. If it isn't we should update the typing instead

workspaces/cli-shared/src/diffs/trail-parsers.ts
JaapRood: These trails are generated by the engine, as a result from the diffing, here: https://github.com/opticdev/optic/blob/ce222d9dd130b6cfd11b8415a5a1a03733e85c5b/workspaces/optic-engine/src/interactions/result.rs#L627-L635.


For a quick overview, there's 4 types of trails you'll find returned by the engine:

- [interaction trail](https://github.com/opticdev/optic/blob/ce222d9dd130b6cfd11b8415a5a1a03733e85c5b/workspaces/optic-engine/src/interactions/result.rs#L493-L495): points to a certain place inside a _specific_ interaction. Examples: it's query parameters, the request body, a response with a certain status code, etc.
- [requests trail / request spec trail](https://github.com/opticdev/optic/blob/ce222d9dd130b6cfd11b8415a5a1a03733e85c5b/workspaces/optic-engine/src/interactions/result.rs#L627-L635) (should actually be called `EndpointSpecTrail` probably): pointing to a place in the spec by their ids on an endpoint level. Examples: a path by id,  a request body by id and content type, query parameters by a query parameters id, etc. They are mostly used in diffs to indicate the interaction matched against this particular part of the spec.
- [json trails](https://github.com/opticdev/optic/blob/ce222d9dd130b6cfd11b8415a5a1a03733e85c5b/workspaces/optic-engine/src/shapes/traverser.rs#L305-L307): point to a location inside a JSON body, most likely coming from an interaction. Example: the third array item nested inside the "author" field of an object. As we're moving of only JSON as the source, this is probably better named "body trail".
- [shape trail](https://github.com/opticdev/optic/blob/ce222d9dd130b6cfd11b8415a5a1a03733e85c5b/workspaces/optic-engine/src/shapes/traverser.rs#L273-L276): points to a location inside the shape part of the spec. Examples: the shape id of an array, the parent shape id of an object field, etc. Like the request spec trails, they are mostly used in diffs (shape diffs in this case) to indicate something matched against the spec.

Hopefully that's useful in some way 😅.

JaapRood (approved):

PR #999, opened 2021-07-13
Fix diff trails

niclim:
## Why

So turns out I added in bugs (classic) - this PR fixes this and now i have a better idea of how these work (only takes breaking the diff flow for me to understand this) - requestId, responseId and queryParameterId are only set if there is an existing, learnt body (when it's unrecognized, it's not set) - these are not needed in most locations, and this is implicitly relied on. I changed this to be explicitly typed so you have to check for the presense of these (rather than blindly casting `!`).

What I think would really really help the diff flow, is to instead have different implementations for each diff, which handle their own usage, and certain diff variants (e.g. new regions or shape diffs) can only be used in certain contexts, where they would have things like "get node id" implemented.

## What

@JaapRood  I'm going to merge this in first, since it fixes the diff flow for query parameters, but if you have any comments, i'll address them as a follow up

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/cli-shared/src/diffs/trail-parsers.ts
+++ b/workspaces/cli-shared/src/diffs/trail-parsers.ts
@@ -42,7 +42,7 @@ export type LocationDescriptor =
   | {
       type: 'path_response';
       statusCode: number;
-      contentType?: string; // TODO find out why contentType is nullable
+      contentType?: string; // Content type is nullable for responses without bodies
     }
   | {
       type: 'path_query';
--- a/workspaces/cli-shared/src/diffs/trail-parsers.ts
+++ b/workspaces/cli-shared/src/diffs/trail-parsers.ts
@@ -19,6 +19,35 @@ type EndpointForTrails = {
   }[];
 };
 
+export type LocationDescriptor =
+  | {
+      type: 'request';
+      requestId: string;
+      contentType: string;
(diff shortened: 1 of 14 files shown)


workspaces/cli-shared/src/diffs/trail-parsers.ts
```diff
@@ -19,40 +19,43 @@ type EndpointForTrails = {
...
+      queryParametersId: string;
+    }
+  | {
+      type: 'path_request';
+      contentType: string;
+    }
+  | {
+      type: 'path_response';
+      statusCode: number;
+      contentType?: string; // TODO find out why contentType is nullable
```
devdoshi, line 45: responses without bodies would not have contentTypes

workspaces/ui-v2/src/lib/shape-diffs/field.ts
```diff
@@ -46,7 +46,7 @@ export function fieldShapeDiffInterpreter(
     copy: [],
     shapes: [],
     isField: true,
-    isQueryParam: shapeDiff.location.descriptor.type === 'query',
+    isQueryParam: shapeDiff.location.isQueryParameter(),
```
JaapRood, line 49: 🙌

PR #1000, opened 2021-07-13
Fix changelog rendering for query parmaters

niclim:
## Why

Now renders without a white background when there's a change - also added a little bit of horizontal padding
Screen Shot 2021-07-13 at 12 58 39 PM

## What

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/ui-v2/src/components/ShapeRenderer/ShapeRowBase.tsx
+++ b/workspaces/ui-v2/src/components/ShapeRenderer/ShapeRowBase.tsx
@@ -119,14 +119,14 @@ export const RenderRootShape = ({
   shape,
   right,
 }: {
-  shape: IShapeRenderer[];
+  shape: IShapeRenderer;
   right?: any[];
 }) => {
   const { depth } = useDepth();
   return (
     <>
-      <ShapeRowBase depth={depth}>
-        <RenderFieldLeadingValue shapeRenderers={shape} />
+      <ShapeRowBase depth={depth} changes={shape.changes}>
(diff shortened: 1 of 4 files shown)


workspaces/ui-v2/src/components/ShapeRenderer/ShapeRowBase.tsx
```diff
@@ -119,22 +119,22 @@ export const RenderRootShape = ({
...
+        <RenderFieldLeadingValue shapeRenderers={[shape]} />
         {right ? (
           <>
             <div style={{ flex: 1 }} />
             {right}
           </>
         ) : null}
       </ShapeRowBase>
-      <RenderFieldRowValues shapeRenderers={shape} />
+      <RenderFieldRowValues shapeRenderers={[shape]} />
```
niclim, line 137: There's more we can do in these components since these components are designed to handle only 1 shape. I'm not sure if it's possible for shape to be optional (i.e. IShapeRenderer | null)

workspaces/ui-v2/src/components/ShapeRenderer/ShapeRowBase.tsx
```diff
@@ -281,7 +281,6 @@ const useStyles = makeStyles((theme) => ({
   },
   row: {
     flex: 1,
-    backgroundColor: '#f8fafc',
```
niclim, line 284: Removing this means that this now inherits the color of where it's rendered.

JaapRood (approved):

PR #1001, opened 2021-07-13
cli: update links for new docs, launching soon

LouManglass:
## Why

Update links for the next version of Optic. Old versions of Optic will still work: all links are redirected in the new site.

## Note

This should not merge until new docs launches

## Validation
* [ ] CI passes

### DIFF
--- a/workspaces/cli-config/src/helpers/default-ignore-rules.ts
+++ b/workspaces/cli-config/src/helpers/default-ignore-rules.ts
@@ -17,6 +17,6 @@ export const defaultIgnoreRules = [
 
 export const defaultIgnoreFile = `
 # Default Ignore Rules
-# Learn to configure your own at http://localhost:4000/docs/using/advanced-configuration#ignoring-api-paths
+# Learn to configure your own at http://localhost:4000/reference/optic-yaml/ignore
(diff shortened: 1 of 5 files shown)


workspaces/cli-config/src/helpers/default-ignore-rules.ts
```diff
@@ -17,6 +17,6 @@ export const defaultIgnoreRules = [
 
 export const defaultIgnoreFile = `
 # Default Ignore Rules
-# Learn to configure your own at http://localhost:4000/docs/using/advanced-configuration#ignoring-api-paths
+# Learn to configure your own at http://localhost:4000/reference/optic-yaml/ignore
```
niclim, line 20: Do we want this to point to the hosted `useoptic.com` link? Since that will probably be up to date - or do we want this to match the current locally running cli?

niclim (approved):

workspaces/cli-config/src/helpers/default-ignore-rules.ts
LouManglass: ooh, nice catch, thanks! Fixing now.

PR #1004, opened 2021-07-14
Consider interactions without query params in generating shape diff affordances

JaapRood:
## Why
Fixes a bug during query params QA, where new query params were not detected as optional correctly, due to interactions without any query parameters at all being ignored in the learning phase.

## What
As with the undocumented body learner, we now consider there being no query parameters as "empty" query parameters for learning purposes. That allows the necessary interactions to be picked up as a place where a new query parameter is missing.

In addition, it makes an adjustment to the example viewer UI, making sure empty bodies are rendered explicitly, rather than "null". In the case of the single line query parameters, this was confusing, as it might as well be a query parameter called "null". 

Before and after:
Screenshot 2021-07-14 at 12 43 15

Screenshot 2021-07-14 at 13 13 29

## Validation
* [ ] CI passes
* [x] End to end test coverage

### DIFF
--- a/workspaces/optic-engine/src/interactions/mod.rs
+++ b/workspaces/optic-engine/src/interactions/mod.rs
@@ -220,7 +220,8 @@ pub fn analyze_documented_bodies(
 
   results.into_iter().filter_map(move |result| match result {
     InteractionDiffResult::MatchedQueryParameters(diff) => {
-      let query_params = &interaction.request.query;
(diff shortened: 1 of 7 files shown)


niclim (approved):

JaapRood:
End of day for me, so not merging, but anyone else feel free if you're keen to (@niclim, @LouManglass).

PR #1007, opened 2021-07-14
Filter out invalid interactions

niclim:
## Why

In https://github.com/opticdev/optic/pull/904/files - we fixed some issues with what Optic expects in terms of request / response bodies.

Optic crashes / does not learn diffs in the following scenario, request or response body is null (i.e. null for shapeHash, asText and asJsonString).

## What

- Created a function isValidHttpInteraction
- Used this function in:
  - InMemoryCaptureService
  - LocalCliCaptureService
  - Http Proxy (which captures interactions)
  - (this function should be used for any other inputs to the diff flow, e.g. from slurp, or design by example)

TODO:
- [ ] Test out the proxy through a side channel, and make sure we aren't crashing anything
- [ ] Add in sentry to the local cli commands, so we can capture errors here (could be a follow up PR)

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/cli-shared/src/diffs/interaction-diff-worker-rust.ts
+++ b/workspaces/cli-shared/src/diffs/interaction-diff-worker-rust.ts
@@ -2,7 +2,11 @@ import {
   CaptureInteractionIterator,
   LocalCaptureInteractionPointerConverter,
 } from '../captures/avro/file-system/interaction-iterator';
-import { Streams } from '@useoptic/optic-domain';
+import {
+  Streams,
+  isValidHttpInteraction,
+  IHttpInteraction,
+} from '@useoptic/optic-domain';
 import { diffInteractions } from '@useoptic/optic-engine-native';
 import fs from 'fs-extra';
 import { fork } from 'stream-fork';
@@ -54,8 +58,13 @@ export class InteractionDiffWorkerRust {
   private async setup() {
     const ignoreFilter = parseIgnore(this.config.ignoreRules);
 
-    const interactionFilter = (i: any) => {
-      return !ignoreFilter.shouldIgnore(i.request.method, i.request.path);
+    const interactionFilter = (interaction: IHttpInteraction): boolean => {
+      const shouldIgnore = ignoreFilter.shouldIgnore(
+        interaction.request.method,
+        interaction.request.path
+      );
+      const isValid = isValidHttpInteraction(interaction);
+      return !shouldIgnore && isValid;
     };
 
     const interactionIterator = CaptureInteractionIterator(
--- a/workspaces/cli-shared/src/httptoolkit-capturing-proxy.ts
+++ b/workspaces/cli-shared/src/httptoolkit-capturing-proxy.ts
@@ -16,6 +16,7 @@ import {
   IArbitraryData,
   IBody,
   IHttpInteraction,
+  isValidHttpInteraction,
 } from '@useoptic/optic-domain';
 
 export interface IHttpToolkitCapturingProxyConfig {
@@ -203,6 +204,11 @@ export class HttpToolkitCapturingProxy {
             getters: true,
           })
         );
+
+        if (!isValidHttpInteraction(sample)) {
+          developerDebugLogger('Filtering out invalid sample', sample);
+          return;
+        }
 
         if (this.config.hostnameFilter) {
           //when filter is set, verify a match
--- a/workspaces/optic-domain/src/interactions.ts
+++ b/workspaces/optic-domain/src/interactions.ts
@@ -1,3 +1,4 @@
+// Types
 export interface IRequest {
   host: string;
   method: string;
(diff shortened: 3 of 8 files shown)


workspaces/optic-domain/src/interactions.ts
```diff
@@ -40,3 +41,31 @@ export interface IInteractionBatch {
...
+    value.shapeHashV1Base64 === null
+  );
+};
+
+export const isValidHttpInteraction = (
+  interaction: IHttpInteraction
+): boolean => {
+  if (
+    interaction.request.body.contentType !== null &&
+    isArbitraryDataNull(interaction.request.body.value)
```
niclim, line 62: When a request has no request body, contentType is null. If it's specified at all, we expect data, otherwise we generate diffs in the UI, but no bodies to learn

workspaces/optic-domain/src/interactions.ts
```diff
@@ -40,3 +41,31 @@ export interface IInteractionBatch {
...
+  ) {
+    return false;
+  }
+  if (isArbitraryDataNull(interaction.response.body.value)) {
+    return false;
+  }
```
niclim, line 68: Likewise as above, but a status code with no body can't be learnt

niclim:
@devdoshi I've added in some code to "fix" the errors we've been seeing, however I'm not sure if this is something we want to support in the future (i.e. learning requests / responses with no bodies). Currently our UI and code does not handle this gracefully, so the best solution for us is to filter them out for now and expand our definition of what a "valid interaction" is

workspaces/ui-v2/src/pages/diffs/ReviewEndpointDiffPage/ReviewEndpointDiffPage.tsx
```diff
@@ -112,6 +113,20 @@ export const ReviewEndpointDiffPage: FC<ReviewEndpointDiffPageProps> = ({
     }
   );
 
+  // We should redirect using useRedirectForDiffCompleted
+  if (allDiffs.length === 0) {
```
niclim, line 117: Our diff interpretations can be filtered out (if there's invalid diffs, for a number of reasons in `ParsedDiff`) - however our code doesn't actually handle this. 

This functionally is still wrong, but doesn't crash the page. We should tackel this when we migrate the diff code to redux (such that we don't show invalid diffs at all to the UI - kinda hard now since we cannot tell until we intepret the diffs to see if they are invalid)

workspaces/cli-shared/src/httptoolkit-capturing-proxy.ts
```diff
@@ -204,6 +205,11 @@ export class HttpToolkitCapturingProxy {
           })
         );
 
+        if (!isValidHttpInteraction(sample)) {
+          developerDebugLogger('Filtering out invalid sample', sample);
+          return;
+        }
```
niclim, line 211: I wanted to add a sentry trigger here, but I don't think our proxy (which runs through api start) actually has sentry instrumented. I'll mark this down as a follow up

workspaces/cli-shared/src/diffs/interaction-diff-worker-rust.ts
```diff
@@ -54,8 +58,13 @@ export class InteractionDiffWorkerRust {
...
+    const interactionFilter = (interaction: IHttpInteraction): boolean => {
+      const shouldIgnore = ignoreFilter.shouldIgnore(
+        interaction.request.method,
+        interaction.request.path
+      );
+      const isValid = isValidHttpInteraction(interaction);
+      return !shouldIgnore && isValid;
     };
 
     const interactionIterator = CaptureInteractionIterator(
```
niclim, line 70: Ok I realize now that there's a couple of other `CaptureInteractionIterator` - I'm not sure, but should we also add / check if it's a valid http interaction in the capture interaction iterator? Other locations are https://github.com/opticdev/optic/search?q=CaptureInteractionIterator

niclim: Looking at this more, I think it's probably correct to exclude any invalid http interaction by default here https://github.com/opticdev/optic/blob/develop/workspaces/cli-shared/src/captures/avro/file-system/interaction-iterator.ts#L55

I think by definition, it's something we shouldn't capture - unless we expand our UI / system to capture more explicitly? Thoughts?

JaapRood (commented):
If I understand the problem correctly, I think this is something the Engine should handle, not the clients of it. There's so many ways to address it and so many path ways, and you would require all of them to know what a valid interaction is and be diligent about filtering for it. Especially if the desired behaviour is dropping invalid interactions, rather than returning validation results for it, the engine should be perfectly capable of dropping and skipping them (as is the case in diffing at the moment, not learning).

If I'm not the only one to think so, I'm happy to take that on.

niclim:
@JaapRood I think you are probably right, it probably belongs in the engine. The biggest problem here is there's a mismatch in logic between the `diff` and the `learn diff` flows, which causes a diff to show up and be generated, but learning this never generates any results.

PR #1010, opened 2021-07-14
Remove contributions in the diff flow

niclim:
## Why

In the diff flow, we cannot contribute to request / response body fields, but for some reason we can contribute to path parameters. This is confusing and distracting - and as part of streamlining / giving better clarity to each page's flow, I'm removing adding path contributions to this flow (naming an endpoint is still here, because I think that's not distracting to have)

Screen Shot 2021-07-14 at 4 22 38 PM
Screen Shot 2021-07-14 at 4 22 29 PM

## What

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


niclim:
@JaapRood not sure if this slipped your notifications - repinging if it did

JaapRood (approved):
Sorry @niclim, this got lost in my notifications. Looking good!

PR #1014, opened 2021-07-16
Fix race condition diff

niclim:
## Why

In the interaction diff worker, we have two outputs to diff results, an InMemorySink, and a fsWriteStream. `listDiffs` and `listUnrecognizedUrls` read from the in memory diff, where learnShapeDiffAffordances reads from the fs. The diff shouldn't be marked as complete until the fsWriteStream completes, otherwise we get into a race condition between the fsWriteStream and the learnShapeDiffAffordances (which reads from the fs)

## What

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/cli-server/src/spectacle/index.ts
+++ b/workspaces/cli-server/src/spectacle/index.ts
@@ -230,7 +230,7 @@ class LocalCliDiff {
     //@TODO: should this be writing output to the file system?
     // Consume stream instantly for now, resulting in a Promise that resolves once exhausted
     this.diffing = AsyncTools.toArray(worker.results);
-    worker.onWriteComplete.then(() => {
+    Promise.all([this.diffing, worker.onWriteComplete]).then(() => {
       this.dependencies.notifications.emit('complete');
     });
   }
--- a/workspaces/cli-server/src/spectacle/index.ts
+++ b/workspaces/cli-server/src/spectacle/index.ts
@@ -230,7 +230,7 @@ class LocalCliDiff {
     //@TODO: should this be writing output to the file system?
     // Consume stream instantly for now, resulting in a Promise that resolves once exhausted
     this.diffing = AsyncTools.toArray(worker.results);
-    this.diffing.then(() => {
+    worker.onWriteComplete.then(() => {
       this.dependencies.notifications.emit('complete');
     });
   }
--- a/workspaces/cli-shared/src/diffs/interaction-diff-worker-rust.ts
+++ b/workspaces/cli-shared/src/diffs/interaction-diff-worker-rust.ts
@@ -8,7 +8,8 @@ import fs from 'fs-extra';
 import { fork } from 'stream-fork';
 import { DiffResult } from '@useoptic/optic-domain/build/streams/diff-results';
 import path from 'path';
-import { PassThrough } from 'stream';
+import stream, { PassThrough } from 'stream';
+import util from 'util';
 import { parseIgnore } from '@useoptic/cli-config/build/helpers/ignore-parser';
 interface WorkerResult {
   results: AsyncIterable<DiffResult>;
@@ -119,17 +120,12 @@ export class InteractionDiffWorkerRust {
     const inMemorySink = new PassThrough();
     const results = Streams.DiffResults.fromJSONL()(inMemorySink);
     const fsSink = fs.createWriteStream(diffsJsonlPath);
+    const finished = util.promisify(stream.finished);
+
     workerProcessOutput.pipe(fork([inMemorySink, fsSink]));
     return {
       results,
-      onWriteComplete: new Promise((resolve, reject) => {
-        fsSink.on('close', () => {
-          resolve();
-        });
-        fsSink.on('error', () => {
-          reject();
-        });
-      }),
+      onWriteComplete: finished(fsSink),
     };
   }
 }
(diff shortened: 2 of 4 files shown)


workspaces/cli-shared/src/diffs/interaction-diff-worker-rust.ts
```diff
@@ -121,6 +122,14 @@ export class InteractionDiffWorkerRust {
...
     return {
       results,
+      onWriteComplete: new Promise((resolve, reject) => {
+        fsSink.on('close', () => {
+          resolve();
+        });
+        fsSink.on('error', () => {
+          reject();
+        });
+      }),
```
niclim, line 132: Is there a better way to promisify a stream?

workspaces/cli-server/src/spectacle/index.ts
```diff
@@ -230,7 +230,7 @@ class LocalCliDiff {
     //@TODO: should this be writing output to the file system?
     // Consume stream instantly for now, resulting in a Promise that resolves once exhausted
     this.diffing = AsyncTools.toArray(worker.results);
-    this.diffing.then(() => {
+    worker.onWriteComplete.then(() => {
```
niclim, line 233: Do you think we need to Promise.all this (i.e. `this.diffing` and `worker.onWriteComplete`?). this.diffing is a promise that can be read + resolved. And it seems like the intent is to conver this.diffing into a direct stream to consume from?

niclim:
If this PR works, we can close the other PR https://github.com/opticdev/optic/pull/1013/files

workspaces/cli-server/src/spectacle/index.ts
devdoshi: This whole bit of code can use a once-over from @JaapRood since it's definitely not taking advantage of all the streaming power. onWriteComplete is probably ok to just use fsSink

workspaces/cli-shared/src/diffs/interaction-diff-worker-rust.ts
devdoshi: we can merge and let @JaapRood weigh in later, for now if we can't reproduce the issue then it should be good

devdoshi (approved):
see comments

workspaces/cli-shared/src/diffs/interaction-diff-worker-rust.ts
JaapRood: If just bothered with the file system sink, you'd use [`stream.finished`](https://nodejs.org/dist/latest-v14.x/docs/api/stream.html#stream_stream_finished_stream_options_callback). That handles a bunch of edge cases and also propagates the error rather than dropping it. In this case, that's what I would say is most appropriate here.

workspaces/cli-server/src/spectacle/index.ts
JaapRood: > Do you think we need to Promise.all this (i.e. this.diffing and worker.onWriteComplete?)

It would be the safe thing to do, as you might create a race condition in for the other fork, otherwise.

As @devdoshi pointed out internally, the worker forks a single result stream into two separate streams, allowing them to be piped to the file system, but consumed in-memory as well. In case subsequent actions are dependent on both of those finishing, we'll want to wait for both of them to finish 😅.

> This whole bit of code can use a once-over from @JaapRood since it's definitely not taking advantage of all the streaming power.

That's true, however as long as we're using the file system directly as a sink it might be hard to make tangible improvements. Streaming systems work very robustly when you write results to a single sink, and all consuming then happens from there. This way, you can make consumption completely consistent and very fault tolerant: if it turned up in the sink, it's correct, otherwise it's not. However, file systems don't generally directly allow you to write and read from it at the same time, at least not without some additional mechanisms some people might call databases 😅. 

That said, I'm sure we can clean this up a bit regardless, for at least it could just be less confusing.

niclim:
Ok, so i dug a little deeper and tried to recreate these. I'm able to (somewhat consistently) reproduce this by:
`task postpull` -> starting which seems to make the fsWriteSink slow enough to cause this race condition. I found that the local cli server hits the spectacle endpoint, which doesn't wait for onComplete and ends up creating it's own `onComplete` capture service class.

This was another cause of the race condition. I am fairly confident that this should work now, i'll merge this in so we can hopefully get more time to recreate this race condition before our next release

workspaces/spectacle/src/spectacle.ts
```diff
@@ -159,9 +159,12 @@ export async function makeSpectacle(opticContext: IOpticContext) {
         context: GraphQLContext
       ) => {
         const { diffId, captureId } = args;
-        await context
+        const {
+          onComplete,
+        } = await context
           .spectacleContext()
           .opticContext.capturesService.startDiff(diffId, captureId);
+        await onComplete;
```
devdoshi, line 167: I don't think this should be necessary, and would stop us from receiving incremental diff results. As far as I understand, the diff queries all wait for the diff onComplete on the server side. So startDiff should always be able to return "immediately" and then queries could wait arbitrarily long. To experiment quickly with race conditions, you can add a stream transformer that adds delays to the fs write stream. @niclim

niclim: The problem is, `onComplete` in the UI is immediately resolved after the startDiff spectacle call resolves https://github.com/opticdev/optic/blob/a4bdd09488ed0d071f1561ffa72212dd9bb61171/workspaces/spectacle-shared/src/local-cli/client.ts#L118-L122

This is causing our race condition with the fsSink. Because this resolves without `onComplete`, the trail learner (which reads from the FS) begins before the fsSink completes (even though the inMemory diff resolves wait for the inMemory passthrough to resolve).

Good point about the delay to the fswrite stream

PR #1016, opened 2021-07-16
feat: build/deploy the website in CI

notnmeyer:
deployment automation for the website. the gha workflow will deploy to either staging or production,
* pushes to a PR open against develop will deploy to staging
* pushes to develop or release (which will happen when changes are merged) will deploy to production

here's a successful deploy to staging from this branch, https://github.com/opticdev/optic/runs/3089397143. i'll verify prod deploys without a problem when this is merged.

i also included website's taskfile in the root taskfile. its tasks are available from the repo root under the "website" prefix,
```
➜ pwd
/Users/nate/code/optic/optic

➜ task -l | grep website
* website:build:                                        Build the website
* website:deploy:                                       Deploy the website to S3
* website:start:                                        Run the website locally
```

and you can still run them from the website dir as well, just drop the prefix,
```
➜ pwd
/Users/nate/code/optic/optic/website

➜ task -l
* build:        Build the website
* deploy:       Deploy the website to S3
* start:        Run the website locally
```

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -9,12 +9,14 @@ on:
       - website/**
       - .github/**
       - "**.md"
+      - "Taskfile.yml"
   pull_request:
     branches: [ 'develop' ]
     paths-ignore:
       - website/**
       - .github/**
       - "**.md"
+      - "Taskfile.yml"
 
 # A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:
--- a/.github/workflows/website.yml
+++ b/.github/workflows/website.yml
@@ -1,0 +1,31 @@
+name: Website
+
+on:
+  pull_request:
+    branches: [develop]
+    paths:
+      - .github/wokflows/website.yml
+      - website/**
+  push:
+    branches: [release]
+    paths:
+      - website/**
+  workflow_dispatch:gs
+
+jobs:
+  run:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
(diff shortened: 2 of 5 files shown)


.github/workflows/run_tests.yml
```diff
@@ -9,12 +9,14 @@ on:
...
       - .github/**
       - "**.md"
+      - "Taskfile.yml"
   pull_request:
     branches: [ 'develop' ]
     paths-ignore:
       - website/**
       - .github/**
       - "**.md"
+      - "Taskfile.yml"
```
notnmeyer, line 19: this workflow ran against my PR and i certainly wouldnt expect it to. i we should exclude changes to the root Taskfile.yml from running this workflow.

notnmeyer:
its still possible to deploy locally as well with something like, `AWS_PROFILE=staging APP_ENV=staging DRY_RUN=true task website:deploy`—but holler if you find yourself needing that since its probably a smell

LouManglass (approved):
I gave a quick review of the aws-actions documentation, and there's one part I don't get: In this GitHub Action yaml, aws-actions configures our credentials, then we assume the staging role if necessary. We never assume the production role. I'm not sure if that's a default that we get when configuring AWS credentials, or if that's yet to come based on the PR comment:

> here's a successful deploy to staging from this branch, https://github.com/opticdev/optic/runs/3089397143. i'll verify prod deploys without a problem when this is merged.

But in either case I'm sure you have this addressed - I'm just not sure how that should work. So, I'm approving.

notnmeyer:
the "Configure AWS credentials" is production credentials. prod creds are used to assumed staging creds when necessary. its set up this way so we can use a single set of creds rather than a set per env.

PR #1017, opened 2021-07-16
Prevent invalid interactions from generating diffs

niclim:
## Why

As a follow up to this https://github.com/opticdev/optic/pull/1007, moved the invalid interactions filtering logic into the diff flow. What happens is we potentially generate interactions that are invalid

An example interaction (with invalid requests and responses) - copy paste this into todos-partial

```json
      {
        "uuid": "aaaf782f-8e44-4e1327-806d-032323fb0a4fd",
        "request": {
          "host": "localhost",
          "method": "GET",
          "path": "/api/lists/list123/completed",
          "query": {
            "shapeHashV1Base64": null,
            "asJsonString": null,
            "asText": null
          },
          "headers": {
            "shapeHashV1Base64": null,
            "asJsonString": null,
            "asText": null
          },
          "body": {
            "contentType": "application/json",
            "value": {
              "shapeHashV1Base64": null,
              "asJsonString": null,
              "asText": null
            }
          }
        },
        "response": {
          "statusCode": 200,
          "headers": {
            "shapeHashV1Base64": null,
            "asJsonString": null,
            "asText": null
          },
          "body": {
            "contentType": ,
            "value": {
              "shapeHashV1Base64": null,
              "asJsonString": null,
              "asText": null
            }
          }
        },
        "tags": []
      },

```

## What

I'm not sure if this is the right place to put this. I thought about putting this logic in the interaction visitor, but I'm also not sure that the visitor is the right place to put this. The visitor visits multiple nodes in the endpoints graph, given a single http interaction, given that the http interaction is "invalid" - i.e. we cannot create and learn from this interaction, it seems right to remove it up front?

The only other thing is, lets say we send a request with valid query + request bodies, but the response is invalid. Should we discard the entire interaction, or should we learn from the valid query and request bodies and discard the response body? If that's the case, we should move this block https://github.com/opticdev/optic/blob/develop/workspaces/optic-engine/src/interactions/traverser.rs#L32-L98 (perhaps as part of the Visitor structs)?

Thoughts?

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/optic-engine/src/interactions/visitors/diff.rs
+++ b/workspaces/optic-engine/src/interactions/visitors/diff.rs
@@ -196,15 +196,16 @@ impl RequestBodyVisitor<InteractionDiffResult> for DiffRequestBodyVisitor {
 
   fn visit(&mut self, interaction: &HttpInteraction, context: &RequestBodyVisitorContext) {
     if let Some(operation) = context.operation {
-      let actual_content_type = &interaction.request.body.content_type;
-      let actual_body: Option<BodyDescriptor> = (&interaction.request.body.value).into();
-      let (request_id, request_body_descriptor) = operation;
-      //dbg!( actual_content_type);
-      //dbg!(&request_body_descriptor);
+      let maybe_interaction_content_type = &interaction.request.body.content_type;
+      let maybe_interaction_body_descriptor: Option<BodyDescriptor> =
+        (&interaction.request.body.value).into();
+      let (request_id, request_descriptor) = operation;
+      //dbg!( maybe_interaction_content_type);
+      //dbg!(&request_descriptor);
       match (
-        &request_body_descriptor.body,
-        actual_content_type,
-        actual_body,
+        &request_descriptor.body,
+        maybe_interaction_content_type,
+        maybe_interaction_body_descriptor,
       ) {
         (None, None, _) => {
           self
@@ -259,7 +260,7 @@ impl RequestBodyVisitor<InteractionDiffResult> for DiffRequestBodyVisitor {
   fn end(&mut self, interaction: &HttpInteraction, context: &PathVisitorContext) {
     if let Some(path_id) = context.path {
       if self.visited_with_matched_content_types.is_empty() {
-        let actual_content_type = &interaction.request.body.content_type;
+        let maybe_interaction_content_type = &interaction.request.body.content_type;
         let mut interaction_trail_components = vec![
           InteractionTrailPathComponent::Url {
             path: interaction.request.path.clone(),
@@ -268,7 +269,7 @@ impl RequestBodyVisitor<InteractionDiffResult> for DiffRequestBodyVisitor {
             method: interaction.request.method.clone(),
           },
         ];
-        if let Some(content_type) = actual_content_type {
+        if let Some(content_type) = maybe_interaction_content_type {
           interaction_trail_components.push(InteractionTrailPathComponent::RequestBody {
             content_type: content_type.clone(),
           });
@@ -316,18 +317,19 @@ impl ResponseBodyVisitor<InteractionDiffResult> for DiffResponseBodyVisitor {
   fn visit(&mut self, interaction: &HttpInteraction, context: &ResponseBodyVisitorContext) {
     //dbg!("visit response body");
     if let Some(response) = context.response {
-      let actual_content_type = &interaction.response.body.content_type;
-      let actual_body: Option<BodyDescriptor> = (&interaction.response.body.value).into();
-      let (response_id, response_body_descriptor) = response;
-      //dbg!("actual response content type", actual_content_type);
+      let maybe_interaction_content_type = &interaction.response.body.content_type;
+      let maybe_interaction_body_descriptor: Option<BodyDescriptor> =
+        (&interaction.response.body.value).into();
+      let (response_id, response_descriptor) = response;
+      //dbg!("actual response content type", maybe_interaction_content_type);
       // dbg!(
       //   "expecting response content type",
-      //   &response_body_descriptor
+      //   &response_descriptor
       // );
       match (
-        &response_body_descriptor.body,
-        actual_content_type,
-        actual_body,
+        &response_descriptor.body,
+        maybe_interaction_content_type,
+        maybe_interaction_body_descriptor,
       ) {
         (None, None, _) => {
           self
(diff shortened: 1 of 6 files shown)


workspaces/optic-engine/tests/snapshots/interaction_diff__can_yield_unmatched_shape-2.snap
```diff
@@ -1,31 +1,5 @@
 ---
-source: workspaces/optic-engine-native/tests/interaction_diff.rs
```
niclim, line 2: This snapshot is correct now - it's in "incompliant" interaction https://github.com/opticdev/optic/blob/develop/workspaces/optic-engine/tests/interaction_diff.rs#L251 - this should not generate a diff

Interestingly, this test has been here for a while, so this is probably something we knew about?

JaapRood (commented):
> the only other thing is, lets say we send a request with valid query + request bodies, but the response is invalid. Should we discard the entire interaction, or should we learn from the valid query and request bodies and discard the response body?

The root of the problem, as I understand it (please correct me if I'm wrong), is that a `Body` is capable of describing an invalid variant: when `content_type` is `Some(_)`, but the `Option<BodyDescriptor>` (derived from the ArbitraryData by trying to parse from it's attributes in order) is `None`. For that variant we want to not visit the body, and thus not generate any results (either matching or non matching), right?

Framed like that, I reckon this is something we want to handle in the interaction traverser: that's where we decide how to the interaction is visited. It should be able to deal with any `HttpInteraction` variant and actually already reasons about its domain specific properties. Technically the `Body` struct has enough information to determine whether there it is representing a domain-invalid variant, so perhaps a method implemented on that makes sense.

With that approach, we take the route of diffing as much of an interaction as possible, despite one of the bodies maybe not having persisted correctly. That definitely seems like the more fault tolerant approach!

niclim:
> All that being said: I actually have trouble generating a crash by pasting the interaction in the description in todos-partial.json: a diff shows up, it can be handled, 🤷‍♂️.

Oh sorry, change the interaction url to something like `"path": "/api/abc/abc",` - you can learn the initial body, and then the diff continues to show up (in subsequent diff matches). Perhaps it's bug in the initial body learner?

niclim:
Invalid interactions (in this definitions, it's diffs that continue to not be learnt) - requests with a content type but _no_ body, or responses with _no_ body

So looking further into this, here's what happens:
- A diff with a "null" `ArbitraryData` matches a path and produces an unmatchedBody (request or response)
- If there is no body set in the specification events, this will produce a diff, which doesn't generate any bodies.
- As a result, when the spec request/response body is `None` and the interaction body is `None`, we should not generate a diff.

I think it's actually more correct to learn this diff, which is an empty body with content types. I've made changes so that the learner will not produce a diff after the learned bodies are generated here (which I think is more correct)

I think there's a little work to do on the UI to display this better, since we don't show empty bodies, but this should prevent users from having a diff that can never be learnt. This includes:
- Update `endpoints/thunks` to not swallow invalid bodies
- Update diff flow to show request body with no body in checkbox
- Update UI components for rendering UI to show empty bodies - polymorphic merged?

I'm updating the spectacle graph to expose an endpoint projection, with that work, the above UI changes should be easier to make.

workspaces/optic-engine/tests/interaction_diff.rs
```diff
@@ -240,15 +255,29 @@ async fn can_yield_unmatched_shape() {
...
     compliant_interaction,
     &DiffInteractionConfig::default(),
   );
   assert_debug_snapshot!(results);
   assert_eq!(results.len(), 0);
+}
 
-  let incompliant_interaction = HttpInteraction::from_json_str(
+#[test]
+fn can_handle_no_request_and_response_body() {
```
niclim, line 268: New test, this includes events that are created when we learn an interaction with no request or response body (i.e. there are no shapes)

workspaces/optic-engine/src/interactions/visitors/diff.rs
```diff
@@ -197,26 +197,36 @@ impl RequestBodyVisitor<InteractionDiffResult> for DiffRequestBodyVisitor {
   fn visit(&mut self, interaction: &HttpInteraction, context: &RequestBodyVisitorContext) {
     if let Some(operation) = context.operation {
       let actual_content_type = &interaction.request.body.content_type;
+      let actual_body: Option<BodyDescriptor> = (&interaction.request.body.value).into();
       let (request_id, request_body_descriptor) = operation;
```
JaapRood, line 201: Nit on a bit of possible clean up while we're touching this: I wonder if we can rename these to be a little bit less confusing, I had to look at this multiple times to see what was what 😅. Took me a bit to realise`actual_body` is actually a `BodyDescriptor` not a `Body`, while the thing called `request_body_descriptor` isn't a `BodyDescriptor`. "actual" is also pretty vague, come to think of it; actually what? Perhaps distinguishing between what comes from the interaction and what comes from the spec is helpful? 🤷‍♂️

workspaces/optic-engine/src/interactions/visitors/diff.rs
```diff
@@ -197,26 +197,36 @@ impl RequestBodyVisitor<InteractionDiffResult> for DiffRequestBodyVisitor {
...
+        (None, Some(content_type), None) => {
+          self
+            .visited_with_matched_content_types
+            .insert(request_id.clone());
+        }
+        (None, Some(content_type), Some(interaction_body)) => {
           self
             .visited_with_unmatched_content_types
             .insert(request_id.clone());
         }
```
JaapRood, line 223: The difference between these two cases is definitely subtle, but written this way also makes sense. The interaction might have been sent with a content type, but in the end we're matching bodies to bodies. The content type is only really interesting if you want to compare 2 present bodies, a `None` matching `None` doesn't require that.

Nice one!

JaapRood (approved):
Great one! This generalised the matching logic nicely. The visitor is exactly where this kind of matching logic should live. Especially since this shows how that logic can be subtle, it's great it prevents the need for some validation layer to live somewhere else. The visitor should be robust enough to handle any variant of `HttpInteraction`, and this achieves that!

workspaces/optic-engine/src/interactions/visitors/diff.rs
niclim: Changed to `interaction_...` and `maybe_...` to indicate `Option`

PR #1018, opened 2021-07-16
Only learn query parameters when observerd

niclim:
## Why

Previously, we were learning query parameters everywhere (even if they were unobserved) - updating this to not do that. This was introduced to fix a previous bug  https://github.com/opticdev/optic/pull/984

## What

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


JaapRood (approved):
That one was definitely my bad 😅, thanks for the fix!

PR #1021, opened 2021-07-19
Expose an endpoint node in spectacle projection

niclim:
## Why

As discussed separately, there's more and more of a case we want to build towards having endpoint modeled in our spectacle graph. This PR is the first step in moving towards this.

## What

- Update the spectacle/endpoints projection in rust to expose an endpoint node. This means that for a method + pathId, there will be a single endpoint node, which all the query, request and response nodes will point to.
- This includes implicitly adding and removing an endpoint node in the spectacle projection (i am not sure if this is fully necessary right now, but it's definitely more correct to have the endpoint creation / removal state from other events).
- Updated the spectacle projection so that we have an endpoints resolver (combined the command deletion generation into this and created a CommandGenerator class - to model command generation in spectacle).

The future tasks from here are to:
- Switch all the consumers that are using `request` resolvers to the endpoint resolvers (this includes the UI, changelog, etc)
- Mark the `request` resolver as deprecated (we definitely want to do thsi in the future, but it might be easier to just skip this step if we don't have any external consumers) - @devdoshi what's your thoughts here?
- Remove the request resolver and associated things in the future
- Update the UI to fix the bug where multiple paths show up https://github.com/opticdev/optic/issues/929#issuecomment-876749434
- Update the UI to surface no body request / responses (currently being swallowed in the endpoints/thunk) - related to https://github.com/opticdev/optic/pull/1017#issuecomment-882845528

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


niclim:
There's a couple of snapshots generated to see what the new projection looks like - `workspaces/optic-engine/tests/snapshot/domain_conflict_scenarios_generated__scenario_9949a8cc69a0063a70f5ae98672d8c23a1068a2b2d08f8bb7b9bbd968ec29f0d__spectacle_endpoints_graph.snap`

I think some of our snapshot data isn't completely correct (taking a look at some, there's orphaned batch commit nodes, indicating the batch commit data or event context data isn't correct). I'll look into this in a future PR (we could use with better tests around the spectacle / rust projection side

workspaces/optic-engine/src/projections/spectacle/endpoints.rs
```diff
@@ -344,28 +349,123 @@ impl EndpointsProjection {
     }
   }
 
+  ////////////////////////////////////////////////////////////////////////////////////////////////////
+  fn get_or_create_endpoint_node_index(
```
JaapRood, line 353: Nit: we've called these get or create methods "ensure_" in the other projections, could be nice to have that consistent :)

JaapRood (approved):
Looks like a solid implementation of a good idea. We've seen the concept of an Endpoint take more and more of a hold in the UI, so inevitably it also because an important unit of partition in the data model. That said, it _is_ a user domain concept at this moment, with the internals of diffing and learning being able to work more directly on query parameters, requests and responses. Adding an endpoint node to the Spectacle graph, seems like the right balance to be struck here, or at least a first good step in search of the correct balance.

PR #1025, opened 2021-07-20
chore(website): add workflow_dispatch

notnmeyer:
adds workflow_dispatch so we can run the workflow whenever

PR #1026, opened 2021-07-20
fix(website): build CI correctly when merged to develop

notnmeyer:
the last (i hope so bad) in my series of "make CI work correctly" PRs. https://github.com/opticdev/optic/pull/1025 fixed a typo that was preventing CI from running when i expected it to. i noticed that after merging that we did a staging deploy when we expected prod. 

quoting the regex when we do the branch comparison causes it to be treated like a string, which is not the desired behavior.

```
bash-5.1$ [[ "develop" =~ "^(develop|release)$" ]] && echo "Hi"
bash-5.1$ [[ "develop" =~ ^(develop|release)$ ]] && echo "Hi"
Hi
```

### DIFF
--- a/.github/workflows/website.yml
+++ b/.github/workflows/website.yml
@@ -28,7 +28,7 @@ jobs:
           then
             echo "Deploying to Production..."
             echo "APP_ENV=production" >> $GITHUB_ENV
-          else=
+          else
             echo "Deploying to Staging..."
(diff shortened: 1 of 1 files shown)


notnmeyer:
https://github.com/opticdev/optic/runs/3115584995?check_suite_focus=true 👍

PR #1027, opened 2021-07-20
UI to use endpoint spectacle

niclim:
## Why

This one is a big one, apologies for the large diff - it started trying to migrate incrementally, but kept finding things we needed to update and change to get this to work "properly".

In moving to the new endpoint projection, its now easier to handle multiple request and response bodies. In doing so, there were a number of places to update in the UI.

This also fixes a number of bugs:
- Requests showing up multiple times because of different content types https://github.com/opticdev/optic/issues/929#issuecomment-876749434
- UI now displays when there is no body requests or responses (to actually generate this is a little weird for requests, since requests only get treated as "no body requests" when they attach a content-type on to it). There's actually more we should do, since if we have a request that accepts _both_ application/json AND no request body, we don't actually show a no body request until we send a request with contentType: Something, body: None. This is something we _kind of_ model, but aren't handling correctly when handling the interaction, but i figured this was a smallish edge case and I've already spent enough time digging into this.

## What

Screen Shot 2021-07-20 at 3 18 09 PM

## Validation

* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/ui-v2/src/components/HttpBodySelector.tsx
+++ b/workspaces/ui-v2/src/components/HttpBodySelector.tsx
@@ -1,0 +1,76 @@
+import React, { useState } from 'react';
+import {
+  FormControl,
+  FormControlLabel,
+  makeStyles,
+  Radio,
+  RadioGroup,
+} from '@material-ui/core';
+
+import { FontFamily } from '<src>/styles';
+
+type HttpBodySelectorProps<T> = {
+  items: T[];
+  children: (body: T) => React.ReactNode;
+  getDisplayName: (item: T) => string;
+};
+
+/**
+ * A component that is used to select / handle different items
+ */
+export const HttpBodySelector: <T extends {}>(
+  props: HttpBodySelectorProps<T>
+) => React.ReactElement = ({ items, children, getDisplayName }) => {
+  const [selectedItem, setSelectedItem] = useState(0);
+  const inBoundsSelected = Math.min(selectedItem, items.length - 1);
+  const classes = useStyles();
+
+  return (
+    <>
+      {items.length === 0 ? (
+        <>No bodies found</>
+      ) : items.length === 1 ? (
+        children(items[0])
+      ) : (
+        <>
+          <FormControl component="fieldset" className={classes.formGroup}>
+            <RadioGroup
+              value={selectedItem}
+              onChange={(e) => {
+                setSelectedItem(Number(e.target.value));
+              }}
+              className={classes.radioGroup}
+            >
+              {items.map((item, i) => (
+                <FormControlLabel
+                  classes={{
+                    label: classes.radio,
+                  }}
+                  key={getDisplayName(item)}
+                  value={i}
+                  control={<Radio />}
+                  label={getDisplayName(item)}
+                />
+              ))}
+            </RadioGroup>
+          </FormControl>
+          {children(items[inBoundsSelected])}
+        </>
+      )}
+    </>
+  );
+};
+
+const useStyles = makeStyles((theme) => ({
+  formGroup: {
+    marginBottom: theme.spacing(2),
+  },
+  radioGroup: {
+    display: 'flex',
+    flexDirection: 'row',
+  },
+  radio: {
+    fontFamily: FontFamily,
+    fontSize: theme.typography.fontSize - 1,
+  },
+}));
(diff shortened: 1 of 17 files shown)


workspaces/ui-v2/src/components/HttpBodySelector.tsx
```diff
@@ -0,0 +1,76 @@
...
+  props: HttpBodySelectorProps<T>
+) => React.ReactElement = ({ items, children, getDisplayName }) => {
+  const [selectedItem, setSelectedItem] = useState(0);
+  const inBoundsSelected = Math.min(selectedItem, items.length - 1);
+  const classes = useStyles();
+
+  return (
+    <>
+      {items.length === 0 ? (
+        <>No bodies found</>
```
niclim, line 31: I don't think we ever hit this branch, but since it's a "generic" component (likely we could rename this to MultiSelector) - we would hit it and handle it. It's poorly styled because we don't hit this path

workspaces/ui-v2/src/pages/diffs/EndpointDocumentationPane.tsx
```diff
@@ -110,65 +111,91 @@ export const EndpointDocumentationPane: FC<
...
-                      />
-                    )}
-                  </ShapeFetcher>
+          </div>
+        </HighlightedLocation>
+      )}
+
+      {selectors
+        .getResponsesInSortedOrder(thisEndpoint.responsesByStatusCode)
+        .map(([statusCode, responses]) => {
```
niclim, line 155: Had to restructure this, such that statusCodes are grouped together. Otherwise we'd have:
200 Response
responseId 1
200 Response
responseId 2

Which isn't easy to read (likely we'd want to group these together)

workspaces/ui-v2/src/pages/docs/components/EndpointTOC.tsx
```diff
@@ -33,47 +34,58 @@ export function EndpointTOC(props: EndpointTOCProps) {
...
         />
       )}
 
-      {props.requests.map((request) => (
+      {props.requests.length > 0 && (
         <EndpointTOCRow
-          key={request.requestId}
           label={'Request Body'}
-          anchorLink={request.requestId}
+          anchorLink={'request-body'}
```
niclim, line 52: Updated the anchor links - in a single endpoint page there are:
- 1 or 0 query parameters
- 1 or 0 request (with multiple / polymorphic request bodies)
- x number of status codes (with multiple request bodies)

workspaces/ui-v2/src/types/endpoints.ts
```diff
@@ -39,23 +41,31 @@ export interface IQueryParameters {
...
   responseId: string;
   statusCode: number;
-  contentType: string;
-  rootShapeId: string;
+  endpointId: string;
   pathId: string;
   method: string;
   description: string;
+  body: IBody | null;
+}
```
niclim, line 66: Currently what happens is, requests and responses created for every new content type (i.e. contentType `application/json` and `text/html` generate a new request/responseId - meaning that bodies is only _ever_ one or none.

I'm not 100% which one we'd want (i.e. a new request / response id per content type, or content types / bodies pointing to the same request node). Either way, this PR is big enough as it is, and should be updated in the spectacle projection -> into the UI

workspaces/ui-v2/src/components/HttpBodySelector.tsx
```diff
@@ -0,0 +1,76 @@
...
+                    label: classes.radio,
+                  }}
+                  key={getDisplayName(item)}
+                  value={i}
+                  control={<Radio />}
+                  label={getDisplayName(item)}
+                />
+              ))}
+            </RadioGroup>
+          </FormControl>
```
JaapRood, line 56: I'm not sure radio buttons are what I would have gone with, but the reasons for that are probably too subtle to not just ship this for now (especially as our future bet is very likely to hit this UI again anyway). 

Just for completeness sake: in my mind, all use of form controls imply a user context where the user is making mutations to state. With these radio buttons you could interpret and read this as choosing which content type is currently active (not just for navigating, but in how it works). Once edit mode + other forms come into play, this issue becomes a bit bigger. With these buttons being used straight up for navigation purposes only, we could alleviate that source of confusion by rendering them with other types of elements.

JaapRood (approved):
Looks good, and the increased support for rendering multiple body types a validation of how this new endpoint concept was worth adding.

workspaces/ui-v2/src/components/HttpBodySelector.tsx
niclim: Hmmmmmm that's a really good point, and something I overlooked. Lets revisit this in our next bet - as this should be fairly easy to change since it's all in one space.

acunniffe: +1 the @JaapRood feedback. Reviewing all PRs in tomorrow's release and this stood out to me. 

since we're touching this area of the UI next I say ship it.

PR #1028, opened 2021-07-20
chore: Dependency updates

niclim:
## Why

Misc chore stuff - updating our dependencies, ripping out unused ones, some other clean up.

Dependabot alerts:
- react-dev-utils
- node-notifier

## What

Moved ui depedencies from dependencies -> devDependencies

Remove invariant dependency (throwing errors is the same thing, provides better type safety and has less indirection)

Breaking changes for dependencies:
- `react-dev-utils` 10 -> 11 - we only use openBrowser, which looks like it still takes the same interface (I couldn't find where to dig up the changelog, since it looks like it's intended as an internal package to create-react-app)?
- `node-notifier` 7 -> 10 - https://github.com/mikaelbr/node-notifier/blob/master/CHANGELOG.md - biggest one is timeout from seconds -> miliseconds, I've updated our usage
- sentry/node 5 -> 6 - https://github.com/getsentry/sentry-javascript/releases/tag/6.0.0 no breaking changes

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


JaapRood (commented):
I don't think we should get rid of `invariant`, if anything I'd say we want to use it _more_, as the invariant errors are a lot specific than generic `Error` instances, especially when involving debugging through Sentry. If there is a more specific error to throw that is caught down the line, that might be preferable, but for invariant errors, invariant is the best tool we have.

niclim:
@JaapRood What sort of extra error messages with invariant do we get? The biggest reason I dislike it is because typescript hasn't been infering type safety through the `invariant` calls, where `if condition -> throw error` does. I'd prefer the second option, as weaker type safety has been one of the reasons we've missed a lot of bugs in the diff flow (with `!` and `variable as Type`) - and I'd prefer to have stronger type safety, over better debugging (i.e. prevent bugs from happening, over fixing bugs after they happen)

JaapRood:
> What sort of extra error messages with invariant do we get? 

It's not the error message, it's that an Invariant Violation is a more specific type of error than a generic `Error` instance. It means the app got into a state we never expected it to be in, akin to an `unreachable!` in Rust vs a generic `panic!`. It's the kind of error that should never be caught and crash the program, as program in an undefined state can't reliably be recovered. 

I'm not quite sure how Typescript safety helps with those kinds of bugs: they are not meant to be caught and handled (which is what I imagine this is about), but should be viewed as panics. 

Now if you reckon any of the bugs that invariant was used for are recoverable, than it's a different story, of course!

JaapRood:
I guess what you mean with type safety is that if you have asserted the invariant, like a property existing on an object, Typescript can't then also infer that property _must_ exist once passed the assertion?

I wonder what else we can do to make that work, other than just dumping invariant violation errors in favour of generic `Error` instances. We could throw our own `InvariantViolationError`, at a minimum? Or perhaps there is some other type of Typescript magic?

niclim:
> I guess what you mean with type safety is that if you have asserted the invariant, like a property existing on an object, Typescript can't then also infer that property _must_ exist once passed the assertion?

Correct. If we do:
```javascript
const shouldNeverBeFalse = false;
if (!shouldNeverBeFalse) {
    throw new Error()
}
// shouldNeverBeFalse will be infered as `true`

// where as
invariant(shouldNeverBeFalse, 'asd');
// typescript can't infer through function calls that shouldNeverBeFalse is only true.
```

Throwing an error with a custom message, makes it just as easy to trace in code (search in editor) - if we really need the visual cue that "this is an invariant" we can create our own custom class to throw, but I'm not convinced that this is that necessary. 

Additionally, I think a lot of the invariants can be cut down by better typings / code paths - something I was planning to do with refactoring the ParsedDiff class. We probably also _don't_ want to crash the UI if we hit an invariant, rather alert sentry, and ignore the diff / continue on. There's an error handling strategy we can probably come up with and have better distinction between code paths (i.e. modules that could raise errors, and consumers of them - components / redux, that call them and catch errors / alert sentry, etc)

niclim:
Ok I added in an `InvariantViolationError` class, which adds more granularity to errors, and hopefully is a good starting point for better subclassing of errors

JaapRood (approved):

LouManglass:
@niclim @JaapRood  Should we expect this in today's release?

JaapRood:
@LouManglass since it's a chore, I'm not sure why would take the risk of bumping an outside dependency destabilising the release. I'd say hold it, because we can. Unless @niclim feels strongly otherwise, it was approved after all!

LouManglass:
> @LouManglass since it's a chore, I'm not sure why would take the risk of bumping an outside dependency destabilising the release. I'd say hold it, because we can. Unless @niclim feels strongly otherwise, it was approved after all!

Sounds reasonable, I'll branch for release now then. We'll release this next cycle, so it'll have some local development testing under it.

niclim:
Yep, I'm going to wait until after the release to merge this, there's no need to add additional risk to the release for this

PR #1038, opened 2021-07-21
doc: plumb share use case

LouManglass:
## Why

Add sharing use cases

## What

Pretty much everything under:

- https://o3c.info/share

The latest push adds some integrations under https://o3c.info/share/scripts, which also appear on the Reference page (https://o3c.info/reference) now as well. Share also appears in the Use Cases drop down.

Still WIP, but ready for comments as I add a few final touches.

## Validation
* [ ] CI passes
* [ ] Verified in staging


acunniffe:
Nice work!

PR #1039, opened 2021-07-21
Remove query parameter FFs

niclim:
## Why

Remove the FFs. Not to be merged before Friday

## What

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


niclim:
@JaapRood I'm going to merge this in first, will address any comments retroactively - there's some weirdness with FF injection in cloud spec viewer and this is the easiest way to "turn on" cloud spec viewer query parameters (i'll fix the root cause tomorrow)

PR #1040, opened 2021-07-22
Polish Query Parameters UI and UX

JaapRood:
## Why
After our first pass of the UI and UX for Query Parameters, there were a couple of things left over to polish. Now that we now we there's no breaking bugs left for release, we can use the last of our appetite to revisit a couple of smaller things.

## What

- While we model multiple occurring query parameters as Lists in the spec, we can't actually say that this is how the server will interpret them. Nor do we support any of the other ways lists are expressed (e.g. through comma separate values). The only thing we can really say is whether the API that's being documented intents to allow keys to be sent multiple times (and knows what to do with them). So instead of rendering these as lists in the shape viewer, documentation and diff messaging, we now render the attribute "multiple". 
Screenshot 2021-07-22 at 13 16 30
Screenshot 2021-07-22 at 11 32 56
- In some cases, the query string was being rendered as structured JSON inside the diff example. While this data is available in the interaction, it's actually intentionally not accepted by the Optic Engine at the moment, always parsing the raw query string itself (that's how we ensure the stability we've got at the moment). So to represent that as JSON in the UI that way, is giving a false sense of Optic's current capabilities. Therefor, all bodies are rendered as raw at the moment. In the future, we could enhance this rendering, but since that will have to be powered by the Optic Engine (as we want to rely on _its_ parsing semantics), that currently goes beyond our appetite.
- The query parameters parsing information was moved to the DiffCard. Especially as we're rendering raw query strings for diffing, it's the most important place to allow users to understand how Optic interprets the example. A stronger approach to this would for both the raw and parsed body to be rendered, allowing a user to see directly how parsing occurred. However, as said above, that goes beyond our current appetite, so we'll stick with the tooltip for now.
Screenshot 2021-07-22 at 13 12 04

## Validation
* [x] CI passes
* [x] All changes are purely local UI, no Optic Engine, Spectacle, cli, etc.
* [x] All `diff-use-cases-with-events` examples manually verified

### DIFF
--- a/workspaces/ui-v2/public/example-sessions/diff-use-cases-with-events.json
+++ b/workspaces/ui-v2/public/example-sessions/diff-use-cases-with-events.json
@@ -2139,6 +2139,145 @@
           "createdAt": "2021-07-12T12:03:24.891+02:00"
         }
       }
+    },
+    {
+      "BatchCommitStarted": {
+        "batchId": "f64b139e-5637-407e-a85c-e4d3a2f01c27",
+        "commitMessage": "Add query parameter tags",
+        "eventContext": {
+          "clientId": "splendid-deer-13",
+          "clientSessionId": "9667c8ae-5c32-4bb5-b442-9170533935fe",
+          "clientCommandBatchId": "f64b139e-5637-407e-a85c-e4d3a2f01c27",
+          "createdAt": "2021-07-22T10:36:59.319+02:00"
+        },
+        "parentId": "65d32f89-4b62-42c8-a657-7a84ba5fe13d"
+      }
+    },
+    {
+      "ShapeAdded": {
+        "shapeId": "shape_rUSV0iS4eM",
+        "baseShapeId": "$string",
+        "parameters": {
+          "DynamicParameterList": {
+            "shapeParameterIds": []
+          }
+        },
+        "name": "",
+        "eventContext": {
+          "clientId": "splendid-deer-13",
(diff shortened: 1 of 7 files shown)


JaapRood:
@LouManglass since we're so close to the release train, I'd appreciate your review on this as well. I'm confident this is all solid and won't impact stability, but we can leave it if you think it brings stability into question.

LouManglass:
> @LouManglass since we're so close to the release train, I'd appreciate your review on this as well. I'm confident this is all solid and won't impact stability (see validations in the PR description), but we can leave it if you think it brings stability into question.

I'm actually going to put this question back in your corner: I don't really have the time to validate this PR and do the release usability testing today, and I see two paths forward:

- If you're confident it's solid, or at least solid enough (seems reasonable), then get a code review and let's merge it this "morning" (Eastern). I'll add it to the release and it'll go through usability testing. We'll have to fix or revert if issues are uncovered.
- If that sounds like a bit too much, let's put this in the next release.

IMO it seems reasonable to get pulled in to this release with a code review, but I can't really give it coverage outside of release testing. If that's worrying, we should hold back.
- If

JaapRood:
@LouManglass I'm confident, with your QA intended as more of a sanity check than anything else. I intentionally limited the scope of these to be of low impact and basically have nothing to do with correctness of results. Also really easily reverted if it does out of the blue break stuff. 

So if I can get someone to code review this, lets ship it.

niclim (approved):

PR #1041, opened 2021-07-22
add intercom support

acunniffe:
## Why
- to send messages to oss users who have had trouble OR made it far in Optic
- to announce product updates to active users 

## What
Respects the users's anonymity by using the random analytics ID as the intercom user_id. Also allows us to target specific users who have had certain classes of issues, or who have used certain features, with in-app messages or questions. 

By default the chat bubble is hidden, our message sequencing/targeting can be set up on intercom's SaaS website. 

## Validation
* [ ] CI passes
* [ ] Verified in staging


devdoshi:
do we want it on demo and examples too?

workspaces/ui-v2/src/contexts/analytics/implementations/localCliAnalytics.ts
```diff
@@ -60,6 +60,17 @@ export const initialize: AnalyticsStoreProps['initialize'] = async (
     });
     Sentry.setUser({ id: metadata.clientAgent });
   }
+
+  // @ts-ignore
+  if (window.Intercom && appConfig.analytics.intercomAppId) {
+    // @ts-ignore
```
niclim, line 66: Throw this in the top of this file - this should fix the ts ignore things
```
declare global {
  interface Window {
    Intercom?: () => any;
  }
}
```

niclim (approved):
There's probably some things we can do to make this clearer that intercom is only for local-cli (in config), but this is fine for now

acunniffe:
> do we want it on demo and examples too?

I don’t think so, but we can bring it in later if we have a story for it

workspaces/ui-v2/src/contexts/analytics/implementations/localCliAnalytics.ts
acunniffe: On it

workspaces/ui-v2/src/contexts/analytics/implementations/localCliAnalytics.ts
```diff
@@ -5,6 +5,12 @@ import { LogLevel } from '@sentry/types';
 import { AnalyticsStoreProps } from '../AnalyticsStore';
 import { getOrSetAgentFromLocalStorage } from '../utils';
 
+declare global {
+  interface Window {
+    Intercom?: () => any;
```
niclim, line 10: oops, realize this needs to be `(...args: any)`

PR #1046, opened 2021-07-23
Update env variable injection for cloud viewer and local builds

niclim:
## Why

Cloudviewer wasn't getting any environment variables being injected into it when built - instead, it had adhoc env variable injection. This was confusing so I updated it to be consistent with how other environment variables are injected.

I also split out the REACT_APP_STAGING and PROD specific environment variables. Instead, made these injectable so the UI does _not_ need to know about different environments - these should be injected at build time instead. This was not fully possible since cloud spec viewer uses the same docker image to run in both staging _and_ production.

As part of this, also made it that:
- any user facing build, alpha/beta/latest environments point to prod environment
- sidechannel and local point to staging environment

## What

The UI deployment with environment variables (cloud viewer and demo) could really use some investment with their build processes. The cloud viewer and demo use a slightly different task build (perhaps we could consolidate the release / deployment flows here) - and we probably want different docker images (different by environment variable) for staging and prod. I think this needs to be in the build stage since the UI creates a static bundle with environment variables _already_ injected into them.

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


Taskfile.yml
```diff
@@ -29,8 +29,13 @@ includes:
     dir: website
 
 tasks:
+  # Set up a local environment
   postpull:
-    deps: ['workspaces:clean', 'workspaces:build']
+    cmds: 
+      - task: 'workspaces:clean'
+      - FLAGS_FILE=.env.local task workspaces:setenv
+      - task: 'workspaces:build'
```
niclim, line 37: Is there a better way to specify env variable setup using task?

Taskfile.yml
```diff
@@ -60,6 +65,7 @@ tasks:
           cp $FLAGS_FILE workspaces/cli-scripts/.env
           cp $FLAGS_FILE workspaces/optic-engine/.env
           cp $FLAGS_FILE workspaces/optic-engine-native/.env
+          cp $FLAGS_FILE workspaces/ui-v2/.env.development.local
```
niclim, line 68: This should be fine to add in - this env variable is used by the create-react-app `yarn start` instance, the production env variable is used for the `yarn build` instance

workspaces/ui-v2/src/entry-points/cloud/cloud-viewer.tsx
```diff
@@ -49,10 +49,12 @@ const appConfig: OpticAppConfig = {
...
     },
     backendApi: {
+      // Because we use the same docker image for both staging, and prod
+      // The UI needs to be context aware and know about _both_ staging and prod
       domain:
         window.location.hostname.indexOf('useoptic.com') >= 0
-          ? process.env.REACT_APP_PROD_API_BASE
-          : process.env.REACT_APP_STAGING_API_BASE,
+          ? 'https://spec.useoptic.com' // production
+          : 'https://api.o3c.info', // staging
```
notnmeyer, line 57: why hardcode these?

notnmeyer: ah, i read the full pr description. i see it called out

Taskfile.yml
notnmeyer: this is fine imo, but there are other syntaxes for env stuff, https://taskfile.dev/#/usage?id=environment-variables

workspaces/ui-v2/src/entry-points/cloud/cloud-viewer.tsx
niclim: This is ideally only going to be temporary. We have to determine these at run time (until we generate two artifacts).

The better solution here is to inject a single environment variable at build time and generate two artifacts (one for staging, one for production). I think it's also confusing to have `REACT_APP_API_BASE`, `REACT_APP_STAGING_API_BASE` and `REACT_APP_PROD_API_BASE` given that we share env variables across different builds (there's multiple entry points to inject environment variables, such as cloud viewer, building npm artifacts).

niclim: If we want to be able to inject / control these, I can readd `REACT_APP_STAGING_API_BASE` and `REACT_APP_PROD_API_BASE` - I just wanted to avoid more confusion with our env variables, with the hope this is temporary.

notnmeyer: got it!

devdoshi: @notnmeyer thoughts on staging+prod artifacts vs 1 artifact with logic to choose env? Either way both need to be tested, 2 artifacts means longer time and more room for error

notnmeyer: i thought the env was baked into application at build time isnt that the cruxt of the problem? if we can set the env on the container instead then i would absolutely prefer to do that.

devdoshi:
I think for now we should just fix the direct issue, putting the envs in our .env.* files and using runtime behavior to decide which env to use. Then @niclim @notnmeyer we can discuss some alternatives that get us the properties we want.

niclim:
> I think for now we should just fix the direct issue, putting the envs in our .env.* files and using runtime behavior to decide which env to use. Then @niclim @notnmeyer we can discuss some alternatives that get us the properties we want.

This does fix the direct issue - the only point of weirdness is this https://github.com/opticdev/optic/pull/1046#discussion_r675868144

We do also want to have the side_channel build point to staging (i.e. uploading and sharing) - which is why I've done the explicit split from PROD + STAGING variables. This allows us to more accurately QA what is going out in each release.

workspaces/ui-v2/src/entry-points/cloud/cloud-viewer.tsx
niclim: It's not possible to inject env variables in FE applications without some work (we could do something where we inject variables where we serve cloudviewer - but making it so we can use the same interface as `process.env` might be trickier). 

I strongly dislike having the UI having access to both environments, and rather these be injectable - if we let the UI have logic to handle both + select the env - there's a lot higher risk of cross contaimination between envs (debugging env errors would require looking in multiple places, wherever the UI implements this selection logic, and the env variable files (which now contain env variables for both).

Either way - this is a larger discussion. In respect to this PR - the hardcoded env variables here are intended to be temporary (there's also a bunch of other instances of hard coded env fallbacks in the UI - not that it justifies this) - any concerns with moving forward with the current implementation - we'll follow up with the staging / prod artifacts discussion separately @notnmeyer @devdoshi

notnmeyer (approved):
i dont care too much about the approach here, i want this to work however the people working on the app want it to work.

Taskfile.yml
JaapRood: Overriding the local flags file with this new `.env.local` has just cost me a whole good of head scratching. A developers local `.env` should remain untouched at all times, it's how developer A can run different local config than developer B (as they are working on different things).

JaapRood: Reverted as part of #1077.

PR #1047, opened 2021-07-23
Misc diff UI enhancements

niclim:
## Why

Couple of minor UI issues including:
- Fixing non-clickable text field for EditableTextField Variant.Small (i.e. can't click on the start of the field, always focuses on the end of the field)
- Darken background on focus for changelog items
- Change : to {} for path parameters
- Update text when editing a staged pending endpoint to `Save Changes`

## What

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/ui-v2/src/utils/pathMatcher.ts
+++ b/workspaces/ui-v2/src/utils/pathMatcher.ts
@@ -20,12 +20,14 @@ export function urlStringToPathComponents(
   return components;
 }
 
-export const makePattern = (components: PathComponentAuthoring[]) => {
+export const createPathFromPathComponents = (
+  components: PathComponentAuthoring[]
+) => {
   return (
     '/' +
     components
       .map((i) => {
-        return i.isParameter ? `:${i.name}` : i.originalName;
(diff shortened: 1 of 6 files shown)


workspaces/ui-v2/src/utils/pathMatcher.ts
```diff
@@ -20,12 +20,14 @@ export function urlStringToPathComponents(
...
-export const makePattern = (components: PathComponentAuthoring[]) => {
+export const createPathFromPathComponents = (
+  components: PathComponentAuthoring[]
+) => {
   return (
     '/' +
     components
       .map((i) => {
-        return i.isParameter ? `:${i.name}` : i.originalName;
+        return i.isParameter ? `{${i.name}}` : i.originalName;
```
niclim, line 30: Previously, this was piped into the `path-to-regexp` lib, however since we now construct our own regex from a pattern, we no longer need to preserve the `:`. This only is used for display purposes, and since we use {} everywhere, updating to do that

JaapRood (approved):

PR #1048, opened 2021-07-23
Fix spectacle returning removed endpoints

niclim:
## Why

This is a bug - spectacle is returning removed endpoints - this fixes it. We probably want to release a patch version for this (10.2.1)

## What

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/spectacle/src/graphql/schema.ts
+++ b/workspaces/spectacle/src/graphql/schema.ts
@@ -113,11 +113,9 @@ type Endpoint {
   responses: [HttpResponse!]!
 
   # Contributions which define descriptions
-  # TODO figure out what this currently maps to in the contributions list
   contributions: JSON!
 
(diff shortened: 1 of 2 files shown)


workspaces/spectacle/src/graphql/schema.ts
```diff
@@ -113,11 +113,9 @@ type Endpoint {
   responses: [HttpResponse!]!
 
   # Contributions which define descriptions
-  # TODO figure out what this currently maps to in the contributions list
   contributions: JSON!
 
   # Is the endpoint removed
-  # TODO figure out how this gets mapped
```
niclim, line 120: I totally meant to check this when i added this todo - apparently this slipped through

JaapRood (approved):
Little bits matter! ☺️

PR #1049, opened 2021-07-26
Fix bug where duplicate diffs are generated for unmatched optional primitive fields

JaapRood:
## Why
While working on the shape diffing of query parameters, we ran into a bug where two diffs would be generated for a single unmatched field shape instead of one.

Scenario:

- an optional field has been documented for a certain shape.
- a shape with this field set to a primitive is diffed. 

Observed result:
- two `ShapeDiffResult`s are generated, only different in their shape:
  - `ObjectFieldTrail > OptionalTrail`.
  - `ObjectFieldTrail > OptionalTrail > OptionalItemTrail`.

Expected result
- A single `ShapeDiffResult` with shape trail `ObjectFieldTrail > OptionalTrail > OptionalItemTrail`.

## What
A new test was added to isolate the problem, to find out later there was a `@BUG` mark with an existing test as well remarking on the same issue (2 results instead of 1). Debugging the shape traverser and visitor, it became apparent that the `BodyPrimitiveVisitor<ShapeDiffResult>` was presented with 2 choices for optional fields, instead of the single one we'd expect. This caused 2 unmatched results instead of 1, with the shape diff results mapping straight over from the choices.

Subsequently, the `ShapeQueries.list_trail_choices`, which is responsible for returning the possible shapes given a spec trail, was changed to only generate the variant with the `OptionalItemTrail`. That makes sense: optionals only make sense within the context of the field of an object, not on their own. An optional without knowing what is described as optional (even if that place is unknown), makes little sense from the perspective of the diff visitors.

However, the `ShapeChoiceMapping` projection, used in testing, _did_ have a use for the now eliminated trail: as a way to encode that the choice for a shape might be `undefined`. To fix this, this is now derived from the `OptionalItemTrail`, instead.

## Validation
* [x] CI passes
* [x] Snapshots updated and verified
* [x] `diff-use-cases-with-events` working as expected

### DIFF
--- a/workspaces/optic-engine/src/shapes/visitors/diff.rs
+++ b/workspaces/optic-engine/src/shapes/visitors/diff.rs
@@ -110,12 +110,15 @@ impl BodyPrimitiveVisitor<ShapeDiffResult> for DiffPrimitiveVisitor {
         _ => unreachable!("should not call primitive visitor without a primitive value"),
       });
     if matched.is_empty() {
-      unmatched.iter().for_each(|&choice| {
-        self.results.push(ShapeDiffResult::UnmatchedShape {
-          json_trail: json_trail.clone(),
-          shape_trail: choice.shape_trail(),
+      unmatched
+        .iter()
+        .filter(|choice| !matches!(choice.core_shape_kind, ShapeKind::OptionalKind)) // prevent generating two diffs for optional primitives
+        .for_each(|&choice| {
+          self.results.push(ShapeDiffResult::UnmatchedShape {
+            json_trail: json_trail.clone(),
+            shape_trail: choice.shape_trail(),
+          });
         });
-      });
     }
   }
 }
--- a/workspaces/optic-engine/tests/shape_diff.rs
+++ b/workspaces/optic-engine/tests/shape_diff.rs
@@ -436,8 +436,9 @@ fn can_diff_optional() {
     &shape_id,
   );
 
+  dbg!(&results);
   assert_debug_snapshot!("can_diff_optional__results", results);
-  assert_eq!(results.len(), 2); //@BUG: this should be 1, right? and the trail looks wrong
+  assert_eq!(results.len(), 1); //@BUG: this should be 1, right? and the trail looks wrong
 }
 
 #[test]
--- a/workspaces/optic-engine/tests/shape_diff.rs
+++ b/workspaces/optic-engine/tests/shape_diff.rs
@@ -436,7 +436,6 @@ fn can_diff_optional() {
     &shape_id,
   );
 
-  dbg!(&results);
   assert_debug_snapshot!("can_diff_optional__results", results);
   assert_eq!(results.len(), 1); //@BUG: this should be 1, right? and the trail looks wrong
 }
--- a/workspaces/optic-engine/tests/shape_diff.rs
+++ b/workspaces/optic-engine/tests/shape_diff.rs
@@ -437,7 +437,7 @@ fn can_diff_optional() {
   );
 
   assert_debug_snapshot!("can_diff_optional__results", results);
-  assert_eq!(results.len(), 1); //@BUG: this should be 1, right? and the trail looks wrong
+  assert_eq!(results.len(), 1);
 }
 
 #[test]
--- a/workspaces/optic-engine/tests/fixtures/shape-diff-use-cases/optional single string or list of strings.json
+++ b/workspaces/optic-engine/tests/fixtures/shape-diff-use-cases/optional single string or list of strings.json
@@ -1,0 +1,326 @@
+{
+  "events": [
+    {
+      "BatchCommitStarted": {
+        "batchId": "28616e96-c9e0-4d13-bff7-1627db17d144",
+        "commitMessage": "Initialize specification attributes",
+        "eventContext": {
+          "clientId": "anonymous",
+          "clientSessionId": "unknown-session",
+          "clientCommandBatchId": "28616e96-c9e0-4d13-bff7-1627db17d144",
+          "createdAt": "2021-07-19T11:30:01.583102-04:00"
+        },
+        "parentId": "root"
+      }
+    },
+    {
+      "ContributionAdded": {
+        "id": "metadata",
+        "key": "id",
+        "value": "8c2d80c1-1ea9-4a14-a5e8-4d6938c49f83",
+        "eventContext": {
+          "clientId": "anonymous",
+          "clientSessionId": "unknown-session",
+          "clientCommandBatchId": "28616e96-c9e0-4d13-bff7-1627db17d144",
+          "createdAt": "2021-07-19T11:30:01.583102-04:00"
+        }
+      }
+    },
+    {
+      "BatchCommitEnded": {
+        "batchId": "28616e96-c9e0-4d13-bff7-1627db17d144",
+        "eventContext": {
+          "clientId": "anonymous",
+          "clientSessionId": "unknown-session",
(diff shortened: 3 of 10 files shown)


workspaces/optic-engine/src/projections/shape.rs
```diff
@@ -175,81 +175,93 @@ impl From<&ShapeProjection> for ShapeChoiceMapping {
...
+              Some(shape_id.clone())
+            }
+            _ => None,
+          })
+          .map(|optional_shape_id| {
+            Primitive(PrimitiveChoice {
+              shape_id: optional_shape_id,
+              json_type: JsonType::Undefined,
+            })
+          });
```
JaapRood, line 191: Instead of mapping from choices 1:1, we generate the primitive `undefined` choice from looking at the parent trail, generating one when `OptionalItemTrail` is the last component of the shape trail. This passes the existing choice mappings snapshot tests.

workspaces/optic-engine/src/queries/shape.rs
```diff
@@ -96,23 +96,16 @@ impl<'a> ShapeQueries<'a> {
...
               });
-            let mut output = vec![ChoiceOutput {
-              parent_trail: shape_trail.clone(),
-              additional_components: vec![ShapeTrailPathComponent::OptionalTrail {
-                shape_id: shape_id.clone(),
-              }],
-              shape_id: shape_id.clone(),
-              core_shape_kind: core_shape_node.descriptor.kind.clone(),
-            }];
-            output.append(&mut self.list_trail_choices(&trail));
```
JaapRood, line 115: The choice of only the `OptionalTrail`, without any description of what the underling optional shape is, is not actually a useful one in the diffing, and was the main source of this bug. I've yet to find anything else beyond the shape choice mapping that relied on this, but I'll dig in deeper through testing to make sure.

workspaces/optic-engine/tests/shape_diff.rs
```diff
@@ -436,8 +436,9 @@ fn can_diff_optional() {
     &shape_id,
   );
 
+  dbg!(&results);
   assert_debug_snapshot!("can_diff_optional__results", results);
-  assert_eq!(results.len(), 2); //@BUG: this should be 1, right? and the trail looks wrong
+  assert_eq!(results.len(), 1);
```
JaapRood, line 441: Turns out this was already a bug we knew about, with this test isolating the same case as the new test.

JaapRood:
@devdoshi affecting the `list_trail_choices`, this is a non trivial change to shape diffing. Your eyes on this would be appreciated.

workspaces/optic-engine/src/queries/shape.rs
devdoshi: the contract of list_trail_choices is to list the choices at a particular trail. Thus the optional trail has meaning, communicating that it's allowed for a value to not be present here. When you compute coverage, for example, you would need to know that to track how often the value is omitted.

workspaces/optic-engine/src/shapes/traverser.rs
```diff
@@ -287,6 +287,10 @@ impl ShapeTrail {
     new_trail.path.push(component);
     new_trail
   }
+
+  pub fn last_component(&self) -> Option<&ShapeTrailPathComponent> {
+    self.path.last()
```
devdoshi, line 292: do you ever want the root to come out here?

workspaces/optic-engine/tests/shape_diff.rs
devdoshi: The diffing should not have emitted a diff result for the OptionalTrail in the 3rd array item (empty object), so to me this indicated diff logic as opposed to list_trail_choices

workspaces/optic-engine/src/queries/shape.rs
JaapRood: > Thus the optional trail has meaning, communicating that it's allowed for a value to not be present her

I guess what I'm trying to assert here is that "optional" is an incomplete choice. You'll want to know _what_ is the thing that is optional for the choice to be complete. There is a 1:1 mapping here, we don't have the case of an optional, without knowing what the subject of the optionality is. Therefor, I think it's right to only return the "full" choice: optional, plus it's subject. That should still be enough to compute coverage from, right? Count `OptionalTrailItem` instead of `OptionalTrail` as the two should always come together?

workspaces/optic-engine/tests/shape_diff.rs
JaapRood: It doesn't do that? Both diffs described a diff with the second array item being a number instead of a string, just with two different shape trails.

workspaces/optic-engine/src/shapes/traverser.rs
JaapRood: Yeah, why not? 🤷‍♂️ It seems to me this would return whatever the last component of the shape trail is, and if that's the root, sure? I don' think there's enough context from this method to limit that, that's up to the caller to decide.

workspaces/optic-engine/src/queries/shape.rs
devdoshi: Optional means there are 2 possibilities: providing nothing (undefined) or providing one of the choices afforded by the optional item. You don't need to know what the optional item is to know you can provide nothing. Similarly, if we instead had no Optional but had Undefined, we would use OneOf(Undefined, ...) and the Undefined would be one of the choices. I'm not understanding what you're trying to say about it being incomplete and something that list_trail_choices shouldn't emit

devdoshi: Coverage counting when it is undefined should increment that Optional shapeid's trail and coverage counting any of the items should count their trails. Each possibly outcome should have its own trail, right?

JaapRood: We got aligned on this in a sync call, with as a conclusion that `list_trail_choices` _should_ contain both choices as they strictly _are_ choices. Instead, we'll deal with some choices not being meaningful in certain contexts downstream.

workspaces/optic-engine/src/shapes/visitors/diff.rs
```diff
@@ -110,12 +110,15 @@ impl BodyPrimitiveVisitor<ShapeDiffResult> for DiffPrimitiveVisitor {
...
+      unmatched
+        .iter()
+        .filter(|choice| !matches!(choice.core_shape_kind, ShapeKind::OptionalKind)) // prevent generating two diffs for optional primitives
+        .for_each(|&choice| {
+          self.results.push(ShapeDiffResult::UnmatchedShape {
+            json_trail: json_trail.clone(),
+            shape_trail: choice.shape_trail(),
+          });
         });
-      });
```
JaapRood, line 118: @devdoshi granted, this is a whole bunch cleaner that the changes in `list_trail_choices`.

workspaces/optic-engine/tests/shape_diff.rs
```diff
@@ -437,7 +437,7 @@ fn can_diff_optional() {
   );
 
   assert_debug_snapshot!("can_diff_optional__results", results);
-  assert_eq!(results.len(), 2); //@BUG: this should be 1, right? and the trail looks wrong
+  assert_eq!(results.len(), 1); //@BUG: this should be 1, right? and the trail looks wrong
```
niclim, line 440: Remove the comment?

niclim (approved):
Can't say I understand the full flow - but the snapshots look correct from what you've explained from the bug.

Perhaps we should merge this in earlier and try do some tests while this is landed (and call it out during the release QA) - I'll also keep an eye out on this

PR #1053, opened 2021-07-26
Filter out removed request and response bodies based on batchCommit comparison

niclim:
## Why
Filter out removed request and response bodies - applying the same logic on endpoints to show or hide bodies if they were removed in a certain range of batch commits.

## What

This PR:
- Exposes the isRemoved flags on request and response bodies
- Exposes changes on the request and response bodies
- Filters out removed query, request and response bodies depending on visibility requirements (i.e. documentation should filter all removed, changelog should only show removed _between_ the specific batch commit ids
- Only returns the latest query parameters node

TODO to follow this up
- [ ] Only show the latest request and response nodes (based on status code and content type)
- [ ] Add in UI indication to highlight request / response body added / updated / removed in changelog view

The biggest thing that is still "buggy" is that if you remove a request and then readd a request with the same content type - the changelog will show them as added and removed - instead of diffing the two shapes. 

The other thing is removing an endpoint and readding the same endpoint when the path of the endpoint is "known" - i.e. there are path events that haven't been removed due to another endpoint (e.g. we remove `/api/todos` but still have `/api/todos/list/:listId`) - the changelog will treat these as the same id.

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


workspaces/spectacle/tap-snapshots/test-index.ts-TAP.test.js
```diff
@@ -100,17 +100,7 @@ exports[
 Object {
   "data": Null Object {
     "endpointChanges": Null Object {
-      "endpoints": Array [
```
niclim, line 103: This is a change so that specifying `sinceBatchCommitId: undefined` will return an empty list of changes, rather than every change that ever happened

niclim: Ok, apparently our changelog feature was relying on sending `sinceBatchCommitId: undefined` to return us every change that happened in the spec. I don't think this makes sense (because the results are, everything changed) - but to reduce the amount of work, I'm just going to leave it as is.

workspaces/graph-lib/src/endpoints-graph/index.ts
```diff
@@ -111,6 +111,36 @@ export type Edge =
 const isNodeRemoved = (node: Node): boolean =>
   node.type !== NodeType.BatchCommit && node.data.isRemoved;
 
+////////////////////////////////////////////////////////////////////////////////
+const getLatestCreatedNodeId = (
```
niclim, line 115: We can use this to also handle requests and responses. I'm planning on handling this tomorrow

workspaces/spectacle/src/helpers.ts
```diff
@@ -85,84 +85,70 @@ export function buildEndpointChanges(
...
-        batchCommit.result.data.createdAt >
-        sinceBatchCommit!.result.data.createdAt
-    );
-  } else {
-    deltaBatchCommits = sortedBatchCommits;
-  }
+  const deltaBatchCommits = getDeltaBatchCommitsForEndpoints(
+    endpointQueries,
+    // In this case specifically, we want _all_ endpoint changes
+    sinceBatchCommitId || ALL_BATCH_COMMITS
```
niclim, line 91: This function is confusing - we want to query _all_ batch commits sometimes (in the changelog view). I want to reconcile this, but I don't know if I'll have time to get around to this. I'm adding this constant so that it's _explicit_ that we want to query all changes in all batches (which results in everything being marked as added)

workspaces/ui-v2/src/store/selectors/__tests__/visibilitySelectors.test.ts
```diff
@@ -0,0 +1,124 @@
+import { ChangeType } from '<src>/types';
```
niclim, line 1: Our first of hopefully many new tests in the UI

workspaces/ui-v2/src/store/selectors/visibilitySelectors.ts
```diff
@@ -0,0 +1,81 @@
+import { ChangeType } from '<src>/types';
+
+export const isItemVisible = <T extends { isRemoved: boolean }>(
```
niclim, line 3: These are generic functions we can hopefully use for shape visibility (and any other removable items)

niclim:
Ok @JaapRood I think this covers _most_ of the buggy behavior with the changelog. I'm going to spend tomorrow trying to fix up the rest of this (i.e. making requests and responses unique via status code and content type) and optimistically fixing the buildEndpointChanges function before rolling back to the field edit bet.

I think if this is to land, I'd be fairly happy with the state of this.

workspaces/graph-lib/src/endpoints-graph/index.ts
niclim: I also wonder if this should live in graph-lib, or if it should be pulled back into the spectacle/projection/endpoints

The reason I hesitate to pull it back into the spectacle projection is because I don't know if only returning the most recent node generally makes sense, and if we need the old nodes in certain cases.

workspaces/ui-v2/src/pages/docs/EndpointRootPage.tsx
```diff
@@ -120,6 +120,14 @@ export const EndpointRootPage: FC<
...
     (path) => path.isParameterized
   );
+  const filteredQuery =
+    thisEndpoint.query && selectors.isItemVisible(thisEndpoint.query)
+      ? thisEndpoint.query
+      : null;
+  const filteredRequests = selectors.filterRemovedItems(thisEndpoint.requests);
+  const filteredResponsesByStatusCode = selectors.filterMapOfRemovedItems(
+    thisEndpoint.responsesByStatusCode
+  );
```
JaapRood, line 130: Little nit you should totally ignore: "filteredX" is one of those variables that makes total sense when you write it, but when you or someone revisits later is met with "filtered.. how?". With an app being 1 massive data pipeline, isn't everything we do filtered in some way? ^^

"existingX" might make sense here, as we clean up removed items. Or maybe something else descriptive?

workspaces/ui-v2/src/store/selectors/visibilitySelectors.ts
JaapRood: Those kinds of selectors will definitely come in handy for the shape editing stuff as well. Good that it's isolated, makes it easier to cover with tests later as well.

JaapRood (approved):

workspaces/ui-v2/src/pages/docs/EndpointRootPage.tsx
niclim: I changed this to `visibleX` - agreed that filteredX could be confusing and less descriptive

PR #1055, opened 2021-07-27
Implement generation of editing shape commands through required shape kinds

JaapRood:
## Why
We're working on allowing fields of objects to be made nullable / non-nullable and optional / required. The Optic Engine must support generating commands for fields, so the UI can preview and apply them. 

## What
This is a first pass at this, so we can iterate and enhance as the needs reveal themselves as we hook this up to an AI. It adds an `edit_shape_commands` method to the internal `ShapesQueries`, which accepts the Shape trail of the field to be edited and the shape kinds that are required for the field. Returned are the commands required to affect these changes once applied to the spec.

For now, the "wrapper" shapes for optional and nullables are always regenerated; re-use of existing shapes is possible, but adds extra complexity we don't want to deal with just yet. We can come back to that later.

As a bonus, this also extracted Spectacle specific shape projections and queries from the internal `ShapesProjection`. That got rid of a lot of confusion about what is meant to be use where, especially as there are some very similar types (the internal `ChoiceOutput` vs the Spectacle specific `ShapeChoice`). 

Follow ups for later: 

- Add a layer on top to approach this editing through Spectacle. Probably with an edit query for the spectacle side so `JsonLike` can be used instead of `ShapeKind` for specifying which properties are required for the field.
- Re-use of existing shapes, rather than generating new optional and nullable shapes from scratch every time.
- Support for editing array items.

## Validation
* [ ] CI passes
* [x] Added tests covering various field changing scenarios.


workspaces/optic-engine/src/queries/shape.rs
```diff
@@ -438,6 +441,96 @@ impl<'a> ShapeQueries<'a> {
     let command = ShapeCommand::remove_field(field_node.field_id.clone());
     Some(std::iter::once(command))
   }
+
+  pub fn edit_shape_commands(
+    &'a self,
+    shape_trail: &ShapeTrail,
+    required_kinds: impl IntoIterator<Item = &'a ShapeKind>,
+    id_generator: &mut impl SpecIdGenerator,
+  ) -> Option<impl Iterator<Item = ShapeCommand>> {
```
JaapRood, line 450: The magic happens here. For now, we're straight up accepting `ShapeKind` instances to detail what a shape should be. For nullable and optionals, that's enough. For more complex types, it might not be, but since those are out of the scope of the first version of field level editing, lets cross that bridge when we get to it.

workspaces/optic-engine/src/queries/shape.rs
```diff
@@ -438,6 +441,96 @@ impl<'a> ShapeQueries<'a> {
...
+    }?; // nothing to update if there isn't a subject
+
+    let required_kinds: BTreeSet<_> = required_kinds
+      .into_iter()
+      .filter(|kind| togglable_kinds.contains(kind))
+      .cloned()
+      .collect();
+
+    let mut new_shape_prototypes = vec![];
+    let mut root_shape_id = subject_shape_id.clone();
```
JaapRood, line 490: When we implemented the trail learners and command generation from that, we figured out how useful it can be (especially as we add more complex editing later on) to generate prototypes of shapes first, rather than commands straight away. Most significantly, it allows new id generation to happen and references to those to be made ahead of time, so we can make sure all hierarchy is setup correctly, making the commands easier to write sequentially.

I actually hope to find a generic type of Shape prototype that can be used by all the shape command generators, so we only have to implement the generation of commands once. That's something to look at later, though.

workspaces/optic-engine/src/queries/shape.rs
```diff
@@ -438,6 +441,96 @@ impl<'a> ShapeQueries<'a> {
...
+      .iter()
+      .map(|choice| &choice.core_shape_kind)
+      .cloned()
+      .collect();
+
+    let togglable_kinds = if shape_trail.is_field() {
+      BTreeSet::from_iter(vec![ShapeKind::OptionalKind, ShapeKind::NullableKind])
+    } else {
+      BTreeSet::new() // only allow edits to fields for now
+    };
```
JaapRood, line 463: I use `BTreeSet` instead of `HashSet` here so we can rely on the id generation and commands order being stable. That really hurt us in testing the learners, and since this is action to happen from direct user interaction, we should have the performance budget for it.

workspaces/optic-engine/src/queries/shape.rs
```diff
@@ -488,7 +633,278 @@ mod test {
...
+    let choice_mapping = updated_spec.shape().to_choice_mapping();
+
+    assert_debug_snapshot!(
+      "can_generate_edit_shape_commands_to_make_optional_field_nullable_and_optional__choice_mapping",
+      choice_mapping
+    );
+  }
+
+  #[test]
+  pub fn can_generate_edit_shape_commands_to_make_polymorphic_field_optional() {
```
niclim, line 819: `nullable`

workspaces/optic-engine/src/queries/shape.rs
```diff
@@ -488,7 +633,278 @@ mod test {
...
+      ShapeTrailPathComponent::ObjectFieldTrail {
+        field_id: String::from("field_1"),
+        field_shape_id: String::from("one_of_shape_1"),
+        parent_object_shape_id: String::from("object_shape_1"),
+      },
+    );
+    let required_kinds = vec![ShapeKind::NullableKind];
+    let edit_shape_commands = shape_queries
+      .edit_shape_commands(&shape_trail, &required_kinds, &mut id_generator)
+      .expect("field should be able to be made nullable and optional")
```
niclim, line 847: `field should be able to be made nullable`

workspaces/optic-engine/src/queries/shape.rs
```diff
@@ -488,7 +633,278 @@ mod test {
...
+      ShapeTrailPathComponent::ObjectFieldTrail {
+        field_id: String::from("field_1"),
+        field_shape_id: String::from("optional_shape_1"),
+        parent_object_shape_id: String::from("object_shape_1"),
+      },
+    );
+    let required_kinds = vec![];
+    let edit_shape_commands = shape_queries
+      .edit_shape_commands(&shape_trail, &required_kinds, &mut id_generator)
+      .expect("field should be able to be made nullable and optional")
```
niclim, line 892: `field should be able to be made required`

workspaces/optic-engine/src/queries/shape.rs
```diff
@@ -488,7 +633,278 @@ mod test {
...
+    let mut id_generator = SequentialIdGenerator { next_id: 1093 }; // <3 primes
+
+    let shape_trail = ShapeTrail::new(String::from("object_shape_1")).with_component(
+      ShapeTrailPathComponent::ObjectFieldTrail {
+        field_id: String::from("field_1"),
+        field_shape_id: String::from("optional_shape_1"),
+        parent_object_shape_id: String::from("object_shape_1"),
+      },
+    );
+    let required_kinds = vec![ShapeKind::StringKind, ShapeKind::NullableKind];
```
niclim, line 754: Should this be string kind? or optional kind?

workspaces/optic-engine/src/shapes/traverser.rs
```diff
@@ -287,6 +289,38 @@ impl ShapeTrail {
     new_trail.path.push(component);
     new_trail
   }
+
+  pub fn with_optional(&self, optional_shape_id: ShapeId, subject_shape_id: ShapeId) -> Self {
```
niclim, line 293: Is this used / should we use this (along side with_nullable) / or should we remove this?

niclim (approved):
Think this makes sense - I think I need to spend some time digging through how optionals / nullables are represented in the rust engine code but the overall logic looks good.

workspaces/optic-engine/src/projections/spectacle/shape.rs
```diff
@@ -0,0 +1,80 @@
...
+use crate::queries::spectacle::spec_choices::{ShapeChoice, ShapeChoiceQueries};
+use crate::shapes::ShapeTrail;
+use crate::state::shape::ShapeId;
+use serde::Serialize;
+use std::collections::BTreeMap;
+use std::iter::FromIterator;
+
+#[derive(Debug, Serialize, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct GraphNodesAndEdges<N, E> {
```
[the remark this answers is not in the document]
JaapRood: Yeah, internally the Spectacle projection uses the internal projection. It then adds on top the `ShapeChoiceQueries`, which introduces the `ShapeChoice` type that Spectacle ends up querying. How mutations happen is still up to us. Delivering back mutated shape choices to the Engine seems like the way to go, until you realise that those require lots of extra knowledge about how to make shape ids fit together. We'll have to try and see how that best fits together.

workspaces/optic-engine/src/queries/shape.rs
JaapRood: That might well very be how that's going to shape out, I'm not 100% sure yet. Regardless, this interface is something we can further iterate on, or leave as is. It does the most core thing without any frills at the moment, so should be a good base either way.

workspaces/optic-engine/src/queries/shape.rs
```diff
@@ -438,6 +441,96 @@ impl<'a> ShapeQueries<'a> {
     let command = ShapeCommand::remove_field(field_node.field_id.clone());
     Some(std::iter::once(command))
   }
+
+  pub fn edit_shape_commands(
+    &'a self,
+    shape_trail: &ShapeTrail,
+    required_kinds: impl IntoIterator<Item = &'a ShapeKind>,
```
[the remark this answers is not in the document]
JaapRood: Naming this one wasn't easy, but hadn't thought about the clash of "required" in the context of optional fields yet. 

- `new_kinds` sounds like you might be adding kinds to an set, but you're not: you're supplying the exact kinds it should be, what's not in there shouldn't happen.
- `valid_kinds` sounds like validation happened and we can blindly apply those kinds, which it hasn't yet. They still need to be filtered for kinds that _can_ actually be toggled given it's a field, array item or some other shape.

Maybe `requested_kinds` is better 😅

workspaces/optic-engine/src/queries/shape.rs
JaapRood: Left in StringKind, to verify the filtering of togglable shape kinds works correctly and we get no panics on additional shape kinds.

workspaces/optic-engine/src/shapes/traverser.rs
JaapRood: Nice catch! Removing, part of a previous iteration of determining which shape to edit by mutating the trail. Didn't work out well 😅

PR #1056, opened 2021-07-27
Orphan spectacle endpoint nodes if not unique

niclim:
## Why
Following up from this comment https://github.com/opticdev/optic/pull/1053#discussion_r677033469 - after thinking about it, I think it is better suited to live in the spectacle projection.

The changes are best reviewed in `bug_scenarios__duplicated_nodes__spectacle_endpoints_graph.snap` - this shows the request and response body nodes being orphaned if there is a newer matching body. The uniqueness conditions are as followed:
In a single endpoint
- 1 query parameter shape
- request body with content type
- response body with content type and status code.

## What

On building the spectacle endpoint projection when inserting a query param, request body or response body, we check if a matching body is present, if so we orphan that body.

What this should fix is the case where the changelog shows multiple of the same content type bodies. We select the latest body. There'll need to be more work / thinking to handle diffing between two shapes (which is what a user would likely want to see in the case there are multiple content type bodies).

To recreate this, you can pull this code and go to `todos-partial`
- Delete endpoint `/api/todos` (it needs to be this one, because it needs to force this to reuse the same endpoint - which is a separate issue)
- relearn `/api/todos`
- go to the changelog view and compare `/api/todos` - go to a commit far enough back that it includes _both_ request/response body creations.

This should also handle when we implement removing query / request / response bodies.

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


workspaces/graph-lib/src/endpoints-graph/index.ts
```diff
@@ -111,36 +111,6 @@ export type Edge =
 const isNodeRemoved = (node: Node): boolean =>
   node.type !== NodeType.BatchCommit && node.data.isRemoved;
 
-////////////////////////////////////////////////////////////////////////////////
-const getLatestCreatedNodeId = (
```
niclim, line 115: This functionality is moved to the spectacle endpoint projection

workspaces/optic-engine/src/projections/spectacle/endpoints.rs
```diff
@@ -529,11 +527,65 @@ impl EndpointsProjection {
...
+      http_content_type: http_content_type.clone(),
       root_shape_id: root_shape_id.clone(),
       is_removed: false,
     });
+
+    // If there is already a set request body with a matching content type, we should orphan the old node
+    // @GOTCHA - this only allows a single content type request parameter to be attached to an endpoint node
+    // This means history is only visible to users for the latest request node with content type
+    // @GOTCHA - event generation does not group together same path + method, meaning we need to search
+    // all request ids attached to an endpoint node (i.e. request -> body is 1:1 since request nodes are never reused)
```
niclim, line 547: So if we want to fix this, I think we will need to change the event generation such that adding a new request looks for existing requests. Right now requests and responses are newly created without context of the current spec, so a request node is always created. I'm not sure where this is created (likely both the UI and rust code).

We'll also probably need to figure out some migration path / handling old event shapes since this code expects a single body per request / response node (how events are currently generated)

workspaces/optic-engine/tests/snapshots/domain_conflict_scenarios_generated__scenario_af20bd1d8e76cf9db0416175232b9b0276447f7148f3261f2d14b006aa7a0229__spectacle_endpoints_json.snap
```diff
@@ -133,10 +133,10 @@ expression: "&spec_projection.spectacle_endpoints_serializable()"
       }
     ],
     [
-      6,
-      4,
+      9,
```
niclim, line 136: This is correct given how the response nodes are set in this fixture (there are two response nodes with `200` status codes).

workspaces/spectacle/src/graphql/schema.ts
```diff
@@ -167,7 +167,7 @@ type QueryParameters {
   id: String!
 
   # Root shape ID for the QueryParameter. Look at the shapeChoices query getting more information about the root shape
-  rootShapeId: String!
+  rootShapeId: String
```
niclim, line 170: Query parameter root shapes _can_ be null (they shouldn't but our query parameters rely on two events)

workspaces/optic-engine/src/projections/spectacle/endpoints.rs
JaapRood: > Right now requests and responses are newly created without context of the current spec, so a request node is always created.

Happens in the undocumented bodies learner in the Engine, giving the UI the commands it can apply. Not sure if the UI mutates those in any way (doubt it / hope not).

JaapRood (approved):
Seems like this is a move for the better. Implicitly removing an old node because a new one is added seems logic similar to "set shape", where only a single "unique" can be set at once, so setting a new one unsets the old. In the ideal case we could be more explicit about things, but if this helps us now, it's totally worth doing.

workspaces/optic-engine/tests/bug-scenarios.rs
```diff
@@ -1958,3 +1958,30 @@ fn scenario_34() {
...
+#[test]
+fn duplicated_nodes_with_deletion() {
+  let events = SpecEvent::from_file(
+    std::env::current_dir()
+      .unwrap()
+      .join("tests/fixtures/events-deleted-duplicated-nodes-spec.json")
+      .to_str()
+      .unwrap(),
+  )
+  .expect("ergast spec should deserialize");
```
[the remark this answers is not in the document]
niclim: I think I'll leave it as is for now - but we'll definitely want to consolidate this when we've got a pattern we want to follow.

PR #1057, opened 2021-07-27
Add stronger typing to graph-lib

niclim:
## Why

We were doing a lot of casting `as ...NodeWrapper` and type assertions. These changes make it so we don't need to do these changes.

## What

- Make NodeListWrapper generic over Node
- Change Node and NodeWrapper to be union types for better type narrowing

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


workspaces/graph-lib/src/endpoints-graph/index.ts
```diff
@@ -631,7 +669,8 @@ export class GraphQueries {
   }
 
   //@TODO move away from null here
-  wrapList(type: NodeType | null, nodes: Node[]): NodeListWrapper {
+  // TODO figure out how to make this generic
+  wrapList(type: NodeType | null, nodes: Node[]): NodeListWrapper<NodeWrapper> {
```
niclim, line 673: I couldn't figure out how or why wrapList and wrap weren't playing nice with generics. There's probably a way to do this, but it probably requires some type juggling. The impact of this is we just need to cast `as NodeListWrapper<NodeTypeToNodeWrapper<T>>` when we call wrapList (which is only used internally in this class) - which is much better than having the consumers have to cast things

workspaces/spectacle/src/spectacle.ts
```diff
@@ -323,15 +323,7 @@ export async function makeSpectacle(opticContext: IOpticContext) {
...
-        let path = parent.endpointNode.path();
-        let parentPath = path.parentPath();
-        const components = [path.value];
-        while (parentPath !== null) {
-          components.push(parentPath.value);
-          path = parentPath;
-          parentPath = path.parentPath();
-        }
-        return Promise.resolve(components.reverse());
+        return Promise.resolve(parent.endpointNode.path().components());
```
niclim, line 326: Major difference between `path().components()` is this creates `path.value`. I've updated the `PathComponent` resolvers to use the Wrapper class

JaapRood (approved):

PR #1059, opened 2021-07-28
fix intercom user id

acunniffe:
## Why
Realized the intercom id was being set to `clientId` not `clientAgent`


LouManglass (approved):
Looks right.

PR #1060, opened 2021-07-28
Shape accumulation to redux

niclim:
## Why

As part of the field level editing work we'll probably want to refer to shapes and modify them. Moving this data into redux and setting it up in a queryable fashion will make this work a lot easier for us (now and in the future)

## What

- Pulled shape accumulation into redux - this is modeled as:
```javascript
rootShape: Record<ShapeId, AsyncStatus<ShapeId>>;
shapeMap: Record<ShapeId, ReduxShape>
```
- There's three different kinds of shapes (spectacle, redux, ishaperender) - before this was one shape so we had a lot of nullable / optional keys
- added selectors to be able to convert this store of data into a shape the UI expects

## Validation
This worked the first time i tried to run the UI after my changes, so I'm still looking for what i missed!

But I'll merge this after the thursday release and hopefully stress test this a little (typing helped out a lot with this refactor)


workspaces/ui-v2/src/components/ShapeRenderer/ShapePrimitive.tsx
```diff
@@ -1,34 +1,27 @@
 import * as React from 'react';
 import makeStyles from '@material-ui/styles/makeStyles';
 import { useSharedStyles } from './SharedStyles';
-import { IShapeRenderer, JsonLike } from './ShapeRenderInterfaces';
-import { useShapeRenderContext } from './ShapeRenderContext';
+import { IShapeRenderer, JsonLike } from '<src>/types';
 import classNames from 'classnames';
 
-export const ShapePrimitiveRender = ({ jsonType, value }: IShapeRenderer) => {
+export const ShapePrimitiveRender = ({ jsonType }: IShapeRenderer) => {
```
niclim, line 7: Value was always undefined - this is something we'll need to add back for show examples, but given that value wasn't typed consistently - it'll be easier to reimagine + update typing where appropriate when we come to implement that feature

workspaces/spectacle/src/spectacle.ts
```diff
@@ -197,7 +197,9 @@ export async function makeSpectacle(opticContext: IOpticContext) {
         args: any,
         context: GraphQLContext
       ) => {
-        return context.spectacleContext().shapeViewerProjection[args.shapeId];
+        return (
+          context.spectacleContext().shapeViewerProjection[args.shapeId] || []
```
niclim, line 201: This was _sometimes_ undefined if we query with a shapeId not defined in the projection.

workspaces/ui-v2/src/components/FieldOrParameter.tsx
```diff
@@ -49,50 +48,6 @@ function summarizeTypes(shapes: IShapeRenderer[]) {
...
 }
 
-export interface IFieldDetails {
-  name: string;
-  contribution: IContribution;
-  shapes: IShapeRenderer[];
-  depth: number;
-}
-
-export function createFlatList(
```
niclim, line 59: Moved to shapeSelectors

workspaces/ui-v2/src/store/selectors/shapeSelectors.ts
```diff
@@ -0,0 +1,126 @@
...
+  const resolveShape = (shapeId: string): IShapeRenderer[] => {
+    const reduxShapes = state.shapes.shapeMap[shapeId];
+    if (!reduxShapes || reduxShapes.length === 0) {
+      console.error(`Could not find shape ${shapeId} in redux store`);
+      Sentry.captureEvent({
+        message: 'Could not find shape in redux store',
+        extra: {
+          shapes: state.shapes,
+        },
+      });
```
niclim, line 39: This might not be worth sentrying, since it's likely there's an invalid spec, but for now it's probably better to be over sensitive and mute / remove this if this becomes too noisy

workspaces/ui-v2/src/store/shapes/thunks.ts
```diff
@@ -0,0 +1,195 @@
...
+
+  if (results.errors || !results.data) {
+    console.error(results.errors);
+    throw new Error(JSON.stringify(results.errors));
+  }
+
+  return results.data;
+};
+
+const fetchShapesAndChildren = async (
```
niclim, line 96: The biggest change here is that we are saving the data in a slightly different format - where we store each shape (but not recursively) - it's up to the selector to join the relevant information.

This is done because most likely want to access shape information by shapeId when we make changes, etc - storing the data in this format makes it easier to query for shapes, rather than traversing all the rootShapes.

I'm not sure if we need an index for fieldId or a fieldId -> shapeId index, but we can definitely add that in later if we find its necessary.

workspaces/ui-v2/src/store/shapes/thunks.ts
```diff
@@ -0,0 +1,195 @@
...
+  let stack: ShapeId[] = [rootShapeId];
+
+  while (stack.length > 0) {
+    const results = await Promise.all(
+      stack.map((shapeId) =>
+        fetchSpectacleShape(spectacle, shapeId, sinceBatchCommitId)
+      )
+    );
+    // Reset the stack and add in more shapes to fetch
+    stack = [];
```
niclim, line 114: The more standard DFS / BFS would be do visit nodes one at a time, but since this is async, we want to be greedy and query _all_ shapes that we _know_ we have to query.

workspaces/ui-v2/src/types/shapes.ts
```diff
@@ -0,0 +1,62 @@
...
+  name: string;
+  contribution: IContribution;
+  shapes: IShapeRenderer[];
+  depth: number;
+}
+
+// Used to render query parameters
+export type QueryParameters = Record<string, IFieldRenderer>;
+
+export type IShapeRenderer =
```
niclim, line 27: What's important to note here is there are three types of shapes (previously we had one, which meant if we added properties after we didn't have great type safety since we had to make things `optional`).
- SpectacleShape (shape queried from spectacle)
- ReduxShape (shape that's stored in redux)
- IShapeRenderer (shape that's used to render in components)

workspaces/ui-v2/src/store/shapes/slice.ts
```diff
@@ -0,0 +1,47 @@
...
+  rootShapes: {},
+  shapeMap: {},
+};
+
+const shapesSlice = createSlice({
+  name: 'shapes',
+  initialState,
+  reducers: {},
+  extraReducers: (builder) => {
+    // TODO we probably need to refetch shapes if they get outdated
```
niclim, line 20: When we come to actually make changes to shapes, we'll need to figure out if we need to refetch shapes (which means we should reset all shapes or a subset if we can be sure only some shapes changed).

workspaces/ui-v2/jest.config.js
```diff
@@ -3,10 +3,7 @@ module.exports = {
   collectCoverageFrom: ['src/**/*.{js,jsx,ts,tsx}', '!src/**/*.d.ts'],
   setupFiles: ['react-app-polyfill/jsdom'],
   setupFilesAfterEnv: ['<rootDir>/src/setupTests.js'],
-  testMatch: [
-    '<rootDir>/src/**/__tests__/**/*.{js,jsx,ts,tsx}',
-    '<rootDir>/src/**/*.{spec,test}.{js,jsx,ts,tsx}',
-  ],
+  testMatch: ['<rootDir>/src/**/*.{spec,test}.{js,jsx,ts,tsx}'],
```
niclim, line 6: Removed the `__tests__` matcher because i wanted to have utils for tests that don't get picked up in the `__tests__` folder

workspaces/ui-v2/src/lib/__tests/diff-helpers/shape-interpretations.test.ts
```diff
@@ -36,7 +31,7 @@ test('a known field is missing.managed', async () => {
   const diff = universe.diffSet.groupedByEndpointAndShapeTrail()[0];
   const result = await shapeDiffPreview(diff, universe.universe);
   commonAssertions(result);
-  await logResult(result.preview);
+  // logResult(result.preview);
```
niclim, line 34: I'm not sure why these were logging, but for future debugging I'm keeping the fn calls here commented out

workspaces/ui-v2/src/store/selectors/shapeSelectors.ts
JaapRood: More so from the perspective that you'd expect selectors to be pure, but 🤷‍♂️

JaapRood (approved):
Looking good from what I can tell from a static review, should make a lot of the future work a whole lot easier to do well!

PR #1062, opened 2021-07-29
chore: shorten CI announce message

notnmeyer:
Missed this in my previous PR. this shortens the Slack announce message to match what we're doing in monorail

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -84,7 +84,7 @@ tasks:
       - yarn wsrun --stages --report --fast-exit --exclude-missing ws:build
 
     sources:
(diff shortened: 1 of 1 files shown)


Taskfile.yml
```diff
@@ -84,7 +84,7 @@ tasks:
       - yarn wsrun --stages --report --fast-exit --exclude-missing ws:build
 
     sources:
-      - !workspaces/*/build/**/*
+      - "!workspaces/*/build/**/*"
```
notnmeyer, line 87: my editor complains when we dont quote negations

PR #1066, opened 2021-07-29
Fix cloud viewer prod url

niclim:
## Why

Cloud viewer wasn't working because prod was pointing to spec, instead of api

## What

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


niclim:
This should be fine to merge to releases, the cli version is correct - we just need a new docker image with this updates (and the github workflow should not publish a new version since `npm version` === `code package.json version`

PR #1069, opened 2021-07-30
Adapt edit of shape commands generation to Spectacle interface

JaapRood:
## Why
We need to bridge the gap between the Engine's lower level ability to edit the spec at a specific location (shape trail), with the more client-friendly interface Spectacle will expose. 

## What
This PR adds an `edit_field_shape` query, accepting a `FieldId` and a set of `JsonTypes` the field should represent (nullable and optional so far). It's designed to be exposed pretty much directly through a Spectacle query. 

It is intentionally more narrowly scoped than editing _any_ shape. Not only don't we support anything else at the moment, editing a field describes a _location_ to edit, not a shape directly. Optic (as its implemented right now) doesn't actually allow you to edit a `Shape` directly, only creating and composing them. So rather than allow arbitrary shape ids to be "edited", we require a place.

The internal shape graph has changed to accommodate that as well. `edit_shape_commands` was renamed to `edit_shape_trail_commands` to better describe what it is what we're actually editing. A new `resolve_shape_trail` accepts any Optic Id (FieldId, ShapeId, etc) and uses the shape graph to figure out the location of what we're trying to edit. For now, the only thing it's capable of resolving are field ids and objects (enough for our current purposes, as well as future purpose of converting an Object to a Map), with further changes being required to the shape graph to make resolving of places like list items possible as well.

Follow ups to be done separately:

- Expose query through `optic-engine-wasm`'s interface
- Expose query through Spectacle's graph
- Hook up UI to use Spectacle query

## Validation
* [x] CI passes
* [x] Can generate commands to make (nested) fields optional or nullable


workspaces/optic-engine/src/queries/spectacle/spec_choices.rs
```diff
@@ -165,4 +166,110 @@ impl<'a> ShapeChoiceQueries<'a> {
...
+
+  pub fn edit_field_commands(
+    &'a self,
+    field_id: &FieldId,
+    requested_json_types: impl IntoIterator<Item = &'a JsonType> + 'a,
+    id_generator: &'a mut impl SpecIdGenerator,
+  ) -> Option<impl Iterator<Item = ShapeCommand> + 'a> {
+    let field_shape_trail = self.shape_queries.resolve_shape_trail(field_id)?;
+
+    dbg!(&field_shape_trail);
```
niclim, line 178: dbg statement

niclim (approved):

JaapRood:
@niclim end of day for me so will hold off on the merge, waiting for CI to pass, but feel free to if you're keen!

PR #1070, opened 2021-07-30
Handle field removal with changelog

niclim:
## Why

I initially tried splitting up the shape projection into an internal projection and a spectacle projection, however I think this made the code more complex, and it's really only to power the changelog for removal (which we might want to think more carefully about how we power this for future changelog features). 

As a result, I'm going to treat `removed` fields by marking them as `isRemoved` and filtering out values conditionally based on whether we want to query removed fields. This seems like the simplest way forward

## What

- Changed `without_field` to change the field to `is_removed` to true
- This is slightly confusing but:
  - If the shape is used for internal projections / diff generation, etc - filter out any removed nodes
  - If the shape is used for spectacle projections / data display, do not filter out these nodes.
- Added in RemovedIn edges for the changelog history for removed fields.
- 100000 snapshot test changes (the majority of them just include `is_removed` as a key - I've commented on the most important snapshot change)

TODO
- [x] Add a test for removed nodes and the shape projection with the return and is_removed fields
- [x] Test whether the internal projection and spectacle choice map returns this correctly

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


workspaces/optic-engine/src/queries/spectacle/spec_choices.rs
```diff
@@ -98,7 +98,7 @@ impl<'a> ShapeChoiceQueries<'a> {
       .into_iter()
       .map(move |choice| match choice.core_shape_kind {
         ShapeKind::ObjectKind => {
-          let object_fields = queries.resolve_shape_field_id_and_names(&choice.shape_id);
+          let object_fields = queries.resolve_shape_field_id_and_names(&choice.shape_id, true);
```
niclim, line 101: In the external / spectacle projection / shapeViewerProjection - we want to include removed nodes, so we can display the data associated with the removed field. (TODO test this out)

workspaces/optic-engine/src/shapes/traverser.rs
```diff
@@ -131,7 +131,7 @@ impl<'a> Traverser<'a> {
               // - find field node by key in object's field node edges
               let field_ids = self
                 .shape_queries
-                .resolve_shape_field_id_and_names(&choice.shape_id);
+                .resolve_shape_field_id_and_names(&choice.shape_id, false);
```
niclim, line 134: In the internal projection, we do _not_ want to resolve removed nodes, the only thing we need the removed nodes for is display / information in the changelog, not for modifying fields

workspaces/optic-engine/src/queries/snapshots/optic_engine__queries__shape__test__can_generate_remove_field_commands__choice_mapping.snap
```diff
@@ -8,7 +8,14 @@ expression: choice_mapping
...
                 json_type: Object,
-                fields: [],
+                fields: [
+                    ObjectFieldChoice {
+                        name: "lastName",
+                        field_id: "field_1",
+                        shape_id: "string_shape_1",
+                        is_removed: true,
+                    },
+                ],
```
niclim, line 18: This is the most important snapshot to review - the others just add `is_removed` to the all the keys. We expect removed nodes to be displayed in the spectacle shapeViewerProjection, however they have the key `is_removed` which can then be appropriately removed from the FE (similar to how endpoints, requests and responses are handled)

workspaces/optic-engine/src/shapes/traverser.rs
```diff
@@ -129,23 +129,23 @@ impl<'a> Traverser<'a> {
           .map(|choice| {
             if let ShapeKind::ObjectKind = &choice.core_shape_kind {
               // - find field node by key in object's field node edges
-              let field_ids = self
+              let field_nodes = self
                 .shape_queries
-                .resolve_shape_field_id_and_names(&choice.shape_id);
+                .resolve_shape_fields_nodes(&choice.shape_id, false);
```
niclim, line 134: In this case, traversing the shape for generating a diff should _not_ query removed shapes

workspaces/optic-engine/src/projections/shape.rs
```diff
@@ -416,26 +434,15 @@ impl ShapeProjection {
...
+      Some(Node::Field(field_node)) => field_node.is_removed = true,
+      _ => {}
     }
-
-    self.node_id_to_index.remove(&field_id); // prevents field node to be looked up by field id
-
-    // GOTCHA: we're not deleting the field node itself, as that would invalidate self.node_id_to_index
-    // as the graph indexes shift.
-    // TODO: figure out the implications of the above, like correctness of queries and
-    // eventual garbage collection.
```
JaapRood, line 438: I don't think we can get away with this. In a graph representation, the entire point is that you can construct queries by walking it, and rely on edges being representative. Adding the "is removed" filter to a couple accessor methods is really brittle, as often the graph is walked directly (as it is with the new edit shape stuff).

JaapRood (changes requested):
My initial thought was: fair enough, lets do what's pragmatic and enhance from there. However, changing the handling of removing field in this way, by not actually removing the node from the node_id's Map and not removing the edge, is not an acceptable solution. It goes against why this projection is modelled as a graph and the queries we have built, are building and will build on top of it.

I'm still happy for a pragmatic solution where we don't actually have to fork these projections, or happy to take on a more formal split between the two projections, but either way this approach is taking on a technical debt that's not worth it.

PR #1071, opened 2021-07-30
make diff fingerprints portable

devdoshi:
## Why
running a diff on one platform may yield fingerprints that do not match those on other platforms. This makes it difficult to decouple the system that performs the diff from the system that consumes the diff.

## What
Fingerprints should now be consistent across platforms, and since we're using the serde_json as input, a fingerprint should only change if there has been an incompatible change made.

The main suboptimal parts are: we clone the trails so we can normalize the jsonTrails, and we don't stream the serde_json bytes into the hasher. @JaapRood if you have thoughts on that let me know!

## Validation
* [ ] CI passes
* [ ] Verified in staging
* [ ] no egregious performance regression


workspaces/snapshot-tests/outputs/fe00636b02f665bf482575684386a6d1a6fca72da0ed9f6a5b8c8a71a88f0a97/af5c6563210c4c3efd24538b6de1d243bbc1ed32a9d107caddb480d2b6efccb3/output.snapshot.txt
```diff
@@ -17,10 +17,38 @@
...
+          }
         }
       }
     },
-    "6005d492e893247f"
+    "4822093850025955565"
+  ],
+  [
+    {
+      "UnmatchedResponseBodyContentType": {
```
devdoshi, line 30: not sure how to interpret this one...

JaapRood (approved):
Simple and effective, nice! Anything else would be a premature optimisation at this point.

workspaces/snapshot-tests/outputs/fe00636b02f665bf482575684386a6d1a6fca72da0ed9f6a5b8c8a71a88f0a97/af5c6563210c4c3efd24538b6de1d243bbc1ed32a9d107caddb480d2b6efccb3/output.snapshot.txt
JaapRood: That is.. odd? Why would any of that change because of the way fingerprints work? Perhaps a latent snapshot from some other PR? I haven't seen any failing PR merges though 🤔

workspaces/optic-engine/src/shapes/result.rs
```diff
@@ -18,10 +19,29 @@ pub enum ShapeDiffResult {
...
+    }
+  }
   pub fn fingerprint(&self) -> String {
-    let mut hash_state = DefaultHasher::new();
-    Hash::hash(self, &mut hash_state);
-    format!("{:x}", hash_state.finish())
+    let normalized = self.normalized();
+    let s = serde_json::to_vec(&normalized).expect("ShapeDiffResult should be json serializable");
+    let hashed = hash(&s);
+    hashed.to_string()
```
JaapRood, line 44: Casting to a string like this works, but it literally encodes the u64 straight to a string. Using `format!("{:x}", hashed)` you get a hexadecimal string, which seems to be the defacto choice for hashes 🤷‍♂️.

LouManglass:
🎉

PR #1072, opened 2021-07-30
Filter removed fields UI

niclim:
## Why

This PR targets https://github.com/opticdev/optic/pull/1070

Handle the removed field and relevant changelog display. Added a removed field event

Screen Shot 2021-07-30 at 3 22 16 PM
Screen Shot 2021-07-30 at 3 22 47 PM
* note - Notice the diff engine generates `unknown` when you add string and optional together - it looks like this is a bug on develop as well 

## What
- Split out changes from shapeMap in shape slice
- Handle rendering isRemoved and fields as render in changelog (using existing plumbing to make this easier).

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


JaapRood:
@niclim delaying the review of this until we get #1070 sorted, as that seems likely to go through significant change. Let me know if you think a review is useful regardless!

niclim:
All the code here is what's required to hook up the UI. But it does require spectacle to return the isRemoved nodes based on events, which is likely to change as part of #1070. I'll take a look at rebasing this to develop so that this can land and we have less in flight. I think it's very likely we'll need to split up the two shape projections

niclim:
Closing - will reimplement this when get around to adding in removed flags in spectacle projections

PR #1076, opened 2021-08-02
add project name, experiment with available fields

LouManglass:
## Why

Improve troubleshooting links, and experiment with improved capabilities to clarify sessions

## What

Duplication of troubleshooting data to prove concept, will pick one method going forward.

## Validation
* [x] CI passes
* [X] Verified in staging
* [x] Review


niclim (approved):
We might also want specId - i think @jshearer was thinking / looking at this as well

LouManglass:
> We might also want specId - i think @jshearer was thinking / looking at this as well

Good call - we're chatting analytics this week, I've added this to my list to go over. Shouldn't be hard, but it's not a trivial task for me today so I'm going to merge this in for now.

PR #1077, opened 2021-08-03
Allow field edit and removal commands to be queried through Spectacle

JaapRood:
## Why
To allow clients to make changes to fields, we'll want to expose a query through Spectacle.

## What
The top level `field` query from the Spectacle GraphQL schema is implemented and a reshuffling of resolvers allows the edit and remove commands to be generated separately. Spectacle implements these command generators through using two new additions to `optic-engine-wasm` interface: `spec_field_edit_commands` and `spec_field_remove_commands`. 

Follow up for later:
- ~~use these queries from the UI~~.

## Validation
* [x] CI passes
* [x] Returns command for fields for example project + standalone server
* [x] Works end to end in the UI.


workspaces/spectacle/src/graphql/schema.ts
```diff
@@ -291,6 +291,7 @@ type ObjectField {
 
 type FieldCommands {
   remove: [JSON!]!
+  edit(requestedTypes: [JsonType]!): [JSON!]!
```
niclim, line 294: Probably want to make `JsonType!`

workspaces/spectacle/src/spectacle.ts
```diff
@@ -679,6 +679,19 @@ export async function makeSpectacle(opticContext: IOpticContext) {
...
+        return commandGenerator.field.remove(parent.fieldId);
+      },
+      edit: async (
+        parent: FieldShape,
+        args: { requestedTypes: [JsonLike] },
+        context: GraphQLContext
+      ) => {
+        // TODO: query engine for actual commands
+        return [];
+      },
```
niclim, line 693: You'll also want to delete the L676-L697.

I think that commandGenerator.field.remove is also not connected up yet

JaapRood:
Got the engine hooked up to both the edit and removal command queries.

```graphql
query GetField($fieldId: ID!) {
  field(fieldId: $fieldId) {
    fieldId
    commands {
      remove
      edit(requestedTypes: [Null, Undefined])
    }
  }
}
```

niclim:
@JaapRood I've merged my changes into develop, if you rebase / merge the latest changes, all the relevant UI code should be there to have an end to end flow

Taskfile.yml
```diff
@@ -33,7 +33,7 @@ tasks:
   postpull:
     cmds:
       - task: 'workspaces:clean'
-      - FLAGS_FILE=.env.local task workspaces:setenv
+      - FLAGS_FILE=.env task workspaces:setenv
       - task: 'workspaces:build'
```
JaapRood, line 37: @niclim I'm sneaking this in, as this little tweak just cost me a whole bunch of head scratching, not understanding why Optic wouldn't compile with the removal buttons showing up. I enabled the feature flag in my local env, yet, wtf? Only to realise that my local `.env` is being overwritten!

A developer's local `.env` should never be overwritten, it's how one developer can enable some features (possibly completely local ones!) or adapt building Optic to their local machine (like an ARM Mac!), while not bothering others with that. If anyone want to mimic a certain environment, they have to opt-in.

niclim: Ah yeah that makes sense - I definitely overlooked this when I implemented this (i.e. FFs, and custom overrides). I still think we might want to have a "default" if `.env` isn't specified, given our env variables do drive some functionality (share). I think there's better env variable strategies that we could spend some time on.

workspaces/spectacle/src/graphql/schema.ts
niclim: Hmm - still probably want this to be `[JsonType!]!` - this will allow an array of `Enum` or `null` instead of just `Enum`

workspaces/spectacle/src/spectacle.ts
```diff
@@ -673,10 +723,17 @@ export async function makeSpectacle(opticContext: IOpticContext) {
         // TODO FLEB - connect up to optic engine
         return false;
       },
-      commands: (parent: FieldShape) => {
-        return {
-          remove: commandGenerator.field.remove(parent.fieldId),
-        };
+    },
+    FieldCommands: {
+      remove: async (parent: FieldShape) => {
```
niclim, line 728: This type is actually just `{fieldId: string}` - since only `commands` gets resolved as the parent

Taskfile.yml
JaapRood: I guess a default config could be copied in place if no local config exists, making that a bit smoother for first time setup / fresh clones. Even better would be a _merge_ instead of a replace, adding in any new settings that aren't set locally.

workspaces/spectacle/src/spectacle.ts
```diff
@@ -232,6 +237,51 @@ export async function makeSpectacle(opticContext: IOpticContext) {
...
+      ) => {
+        const fieldNodeWrapper = context
+          .spectacleContext()
+          .shapeQueries.findNodeById(fieldId);
+
+        if (
+          !fieldNodeWrapper ||
+          fieldNodeWrapper.result.type !== shapes.NodeType.Field
+        )
+          return Promise.resolve(null);
```
niclim, line 253: I don't actually think we need Promise.resolves here - but that's how the rest of the resolvers are set up - just an fyi

workspaces/spectacle/src/spectacle.ts
```diff
@@ -232,6 +237,51 @@ export async function makeSpectacle(opticContext: IOpticContext) {
...
+
+        let shapeNodeWrapper = context
+          .spectacleContext()
+          .shapeQueries.listIncomingNeighborsByType(
+            fieldId,
+            shapes.NodeType.Shape
+          )
+          .results.find((shapeNode) => {
+            return shapeNode.result.type === shapes.NodeType.Shape;
+          });
```
niclim, line 265: I _think_ I've updated the typing such that `listIncomingNeighborsByType` correctly returns the right type, so you only don't need to find (just check length >=1) - similar to https://github.com/opticdev/optic/blob/develop/workspaces/graph-lib/src/endpoints-graph/index.ts#L333-L341

You could also add this query (shapeFromField) in the FieldNodeWrapper class in graph-lib

workspaces/spectacle/src/spectacle.ts
```diff
@@ -232,6 +237,51 @@ export async function makeSpectacle(opticContext: IOpticContext) {
...
+        if (
+          !shapeNodeWrapper ||
+          shapeNodeWrapper.result.type !== shapes.NodeType.Shape
+        ) {
+          throw new Error('field should describe a shape');
+        }
+
+        let shapeNode = shapeNodeWrapper.result;
+
+        return Promise.resolve({
```
niclim, line 276: I think field is the only node we don't have as a Wrapper node (because we use the spectacleShapeViewerProjection)

Taskfile.yml
niclim: Ah yeah, that was the original intent, just implemented poorly - i'll take a stab at this sometime in the future (will matter when we have new team members / open source contributors)

workspaces/optic-domain/src/shapes.ts
```diff
@@ -30,6 +30,8 @@ export enum JsonLike {
   UNDEFINED = 'Undefined',
 }
 
+export type JsonType = JsonLike;
```
niclim, line 33: Why do we need this? Can't we use the enum directly as a type?

niclim (approved):

workspaces/optic-domain/src/shapes.ts
niclim: I think there's some type juggling that you can do to convert an enum into a type union (I spend a bit of time digging into this before going with a different approach) if this is what you were trying to do:
https://stackoverflow.com/a/52396706

workspaces/spectacle/src/graphql/schema.ts
JaapRood: Huh? Must have gotten lost in a squash or revert somewhere 🤷‍♂️.

workspaces/spectacle/src/spectacle.ts
JaapRood: I was just assuming all resolvers are supposed to be async as they all seem to be, but 🤷‍♂️.

workspaces/spectacle/src/spectacle.ts
JaapRood: You're right, it does return the more specific class, but checking for length and returning an array index breaks my nice functional chain. If anything, `NodeListWrapper` needs a `first` method hiding all that nastiness, making up for `Array`s lack of one. 

Feel free to contribute a clean up for this, but I'm going to pass on it for today (might pick up in the morning!).

workspaces/optic-domain/src/shapes.ts
JaapRood: Yay, Typescript 🥲.

niclim: Hmm if it's just for renaming, we can just find replace in all typescript files `JsonLike` -> `JsonType` - or change optic domain and follow the trail where typescript starts yelling

workspaces/spectacle/src/spectacle.ts
JaapRood: Turns out it still returns the generic `shapes.Node`, with further assertion being necessary to limit it down to a `shapes.ShapeNode`. I attempted a version where `NodeListWrapper` is a class with a `first` method, but that didn't really allow for any tangible improvements. It wasn't enough of a direction to leave it, only muddling up the water further. The typing to this graphing definitely needs some love, but that's outside of the scope of this PR.

workspaces/optic-domain/src/shapes.ts
JaapRood: Renamed in https://github.com/opticdev/optic/pull/1077/commits/84241850d5f563147868d290f020588563ea6f41

PR #1078, opened 2021-08-03
Filter out contributions that affect removed fields

niclim:
## Why
Describe the motivation for the changes

Filter out contributions that affect removed fields as they are effectively "invalid" contributions - we could make this more UI friendly by clearing / setting the initial value if the field becomes deleted, but I think it's fairly minor

## What

Some minor changes to the redux store - we _are_ duplicating field (in field map and shape). We could reduce data duplication by changing `shape.asObject.fields` -> `shape.asObject.fieldIds`. I haven't in this case but if you think it's worth doing I can include this in this PR

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


workspaces/ui-v2/src/store/shapes/slice.ts
```diff
@@ -30,11 +30,11 @@ const shapesSlice = createSlice({
         loading: false,
         data: action.meta.arg.rootShapeId,
       };
-      const { shapeMap, fieldIdToShapeId } = action.payload;
+      const { shapeMap, fieldMap } = action.payload;
       for (const [shapeId, reduxShapes] of Object.entries(shapeMap)) {
         state.shapeMap[shapeId] = reduxShapes;
       }
-      state.fieldIdToShapeId = fieldIdToShapeId;
+      state.fieldMap = fieldMap;
```
niclim, line 37: @niclim update so that this doesn't get overridden since we call this in multiple steps

niclim: thanks for the changes

JaapRood (approved):

PR #1079, opened 2021-08-03
Connect up remove field ui to command generation

niclim:
## Why

Connect up the remove fields to have an end to end flow

## What

- I decided against trying to implement more complicated ui logic to dedupe remove field commands, as long as we apply them in order (remove field -> remove endpoint) there shouldn't be any issue . Presumably in the future, we could reuse shapes - this means shapes aren't necessarily tied to an endpoint so it doesn't make sense to make this a requirement (i.e. removing a field only if the shape is tied to an endpoint
- This is not going to work until the spectacle code is joined together, however since the UI is gated (i.e. users cannot remove a field without the FF), we should be fine to merge this non-working spectacle query

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/ui-v2/src/store/documentationEdit/slice.ts
+++ b/workspaces/ui-v2/src/store/documentationEdit/slice.ts
@@ -92,7 +92,8 @@ const documentationEditSlice = createSlice({
       state.removedEndpoints = newRemovedEndpoints;
     },
     removeField: (state, action: PayloadAction<{ fieldId: string }>) => {
-      // TODO FLEB filter any other removed fields that is now a child of the newly added field.
+      // @GOTCHA - selecting a field, and then selecting a parent of this field (e.g. an object that contains this removed field)
+      // will count as a two selected fields. This is still correct, just extra commands
       state.fieldEdits.removedFields.push(action.payload.fieldId);
     },
     unremoveField: (state, action: PayloadAction<{ fieldId: string }>) => {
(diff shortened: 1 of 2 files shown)


workspaces/ui-v2/src/store/documentationEdit/slice.ts
```diff
@@ -92,7 +92,8 @@ const documentationEditSlice = createSlice({
       state.removedEndpoints = newRemovedEndpoints;
     },
     removeField: (state, action: PayloadAction<{ fieldId: string }>) => {
-      // TODO FLEB filter any other removed fields that is now a child of the newly added field.
+      // @GOTCHA - selecting a field, and then selecting a parent of this field (e.g. an object that contains this removed field)
```
niclim, line 95: I decided I'm not going to fix this as it's not trivial to fix with how we've modeled the store. This slice only has access to documentationEdits - and this logic needs to look at the shape tree. The handling should be in this action. To solve this we could:
- include the current store shapeMap in the action payload
- Restructure the redux store

Since this is only going to generate more commands, but the commands should be valid regardless, this is fine to keep in.

JaapRood: Minimising the amount of commands / resulting events is definitely an optimisation we can skip for now!

JaapRood (approved):

PR #1080, opened 2021-08-03
Add in simulated store for documentation edit view

niclim:
## Why

As part of making changes, users will want to see the preview of changes they have made in these views.

See how task is removed - it's removed on the left side, and disappeared from the right (ideally, we'd have the red highlight with removed, but that's work we can do to follow up with this bet)

Screen Shot 2021-08-03 at 3 31 23 PM

## What

Moved simulated command context into components and split out diff specific functionality to be used across both the diff and documentation flows

TODO
- [x] Merge in https://github.com/opticdev/optic/pull/1078 as this includes a fix where the fieldMap is sparsely populated
- [x] Merge in https://github.com/opticdev/optic/pull/1079 and create a shared spectacle client instance for generating commands

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


workspaces/ui-v2/src/clients/spectacle.ts
```diff
@@ -0,0 +1,90 @@
+import { CQRSCommand } from '../../../optic-domain/build';
+import { IForkableSpectacle } from '../../../spectacle/build';
+
+export class SpectacleClient {
```
niclim, line 4: There's more to pull into spectacle client, but this is a first step since we've been sharing calls across two places

workspaces/ui-v2/src/pages/docs/components/SimulatedBodyPreview.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+  useEffect(() => {
+    let isStale = false;
+    (async () => {
+      // TODO FLEB add in edited fields
+      const removeFieldCommandsPromise: Promise<CQRSCommand[]> = Promise.all(
+        removedFields.map((fieldId) =>
+          spectacleClient.fetchFieldRemoveCommands(fieldId)
+        )
+      ).then((fieldCommands) => fieldCommands.flat());
```
niclim, line 40: We call this twice (command preview generation, and in the save documentation changes), which is not ideal, but it is simpler to implement. Otherwise we would have to add in async state into redux, check for loading states and only allow submitting this was "loaded".

Since this is fairly low overhead and only 1 extra call, I think this is acceptable until we see we want to invest in performance

niclim:
This is ready to review, however it does require _both_ https://github.com/opticdev/optic/pull/1078 and https://github.com/opticdev/optic/pull/1079 PRs to be merged in, so it will include those changes as well. Going to open this up for review as it'll be useful for end to end testing (I also tried to select the commit range across the merge commits, but that wasn't working well)

workspaces/ui-v2/src/clients/spectacle.ts
JaapRood: I can see how for something as specific as fetching only a single deeply-nested field (commands in this case) creating a client class can make sense, but with how this is named, it seems to suggest all calls to Spectacle must go through the Spectacle client. That would seem to go against one of the major benefits of a GraphQL service: the ability for multiple consumers all to request exactly their own version of a shape they need. 

Not to say that there couldn't be a SpectacleClient (to centralise error handling for example, or for there to be predefined queries), but I'd expect it to be a whole lot more flexible, in line with what GraphQL provides.

JaapRood (approved):
Bit hard to review with all the other changes mixed in, so feel free to request another later.

workspaces/ui-v2/src/clients/spectacle.ts
niclim: Hmm yeah, that's definitely not the intent to be prescriptive. We do need to do some work to make the queries have better typing (such that the string you define, spits out the typing you expect). Perhaps error handling, or raising should live in the spectacle instance (in query or mutate) and that better typings are automatically inferred. 

Before we have that though, I think needing to define specific queries and add specific typings is much more important than having the ability for multiple consumers to request their specific data - given most things are through a local http request

Basically, yes I agree, but I think for now as it is implemented this provides more value and we can evolve spectacle to either handle these issues and we remove the need for a spectacle client, or we update the spectacle client to be more graphql friendly.

PR #1085, opened 2021-08-05
Improve how Typescript compiles across the monorepo

JaapRood:
## Why

Optic is structured as a monorepo, with multiple interdependent packages individually published and built in one repo. However, making changes across Typescript packages has been very painful so far, basically sourcing from every package being compiled in isolation by a separate Typescript instance. 

Generically, when working on package A, that depends on B, which depends on C, making a change in one of the dependencies doesn't automatically propagate up to package A: you have to manually make sure everything is compiled in the right order. 

Often, to make sure things are done correctly, you end up rebuilding everything, which is super slow. And even then you have to reload the VSC window, or restart the Typescript language server.

## What
In essence, we're connecting the isolated packages and making the Typescript compiler aware of dependencies living locally and are compilable as well.

It should enable :

- [x] Compiling all Typescript from the project root
- [x] Watching and compiling all the Typescript from the project root
- [x] Compiling a single package with all it's dependencies auto-compiling when necessary
- [x] Support TypeScript tooling across packages (e.g. Go to definition, refactor, etc.)
- [x] Allow for source maps debugging across all packages
- [x] Shared configurations and setup to consolidate setup.
- [ ] making ui-v2 webpack-dev-server automatically recompile when one of the dependencies changes (might not work out through a lack of support: https://github.com/facebook/create-react-app/issues/6799)

It manages to do this through:

- [Typescript Project References](https://www.typescriptlang.org/docs/handbook/project-references.html), providing the links between packages.
- [TypeScript Path Mapping](https://www.typescriptlang.org/docs/handbook/module-resolution.html#path-mapping), to allow use to import using `@useoptic/*`, while allowing the compiler to still find the local versions it could compile.

It adds a task:

- `workspaces:watch`, to watch the source of all Typescript and WASM targets and recompile them automatically.

Additional todos: 
- [x] Reconsider building from root with one TSC instance for `workspaces:build` (possibly more efficient): not necessary. As all Typescript packages are now connected, even building them one by one causes incremental builds, only recompiling what's necessary.

## Validation
* [x] CI passes
* [x] All tasks around compiling packages work in various states


niclim (approved):
🎉

Taskfile.yml
```diff
@@ -109,6 +109,22 @@ tasks:
...
+  workspaces:watch:typescript:
+    desc: Watch all the source of Typescripts workspaces and automatically recompile (with a single watcher)
+    cmds:
+      - yarn tsc -b --watch
+
+  workspaces:watch:wasm:
+    desc: Watch the source of WASM targets and automatically recompile
+    cmds:
+      - yarn wsrun --parallel --exclude-missing ws:watch-wasm
+
```
JaapRood, line 127: @notnmeyer this is the best way I could imagine doing watchers with Taskfile. Do you see any problems with this / know of a better way?

workspaces/optic-engine-wasm/scripts/watch.js
```diff
@@ -0,0 +1,36 @@
...
+  },
+};
+
+let wasmPack = createWasmPackPlugin();
+
+wasmPack.apply(compiler);
+
+beforeCompileHook().catch((err) => {
+  throw err;
+});
```
JaapRood, line 36: How hacky is this? Pretty hacky! But given how incredibly small the integration between the wasm-pack watcher is, there being no alternatives and the complexity of forking / writing it ourselves being more complicated than this, I say it's worth it (for now, at least).

niclim:
Seems reasonable - watching on optic-engine-native would be nice, but this solves a large painpoint with the current TS compilations

JaapRood:
Developed against this for two days, all seems to working smoothly, passing CI as well. Lets get this out to develop!

JaapRood:
@LouManglass these changes are definitely an added risk for the next release train. Perhaps we should get a side channel build out there to verify it's all functional?

LouManglass:
> @LouManglass these changes are definitely an added risk for the next release train. Perhaps we should get a side channel build out there to verify it's all functional?

Yeah, I'll check that today. Better to find out sooner.

Taskfile.yml
notnmeyer: sorry, been out. this looks fine to me though. i dont think its a better solution in this case, but taskfile also supports fingerprinting source files and watching a task too,

https://github.com/go-task/task/blob/master/docs/usage.md#watch-tasks
https://github.com/go-task/task/blob/master/docs/usage.md#by-fingerprinting-locally-generated-files-and-their-sources

PR #1090, opened 2021-08-06
Query param not showing up in endpoint changes

niclim:
## Why

Query parameters weren't being considered here and weren't causing endpoints to show they have changes

## What

- Refactored buildEndpointChanges to be a little dryer
- Added query parameters to this function
- fixed a minor issue with color

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/spectacle/src/helpers.ts
+++ b/workspaces/spectacle/src/helpers.ts
@@ -93,64 +93,38 @@ export function buildEndpointChanges(
 
   const changes = new Changes();
 
-  [...deltaBatchCommits.values()].forEach(
-    (batchCommit: endpoints.BatchCommitNodeWrapper) => {
-      // Only createdIn and removedIn edges pointing to a request is treated as
-      // an added or removed change, everything else is updated.
-      // Bodies, and shape changes are handled below
-      batchCommit
-        .createdInEdgeNodes()
-        .results.forEach((node: endpoints.NodeWrapper) => {
-          if (node instanceof endpoints.RequestNodeWrapper) {
-            const endpoint = endpointFromRequest(node);
-            if (endpoint) {
-              changes.captureChange('added', endpoint);
+  for (const [_, batchCommit] of deltaBatchCommits) {
+    for (const { edgeType, nodes } of [
+      { edgeType: 'created', nodes: batchCommit.createdInEdgeNodes().results },
+      { edgeType: 'updated', nodes: batchCommit.updatedInEdgeNodes().results },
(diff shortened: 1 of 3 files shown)


workspaces/spectacle/src/helpers.ts
```diff
@@ -169,51 +143,63 @@ export function buildEndpointChanges(
...
+  }
+
+  const filteredRootShapesWithEndpoint = rootShapesWithEndpoint.filter(
+    ({ rootShapeId }) => {
       if (batchCommitNeighborIds.has(rootShapeId)) {
-        results.push(rootShapeId);
-        return results;
+        return true;
       }
+      // TODO this does not handle array children and polymorphic type changes
```
niclim, line 185: This is why i think certain changes aren't showing up

workspaces/spectacle/src/helpers.ts
```diff
@@ -93,64 +93,38 @@ export function buildEndpointChanges(
...
-            );
-            if (endpoint) {
-              changes.captureChange('updated', endpoint);
+  for (const [_, batchCommit] of deltaBatchCommits) {
+    for (const { edgeType, nodes } of [
+      { edgeType: 'created', nodes: batchCommit.createdInEdgeNodes().results },
+      { edgeType: 'updated', nodes: batchCommit.updatedInEdgeNodes().results },
+      { edgeType: 'removed', nodes: batchCommit.removedInEdgeNodes().results },
+    ]) {
+      for (const node of nodes) {
```
JaapRood, line 102: Three-nested for-loop! Time for a computational complexity-check!

The second loop is constant in size of 3 so can be almost disregarded. The outer loop can be of arbitrary size, but is driven by human input, so not expected to grow to any meaningful size fast.The most inner loop's nodes _should_ all be nodes we have to actually visit, and also describe a human generated entity (endpoints).

Conclusion: seems reasonable, shouldn't present much of an issue until we get to deal with legacy API specs that haven't been garbage collected, if ever. If it ever does, there's plenty of paths forward to optimise (intermediate projections, async compute to prevent blocking the event loop, etc).

JaapRood (approved):

workspaces/spectacle/src/helpers.ts
niclim: At the current point in time, the minimum work we need to do is visit every node in these batch commits, which you've pointed out is fine but likely an issue in the future when specs grow. 

We could also precompute endpoint changes or other derived data during spec writing as another option.

PR #1093, opened 2021-08-09
fix: Plumb the spec ID through correctly

jshearer:
## Why
This looked like it was correct, but turns out that in this particular case the specID was getting swallowed :(


jshearer:
Also add using `v3` endpoint which allows for sending analytics DI, which ended up never getting deployed to the frontend here :(

niclim (approved):

PR #1094, opened 2021-08-10
Fix Typescript compilation for ui-v2

JaapRood:
## Why
The changes in #1085 broke compilation for ui-v2.

## What
There were two issues:

- now that we're incrementally compiling between packages, `ui-v2` needed to also clean it's build cache to avoid stale builds.
- `ui-v2` depends on `spectacle-shared`, but wasn't specified to do so. For runtime this worked fine as `spectacle` pulled it in making it available to ui-v2 too, but during compile time the types could not be resolved.

## Validation
* [ ] CI passes
* [x] fresh `task postpull` works
* [x] compiling only ui-v2 from clean slate works

### DIFF
--- a/workspaces/ui-v2/package.json
+++ b/workspaces/ui-v2/package.json
@@ -21,6 +21,7 @@
     "start:demo": "env OPTIC_WEBPACK_ENTRYPOINT=demo node scripts/start.js",
     "build": "yarn run build:local",
     "ws:test": "jest",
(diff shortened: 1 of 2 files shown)


JaapRood:
Gonna go ahead and merge this as soon as CI passes, as fresh pulls of `develop` are broken without it. Anything found in review can be handled with new PRs.

niclim (approved):

PR #1095, opened 2021-08-10
Fix data-fieldId error

niclim:
## Why

data-fieldId was throwing an error, to suppress this data-attributes should be all lowercase - which this all follows

## What

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


niclim:
Oh - I noticed you cherry-picked one of my commits for this fix in your branch - if it's easier we can just include it as part of your PR (and i can close this) - feel free to close this out if that's the case @JaapRood

niclim:
Will be addressed as part of https://github.com/opticdev/optic/pull/1101

PR #1096, opened 2021-08-10
Implement redux actions for edit fields

niclim:
## Why

Exposes redux actions for editing fields (specifying nullable / optional).

## What

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/ui-v2/src/clients/spectacle.ts
+++ b/workspaces/ui-v2/src/clients/spectacle.ts
@@ -1,4 +1,4 @@
-import { CQRSCommand } from '@useoptic/optic-domain';
+import { CQRSCommand, JsonType } from '@useoptic/optic-domain';
 import { IForkableSpectacle } from '@useoptic/spectacle';
 
 export class SpectacleClient {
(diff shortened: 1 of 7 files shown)


workspaces/ui-v2/src/clients/spectacle.ts
```diff
@@ -47,6 +47,50 @@ export class SpectacleClient {
...
+      },
+    });
+    if (results.errors) {
+      console.error(results.errors);
+      throw new Error(JSON.stringify(results.errors));
+    }
+    if (!results.data || !results.data.field) {
+      const message = `Could not generate edit commands for field: ${fieldId}`;
+      console.error(message);
+      throw new Error(message);
```
JaapRood, line 89: Just curious: why both `console.error` and throwing? Wouldn't doing just the latter also show up in console?

workspaces/ui-v2/src/store/documentationEdit/slice.ts
```diff
@@ -68,6 +76,45 @@ const documentationEditSlice = createSlice({
...
+      action: PayloadAction<{
+        fieldId: string;
+        options: {
+          isOptional: boolean;
+          isNullable: boolean;
+        };
+      }>
+    ) => {
+      const { fieldId, options } = action.payload;
+      state.fields.edited[fieldId] = options;
```
JaapRood, line 104: Is there any benefit to checking whether the field has been removed before accepting edits, or is that something we don't want to constrain here?

workspaces/ui-v2/src/store/documentationEdit/thunks.ts
```diff
@@ -27,6 +28,21 @@ const fetchRemoveEndpointCommands = async (
...
+  spectacle: IForkableSpectacle,
+  fieldId: string,
+  requestedTypes: JsonType[]
+): Promise<CQRSCommand[]> => {
+  const spectacleClient = new SpectacleClient(spectacle);
+  try {
+    return spectacleClient.fetchFieldEditCommands(fieldId, requestedTypes);
+  } catch (e) {
+    console.error(e);
+    Sentry.captureException(e);
```
JaapRood, line 41: We don't have a Sentry handler anywhere further at the "edge" of our program, forwarding all thrown errors, do we? Seems weird to see vendor specific stuff so deeply integrated, but I guess it's pragmatic?

JaapRood (approved):

workspaces/ui-v2/src/store/documentationEdit/slice.ts
niclim: This is more of an upsert, rather than an add. We could presumably set isOptional: true and then also isNullable true. It's only if we have the default values, we want to remove this in the edited state (since we have no commands to generate)

workspaces/ui-v2/src/store/documentationEdit/thunks.ts
niclim: Hmmm, we don't want the errors to be propagated past the redux store, but there is probably some middleware we can listen to rejected promises and sentry error, rather than add this to each thunk. I'll look into this and put up a separate PR

workspaces/ui-v2/src/clients/spectacle.ts
niclim: It depends on how we error handle, redux catches these thrown errors and converts them into a rejected action. Uncaught errors (which crash the page) will show up in the console, we probably want to avoid this, but there's probably a better integration we can add in here

PR #1097, opened 2021-08-10
Fix/diff navigation prompting

niclim:
## Why

Previously, two things weren't picked up with the diff prompt navigate away (i.e. a prompt when users are about to lose their data)
- Any path pattern specification (i.e. parameterized paths)
- Navigating back to the diff root (which causes a new diff to generate, and we lose our data)

## What

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/ui-v2/src/pages/diffs/components/DiffAccessoryNavigation.tsx
+++ b/workspaces/ui-v2/src/pages/diffs/components/DiffAccessoryNavigation.tsx
@@ -18,11 +18,16 @@ export function DiffAccessoryNavigation() {
     handledCount,
     hasDiffChanges,
     getUndocumentedUrls,
+    wipPatterns,
   } = useSharedDiffContext();
   const diffReviewCapturePage = useDiffReviewCapturePageLink();
   const undocumentedUrlsPageLink = useDiffUndocumentedUrlsPageLink();
   const history = useHistory();
   const [handled, total] = handledCount;
+  const hasChanges =
+    hasDiffChanges() ||
+    Object.values(wipPatterns).filter((pattern) => pattern.isParameterized)
+      .length > 0;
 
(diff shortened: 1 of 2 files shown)


workspaces/ui-v2/src/pages/diffs/components/DiffAccessoryNavigation.tsx
```diff
@@ -18,11 +18,16 @@ export function DiffAccessoryNavigation() {
...
+    wipPatterns,
   } = useSharedDiffContext();
   const diffReviewCapturePage = useDiffReviewCapturePageLink();
   const undocumentedUrlsPageLink = useDiffUndocumentedUrlsPageLink();
   const history = useHistory();
   const [handled, total] = handledCount;
+  const hasChanges =
+    hasDiffChanges() ||
+    Object.values(wipPatterns).filter((pattern) => pattern.isParameterized)
+      .length > 0;
```
JaapRood, line 30: Total nit: this naming confuses me, do the `wipPatterns` have anything to do with whether there are any changes? Seems more like `shouldAskForCommit = hasDiffChanges() || isDiffsPage`?

JaapRood (approved):

workspaces/ui-v2/src/pages/diffs/components/DiffAccessoryNavigation.tsx
niclim: wipPatterns refers to the parameterized path patterns here 
Screen Shot 2021-08-11 at 9 57 21 AM

There is user input, but nothing to "save" until you've learnt a path. We've seen users navigate away while this is in progress and then come back and lose their data - therefore we should have a prompt (partially completed form)

PR #1098, opened 2021-08-10
Check if daemon running

niclim:
## Why

The daemon can crash for a number of reasons. Sometimes, there's a recurring bug, other times we can power through this. We should ideally figure out why daemons crash - but for a user, they can retry and restart the daemon. This is not immediately obvious for users who are not familiar with how optic works. As a result, i've added a check to see if the daemon is reachable for the local cli ui builds.

## What

Screen Shot 2021-08-10 at 3 56 09 PM

- Add `/api/health` endpoint as a health check
- Add a EnsureDaemonRunning component

In the future, we could also validate expected versions such that the daemon and UI are running in sync in the same versions. Although there's probably different work to do to handle the case where we could have multiple daemons running in the background.

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/cli-server/src/server.ts
+++ b/workspaces/cli-server/src/server.ts
@@ -98,6 +98,10 @@ class CliServer {
 
     const anonIdPromise = getOrCreateAnonId();
 
+    app.get('/api/health', async (req, res) => {
+      res.json({ status: 'ok' });
+    });
+
     app.get('/api/identity', async (req, res: express.Response) => {
       res.json({ user, anonymousId: await anonIdPromise });
     });
--- a/workspaces/cli-server/src/server.ts
+++ b/workspaces/cli-server/src/server.ts
@@ -98,10 +98,6 @@ class CliServer {
 
     const anonIdPromise = getOrCreateAnonId();
 
-    app.get('/api/health', async (req, res) => {
-      res.json({ status: 'ok' });
-    });
-
     app.get('/api/identity', async (req, res: express.Response) => {
       res.json({ user, anonymousId: await anonIdPromise });
     });
--- a/workspaces/ui-v2/src/entry-points/local/EnsureDaemonRunning.tsx
+++ b/workspaces/ui-v2/src/entry-points/local/EnsureDaemonRunning.tsx
@@ -1,19 +1,34 @@
 import React, { FC, useEffect, useState } from 'react';
 import { makeStyles } from '@material-ui/core';
+import * as Sentry from '@sentry/react';
+
 import { Client } from '@useoptic/cli-client';
 
 import { FullPageLoader } from '<src>/components';
 import * as SupportLinks from '<src>/constants/SupportLinks';
+const packageJson = require('../../../package.json');
+const uiPackageVersion = packageJson.version;
 
 export const EnsureDaemonRunning: FC = ({ children }) => {
   const [loading, setLoading] = useState(true);
   const [error, setError] = useState(false);
+  const [hasMismatchedVersion, setHasMismatchedVersions] = useState(false);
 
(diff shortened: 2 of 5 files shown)


workspaces/ui-v2/src/entry-points/local/EnsureDaemonRunning.tsx
```diff
@@ -0,0 +1,73 @@
...
+      <p>
+        You can restart the daemon by running <code>api daemon:start</code> or
+        rerunning <code>api start</code>.
+      </p>
+      <p>
+        If this continues to happen, please reach out to{' '}
+        <a href={SupportLinks.Contact('Optic App crash report')}>our team</a>{' '}
+        for assistance. Further debug information can be found from our{' '}
+        <a
+          href="https://useoptic.com/reference/optic-cli/commands/debug"
```
LouManglass, line 53: Looks fine, though this might be reusable as a supportLinks.ts constant https://github.com/opticdev/optic/blob/develop/workspaces/ui-v2/src/constants/SupportLinks.ts

LouManglass (approved):

LouManglass:
Approved, with one non-blocker comment/thought.

workspaces/ui-v2/src/entry-points/local/EnsureDaemonRunning.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+  return (
+    <div className={classes.errorContainer}>
+      <h2>Error reaching daemon</h2>
+      <p>
+        It appears the cli daemon is unreachable. This usually means that the
+        daemon crashed and needs to be restarted.
+      </p>
+      <p>
+        You can restart the daemon by running <code>api daemon:start</code> or
```
devdoshi, line 45: I would say `api spec` or `api start` as opposed to daemon:start because spec/start etc also create a session so the page they are on will still probably work. If they do daemon:start the session will not exist and the page will not be able to reload properly. Essentially, it's not clear what you want to happen once they restart the daemon given they have this page open. Thoughts?

devdoshi (approved):
NIce!

workspaces/cli-server/src/server.ts
```diff
@@ -98,6 +98,10 @@ class CliServer {
 
     const anonIdPromise = getOrCreateAnonId();
 
+    app.get('/api/health', async (req, res) => {
+      res.json({ status: 'ok' });
```
devdoshi, line 102: may as well stick the version in here now. i'm curious how often we run into client version being out of sync with daemon version

workspaces/ui-v2/src/entry-points/local/EnsureDaemonRunning.tsx
niclim: Hmmm api spec seems to be better - I think there might be some work we want to do to change the daemon behavior such that you need to run api _abc_ to run the UI or daemon explicitly + redraw the boundaries of how we use this. It seems we've been having some mismatched UI versions hitting older running daemons (with old running code) - such that I think it's more explicit and more intuitive that to see the UI, you need to run `api spec` or something similar, and closing that will close both the UI and daemon

Something for outside of this PR - but I think that we more issues with mismatched versions if we don't address this

workspaces/cli-server/src/server.ts
niclim: I'll add this in and get sentry to report on errors

niclim:
Ok - I updated this so we now also check the version. There also conveniently a `/api/daemon/status` endpoint already there.

Screen Shot 2021-08-10 at 4 42 47 PM

PR #1099, opened 2021-08-10
Deps/graphql

LouManglass:
## Why
Discussion with mockttp maintainer to resolve graphql peerDependencies for NPM 7

## What

- mockttp dependency version
- explicit graphql dependency in local-cli

## Validation
* [ ] CI passes
* [ ] Verified with side-channel build on NPM 7


workspaces/ui-v2/src/entry-points/cloud/cloud-viewer.tsx
```diff
@@ -163,36 +163,6 @@ export default function CloudViewer() {
   );
 }
 
-class CloudInMemorySpectacle
```
niclim, line 166: Removed this since it's a full copy paste of `InMemorySpectacle` and to fix the type signature here would have involved adding graphql as a dependency for the ui-v2 instance (for reasons)

workspaces/ui-v2/src/lib/__tests/diff-helpers/universes/makeCurrentSpecContext.ts
```diff
@@ -16,11 +14,11 @@ delete global.self;
 
 export async function makeCurrentSpecContext(
   events: any[],
-  query: IBaseSpectacle['query']
```
niclim, line 19: This is actually the wrong type signature, since we actually need the queryWrapper signature, however it's harder to extract, and this is just a test so 🤷

LouManglass:
Preliminary testing looks good on Node v12 (NPM 6.x), 14 (NPM 6.x), 16 (NPM 7.x) all latest from NVM. More testing required, on both versions of NPM, then ready for review. Putting aside to test some other work.

niclim (approved):

LouManglass:
Tested multiple commands after installing with both NPM 6 and 7, each on Node 12, 14, 16. No errors seen, expected output from commands (like generate:oas, scripts, status after a test run). This looks ready, full usability testing with release.

JaapRood:
Could this PR be why I'm getting errors about graphQL modules? Running `apidev run baseline` for `optic-example-projects/todo-js`:

```
[optic] Running dependent task start...
    Error: 500 Internal Server Error
    {"errors":[{"message":"Cannot use GraphQLSchema
    \"[object GraphQLSchema]\" from another module or
    realm.\n\nEnsure that there is only one instance
    of \"graphql\" in the node_modules\ndirectory. If
    different versions of \"graphql\" are the
    dependencies of other\nrelied on modules, use
    \"resolutions\" to ensure only one version is
    installed.\n\nhttps://yarnpkg.com/en/docs/selectiv
    e-version-resolutions\n\nDuplicate \"graphql\"
    modules cannot be used at the same time since
    different\nversions may have different
    capabilities and behavior. The data from
    one\nversion used in the function from another
    could produce confusing and\nspurious results."}]}
 ```

Running against [latest](https://github.com/opticdev/optic/commit/f901f6de53aa7b43393b59a0cf59c2b2fec81b6b) `develop`, completely wiped all my node modules and created a fresh build.

LouManglass:
Heh, that's exactly what this patch is supposed to fix...in NPM 7. Apparently fixing NPM broke Yarn. I've got some meetings this morning but I'll back out these changes as soon as I can so develop is back in a good state.

LouManglass:
Oh hey there's this handy Revert button for the PR...

PR #1101, opened 2021-08-11
Implement new Shape Editor UI

JaapRood:
## Why
To properly support field level edits we need a place where the edit UI lives.

## What
A new `ShapeEditor` component has been added, designed to rendered as part of the endpoint documentation page while in edit mode. It's not yet done, nor does it fulfil all the design constraints we've set ourselves, but the foundations seem strong enough to merge, so we can wire it up and iterate on it with separate PRs.

Kapture 2021-08-11 at 18 15 49

## Validation
* [ ] CI passes
* [x] Only rendered with feature flag enabled

### DIFF
--- a/workspaces/ui-v2/src/pages/docs/EndpointRootPage.tsx
+++ b/workspaces/ui-v2/src/pages/docs/EndpointRootPage.tsx
@@ -295,7 +295,8 @@ export const EndpointRootPage: FC<
                           }
                         }
 
-                        return (
+                        return process.env.REACT_APP_FF_FIELD_LEVEL_EDITS !==
+                          'true' ? (
                           <DocFieldContribution
                             key={
                               field.contribution.id +
@@ -312,6 +313,8 @@ export const EndpointRootPage: FC<
                             initialValue={field.contribution.value}
                             required={field.required}
                           />
+                        ) : (
+                          <ShapeEditor fields={fields} />
                         );
                       }}
                       fieldDetails={fields}
@@ -376,28 +379,37 @@ export const EndpointRootPage: FC<
                           {(selectedFieldId, setSelectedFieldId) => (
                             <div className={classes.bodyDetails}>
                               <div>
-                                <ContributionsList
-                                  renderField={(field) => (
-                                    <DocFieldContribution
-                                      key={
-                                        field.contribution.id +
-                                        field.contribution.contributionKey
-                                      }
-                                      endpoint={{
-                                        pathId,
-                                        method,
-                                      }}
-                                      name={field.name}
-                                      shapes={field.shapes}
-                                      depth={field.depth}
-                                      id={field.fieldId}
-                                      initialValue={field.contribution.value}
-                                      required={field.required}
-                                      setSelectedField={setSelectedFieldId}
-                                    />
-                                  )}
-                                  fieldDetails={fields}
-                                />
+                                {process.env.REACT_APP_FF_FIELD_LEVEL_EDITS !==
+                                'true' ? (
+                                  <ContributionsList
+                                    renderField={(field) => (
+                                      <DocFieldContribution
+                                        key={
+                                          field.contribution.id +
+                                          field.contribution.contributionKey
+                                        }
+                                        endpoint={{
+                                          pathId,
+                                          method,
+                                        }}
+                                        name={field.name}
+                                        shapes={field.shapes}
+                                        depth={field.depth}
+                                        id={field.fieldId}
+                                        initialValue={field.contribution.value}
+                                        required={field.required}
+                                        setSelectedField={setSelectedFieldId}
+                                      />
+                                    )}
+                                    fieldDetails={fields}
+                                  />
+                                ) : (
+                                  <ShapeEditor
+                                    fields={fields}
+                                    selectedFieldId={selectedFieldId}
+                                    setSelectedField={setSelectedFieldId}
+                                  />
+                                )}
                               </div>
                               <div className={classes.panel}>
                                 {isEditing ? (
--- a/workspaces/ui-v2/src/pages/docs/EndpointRootPage.tsx
+++ b/workspaces/ui-v2/src/pages/docs/EndpointRootPage.tsx
@@ -26,7 +26,7 @@ import {
   selectors,
   documentationEditActions,
 } from '<src>/store';
-import { IShapeRenderer } from '<src>/types';
+import { IShapeRenderer, IFieldDetails } from '<src>/types';
 import { getEndpointId } from '<src>/utils';
 import { useRunOnKeypress } from '<src>/hooks/util';
 import {
@@ -38,6 +38,7 @@ import {
   MarkdownBodyContribution,
   DeleteEndpointConfirmationModal,
   SimulatedBody,
+  ShapeEditor,
 } from '<src>/pages/docs/components';
 import { useAnalytics } from '<src>/contexts/analytics';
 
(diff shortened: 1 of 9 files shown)


workspaces/ui-v2/src/styles/theme.ts
```diff
@@ -34,6 +36,16 @@ export const methodColorsDark: { [key: string]: string | undefined } = {
   DELETE: '#be5353',
 };
 
+export const jsonTypeColors = {
```
niclim, line 39: 🎉

niclim: I'm not sure what the type representation looks like here, but I think we can set exhaustive type checking by specifying the type signature as `Record<JsonType, string>`

workspaces/ui-v2/src/pages/docs/EndpointRootPage.tsx
```diff
@@ -471,28 +472,40 @@ export const EndpointRootPage: FC<
...
-                                        required={field.required}
-                                        setSelectedField={setSelectedFieldId}
-                                      />
-                                    )}
-                                    fieldDetails={fields}
-                                  />
+                                  {process.env
+                                    .REACT_APP_FF_FIELD_LEVEL_EDITS !==
+                                  'true' ? (
+                                    <ContributionsList
```
niclim, line 478: It's probably worth setting this for the `QueryParametersPanel` and the request `HttpBodyPanel` in this component as well

workspaces/ui-v2/src/pages/docs/components/ShapeEditor.tsx
```diff
@@ -0,0 +1,465 @@
...
+import * as Theme from '<src>/styles/theme';
+
+export const ShapeEditor: FC<{
+  fields: IFieldDetails[];
+  selectedFieldId: string | null;
+  setSelectedField: (fieldId: string | null) => void;
+}> = ({ fields, selectedFieldId, setSelectedField }) => {
+  const classes = useStyles();
+
+  let onChangeFieldDescription = useMemo(
```
niclim, line 26: We can use useCallback here to avoid the first `() => ..`

workspaces/ui-v2/src/pages/docs/components/ShapeEditor.tsx
```diff
@@ -0,0 +1,465 @@
...
+};
+
+const Row: FC<{
+  field: IFieldDetails;
+  description: string;
+  selected: boolean;
+
+  onChangeDescription?: (fieldId: string, description: string) => void;
+  onSelect?: (fieldId: string) => void;
+}> = function ShapeEditorRow({
```
niclim, line 61: Any reason you set the Row variable to a named function? I think just naming `const ShapeEditorRow` sets the name of the function, and makes it easier to trace function calls between things

workspaces/ui-v2/src/pages/docs/components/ShapeEditor.tsx
```diff
@@ -0,0 +1,465 @@
...
+  description,
+  selected,
+  onChangeDescription,
+  onSelect,
+}) {
+  const classes = useStyles();
+
+  const onClickFieldHeader = useMemo(
+    () => () => {
+      if (onSelect) onSelect(field.fieldId);
```
niclim, line 72: We'll want to unselect here - so `onSelect(selected ? null : field.fieldId)` which will unhighlight when dismissed (such that the highlight action is undoable)

workspaces/ui-v2/src/pages/docs/components/ShapeEditor.tsx
```diff
@@ -0,0 +1,464 @@
+import React, { FC, useMemo } from 'react';
```
niclim, line 1: Thoughts on splitting this component up into multiple files? We could create a folder `ShapeEditor` with an index.ts file to match the export structure. Think this might continue to grow and a separation of files might be helpful here

workspaces/ui-v2/src/pages/docs/components/ShapeEditor.tsx
```diff
@@ -0,0 +1,464 @@
...
+    },
+    [onSelect, field.fieldId]
+  );
+
+  const onChangeType = useMemo(
+    () => (type: JsonType, enabled: boolean) => {
+      console.log('changed type for field', type, enabled, field.fieldId);
+    },
+    [field.fieldId]
+  );
```
niclim, line 82: This function signature might need to change, but I'll tackle this when i hook this up into redux actions that I've set up.

niclim (approved):
This looks great! Mainly minor comments (we can use `useCallback` instead of `useMemo(() => () => `) and other minor things, but nothing important

workspaces/ui-v2/src/styles/theme.ts
JaapRood: It seems to be exhaustive so far, with TSC complaining when I didn't have `OBJECT` and `ARRAY` covered. But go nuts if I don't have to complete picture!

workspaces/ui-v2/src/pages/docs/components/ShapeEditor.tsx
JaapRood: Short names for internal use, while more explicit names for the debugger. Ever since fat arrow became a thing people don't use named functions enough anymore, they are so much nicer to debug rather than a massive stack trace of "anonymous". Or maybe debuggers have caught up recently and resolve variable names? I'm not sure, just a bit of a habit. 🤷‍♂️

workspaces/ui-v2/src/pages/docs/components/ShapeEditor.tsx
JaapRood: Sure, that's probably fine. Just make sure to stop propagation on the remove control, so we don't close inadvertently 😅.

workspaces/ui-v2/src/pages/docs/components/ShapeEditor.tsx
JaapRood: It's all tightly connected now, not yet stable and there's no reason why anything other component should use some of these internal components directly. So lets leave that for now.

workspaces/ui-v2/src/pages/docs/components/ShapeEditor.tsx
JaapRood: If it does, then it does. Most important to me is that all of this stuff bubbles up to `ShapeEditor` (and preferable in a way that makes sense in isolation), and have all child components be "unconnected" in oldskool Redux parlance.  You end up handing down props a bit more, but it makes it so much easier to evolve and re-use parts if you don't have to worry about all that context. Preferably, even `ShapeEditor` is unaware of redux or context, and could be hooked up to other state solutions (or mocks) just as easily.

workspaces/ui-v2/src/pages/docs/components/ShapeEditor.tsx
niclim: Sure, I don't think we'd intend any of these internal components directly even if we did split out these into different files. But there is probably a point where this shape editor might be easier to reason able in multiple files - since we are still iterating on this, it seems fine to keep as is

workspaces/ui-v2/src/styles/theme.ts
niclim: This is more if we miss a variant of JsonType in this object, typescript will start yelling

workspaces/ui-v2/src/pages/docs/components/ShapeEditor.tsx
niclim: I'd agree with this, however if we do that and we change the field, that will result in all fields rerendering on a field change (if we pass down top level props) - so in this case it might be better to have access to the redux store. The biggest point of confusion is if we reuse this component across pages; needing to be renderered in a redux store can get messy unless the assumption is that there is _always_ a redux store.

There's definitely other ways to handle this, but it is probably easier to access props via redux selectors passing down props + handling render trigger behavior

niclim: Sure, we can definite do performance optimization in a bunch of different ways - and the fact that we already have functions + children that already cause re-renders does not mean we shouldn't do this. The ideal is that the component is portable with a clear split between them, and you can definitely keep that interface while optimizing performance, but that comes with a higher cost than using redux selectors (which have this logic built in). 

Context has advantages (in this example, redux) over passing props with render optimization.

> If that were to become a performance problem (which it really won't)

We have had performance issues with this page in the past where a change in a large shape caused a large amount of nodes to be rerendered - causing noticeable UI jank - so I wouldn't say it's something that will never happen and we never need to worry about this.

PR #1105, opened 2021-08-12
Only render the ShapeEditor and selected field highlights in edit mode

JaapRood:
## Why
We want the viewing and editing of the documentation to be different, as they serve different goals. To make sure the editor is understood to be interactive and capable of making changes, we don't just want to render it in an inactive, read-only mode.

## What
We render the original list of properties for viewing the documentation, while rendering the new editor when editing.

Kapture 2021-08-12 at 10 21 50

Note: there's ways to further optimise the transition to go from viewing into editing mode, but we'll consider that separately.

## Validation
* [ ] CI passes


niclim (approved):
Seems fine for now, but I presume we'd want to create a new component for the "non-editing" mode (changelog and not-editing docs state)?

PR #1107, opened 2021-08-12
Provide stronger contextual link between Shape's preview and editor

JaapRood:
## Why
We want there to be a strong link between the shape editor and preview in documentation edit mode, making field level editing as discoverable and easy to access as possible.

## What
The currently edited field is highlighted. Fields are selectable from the preview while in edit mode.

Kapture 2021-08-12 at 13 43 10

## Validation
* [ ] CI passes


niclim (approved):

niclim:
LGTM - the only change i made here was renaming `selectableFields` -> `fieldsAreSelectable` - previous was a little ambiguous where selectableFields seemed like an array of fieldIds that were selectable.

PR #1108, opened 2021-08-12
Fix daemon version mismatch

niclim:
## Why

In side channel builds:
```
Request URL: http://localhost:34444/api/daemon/status > {isRunning: true, version: "10.2.4-alpha.b223"}

Looks like this will always fail on side-channel/beta/etc. builds. ui-v2 workspace package.json:

"version": "10.2.4",

and it looks like EnsureDaemonRunning.ts is loading the package.json file and using that for version.
```

## What

Ideally we inject the side channel version into the UI - but for making the side channel work, this is the quickest solution

Side channel action to test this:
https://github.com/opticdev/optic/actions/runs/1124601725

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


niclim:
Merging into develop - tested the side channel and this looks good right now - will spend some time later figuring out why the UI isn't getting the alpha version tag in the side channel

PR #1112, opened 2021-08-12
Connect redux actions to shape editor

niclim:
## Why

New shape editor contains a bunch of actions that needed to be hooked up. This included:
- contributions
- Removing fields
- Editing field types

## What

- Connects up these components from the endpointRootPage - exposed through components + functions
- Fixed a bunch of misc react key errors
- Updates the isRemoved state (it's currently just a strikethrough - but theres likely better styling we can add in here)

Field that is the root removed field
Screen Shot 2021-08-12 at 7 17 13 PM

Field that is a child of a removed field
Screen Shot 2021-08-12 at 7 17 15 PM

Things to follow this:
- Query Parameters can't be edited right now - query parameters can also only be set as optional, instead of nullable - since we can never infer nullable query parameters. 
- Add a loading indicator to the simulated body component

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/ui-v2/src/pages/docs/EndpointRootPage.tsx
+++ b/workspaces/ui-v2/src/pages/docs/EndpointRootPage.tsx
@@ -105,6 +105,33 @@ export const EndpointRootPage: FC<
   const unremoveEndpoint = () =>
     dispatch(documentationEditActions.unremoveEndpoint({ method, pathId }));
 
+  const onChangeFieldType = useCallback(
+    (
+      fieldId: string,
+      requestedFieldTypes: Set<JsonType>,
+      isFieldTypeDifferent: boolean
+    ) => {
+      if (isFieldTypeDifferent) {
+        dispatch(
+          documentationEditActions.addFieldEdit({
+            fieldId,
+            options: {
+              isOptional: requestedFieldTypes.has(JsonType.UNDEFINED),
+              isNullable: requestedFieldTypes.has(JsonType.NULL),
+            },
+          })
+        );
+      } else {
+        dispatch(
+          documentationEditActions.removeFieldEdit({
+            fieldId,
+          })
+        );
+      }
+    },
+    [dispatch]
+  );
+
   const onFieldDescriptionChanged = useCallback(
     (fieldId: string, description: string, isDescriptionDifferent: boolean) => {
       if (isDescriptionDifferent) {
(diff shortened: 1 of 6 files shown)


workspaces/ui-v2/src/pages/docs/components/ShapeEditor.tsx
```diff
@@ -58,34 +85,89 @@ export const ShapeEditor: FC<{
...
   selected,
   onChangeDescription,
+  onChangeFieldType,
   onSelect,
+  isFieldRemoved,
+  onToggleRemove,
 }) {
   const classes = useStyles();
+  const initialDescription = field.contribution['value'];
+  const initialFieldTypes = getInitialTypes(field);
```
niclim, line 126: We need the initial description and field types to compare to the current input (such that we know if we need to reset and remove state) - this is the simplest way to achieve this and isn't a huge amount of extra work

niclim:
Planning on merging this in tomorrow, but will address and post-vacation comments you have!

workspaces/ui-v2/src/pages/docs/components/ShapeEditor.tsx
```diff
@@ -149,32 +231,21 @@ const FieldEditor: FC<{
...
+              currentJsonTypes.includes(editableType) ? 'contained' : 'outlined'
+            }
+            startIcon={
+              currentJsonTypes.includes(editableType) ? <CheckIcon /> : null
+            }
+            onClick={onClickTypeButton(editableType)}
+          >
+            {editableType === JsonType.UNDEFINED ? 'Optional' : editableType}
+          </Button>
+        ))}
```
JaapRood, line 248: I intentionally kept this code wet, as the UI for this is a place holder, and being explicit is well worth it over being dry, especially this far as the edge of the UI. So you might see this undone in the near future 😅.

workspaces/ui-v2/src/pages/docs/EndpointRootPage.tsx
```diff
@@ -105,6 +110,80 @@ export const EndpointRootPage: FC<
...
+      if (userRemovedFields.includes(fieldId)) {
+        return 'root_removed';
+      } else if (allRemovedFields.has(fieldId)) {
+        return 'removed';
+      } else {
+        return 'not_removed';
+      }
+    },
+    [userRemovedFields, allRemovedFields]
+  );
```
JaapRood, line 124: This totally works, but I wonder why pass down a function rather than just an `removedFields` array / set (or something similar)?  My first thought maybe this was a performance optimisation, but this will be called for every field regardless, so that's probably not right. Curious to know!

JaapRood (commented):
Looks great, thanks for hooking that all up and keeping `ShapeEditor` and its children as unconnected components  🎉 😊.

workspaces/ui-v2/src/pages/docs/components/ShapeEditor.tsx
niclim: That's completely fine - I think the reason I update this is because there's other variables to make this work (i.e. filtering our specific keys) - so moving these to a single place.

I agree though, some places it's better to be explicit, and this might be one of them.

workspaces/ui-v2/src/pages/docs/EndpointRootPage.tsx
niclim: I think there's the same number of calls per field? It's just the consumer of this specifies the behavior. Not a huge amount of reasoning why this over passing in the removedFields array / set - so feel free to update later on

JaapRood: Fair enough, good to know whether you feel passionately or not. 

I don't see any risk or adverse performance effects from it (especially as we're also passing callbacks, so re-render optimisations are already limited somewhat), so no concern there. I've just tended to stay away from this pattern as it makes it a bit harder to debug by logging or inspecting props through the debugger. I do like not having to pass in weird data structures, though, so 🤷‍♂️. Lets live with it and find out!

PR #1113, opened 2021-08-13
Query parameter field editing with shapeviewer

niclim:
## Why

## What

- Changed `selectedFieldId` and `setSelectedFieldId` to be non-optional as it powers the functionality of the shape editor
- Moved out array splicing logic into a selector to use in both the old and new versions
- Connected up the query parameter edit field
- Removed the ability to make a query parameter nullable from the shape editor

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...


JaapRood (commented):
👏

PR #1117, opened 2021-08-13
Build FieldViewer component

niclim:
## Why

We have a ShapeEditor component - but we need a non-editable version.

After we:
- launch this feature + remove the FF
- Modify PathParameters to not use FieldOrParameter
- Modify the endpoint name, and body description components to not use FieldOrParam

We can remove all instances of FieldOrParam, EditableTextField, etc

## What

- Created a FieldViewer component, which is a non-editable component that renders the field and contributions
- Moved summarizeTypes into a component file and fixed the display of 2 object items (i.e. `object, or array` now displays as `object or array`)

This is behind the `REACT_APP_FF_FIELD_LEVEL_EDITS` FF.

## Validation
* [ ] CI passes
* [ ] Verified in staging
* etc...

### DIFF
--- a/workspaces/ui-v2/src/components/ShapeTypeSummary.tsx
+++ b/workspaces/ui-v2/src/components/ShapeTypeSummary.tsx
@@ -7,10 +7,18 @@ import * as Theme from '<src>/styles/theme';
 export const ShapeTypeSummary: FC<{
   shapes: IShapeRenderer[];
   required: boolean;
-}> = ({ shapes, required }) => {
+  hasColoredFields?: boolean;
+}> = ({ shapes, required, hasColoredFields }) => {
   const optionalText = required ? '' : ' (optional)';
   const components = shapes.map(({ jsonType }: { jsonType: JsonType }) => (
-    <span key={jsonType} style={{ color: Theme.jsonTypeColors[jsonType] }}>
+    <span
+      key={jsonType}
+      style={{
+        color: hasColoredFields
+          ? Theme.jsonTypeColors[jsonType]
+          : Theme.GrayText,
+      }}
+    >
       {jsonType.toString().toLowerCase()}
     </span>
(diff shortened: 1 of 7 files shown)


workspaces/ui-v2/src/components/ShapeTypeSummary.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { IShapeRenderer } from '<src>/types';
+import * as Theme from '<src>/styles/theme';
+
+export const ShapeTypeSummary: FC<{
+  shapes: IShapeRenderer[];
+  required: boolean;
+}> = ({ shapes, required }) => {
+  const optionalText = required ? '' : ' (optional)';
+  const components = shapes.map(({ jsonType }: { jsonType: JsonType }) => (
+    <span key={jsonType} style={{ color: Theme.jsonTypeColors[jsonType] }}>
```
JaapRood, line 13: I'm not sure we want to render type colours of types in the viewer mode. The colour shift is one of the key differences we rely on to clearly differentiate the editor from the viewer, making it appear interactive rather than the same static UI they see in documentation view. It _might_ hold up without that differentiation, but as this is a new interaction paradigm we're introducing, I'd rather differentiate too much rather than too little.

JaapRood (approved):
Looks good generally, brining the viewer component into a similar implementation style as the editor and getting rid of quite a bit of cruft. Visualising what I can from looking at the mark up there's a small tweak I'd make appearance wise, but happy to address that later on in a final UI / UX pass before we ship.

workspaces/ui-v2/src/components/ShapeTypeSummary.tsx
niclim: Good point, I've reverted / made the color changes only in the edit view

PR #1122, opened 2021-08-17
Add evidence blog post

smizell:
This is PR adds the blog post for making design first and code first work for everyone. I'd love any help reviewing, especially:

* The title. I know there's something better.
* Graphic ideas.
* Trimming any of it down if it's too long/wordy.


LouManglass:
I like the approach - it does a good job of introducing proposals and evidence, and sets up a strong case for adopting incrementally through a forward-only approach. 

I added a truncate tab to give a blurb for the blog page, feel free to move that around if you think the splash should be different. I also added a few tweaks to some spelling/language.

smizell:
@LouManglass thanks a ton. I appreciate that and the editing help.

PR #1125, opened 2021-08-18
Fix bug where fields with unknown shapes can't be edited

JaapRood:
## Why
The possibility exists for a field to be learned as `null` and without any information of the non-null shape. The way editing commands generation has been implemented currently, does not allow for this, not generating any commands at all.

## What
Tests have been added to verify the limitation in the command generation. 

The root of the issue stemmed from the `list_trail_choices` not returning any results for `Unknown` shapes, not giving us the `shape_id` of the shape we are trying to wrap with nullables / optionals. However, the diff and choice mapping code relies on that fact, for which an unknown choice isn't a choice at all. 

As a fix, `list_trail_choices` now returns a result for `Unknown` shapes, and a new `list_known_trail_choices` is introduced for the diffing and spec choice mapping, where unknown choices are filtered out. That allows the resolving of trail choices to remain in the single place and power all our choice related functionality.

## Validation
* [ ] CI passes
* [x] New tests pass
* [x] Validated end-to-end against test project

### DIFF
--- a/workspaces/optic-engine/src/queries/shape.rs
+++ b/workspaces/optic-engine/src/queries/shape.rs
@@ -62,7 +62,6 @@ impl<'a> ShapeQueries<'a> {
           _ => unreachable!("expected to be a core shape node"),
         };
         let trails: Vec<ChoiceOutput> = match core_shape_node.descriptor.kind {
-          // ShapeKind::UnknownKind => vec![],
           ShapeKind::NullableKind => {
             let nullable_parameter_id = core_shape_node
               .descriptor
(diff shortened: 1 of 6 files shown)


workspaces/optic-engine/src/queries/shape.rs
```diff
@@ -62,7 +62,7 @@ impl<'a> ShapeQueries<'a> {
           _ => unreachable!("expected to be a core shape node"),
         };
         let trails: Vec<ChoiceOutput> = match core_shape_node.descriptor.kind {
-          ShapeKind::UnknownKind => vec![],
+          // ShapeKind::UnknownKind => vec![],
```
niclim, line 65: Should this be uncommented?

niclim (approved):

workspaces/optic-engine/src/queries/shape.rs
JaapRood: It should be removed!

PR #1126, opened 2021-08-18
Revert "Deps/graphql"

LouManglass:
Reverts opticdev/optic#1099


LouManglass:
@niclim FYI, the peerdepenency upgrade broke yarn installs. I'm reverting for now so we're in a good state for builds/deploy, and I wanted to follow up on this commit below: it looks like we not only fixed typing, but perhaps also cleaned up some copypasta which we might still want to preserve. So, note, these changes are no longer in develop:

https://github.com/opticdev/optic/pull/1099/commits/74526a372c538db772699f4d6771a3377ac736d8

LouManglass:
Quick check against https://github.com/opticdev/optic/pull/1099#issuecomment-900941332, forgot to make sure I had a clean env, but we should be merging soon.

PR #1141, opened 2021-10-06
Typo fix in 'remember'

rogerd330:
## Why
Reading the documentation as a new user and stopped a typo

## What
Missing letter in 'remember'.

## Validation
Re-read after the change.


acunniffe:
Good find, thank you @rogerd330! Anything else in the docs or project that's confusing which we could help with?

LouManglass:
Thanks @rogerd330 , I've added credit in #1143. We'll get it up in a bit.

PR #1151, opened 2022-01-06
website: remove interactive demo

LouManglass:
## Why

Removing interactive demo for Optic 10

## What

- Interactive demo removed from landing page, docs first load page
- References to interactive demo removed

## Validation

* [ ] CI passes
* [ ] Verified in staging


devdoshi (approved):
Looks good - but I didn't check to see if there are other references to it.

LouManglass:
> Looks good - but I didn't check to see if there are other references to it.

It should be fully inactive on the site, which I checked on the staging pages, but I did not remove some of the plumbing (such as the functions to create the interactive demo). I can do a sweep for dead code, because there's undoubtedly some.

I also removed some of the demo container build tasks as part of cleanup.

LouManglass:
Letting this last build/check complete, I cleaned up dead interactive demo code in response to PR as well.

PR #1276, opened 2022-08-03
chore: rm everything ahead of repo migration

notnmeyer:
We're planning on moving primary development of our open source components back to this repo. To kick that off, we're going to bring over the history from another repo (very soon).

This PR sets us up for that migration.

PR #1279, opened 2022-08-03
Revert "chore: rm everything ahead of repo migration"

notnmeyer:
Reverts opticdev/optic#1276


JaapRood (approved):
Man doesn't like force pushes and goes revert, can respect it

PR #1285, opened 2022-08-03
chore: add dependabot

notnmeyer:
adds the equivalent dependabot config from monorail

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,18 @@
+version: 2
+updates:
+  - package-ecosystem: npm
+    directory: /
+    allow:
+      - dependency-type: production
+    schedule:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,18 @@
+version: 2
+updates:
+  - package-ecosystem: npm
+    directory: /
+    allow:
+      - dependency-type: production
+    schedule:
+      interval: monthly
+    rebase-strategy: auto
```
notnmeyer, line 9: this was previously `disabled`, preventing automatic rebasing when the default branch changed. since this repo should be lower volume than monorail, its probably fine to enable.

niclim (approved):

PR #1293, opened 2022-08-03
Update dependencies

niclim:
Updates https://github.com/opticdev/optic/pull/1290 https://github.com/opticdev/optic/pull/1289 and https://github.com/opticdev/optic/pull/1288

Realized we were using NodeClient instead of init, which had breaking changes. Updated to use `init` instead of `NodeClient` (which I did for a reason but I can't remember why)


JaapRood (approved):
I remember running into the same with `openapi-cli` and also forgot why it was exactly 😅

PR #1297, opened 2022-08-04
feat: GitHub spec resolver

niclim:
This PR:
- Factors out common spec resolution code
- Typescriptifies some of the insourced code
- Creates a github spec resolution 

TODO:
- [x] Do some testing with a pre-release for an end to end flow here 
  - tested running a couple of `optic diff` on specs that had things to resolve against - looks like this is still working for the older cases


acunniffe (approved):

acunniffe:
Reviewed 
- Factors out common spec resolution code
- Creates a github spec resolution

Looks correct to me for public repositories. 

For private repositories this is missing a way to set the token, so I'd imagine we'd factor the code in a way where that was possible.

acunniffe:
You should also write a test where there's a circular `$ref` to make sure it still quits on that. Otherwise we'll hit rate limits

niclim:
> For private repositories this is missing a way to set the token, so I'd imagine we'd factor the code in a way where that was possible.

You pass in an octokit instance, you can instantiate an octokit instance with different authentication methods.

As for tests, I need to figure out how to:
- add a test for git repos
- add a test for github instances

Circular refs should work, but i'll see what I can do with adding a test

acunniffe:
Great re: octokit, I wasn't aware how those credentials worked. Thanks.

niclim:
Ok I added a test that uses a stubbed external file resolver. To actually test the github resolver and git file resolver, we'd need to setup a:
- github app / token and make requests with that
- a git fs mock

these are both better suited as integration tests

PR #1298, opened 2022-08-05
bump version

niclim:


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "openapi-workspaces",
   "private": true,
-  "version": "0.27.0",
```
niclim, line 4: @JaapRood it looks like `yarn run release --prerelease` defaulted to `0.26.29-0` - is this because of the stableVersion in the package.json? 

It looks like you bumped `0.27.0` before - does stable version need to be bumped manually?

I'll bump the version manually for the prerelease, but just curious what we're missing from this flow

JaapRood: I'm confused with what's going on here. I did notice that `yarn run version prerelease` doesn't increment the prerelease counter properly, yet to figure out why

niclim: I think it uses the stable version - which I am not sure how it was set in the first place. If it's part of the yarn version flow or manually updated

JaapRood: Ahh yeah I just ran into the same thing myself. When you prerelease using `yarn version` it adds `stableVersion`. It must have not been removed in a subsequent versioning.

PR #1299, opened 2022-08-08
Fix input bugs for new `oas` commands and tweak output

JaapRood:
With the release of `oas` and its new commands, we've found some bugs feeding it different types of input.

- Fixed running `oas <spec-file>` without further command. Now includes instructions and help on commands that can be used.
- `oas new` now accepts `.json` filenames, and cleanly warns about other formats we don't support.
- correctly configures analytics and error tracking, which got lost in the move back from closed to open source repos.
- provides usage instructions for `add` to end of `status` output: 
Screenshot 2022-08-08 at 16 01 07

### QA:
 - [x] run the described scenarios above and observe how there's no more nasty stack traces.


niclim (approved):

JaapRood:
@acunniffe snuck in your suggestion for `add` advertising in the output of `status`:

Screenshot 2022-08-08 at 16 01 07

PR #1301, opened 2022-08-11
adding range change rules in breaking-changes

acunniffe:
Addressing a request for additional breaking-change rules here:
- https://github.com/opticdev/optic/discussions/1268
- https://github.com/opticdev/optic/discussions/1264

For `number` or `integer` properties that have`maximum` or `minimum` keywords set
- expanding the size of the range (maximum increases) is breaking
- expanding the size of the range (minimum decreases) is breaking

For `array` properties that have `maxItems` or `minItems` keywords set
- increasing the upper bound (maxItems) is breaking
- decreasing the lower bound (minItems) is breaking

These rules will not be triggered if you add these keywords to properties that do not currently use them.


projects/standard-rulesets/src/breaking-changes/preventArrayItemLengthChanges.ts
```diff
@@ -0,0 +1,77 @@
...
+  arrayMaxItemsShouldNotIncrease,
+  arrayMinItemsShouldNotDecrease,
+  numericMaximumRangeShouldNotExpand,
+  numericMinimumRangeShouldNotExpand,
+} from './helpers/rangeChange';
+
+export const preventRequestArrayLengthChanges = new RequestRule({
+  name: 'prevent request array length changes',
+  rule: (requestAssertions) => {
+    requestAssertions.body.changed((before, after) => {
```
acunniffe, line 16: @niclim is this correct? I want both of these to run independently of each other, ie if Max throws, min still runs.

projects/standard-rulesets/src/breaking-changes/preventParameterTypeChange.ts
```diff
@@ -1,7 +1,7 @@
 import { OperationRule, RuleError } from '@useoptic/rulesets-base';
 import { getOperationAssertionsParameter } from './helpers/getOperationAssertionsParameter';
 import { ParameterIn } from './helpers/types';
-import { didTypeChange } from './helpers/type-change';
+import { didTypeChange } from './helpers/typeChange';
```
acunniffe, line 4: I renamed this helper file to be consistent with other filenames in the repo

projects/standard-rulesets/src/breaking-changes/preventArrayItemLengthChanges.ts
```diff
@@ -0,0 +1,77 @@
...
+  RequestRule,
+  ResponseBodyRule,
+  RuleError,
+} from '@useoptic/rulesets-base';
+import {
+  arrayMaxItemsShouldNotIncrease,
+  arrayMinItemsShouldNotDecrease,
+  numericMaximumRangeShouldNotExpand,
+  numericMinimumRangeShouldNotExpand,
+} from './helpers/rangeChange';
```
niclim, line 11: nit: Can we move these functions into this file? The implementations aren't generic enough to need to be used elsewhere since they contain both the logic of whether a schema has a range increase, and the rule error

projects/standard-rulesets/src/breaking-changes/index.ts
```diff
@@ -67,6 +75,10 @@ const breakingChangesRules = [
   preventResponsePropertyRemoval,
   preventResponsePropertyTypeChange,
   preventResponseStatusCodeRemoval,
+  preventRequestPropertyRangeChanges,
+  preventResponsePropertyRangeChanges,
+  preventRequestArrayLengthChanges,
+  preventResponseArrayLengthChanges,
```
niclim, line 81: How should we handle new rules. What you've got here will start failing on things that weren't failing before. 

We have two options:
- We expand the breaking change definition as part of upgrading, and you can turn off rules as you upgrade
- We can "freeze" the current set of breaking change rules, and you can opt into extra breaking change rules. When we "major version" we can turn on all breaking changes.

projects/standard-rulesets/src/breaking-changes/preventArrayItemLengthChanges.ts
```diff
@@ -0,0 +1,77 @@
...
+        after.value.flatSchema
+      );
+    });
+
+    requestAssertions.body.changed((before, after) => {
+      arrayMinItemsShouldNotDecrease(
+        before.value.flatSchema,
+        after.value.flatSchema
+      );
+    });
```
niclim, line 28: I don't understand this - why is expanding the range a breaking change? Shouldn't it be the opposite?

projects/standard-rulesets/src/breaking-changes/__tests__/range-changes.test.ts
```diff
@@ -0,0 +1,173 @@
+import { OpenAPIV3 } from '@useoptic/openapi-utilities';
+import { TestHelpers } from '@useoptic/rulesets-base';
+import { BreakingChangesRuleset } from '../index';
+
+describe('range changes', () => {
+  test('numeric range expanded', () => {
```
niclim, line 6: It feels like we're missing test coverage here with different variants (increase / decrease, numeric / array, request / response). We can factor our a lot of the common test logic using `test.each` to handle a lot of the different variants of test. Example of this https://github.com/opticdev/optic/blob/main/projects/standard-rulesets/src/breaking-changes/__tests__/parameter-enum-breaking-changes.test.ts#L6-L8

projects/standard-rulesets/src/breaking-changes/preventArrayItemLengthChanges.ts
niclim: yes both will run

projects/standard-rulesets/src/breaking-changes/preventArrayItemLengthChanges.ts
acunniffe: to confirm, just the array ones, right? The numeric ones are used in multiple places

projects/standard-rulesets/src/breaking-changes/index.ts
acunniffe: That's worth a longer convo on. These are change rules, so nothing that was passing will fail. But subsequent classes of changes you make might start failing

projects/standard-rulesets/src/breaking-changes/preventArrayItemLengthChanges.ts
niclim: It looks like the numeric ones aren't used in this file.

projects/standard-rulesets/src/breaking-changes/preventArrayItemLengthChanges.ts
acunniffe: If I'm expecting at least 3 items in the array -- let's say to make a "Top Destinations" view, I don't care if that list can have 5, 10, 20 items because I'm pulling the top 3. 

If the minimum goes down that breaks me...? maybe. 

it's a weird use case. I don't think most people use these keywords, but those who do are probably building expectations around the min / max numbers into their code.

projects/standard-rulesets/src/breaking-changes/helpers/rangeChange.ts
```diff
@@ -0,0 +1,100 @@
...
+      if (before.maximum < after.maximum) {
+        throw new RuleError({
+          message: `range maximum increased: ${before.maximum} -> ${after.maximum}`,
+        });
+      }
+    }
+  }
+}
+
+export function numericMinimumRangeShouldNotExpand(
```
savage-alex, line 35: Would ShouldNotReduce be a more accurate function name?

projects/standard-rulesets/src/breaking-changes/helpers/rangeChange.ts
```diff
@@ -0,0 +1,100 @@
...
+  after: OpenApi3SchemaFact
+) {
+  if (isNumericProperty(before, after)) {
+    if (
+      typeof before.maximum !== 'undefined' &&
+      typeof after.maximum !== 'undefined'
+    ) {
+      if (before.maximum < after.maximum) {
+        throw new RuleError({
+          message: `range maximum increased: ${before.maximum} -> ${after.maximum}`,
```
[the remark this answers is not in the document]
savage-alex: this looks fine to me for max range increase. Is that a breaking change for clients if they were to receive more than they are expecting? As a server I would be fine with it but I belive a client might not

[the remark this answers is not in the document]
savage-alex: Error messages are clear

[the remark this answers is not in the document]
acunniffe: @savage-alex so it sounds like we need to branch the logic here depending on whether the range is in the request or response?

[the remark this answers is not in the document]
acunniffe: @savage-alex so it sounds like we need to branch the logic here depending on whether the range is in the request or response?

projects/standard-rulesets/src/breaking-changes/helpers/rangeChange.ts
acunniffe: Yeah, good callout. I'll update before merging

projects/standard-rulesets/src/breaking-changes/helpers/rangeChange.ts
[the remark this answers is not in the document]
savage-alex: I think so. Depending on the direction it would hurt the server or the client. Even if we called it out to let someone know they can then think about whether its desirable or compatible or not.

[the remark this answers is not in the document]
acunniffe: @savage-alex  -- I talked to some other teams and I think the expected behavior here for request / response is a bit different depending on the team. I updated the code, and they currently both will run, but if you want to turn one of them off you could exempt the rules by name `prevent response property numeric range changes` or  `prevent request property numeric range changes`

PR #1302, opened 2022-08-12
Wrap user errors 

niclim:
These functions call are dependent on a user's FS - we don't need to be notified via sentry for user errors here since they're feedback errors rather than application errors

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,5 @@
 node_modules
 build
 .nyc_output
+
+.env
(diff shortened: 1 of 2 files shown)


JaapRood (approved):
Seems to fit in what's already been established in the rest of the package, throwing a `UserError` and handling that specifically. Zooming out a teensy bit, I'd say the preferred approach, in my eyes, would be returning a `Result` type here (e.g. from `ts-results`), indicating an expected fallible result, so every _thrown_ error can be designated as a panic and lead to an instant crash and logging by Sentry. Mostly because catching and handling errors in Javascript and with the GC can lead to unknown app states and funky bugs, generally way before you imagine it would.

PR #1322, opened 2022-08-24
Feat/specify which config is used

Alexandre-Herve:
Specify that local config is being used in `optic diff` and  `README.md`

### DIFF
--- a/projects/optic/src/commands/diff/diff.ts
+++ b/projects/optic/src/commands/diff/diff.ts
@@ -186,10 +186,10 @@ const getBaseAndHeadFromFiles = async (
   try {
     return Promise.all([getFileFromFsOrGit(file1), getFileFromFsOrGit(file2)]);
   } catch (e) {
-    console.error(e)
+    console.error(e);
     throw new UserError();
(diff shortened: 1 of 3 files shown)


projects/optic/src/commands/diff/diff.ts
```diff
@@ -241,6 +241,9 @@ const runDiff = async (
     }
 
     logComparison(specResults, { output: 'pretty', verbose: false });
+
+    console.log('');
+    console.log(`Configure check rulesets in your local optic[.dev].yml file.`);
```
niclim, line 246: I think it might be easier for us to just name this as `optic.dev.yml` - we'll be throwing in some deprecation warnings later for optic.yml, although I dont think it matters that much -

niclim (approved):

projects/optic/src/commands/diff/diff.ts
Alexandre-Herve: Yeah actually I went on changing things and adding the deprecation warning in the next pr, so I can change it now

PR #1323, opened 2022-08-24
refactor(optic): have init create an optic.dev.yml file

Alexandre-Herve:
- have init create an optic.dev.yml file instead of optic.yml
- use `optic.dev.yml` everywhere in doc by default
- add deprecation warning when accessing `optic.yml`

### DIFF
--- a/projects/optic/src/commands/init/init.ts
+++ b/projects/optic/src/commands/init/init.ts
@@ -117,7 +117,7 @@ export const getInit =
           'File IDs are stable identifiers for your API specifications that will appear in Optic.'
         );
         console.log(
-          'You can change them now before you check in the optic.yml file.'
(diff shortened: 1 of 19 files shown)


projects/optic/src/commands/init/init.ts
```diff
@@ -60,7 +60,7 @@ export const getInit =
     }
 
     const gitRoot = await getRootPath();
```
niclim, line 62: With the optic.dev.yml file, we dont need to do file detection as part of init.

niclim (commented):
This PR looks good, except we shouldn't merge this until our new cloud stuff launches, since init will generate a optic.dev.yml, which the github action flow doesn't respect

niclim (approved):

PR #1324, opened 2022-08-24
feat: generate self-signed CA and use it for proxy

JaapRood:
The first step in supporting TLS for the capturing proxy. 

TODO: expand on this

### DIFF
--- a/projects/openapi-cli/package.json
+++ b/projects/openapi-cli/package.json
@@ -64,6 +64,7 @@
     "json-schema-traverse": "^1.0.0",
     "mockttp": "^3.0.0",
     "node-abort-controller": "^3.0.1",
+    "node-forge": "^1.2.1",
     "node-machine-id": "^1.1.12",
     "ora": "^6.1.2",
     "semver": "^7.3.7",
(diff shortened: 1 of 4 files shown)


JaapRood:
We've got a proxy with a self generated CA working, verified through a new test with the live proxy. Next up is the mechanism that stores the certificate and private key for use across runs.

JaapRood:
I added the mechanism that generates and stores the CA when running `oas capture --proxy` (`oas update --proxy` will be supported later). I've also added the ability to store the cert to a file of your choosing using the `--ca <filepath>` option. I'm not 100% sure whether that's the right way to do it (it's own command might be cleaner and make more sense), but I just wanted the ability myself to able to test it with Postman. There, when you manually add the CA cert through their settings menu, you can proxy and capture HTTPS requests successfully.

As the method of storing the cert and private key, I've chosen to use [`conf`](https://www.npmjs.com/package/conf) for now. It stores the values in a file, in the right place and does so atomically, and supports JSON Schema to validate shapes (using AJV, which we already depend on anyway), as well as migrations between versions. Eventually, we'll want to use the system's secure store at least the private key, like KeyChain on macOS, but that's something we'll look into separately. For now, the CA's short expiry and the fact you have to manually configure it should make it okay the private key can be found to those who know where to look.

With all of that, this PR is pretty much feature complete. I'll take off some of the rough edges and properly document what we're doing, so we can move towards merging it.

niclim (approved):

PR #1326, opened 2022-08-26
Chore/housekeeping remove unused code

niclim:
Removes some unused code in openapi workspaces including:
- openapi-io - watchDependencies, and roundtrip reconciler
- openapi-utilities - the `render` folder, which I'm not entirely sure what it's for, but we don't use it.


JaapRood (approved):
Double checked we're clear of all the reconcilers in `openapi-io`. The last thing these powered were the `openapi-diff-experiment`, which we got rid off. The rest that was applicable made it to `openapi-cli`.

Issue #1320, opened 2022-08-23
Bring back `addedOrChanged` listener for custom rules


acunniffe:
API Enablement teams I've been working with to define their own custom Optic rules have been duplicating a lot of code. 

The rules engine had separate listeners for `added` and `changed` parts of the OpenAPI file. The intent of most of these teams is to write rules that applies to new surface area only. They want everything already in their specs to be allowed because changing them will break consumers. 

Instead of writing the same rule as `added` and then again as `changed` -- requests have come in for `addedOrChanged`. We've had this before but no one was using it. Looking back that's because all those teams were building APIs that were net-new. For legacy APIs, this is much more prominent and we should bring it back. Smarter automated governance was a big part of the value here.

PR #1327, opened 2022-08-26
Support `addedOrChanged` lifecycle in `rulesets-base`

niclim:
Addresses: https://github.com/opticdev/optic/issues/1320

Add support for addedOrChanged assertions on all rules. This can be used as a lifecycle to assert changes wherever added or changed would trigger.


projects/rulesets-base/src/rule-runner/assertions.ts
```diff
@@ -77,13 +77,19 @@ class AssertionRunner<T extends AssertionType> implements Assertions<T> {
...
     const registerAssertion = (...args) => {
       const assertion = args[1] || args[0];
       if (assertionKey === 'changedAssertions') {
         this.changedAssertions.push((before, after) => {
           assertion(after);
         });
+      } else if (assertionKey === 'addedAndChangedAssertions') {
+        this.addedAssertions.push(assertion);
+        this.changedAssertions.push((before, after) => {
+          assertion(after);
```
acunniffe, line 91: @niclim  -- this is the part that's really most consequential right? `addedOrChanged` rules always get the `after`? 

That seems like the most useful behavior to me. You could write the rule to assert the final state of something in the spec. 

The one *weird thing that may happen is that the rule triggers when you change something completely unrelated. Like I change description and now it's enforcing the name of an existing property. That can happen with changed rules too, but you can be smarter about it since you have both. 

I don't think it's a problem that should block this, but we should keep our ears open

acunniffe (approved):

projects/rulesets-base/src/rule-runner/assertions.ts
niclim: Yes, you care about the final state of the spec. I think if you did want to enforce more specific change assertions, you would split out added and changed into separate assertions.

This is really a shorthand of what you would be doing for added or changed.

PR #1329, opened 2022-08-30
Adapt proxy to work as browser proxy by transparently tunnelling non-targeted hosts

JaapRood:
Thus far, the proxy only worked as a reverse proxy: all traffic sent to it would be forwarded to the target host. This made it impossible to configure it as a proxy directly in browser settings (or even system wide), and required clients to be configured  directly.

This PR adapts the proxy to transparently forward all traffic for non-targeted hosts, completely untouched, directly to its original destination. Only traffic for the targeted host is redirected for decoding, parsing and capturing. These changes allow the proxy to be set as a browser or system HTTP(S) proxy.

To achieve this, we spin up an additional server that responds to the http `CONNECT` method, which is what browsers (in addition to other http clients) use to tunnel TLS connections through http proxies. The only thing visible to the optic proxy is the hostname, and a TCP socket the client will start to use as a tunnel if the proxy sends a 200-299 ranged response. Only upon the host matching the provided target host does the proxy pipe the socket to itself, instead of piping it to it's original destination.

Previous usage of sending traffic directly to the proxy, rather than tunnelling, is still supported.

The `-d` option is added to `oas capture` to enable debug logging output. When used with `LOG_LEVEL=info` env variable, tunnelling / capturing behaviour of each connection can be observed.

There are some improvements that can be be made that we'll look into separately: 
- http2 support (overwhelming majority of http2 clients downgrade to 1.1 automatically)
- increase test coverage for transitive error handling

### QA

- [x] `HTTP_PROXY` with `newman` sees traffic and forwards TLS, only failing TLS cert (when enabled) for target host
- [x] Postman manual proxy config can proxy through, only failing TLS certification (when enabled) for target host
- [x] Firefox manual proxy config can proxy through, only failing TLS certification for target host
- [x] Safari and Chrome can proxy through when system proxy set, only failing TLS certification for target host
- [x] all other apps continue working correctly when system proxy set, only failing TLS certification for target host
- [x] no crashes when used as system proxy on work machine for ~48 hours.


JaapRood:
I got a drafted version working to a reasonable degree for TLS traffic: while set as a system proxy, browsers manage to tunnel all kinds of traffic through it, as well as other apps that respect these settings. For the targeted host, a TLS breach is detected (as expected as cert isn't in system chain), but when continuing on, the socket is correctly forwarded to the capturing proxy and interactions captured.

From my limited tested so far, there's a bunch of issues to resolve:
- Direct traffic forwarding isn't implemented yet, rendering all HTTP (non-TLS) calls to a 400.
- Sockets aren't cleaned / closed up after completion, causing the process to stay open and proxy to live on.
- Configuring directly as a proxy in some browsers / clients doesn't work yet (not sure why, maybe because of the HTTP calls).

JaapRood:
In the version I just pushed we're now dealing with socket clean-up a whole lot better, allowing the proxy (and by extension the capture command) to finish. 

My first try was to use `socket.unref()`, which allowed the command to finish (with sockets no longer being considered as a reason to keep the process open), but didn't actually close all the connections. Just calling `proxy.close()` wouldn't work either, as the proxy would simply way for all the connections to end.

In the end, I introduced a second `AbortController`, so we could tear down the proxy with precision:
1. stop the capturing proxy, making sure none of those request / responses are interrupted and can be written cleanly.
2. (could actually be concurrent with step 1 🤔) stop accepting new connections with transparent proxy
3. close all open tunnels that haven't been closed yet.

For the last, I looked at doing it the "right" TCP way using RST, but the only real practical way I see for that has only landed in Node v16.7 (`socket.resetAndDestroy`), so instead we just destroy the socket outright and let the other side figure that out for themselves. Once v14 is out of LTS, we can clean that up.

Next up, I'll look at proxying direct HTTP calls.

JaapRood:
The proxy now deals with HTTP 1.x (non-tis) calls directly. After trying a coupe of different approaches, it was once again confirmed to me how tricky correctly forwarding traffic is, so I went with the direction of tying it directly into the mockttp capturing proxy. That's fully setup to handle this and has been stabilised over a lot of time to handle pretty much anything you throw at it. 

While this approach works and seems pretty robust, it _does_ mean the capturing proxy will get to see some non-target traffic. Realistically, though, if you're using plain old HTTP 1.x without TLS, that's kind of what you're signing up for (and why browsers nowadays actively mark these sites as insecure). We're still only capturing interactions that match the target host, passing along everything else untouched.

Now this is implemented, the proxy seems to work well with every client I've tried so far.

I also had a look at HTTP/2 support. Mockttp can deal with it, but our transparent proxy would require manual implementation for that. Thanks to ALPN protocols, as good as every client / user-agent out there falls back to HTTP/1.1 with TLS, which means we have an opportunity to cut that from scope for now and address that separately.

With that, we're functionally where we need to be, and we can start to focus on stabilising and testing it. I'll look at that next.

JaapRood:
In the version I just pushed we're attempting to very precisely handle errors in the tunnelling of requests to the origin server that are transitive and part of normal socket operations. 

- The server will respond with a `502` or `504` to the `CONNECT` request if it can't connect to the requested host. This includes things like DNS not resolving, but also network timeouts. 
    - Browsers differ in how they handle this, but at least all of them seem to realise the connection they are hoping for never materialised. Maybe sending some content (html?) back could help, but I don't think we have to worry too much.
- Any interruptions in the pipe between the two sockets, because of a timeout or because someone hangs up, are handled without notifying the user or any further logging (at least until we add a logging system that's can deal with such verbosity). They are part of the inter webs, and something we can simply recover from by closing the sockets involved.

Next up, I'm going to look at making sure we're starting the transparent proxy on an available port.

JaapRood:
I pushed a version that sets up the transparent proxy start properly, with eventual recovery of used ports, iterating until a usable port is found. I've also introduced a basic logging facility; we want to be able to give user insight in what the proxy is doing exactly, especially given the system-level scope it could be used on. On top of that, it'll be the main way we'll be able to deal with any stability issues.

With this, the proxy is looking good stability and functionality wise. The one real thing that's missing, before we roll this out, is automated testing. That's a little bit more tricky than normal, given all the I/O, but we'll at least want to cover the basics. That's probably also a decent time to slightly clean up code-org wise: it's all super hacky in the one function that is hard to follow, even for me having written it very recently. Even the most simple step of wrapping that the transparent proxy into single entity should greatly improve things.

JaapRood:
Note to self: `EADDRNOTAVAIL` socket error seen on proxy exit, look into handling.

```
Error: read EADDRNOTAVAIL
    at TCP.onStreamRead (node:internal/stream_base_commons:217:20) {
  errno: -49,
  code: 'EADDRNOTAVAIL',
  syscall: 'read'
}
```

JaapRood:
We now have passing tests, included the one added for new coverage of the tunnelling proxy. Especially once I put the transparent proxy actually in front of the capturing proxy, the existing tests did a great job of highlighting how the transparent proxy wasn't behaving correctly in quite a few ways. While the new tunnelling capability was fine, dealing with direct traffic (both with and without TLS enabled) and capturing the right interactions required a bunch more work. 

What's not covered by tests yet is the error-recovery capabilities we've been adding by handling the right errors at the right times, in the right way. I had a go at one of these tests, but they're quite tricky to setup and not be flaky. Rather than spending the time to do that for every single error case, we will instead rely on monitoring errors properly and adding coverage to proven problems.

With that, it's time to get this reviewed and out the door!

acunniffe (approved):
Looks good to me / works when I tried it.

PR #1332, opened 2022-08-31
chore: latest tag for container image

notnmeyer:
- only publishes to docker hub for release builds
- tags and pushes `latest`


nalanj (approved):

notnmeyer:
actually, this isnt correct — it won't account for multiple platforms

PR #1345, opened 2022-09-02
wip: oas2 traverser & tests

acunniffe:
**Description** - A few users have been asking about when OAS 2 support is coming. We were not sure how hard supporting multiple versions would be, but as we discovered when we released 3.1 support last month, the facts model effectively normalizes out the structure of the OpenAPI specifications. 

The OAS 2 traverser will traverse the unique structure of these documents and map all of their types onto our standard facts. That mean OOTB our built-in rules will work. If a company has written custom rules that look at the raw values of different OpenAPI entities, those may break when encountering an unexpected spec version. That should be pretty rare. If you want to support both major versions, you need to update the custom rules anyway and that is not the worst way to find out you need to. 

- replaces `content` `contentType` with `produces` and `consumes`
- replaces `requestBody` with the request parameter that has `in: body`
- removes `nullable`, `oneOf`, `anyOf`, `allOf` -- not supported in oas 2
- includes on some known limitations
  - omits `formData` and `examples` (these can come later if needed) 
- adds OpenAPI v2 validation and semver parsing
- tests the facts produced by 3 OAS 2 specifications from Azure, & Petstore. Manually reviewed and snapshotted 
   - looked for multiples of bodies for each `produces` and `consumes`
   - path parameters get generated properly 
   - schema facts are correct 

**questions?**

- All over our code we use `ParseOpenAPIResult`, and `jsonLike` is oas3 typed. That type here has never been correct. We say it is oas3, but there's no validation at this stage. The sourcemap parser will only fail if a `$ref` can't be resolved. It could be any data that is YAML and JSON. We have the validator, I propose that is where the typing gets applied. Ie validator would return `errors?: string[] , 3_0?: OpenAPI_V3, 3_1?: OpenAPI_V3_1, 2_0?: OpenAPI_V2_0
- Does the changelog or anything in Optic Cloud strongly expect the specs to be OpenAPI 3? What does the path to supporting OAS 2 look like in the changlog / cloud? 
- What else could we write tests for? 

**current state?**
- the validator is still set to fail on oas2, so we could merge these changes without the support being enabled 
- traverser is working great. I can keep adding more tests.  
- we can't turn it on until we address the `ParseOpenAPIResult` issue and do some retyping. 
- we can't turn it on until we validate Changelog works OR decide not to support that to start
   - **what if we choose not to upload oas2 to cloud / changelogs? That lets us give this to customers and validate interest in full story before going further, and takes pressure off the cloud team who have a lot of other things going on right now**


projects/openapi-utilities/src/openapi/openapi2/openapi-2-traverser.ts
```diff
@@ -0,0 +1,891 @@
...
+      };
+
+      yield* traverseIfPresent(OpenAPIV2.HttpMethods.GET);
+      yield* traverseIfPresent(OpenAPIV2.HttpMethods.PATCH);
+      yield* traverseIfPresent(OpenAPIV2.HttpMethods.POST);
+      yield* traverseIfPresent(OpenAPIV2.HttpMethods.PUT);
+      yield* traverseIfPresent(OpenAPIV2.HttpMethods.DELETE);
+      yield* traverseIfPresent(OpenAPIV2.HttpMethods.HEAD);
+      yield* traverseIfPresent(OpenAPIV2.HttpMethods.OPTIONS);
+    }
```
acunniffe, line 79: this is all the same

projects/openapi-utilities/src/openapi/openapi2/openapi-2-traverser.ts
```diff
@@ -0,0 +1,891 @@
...
+
+    if (this.input.definitions) {
+      for (let [name, schema] of Object.entries(this.input.definitions)) {
```
acunniffe, line 82: components is now definitions

projects/openapi-utilities/src/openapi/openapi2/openapi-2-traverser.ts
```diff
@@ -0,0 +1,891 @@
...
+
+    yield* this.traverseParameters(
+      operation,
+      jsonPointer.append(jsonPath, 'parameters'),
+      [...conceptualPath, 'parameters'],
+      location
+    );
+
+    const requestBody = operation.parameters?.find((i) => {
+      if (isNotReferenceObject(i) && i.in === 'body') return true;
```
acunniffe, line 115: request body in oas 2 is 
- the first parameter
- with `in: body`

projects/openapi-utilities/src/openapi/openapi2/openapi-2-traverser.ts
```diff
@@ -0,0 +1,891 @@
...
+    if (requestBody && requestBodyIndex) {
+      if (isNotReferenceObject(requestBody)) {
+        for (let contentType of this.getConsumesContentTypes(operation) || [])
+          yield* this.traverseBody(
+            requestBody,
+            contentType,
+            jsonPointer.append(
+              jsonPath,
+              'parameters',
+              requestBodyIndex.toString()
```
acunniffe, line 132: json pointer for request body needs to have the index of the parameter

projects/openapi-utilities/src/openapi/openapi2/openapi-2-traverser.ts
```diff
@@ -0,0 +1,891 @@
...
+  format = 'openapi2';
+
+  input: OpenAPIV2.Document | undefined = undefined;
+
+  traverse(input: OpenAPIV2.Document): void {
+    this.input = input;
+  }
+
+  // see https://github.com/OAI/OpenAPI-Specification/issues/1259
+  getConsumesContentTypes(operation: OpenAPIV2.OperationObject): string[] {
```
acunniffe, line 45: see https://github.com/OAI/OpenAPI-Specification/issues/1259

consumes = content type for request bodies. It is set in the operation || globally || []

projects/openapi-utilities/src/openapi/openapi2/openapi-2-traverser.ts
```diff
@@ -0,0 +1,891 @@
...
+    return operation.consumes || this.input?.consumes || [];
+  }
+  getProducesContentTypes(operation: OpenAPIV2.OperationObject): string[] {
```
acunniffe, line 48: see https://github.com/OAI/OpenAPI-Specification/issues/1259

produces = content type for response bodies. It is set in the operation || globally || []

projects/openapi-utilities/src/openapi/openapi2/openapi-2-traverser.ts
```diff
@@ -0,0 +1,891 @@
...
+    operation: OpenAPIV2.OperationObject,
+    jsonPath: string,
+    conceptualPath: IPathComponent[],
+    location: OperationLocation
+  ): IterableIterator<IFact> {
+    const locationForParameter = (parameter: OpenAPIV2.ParameterObject) => {
+      let paramLocation:
+        | PathParameterLocation
+        | QueryParameterLocation
+        | HeaderParameterLocation;
```
acunniffe, line 242: cookies do not exist in oas 2

projects/openapi-utilities/src/openapi/openapi2/openapi-2-traverser.ts
```diff
@@ -0,0 +1,891 @@
...
+    this.checkJsonTrail(jsonPath, schema);
+    yield this.onField(
+      key,
+      schema,
+      required,
+      jsonPath,
+      conceptualPath,
+      location
+    );
+    yield* this.traverseSchema(schema, jsonPath, conceptualPath, location);
```
acunniffe, line 530: all the same

projects/openapi-utilities/src/openapi/openapi2/openapi-2-traverser.ts
```diff
@@ -0,0 +1,891 @@
...
+
+  // TODO discriminate between ArraySchemaObject | NonArraySchemaObject
+  *traverseSchema(
+    schema: OpenAPIV2.SchemaObject,
+    jsonPath: string,
+    conceptualPath: IPathComponent[],
+    location: FieldLocation
+  ): IterableIterator<IFact> {
+    this.checkJsonTrail(jsonPath, schema);
+    // note: OAS 2 does not support oneOf | anyOf | allOf
```
acunniffe, line 541: removed code to traverse polymorphic variants.

projects/openapi-utilities/src/openapi/openapi2/openapi-2-traverser.ts
```diff
@@ -0,0 +1,891 @@
...
+      },
+      value: this.getSpecificationFact(specification),
+    };
+  }
+
+  getSpecificationFact(
+    specification: OpenAPIV2.Document
+  ): OpenApiSpecificationFact {
+    const { paths, schemes, security, ...specificationFact } = specification;
+    return { ...specificationFact, openapi: this.input?.swagger || '2.0' };
```
acunniffe, line 669: weird and hacky...openapi version is called `swagger` in swagger. not sure what the right path is here, but it's not super material

projects/openapi-utilities/src/openapi/openapi2/openapi-2-traverser.ts
```diff
@@ -0,0 +1,891 @@
...
+      } else {
+        console.warn(
+          `Expected a flattened spec, found a reference at: ${getReadableLocation(
+            nextJsonPath
+          )}`
+        );
+      }
+    }
+
+    // @ limitation no example support
```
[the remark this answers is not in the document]
acunniffe: > Is this a limitation because we haven't coded it here 
yes 

> ...or we don't handle examples in a way that works with OAS2
not that, but our current oas3 piece is incomplete too. I wouldn't feel safe saying optic supports examples in either spec yet. It's not 100% clear to me that we need seperate examples or if they're part of the entity they are attached to. 

The sorts of rules teams want for examples are things like
- example matches schema 
- has example
- example includes optional properties (not just required) 

We've never really supported examples fully in oas3. We allowed them in response and request bodies, but not on individual properties and a handful of other places the spec allows for. They work differently in oas2 -- we should probably make some decisions about how much we want to support in both versions. 

https://blog.stoplight.io/difference-between-open-v2-v3-v31

niclim:
closing for now, we have a different approach we want to take to support oas2

PR #1346, opened 2022-09-06
fix(ci): skipped image build workflow

notnmeyer:
## 🍗 Description

the image build was inadvertently skipped in this run https://github.com/opticdev/optic/actions/runs/2979898869/workflow

looking at [the docs](https://docs.github.com/en/rest/releases/releases#get-a-release) the comparison on running the workflow should have been against a bool rather than a string. there's not much logging in actions to confirm specifically this was the issue, but i'm fairly certain it is because its the only condition :)

## 📚 References
closes https://github.com/opticdev/issues/issues/514

## 👹 QA

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -154,7 +154,7 @@ jobs:
 
   publish-container-image:
     needs: release
-    if: github.event.release.prerelease == 'false'
+    if: github.event.release.prerelease == false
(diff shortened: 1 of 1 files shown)


notnmeyer:
manually triggered CI run for the missing image is here: https://github.com/opticdev/optic/actions/runs/3001436114/workflow

PR #1350, opened 2022-09-08
provide a nicer error for invalid OpenAPI

acunniffe:
## 🍗 Description
Let's take a look at the pipeline for an diff/check run. We have various different places it can error out right now. 

1. is valid yaml or json (fails with a nice error)
Screen Shot 2022-09-08 at 3 12 43 PM

2. resolve and deference spec (fails with a reasonable error)
Screen Shot 2022-09-08 at 3 14 25 PM

3. validate OpenAPI (gives you an ugly error, raw AJV) 
Screen Shot 2022-09-08 at 3 16 54 PM

4. Optic CI Results! (These are nice, that's the whole reason we do everything else!) 

Step 3 has always given users trouble. It looks like an optic error, and it distracts folks from fixing their own OpenAPI specifications (where the problem actually is). Even if you knew it was an error with your spec, it's hard to know how to fix it. 

---

This PR only changes step 3 -- it makes errors in your OpenAPI readable and very clearly something you (the openapi author) has to address. 
Screen Shot 2022-09-08 at 10 05 29 AM

misc -- tighten our JSON Schemas 

## 👹 QA
- I updated the validator tests
- (manual) go break something in openapi and try it yourself :)


projects/openapi-utilities/src/openapi3/implementations/openapi3/log-json-pointer.ts
```diff
@@ -0,0 +1,72 @@
...
+        );
+      }
+    },
+  };
+}
+
+function rangeFromLine(
+  line: number,
+  before: number = 4,
+  after: number = 3
```
niclim, line 37: I don't understand this - why are these numbers 4 and 3?

niclim (approved):
Looks great - can we also add in an integration test (see folder https://github.com/opticdev/optic/tree/main/projects/optic/src/__tests__/integration)

It's a little hard to read the snapshot under `[__snapshots__/validator.test.ts.snap](https://github.com/opticdev/optic/pull/1350/files#diff-2c08142e7d99e444128a5a2f14bda9c417dffc5f4f47419ff036e9a0606f9615)` - perhaps it can be rendered with the spacing?

One question some magic numbers but overall think this is good

acunniffe:
Will do. 

What's this? 
> One question some magic numbers but overall think this is good

niclim:
> What's this? One question some magic numbers but overall think this is good


https://github.com/opticdev/optic/pull/1350#discussion_r966347101

Also looks like the imports in tests are failing - it might be worth dumping out the sourcemap generated from loading the `bad openapi spec` so that you dont need a circular import from openapi-utilities (openapi-io relies on openapi-utilities)

projects/openapi-utilities/src/openapi3/implementations/openapi3/log-json-pointer.ts
acunniffe: It's just a preference. Lines above the relevant line and line after. 

it looked nice. I can make them constants

projects/openapi-utilities/src/openapi3/implementations/openapi3/log-json-pointer.ts
```diff
@@ -0,0 +1,72 @@
+import { JsonSchemaSourcemap } from '@useoptic/openapi-io';
+import { sourcemapReader } from './sourcemap-reader';
+import chalk from 'chalk';
+
+export function jsonPointerLogger(sourceMap: JsonSchemaSourcemap) {
+  const reader = sourcemapReader(sourceMap);
+
+  return {
+    log: (pointer: string): string | undefined => {
```
niclim, line 9: Can we return an empty string instead of just undefined?

projects/openapi-utilities/src/openapi3/implementations/openapi3/log-json-pointer.ts
niclim: Oh I see what it's used for. The function name was a little hard to understand. How about `getDisplayedLineNumbers`

acunniffe:
> so that you dont need a circular import from openapi-utilities (openapi-io relies on openapi-utilities)

haha I see. when I helped with the poc it was the other way around. I'll update

PR #1353, opened 2022-09-13
Fix/openapi io/export check openapi version

Alexandre-Herve:
## 🍗 Description

Re export `checkOpenAPIVersion` after e2329ebfd moved it

## 📚 References

## 👹 QA


niclim (approved):
we could also prerelease here - but patch release also works

PR #1361, opened 2022-09-23
Built-in ruleset for examples

acunniffe:
## 🍗 Description
This ruleset has been requested by several users and since it's pretty generic / useful we've decided to bring it into the built-ins. 

The `examples` built-in does two things:
- enforce a requirement that every entity in your OpenAPI spec has a valid example. You can configure this, and also set `applies: addedOrChanged` if you only want this to apply to new APIs
- enforces that each example matches its schema. This works at the property, request, response and parameter level and cannot be configured / turned off. If the example is there, it must match the schema. 

## 📚 References
Config:
```
examples: 
  require_parameter_examples: true
  require_request_examples: true
  require_response_examples: true
  applies: 'always' | 'addedOrChanged'
```

## 👹 QA
There are a lot of tests, but go ahead and give it a spin! 

My biggest concern is the `ajv` parser failing in weird ways depending on keywords you set.

### DIFF
--- a/projects/standard-rulesets/src/examples/constants.ts
+++ b/projects/standard-rulesets/src/examples/constants.ts
@@ -1,0 +1,1 @@
+export const appliesWhen = ['addedOrChanged', 'always'] as const;
--- a/projects/standard-rulesets/src/examples/requireExample.ts
+++ b/projects/standard-rulesets/src/examples/requireExample.ts
@@ -1,0 +1,136 @@
+import {
+  OperationRule,
+  RequestRule,
+  ResponseBody,
+  ResponseRule,
+  RuleError,
+} from '@useoptic/rulesets-base';
+import { appliesWhen } from './constants';
+import { OpenAPIV3 } from 'openapi-types';
+
+export const requireResponseExamples = (applies: typeof appliesWhen[number]) =>
+  new ResponseRule({
+    name: 'require response body examples',
+    rule: (responseAssertions) => {
+      function responseToNumberOfExamples(body: ResponseBody) {
+        const numberOfExamples =
+          (body.raw.example ? 1 : 0) +
+          Object.keys(body.raw.examples || {}).length;
+        return numberOfExamples;
+      }
+
+      if (applies === 'always') {
+        responseAssertions.requirement((value) => {
+          value.bodies.forEach((body) => {
+            if (responseToNumberOfExamples(body) < 1) {
+              throw new RuleError({
+                message: `a valid example is required for every documented response body`,
+              });
+            }
+          });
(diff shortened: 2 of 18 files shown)


projects/standard-rulesets/src/examples/constants.ts
```diff
@@ -0,0 +1 @@
+export const appliesWhen = ['addedOrChanged', 'always'] as const;
```
niclim, line 1: Is there a ready we wouldn't have added or changed? This is inconsistent with our naming checks ruleset

projects/standard-rulesets/src/examples/requireExample.ts
```diff
@@ -0,0 +1,136 @@
...
+      }
+    },
+  });
+
+export const requireParameterExamples = (applies: typeof appliesWhen[number]) =>
+  new OperationRule({
+    name: 'require parameter examples',
+    rule: (operation) => {
+      if (applies === 'always') {
+        operation.headerParameter.requirement((header) => {
```
niclim, line 89: Do we need to handle path parameters?

projects/standard-rulesets/src/examples/requireExample.ts
```diff
@@ -0,0 +1,136 @@
...
+        });
+      }
+    },
+  });
+
+export const requireParameterExamples = (applies: typeof appliesWhen[number]) =>
+  new OperationRule({
+    name: 'require parameter examples',
+    rule: (operation) => {
+      if (applies === 'always') {
```
niclim, line 88: You could do a map here:
```
const lifecycle = applies === 'always' ? 'requirement' : 'addedOrChanged'

// then
operation.queryParameter[lifecycle]
```

niclim (approved):
Overall this looks great - summary of my questions here:
- Should we also have `added` and `changed` lifecycle methods? This is consistent with naming changes
- Should we also handle path parameters?

projects/standard-rulesets/src/examples/constants.ts
acunniffe: for examples the changes case doesn't really make sense. That's equivalent to always 

same is actually true for naming. Since name is part of the conceptual path there's no way a change rule would ever trigger for changing a name -- ironically. we should just remove it there too

projects/standard-rulesets/src/examples/requireExample.ts
acunniffe: I considered it but don't think so -- those are always going to be string IDs and it's likely to create hundreds of errors in a mature API spec to suddenly ask for that

PR #1363, opened 2022-09-27
Fix issue with examples built-in

acunniffe:
## 🍗 Description
Yesterday we released the examples built in #1361 -- issues were reported that this PR addresses 
- works with `format` keyword on strings
- does not crash if a regex is invalid js snyntax

## 📚 References

## 👹 QA

### DIFF
--- a/projects/standard-rulesets/src/examples/requireValidExamples.ts
+++ b/projects/standard-rulesets/src/examples/requireValidExamples.ts
@@ -7,9 +7,11 @@ import {
 } from '@useoptic/rulesets-base';
 import { OpenAPIV3 } from 'openapi-types';
 import Ajv, { SchemaObject } from 'ajv';
+import addFormats from 'ajv-formats';
 
 const ajv = (() => {
   const validator = new Ajv({ strict: false });
+  addFormats(validator);
   // override pattern keyword when invalid regex
   validator.removeKeyword('pattern');
   validator.addKeyword({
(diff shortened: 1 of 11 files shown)


projects/standard-rulesets/src/examples/requireValidExamples.ts
```diff
@@ -7,10 +7,36 @@ import {
...
+        return `pattern not matched ${cxt.schema}`;
+      },
+    },
+    compile: (pattern) => {
+      let regex;
+      try {
+        regex = new RegExp(pattern);
+      } catch (e) {
+        return (data) => true;
+      }
```
niclim, line 32: Is there any way we report that the format / pattern is invalid? I'm not sure how we would do this but it feels confusing to have formats typoed and not validating silently

niclim (approved):
I think this is ok to fix the bug but I feel like we need some way to warn that a rule isn't running as expected

projects/standard-rulesets/src/examples/requireValidExamples.ts
acunniffe: @niclim I agree we should have a way to allow warnings to come out of rules, esp as more and more teams write their own. It's useful to distinguish problem with ruleset execution from problem with API. 

I feel like this is a reasonable (documented) compromise and is ok for now, but should be part of a larger conversation about the warnings.

PR #1373, opened 2022-10-06
optic diff should exit 1 when checks fail

acunniffe:
## 🍗 Description
A user reported that `optic diff` is not exiting `1` when checks fail the same way the now (deprecated) `optic compare` commands did. This was an omission  and is being fixed with this PR

## 👹 QA
Try running `optic diff` with `--check` and then run `echo $` after it finishes.


projects/optic/src/commands/diff/diff.ts
```diff
@@ -215,6 +215,14 @@ const runDiff = async (
     console.log(log);
   }
 
+  const diffResults = {
+    checks: {
+      total: specResults.results.length,
+      passed: specResults.results.filter((check) => check.passed).length,
+      failed: specResults.results.filter((check) => !check.passed).length,
+    },
+  };
```
niclim, line 224: do we need to handle exemptions (i think here it's inline exemptions)

niclim (approved):

projects/optic/src/commands/diff/diff.ts
acunniffe: good note. I've updated the failed predicate to check if exempted is set

PR #1380, opened 2022-10-18
Implement `optic ruleset publish` command

niclim:
## 🍗 Description

This PR implements the optic ruleset publish command. It's currently hidden, will be shown when we've finished implementing the custom rules flow. 

I'm intending usage to look something like: `optic ruleset publish file_path --token <token>`. The name of the ruleset can be included in the JS file. This allows us to be agnostic of the wider environment (e.g. want to set up and publish multiple rules in different repos, we aren't tied to any repo level construct).

In follow up PRs:
- Will connect up this command to backend endpoints

## 📚 References

## 👹 QA

### DIFF
--- a/projects/optic/src/__tests__/integration/ruleset-publish.test.ts
+++ b/projects/optic/src/__tests__/integration/ruleset-publish.test.ts
@@ -1,0 +1,31 @@
+import { runOptic, setupWorkspace, normalizeWorkspace } from './integration';
+
+jest.setTimeout(30000);
+// TODO replace this with real mocks when connected to backend
+jest.mock('node-fetch');
+
+describe('optic ruleset publish', () => {
+  afterEach(() => {
+    jest.resetAllMocks();
+  });
+
+  test('can publish a ruleset', async () => {
+    const workspace = await setupWorkspace('ruleset-publish/valid-js-file');
+    const { combined, code } = await runOptic(
+      workspace,
+      'ruleset publish ./rules.js --token atoken'
+    );
+    expect(normalizeWorkspace(workspace, combined)).toMatchSnapshot();
+    expect(code).toBe(0);
+  });
+
+  test('exits if ruleset file shape is not valid', async () => {
+    const workspace = await setupWorkspace('ruleset-publish/invalid-js-file');
+    const { combined, code } = await runOptic(
+      workspace,
+      'ruleset publish ./rules.js --token atoken'
+    );
+    expect(normalizeWorkspace(workspace, combined)).toMatchSnapshot();
+    expect(code).toBe(1);
+  });
+});
--- a/projects/optic/src/__tests__/integration/ruleset-publish.test.ts
+++ b/projects/optic/src/__tests__/integration/ruleset-publish.test.ts
@@ -5,15 +5,22 @@ jest.setTimeout(30000);
 jest.mock('node-fetch');
 
 describe('optic ruleset publish', () => {
+  let oldEnv: any;
+  beforeEach(() => {
+    oldEnv = {...process.env}
+  })
+
   afterEach(() => {
     jest.resetAllMocks();
+    process.env = {...oldEnv}
   });
 
   test('can publish a ruleset', async () => {
     const workspace = await setupWorkspace('ruleset-publish/valid-js-file');
+    process.env.OPTIC_TOKEN = '123'
     const { combined, code } = await runOptic(
       workspace,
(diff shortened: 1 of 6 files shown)


projects/optic/src/init.ts
```diff
@@ -63,14 +63,13 @@ export const initCli = async () => {
...
-    .command('cloud')
+  const rulesetSubcommands = cli
+    .command('ruleset')
     .description(
-      'Commands to interact with Optic Cloud. See `optic cloud --help`'
+      'Commands to build your own optic rulesets. See `optic ruleset --help`'
     )
     .addHelpCommand(false);
-  registerCreateGithubContext(cloudSubcommands, true);
-  registerCreateManualContext(cloudSubcommands);
```
niclim, line 73: Removed these two since we're no longer running CI triggered comparisons, and therefore these are not used for anything in the `optic` binary

projects/optic/src/commands/ruleset/publish.ts
```diff
@@ -0,0 +1,123 @@
...
+
+const fileIsValid = (
+  file: any
+): file is {
+  default: {
+    name: string;
+    rules: any[];
+  };
+} => {
+  const result = validateRulesFile(file);
```
niclim, line 113: There's more validation we can move into this part. Moving more validation here means that users would get faster feedback (rather than waiting for this to be uploaded, and then trying out the new ruleset.

projects/optic/src/__tests__/integration/ruleset-publish.test.ts
```diff
@@ -0,0 +1,31 @@
...
+describe('optic ruleset publish', () => {
+  afterEach(() => {
+    jest.resetAllMocks();
+  });
+
+  test('can publish a ruleset', async () => {
+    const workspace = await setupWorkspace('ruleset-publish/valid-js-file');
+    const { combined, code } = await runOptic(
+      workspace,
+      'ruleset publish ./rules.js --token atoken'
```
nalanj, line 16: I think it'd be better to lean on an env var than a token flag on the CLI, just so someone can opt into having their token in cleartext on their command line.

projects/optic/src/commands/ruleset/publish.ts
```diff
@@ -0,0 +1,123 @@
...
+
+    const name = userRuleFile.default.name;
+
+    const fileBuffer = await fs.readFile(absolutePath);
+    const rulesetUpload: {
+      id: string;
+      uploadUrl: string;
+    } = await (async (name: any, token: string) => ({
+      id: '',
+      uploadUrl: 'https://example.com/placeholder',
```
nalanj, line 72: This is `upload_url` now.

projects/optic/src/commands/ruleset/publish.ts
```diff
@@ -0,0 +1,123 @@
...
+
+import { OpticCliConfig } from '../../config';
+
+const uploadFileToS3 = async (signedUrl: string, file: Buffer) => {
+  await fetch(signedUrl, {
+    method: 'PUT',
+    headers: {
+      'x-amz-server-side-encryption': 'AES256',
+    },
+    body: file,
```
nalanj, line 18: Do you think we'd want to go ahead and compress this with something like brotli? I know it's preoptimization, but it should drastically reduce storage for us and doesn't feel like it has much downside.

nalanj (approved):
Couple of small things, but looks great overall.

projects/optic/src/__tests__/integration/workspaces/ruleset-publish/valid-js-file/rules.js
```diff
@@ -0,0 +1,4 @@
+module.exports = {
+  name: 'asdasdasdasdasdasd - the best ruleset', // this name is so that this is compressable
```
niclim: @nalanj Ok this is fun, the brotli library we use sometimes returns `null` if the data we pass it cannot be compressed. The reason we haven't run into this before is because the data we try compress (in CLIDiff) is _always_ compress-able. Likely rulesets will also always be compress-able as well, but not guaranteed.

I'm going to address this in a follow up PR

nalanj: Where's the info about it sometimes returning null?

nalanj: See https://github.com/foliojs/brotli.js/issues/36. Feels like that might be what you're hitting.

PR #1382, opened 2022-10-19
Handle uncompressable ruleset file in optic ruleset publish

niclim:
## 🍗 Description

A follow up from https://github.com/opticdev/optic/pull/1380#discussion_r999789380

After thinking about this, I don't think we need to update the CLIDiff flow, the data in that flow should _always_ be compressable (since there's two openapi specs uploaded along with results). If we optimize the changelog format such that it no longer becomes compressable, we will need to update that flow to also handle the null case.

## 📚 References

## 👹 QA


niclim:
superseceded by https://github.com/opticdev/optic/pull/1386

PR #1384, opened 2022-10-19
Add MIT licenses

niclim:
## 🍗 Description

We have a [small thing in our README](https://github.com/opticdev/optic#license) that points to MIT license, but this is not where npm or github looks at licenses. Adding LICENSE files and license to our package.

## 📚 References

## 👹 QA


nalanj (approved):
Just as a quick check, I'm pretty sure most js monorepos are putting one license file up top for the entire body of work, and then putting the license in each `package.json` and leaving it at that.

PR #1386, opened 2022-10-20
Use node brotli over npm package

niclim:
## 🍗 Description

Uses the node package of brotli over the npm package. Fixes the issue around null returned if a payload is not compressable.

This still works with CLIDiff `yarn run local:run diff spec1.yml spec2.yml --web` (tested to sanity check any weirdness)

## 📚 References

## 👹 QA


nalanj (approved):
Of course pending tests passing.

PR #1390, opened 2022-10-27
Example validation fix

acunniffe:
## 🍗 Description
Our logic for parameter example required is out of sync with parameter example schemas match

One allows you to look at `schema.example` or `.example` while the other does not. This brings them into alignment so users see the behaviors they expect

## 📚 References

## 👹 QA

### DIFF
--- a/projects/standard-rulesets/src/examples/requireExample.ts
+++ b/projects/standard-rulesets/src/examples/requireExample.ts
@@ -87,22 +87,32 @@ export const requireParameterExamples = (applies: typeof appliesWhen[number]) =>
     rule: (operation) => {
       const lifecycle = applies === 'always' ? 'requirement' : 'addedOrChanged';
       const errorMessageType = applies === 'always' ? 'every' : 'added';
+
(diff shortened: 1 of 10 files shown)


projects/standard-rulesets/src/examples/requireExample.ts
```diff
@@ -87,22 +87,32 @@ export const requireParameterExamples = (applies: typeof appliesWhen[number]) =>
...
       const errorMessageType = applies === 'always' ? 'every' : 'added';
+
+      function schemaHasExample(schema: OpenAPIV3.SchemaObject | OpenAPIV3.ReferenceObject | undefined) {
+        if (schema) {
+          if ('example' in schema) {
+            return typeof schema.example !== 'undefined'
+          }
+        }
+        return false
+      }
```
niclim, line 98: Do we need to handle examples as well?

And can we also update the tests

niclim (approved):

projects/standard-rulesets/src/examples/requireExample.ts
acunniffe: That's handled in the first part of the if statement 101 108 115

PR #1400, opened 2022-11-02
Support running custom rules in optic diff

niclim:
## 🍗 Description

Supports running custom rules locally. This includes:
- referring to a custom ruleset in optic-cloud
- referring to a locally created ruleset
- caching of rulesets by uploaded_at

TODO:
- [x] Write documentation
- [x] Publish pre-release + update monorail code to match the ruleset resolution code

## 📚 References

## 👹 QA

Download the pre-release build `0.34.6-0` + run it against a local build with the [fetch ruleset endpoint](https://github.com/opticdev/monorail/pull/2338)


projects/optic/src/utils/git-utils.ts
```diff
@@ -6,7 +6,7 @@ export const hasGit = async (): Promise<boolean> =>
       if (err || stderr || !stdout) resolve(false);
       resolve(true);
     };
-    const command = `git -v`;
+    const command = `git --version`;
```
niclim, line 9: This wasn't working for me for some reason - not sure if there's version differences but `git --version` runs correctly, before this wasn't picking up the correct config

projects/rulesets-base/src/custom-rulesets/prepare-rulesets.ts
```diff
@@ -0,0 +1,91 @@
...
+    {
+      uploaded_at: string;
+      url: string;
+    }
+  >;
+  standardRulesets: Record<
+    string,
+    { fromOpticConfig: (config: unknown) => Ruleset | string }
+  >;
+};
```
niclim, line 24: Couple of changes from the old version:
- local rulesets can be defined and will just be resolved
- standad rulesets are injected (to avoid circular deps since this is in ruleset-base)
- hosted rulesets also include uploaded_at to allow for caching

nalanj: Is this updated in what goes over to rules orchestrator?

nalanj (approved):

projects/rulesets-base/src/custom-rulesets/prepare-rulesets.ts
niclim: Yep, I updated rules-orchestrator to include these changes where localrulesets is {} and we're matching the shapes

PR #1405, opened 2022-11-10
example requirements only run on specific content types

acunniffe:
## 🍗 Description
Requiring examples only makes sense for XML, JSON, and text. Images, binary, calendar content types shouldn't require them by default

## 📚 References

## 👹 QA

### DIFF
--- a/projects/standard-rulesets/src/examples/qualifiedContentType.ts
+++ b/projects/standard-rulesets/src/examples/qualifiedContentType.ts
@@ -7,8 +7,7 @@ export function qualifiedContentType(contentType: string): boolean {
       parsedType.essence === 'application/json' ||
       parsedType.essence === 'application/xml' ||
       parsedType.essence === 'text/json' ||
-      parsedType.essence === 'text/plan' ||
-      parsedType.essence === 'text/plan' ||
+      parsedType.essence === 'text/plain' ||
       parsedType.subtype.endsWith('+json') ||
       parsedType.subtype.endsWith('+xml')
     ) {
--- a/projects/standard-rulesets/src/examples/qualifiedContentType.ts
+++ b/projects/standard-rulesets/src/examples/qualifiedContentType.ts
@@ -1,0 +1,20 @@
+import MIMEType from 'whatwg-mimetype';
+
+export function qualifiedContentType(contentType: string): boolean {
+  let parsedType = contentType && MIMEType.parse(contentType);
+  if (parsedType) {
+    if (
+      parsedType.essence === 'application/json' ||
+      parsedType.essence === 'application/xml' ||
+      parsedType.essence === 'text/json' ||
+      parsedType.essence === 'text/plan' ||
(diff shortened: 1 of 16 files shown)


projects/standard-rulesets/src/examples/qualifiedContentType.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export function qualifiedContentType(contentType: string): boolean {
+  let parsedType = contentType && MIMEType.parse(contentType);
+  if (parsedType) {
+    if (
+      parsedType.essence === 'application/json' ||
+      parsedType.essence === 'application/xml' ||
+      parsedType.essence === 'text/json' ||
+      parsedType.essence === 'text/plan' ||
+      parsedType.essence === 'text/plan' ||
```
niclim, line 11: duplicate?

projects/standard-rulesets/src/examples/requireExample.ts
```diff
@@ -7,6 +7,7 @@ import {
 } from '@useoptic/rulesets-base';
 import { appliesWhen } from './constants';
 import { OpenAPIV3 } from 'openapi-types';
+import { qualifiedContentType } from './qualifiedContentType';
 
 export const requireResponseExamples = (applies: typeof appliesWhen[number]) =>
   new ResponseRule({
```
niclim, line 13: Now that I read this again, wouldn't this be better as a ResponseBody type? Otherwise this would trigger on any response change, not just a body change?

And that way, you can scope this rule to specific content types

projects/standard-rulesets/src/examples/requireExample.ts
```diff
@@ -57,6 +62,8 @@ export const requireRequestExamples = (applies: typeof appliesWhen[number]) =>
 
       if (applies === 'always') {
         requestAssertions.body.requirement((value) => {
+          if (!qualifiedContentType(value.contentType)) return;
```
niclim, line 65: Isn't this better to go into the matches block?

niclim (approved):
Approving, but I think this is better off as adding to a matches block + using ResponseBody over Response rules

acunniffe:
> Approving, but I think this is better off as adding to a matches block + using ResponseBody over Response rules

So I thought to do that first, but you still end up checking it twice (at the inner level) too because each response/request can have multiple content types documented. Matches turn off the rule for a response that NEVER qualified, but it couldn't turn it on for just some of those bodies. If we had a ResponseBodyRule or RequestBodyRule we would use matches for sure

projects/standard-rulesets/src/examples/requireExample.ts
acunniffe: I was able to make this change for responses because we have a ResponseBody rule, but not a RequestBody rule. Each request can have multiple content types so I can't do this all in matches -- well I could, but it'd have to go in both places

PR #1409, opened 2022-11-16
Create a SpectralRule

niclim:
## 🍗 Description

Creates a SpectralRule that can be used in a custom rules package to run spectral rules with additional lifecycles.

Up next is hooking up the ExternalRules to the rule runner (and writing documentation)

## 📚 References

## 👹 QA


projects/rulesets-base/src/rules/spectral-rule.ts
```diff
@@ -0,0 +1,137 @@
...
+    passed: false,
+    error: spectralResult.message,
+    isMust: true,
+    isShould: false,
+    where: `${lifecycle} `,
+    change: relevantChangeOrFact,
+  };
+}
+
+function constructFactTree(facts: IFact[]): FactTree {
```
niclim, line 35: Its currently implemented in ruleset-base, but this function will become useful for translating json paths to "openapi nodes"

This takes in facts, which in the future may have different nodes for different api versions.

projects/openapi-utilities/src/openapi3/json-path-interpreters.ts
```diff
@@ -0,0 +1,54 @@
...
+// A trie datastructure keyed by json path parts
+export type FactTree = Record<
+  string,
+  { fact: IFact | null; children: FactTree }
+>;
+
+// We need a special json path decoder since the leading / is significant and points to the root
+const decodePathWithLeadingTrail = (pointer: string) => pointer.split('/').map(jsonPointerHelpers.unescape);
+
+export function constructFactTree(facts: IFact[]): FactTree {
```
niclim, line 13: This function can be adapted to just pass in list of json paths + locators instead of facts (when we remove facts at some point in the future).

This is going to be important as we lean more heavily on json path interpretations (I think this implementation is OpenAPI version agnostic, the version handling differences will be handled based on the facts that are passed in here).

projects/openapi-utilities/src/openapi3/json-path-interpreters.ts
```diff
@@ -0,0 +1,54 @@
...
+      }
+      // Set up the next loop
+      tree = tree[part].children;
+    }
+  }
+
+  return factTree;
+}
+
+export function getFactForJsonPath(
```
niclim, line 36: i need to think a little more about the implications if you pass a json path that is completely wrong in the current spec. The intention is that the json path is _in_ the spec/fact tree, but if it's not you could get a nonsense fact

acunniffe: I bet we'll get a lot of things that are in the spec, but not in our fact tree. it makes sense -- we don't have facts for everything in the spec. 

Spectral is matching based on Json Path (the real one  https://goessner.net/articles/JsonPath/ ) so all the JSON Pointers we get out the other end should at least exist.

projects/rulesets-base/src/extended-rules/spectral-rule.ts
```diff
@@ -0,0 +1,107 @@
...
+    );
+
+    const spectralResults = await this.spectral.run(inputs.nextJsonLike);
+    const results: Result[] = [];
+
+    for (const spectralResult of spectralResults) {
+      const path = jsonPointerHelpers.compile(
+        spectralResult.path.map((p) => String(p))
+      );
+      const fact = getFactForJsonPath(path, factTree);
```
acunniffe, line 78: 95% of spectral rules will point at attributes, not what we call facts. Ie they point at `operationId` not the operation

Is this doing strict equality? child of?

acunniffe (approved):
Looks correct -- assuming the getFact logic will work for spectral rules about attributes of the parts of the specs we have facts for.

projects/rulesets-base/src/extended-rules/spectral-rule.ts
niclim: It's finding the closest parent fact from a json path.

E.g. `/api~1users/get/operationId` will return the Fact at `/api~1users/get`

acunniffe: Awesome -- it looked like it might, but wasn't sure w/o running it.

PR #1410, opened 2022-11-17
Introduced more strict OpenAPI Validation

acunniffe:
## 🍗 Description
OpenAPI can validate against a schema and still be incorrect -- ie certain combinations of keywords are allowed, paths must have matching path components, etc...

After a few customers reported their teams had invalid OpenAPI I surveyed Stoplight, Postman and Swagger Hub and realized the state of affairs is pretty bad. Lots of tools are lettings lots of bad OpenAPI through. 

This PR is the first of many -- it establishes a pattern for annotating our validation schemas with additional code rules that can be used to apply these more strict forms of openapi validation

This PR also sets the first rules by adding exclusive use requirements for the keywords for object, array, and oneOf, allOf, anyOf

## 📚 References

## 👹 QA

### DIFF
--- a/projects/openapi-io/src/validation/validation-schemas.ts
+++ b/projects/openapi-io/src/validation/validation-schemas.ts
@@ -1,5 +1,6 @@
 const openapi3_0_schema_object = {
   type: 'object',
+  'x-custom-validator': 'validateSchema',
   description:
     'The Schema Object allows the definition of input and output data types. These types can be objects, but also primitives and arrays. This object is an extended subset of the JSON Schema Specification Wright Draft 00.  For more information about the properties, see JSON Schema Core and JSON Schema Validation. Unless stated otherwise, the property definitions follow the JSON Schema.',
   // additionalProperties: false,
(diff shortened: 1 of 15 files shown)


projects/openapi-io/src/validation/validator.ts
```diff
@@ -68,7 +71,8 @@ export const processValidatorErrors = (
     if (
       pathsWithErrors.every(
         (addedError) => !addedError.instancePath.startsWith(error.instancePath)
-      )
```
acunniffe, line 71: right now we reduce polymorphic branch errors down to the first case. This isn't correct, but it's reasonable to give users a single thing to do instead of N conflicting things to do. Either way they have to go back and decide what they are trying to achieve. 

This lets any of our customer validator rules log their results too

projects/openapi-io/src/validation/validation-schemas.ts
```diff
@@ -1,5 +1,6 @@
 const openapi3_0_schema_object = {
   type: 'object',
+  'x-custom-validator': 'validateSchema',
```
acunniffe, line 3: whenever we add a new custom validator, we can anchor them using an extension in their validation schemas

niclim (approved):

acunniffe:
Pre-release coming https://github.com/opticdev/optic/releases/tag/v0.35.2-0

PR #1423, opened 2022-11-28
Flat OAS Types

acunniffe:
## 🍗 Description
We use Kong's `openapi-types` type definitions which assume `$ref` is allowed. By the time we've parsed, validated and cast an object to one of these types it will never have `$ref`. The practical effect of this is that our code ends up being more complex -- constantly checking for $ref and casting everything unnecessarily. This affects user rule writing and our ability to ship more advanced changelogs / docs in optic. We do not need to replace every usage at once, but we should starting moving towards using these flat types and phase out the old ones. 

`DereferencedOpenAPIV3.Document`

## 📚 References

## 👹 QA

### DIFF
--- a/projects/openapi-utilities/src/index.ts
+++ b/projects/openapi-utilities/src/index.ts
@@ -1,6 +1,7 @@
 import { OpenAPITraverser } from './openapi3/implementations/openapi3/openapi-traverser';
 import { OpenAPIV3 } from 'openapi-types';
 import { factsToChangelog } from './openapi3/sdk/facts-to-changelog';
+export { DereferencedOpenAPI } from './flat-openapi-types'
 import {
   ConceptualLocation,
   IChange,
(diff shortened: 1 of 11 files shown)


projects/openapi-utilities/src/index.ts
```diff
@@ -1,6 +1,8 @@
 import { OpenAPITraverser } from './openapi3/implementations/openapi3/openapi-traverser';
 import { OpenAPIV3 } from 'openapi-types';
 import { factsToChangelog } from './openapi3/sdk/facts-to-changelog';
+export { DereferencedOpenAPIV2, DereferencedOpenAPIV3, DereferencedOpenAPIV3_1 } from './flat-openapi-types'
```
acunniffe, line 4: Do we like these names or do we want something shorter? 

`FlatOpenAPI`? `OAS_3_1`?

nalanj: `DerefOpenAPI`?

nalanj (approved):
Looks good. I agree naming maybe needs some love, but I'm also not sure how to really shorten the name.

projects/openapi-utilities/src/index.ts
acunniffe: went with Flat

PR #1432, opened 2022-12-08
JSON output flag for diff command

acunniffe:
## 🍗 Description
Two customers has asked for a `--json` output to help generate their company's public changelogs. 

This PR adds a `--json` flag that serializes the terminal changelog from `optic diff`. 

## 📚 References

## 👹 QA

### DIFF
--- a/projects/openapi-utilities/src/utilities/json-changelog.ts
+++ b/projects/openapi-utilities/src/utilities/json-changelog.ts
@@ -193,6 +193,7 @@ function getEndpointLogs(
     responseChanges.push(getResponseChangeLogs(response, key));
   }
 
+  const requestChangelogs = getRequestChangeLogs(request);
   return {
     name: `${method} ${path}`,
     change: getChange(change),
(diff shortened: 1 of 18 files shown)


projects/openapi-utilities/src/utilities/json-changelog.ts
```diff
@@ -0,0 +1,333 @@
...
+  return {
+    name: `${method} ${path}`,
+    change: getChange(change),
+    attributes: change
+      ? getDetailLogs(change, {
+          excludeKeys: ['pathPattern', 'method'],
+        })
+      : [],
+    parameters: parameterChanges,
+    requestBody:
```
acunniffe, line 205: if !added !change !removed -- is this always null?

niclim: Yeah it should be

projects/openapi-utilities/src/utilities/json-changelog.ts
```diff
@@ -0,0 +1,333 @@
...
+  //   yield * indent(getParameterLogs('path', name, parameterChange));
+  // }
+  //
+  // for (const [name, parameterChange] of headers.changes) {
+  //   yield * indent(getParameterLogs('header', name, parameterChange));
+  // }
+  //
+  //
+  //
+  // yield '';
```
niclim, line 245: can we remove the dead code here?

projects/optic/src/commands/diff/diff.ts
```diff
@@ -198,6 +188,21 @@ const runDiff = async (
     },
   };
 
+  if (specResults.changes.length === 0) {
+    console.log('No changes were detected');
+  } else {
+    console.log('');
+  }
+
+  if (options.json) {
```
niclim, line 197: If we're running this in JSON mode, should there be any other output? I presume we only want to dump a single json object out if `--json` is passed in?

Additionally, do you think it's valid to allow `--json` + `--web` opts to be specified at the same time?

niclim (approved):

PR #1439, opened 2022-12-20
Create new generic diff function

niclim:
## 🍗 Description

This is a new refactored version of our openapi diff function. This outputs diffs in the format of:
```typescript
export type ObjectDiff =
  | {
      // Added
      before?: undefined;
      after: string;
    }
  | {
      // Changed
      before: string;
      after: string;
    }
  | {
      // Removed
      before: string;
      after?: undefined;
      // Required to maintain pointers from reordering array keys
      pathReconciliation?: [number, number][];
    };
```

The plan is to:
- use this function to generate a slimmed down changes format (this PR)
- update rules runner to take in this new results format and trigger rules based on this newly changelog output
- create a `generateSpecResults` to generate a new version of this output
- create a `convertResultsToV1` function that will allow us to convert the new changes + results formats into the old `IChange` and `Result` shapes. OR publish both result formats so we can switch over in the future

## 📚 References

## 👹 QA


projects/openapi-utilities/src/diff/diff.ts
```diff
@@ -0,0 +1,205 @@
...
+        const beforeParts = jsonPointerHelpers.decode(beforePath).slice(0, -1)
+        const afterParts = jsonPointerHelpers.decode(afterPath).slice(0, -1)
+        const pathReconciliation: [number, string][] = [];
+        for( let i = 0; i < beforeParts.length; i++) {
+          const before = beforeParts[i];
+          const after = afterParts[i];
+          if (before !== after) {
+            pathReconciliation.push([i, after]);
+          }
+        }
```
niclim, line 153: See test case https://github.com/opticdev/optic/pull/1439/files#diff-e592c7df8403844274bc28cdb28360de2572020a2fe6c638df9bd70490d7280aR22 for example and https://www.notion.so/useoptic/Optimize-Changelog-Results-File-8fcd111e08904cf882af6b7a0103f977#d41d1ce10746419da30d6678824b8add for details

projects/openapi-utilities/src/diff/diff.ts
```diff
@@ -0,0 +1,209 @@
...
+  while (stack.length > 0) {
+    const [before, after] = stack.pop()!;
+    const comparisons: {
+      beforeValue: JSONValue | undefined;
+      beforePath: string;
+      afterValue: JSONValue | undefined;
+      afterPath: string;
+    }[] = [];
+
+    // TODO in the future, skip adding comparisons based on diff preprocessing step
```
niclim, line 68: When we are ready for this, we can modify this fn to take in merkel trees to optimize the speed of diff here

PR #1443, opened 2023-01-02
optic verify command

acunniffe:
## 🍗 Description
Verifying your API is working as documented (aka your OpenAPI and implementation are in sync) is an important part of shipping a great API. Writing contract tests manually is really expensive and brittle -- now you have a spec, tests, and code to keep aligned. OpenAPI has thousands of assertions about your API behavior already written down. Traffic * those assertions in the spec can be used as basic contract tests. 

This PR removes `oas` `add`, `update`, and `status` and replaces them with a new command `verify` 

This command compares traffic you've collected and the current version of the spec, to produce: 
- a list of diffs
- a list of undocumented operations 
- a list of undocumented bodies, status codes, etc. 
- a coverage report showing you what parts of your API have been tested and a final %coverage 

It exits 1 if your API and spec do not match, and 0 if they do. 

The old commands were optimized for documenting new operations (old `add` command) and patching existing ones (old `update` command). This is still possible, but now that functionality is done with flags. If a user sees diffs their next step would be to determine if the diff is a bug OR a planned change that needs to be documented. 
- if it is a bug, they need to change their API implementation and re-run verify
- if it is a planned change that needs to be documented, they can use `--patch` or `--document` to get generate the OpenAPI, or write it manually. 

Other improvements:
- copied commands into `optic` cli for more consistency 
- system proxy capture method
- lots of bug fixes in the diff engine
- undocumented path inference (auto parameterize) ie `/todos/{todoId}`
- support for ignores via:  x-optic-path-ignore

## 📚 References

https://user-images.githubusercontent.com/5900338/210245036-31a88579-66fc-433f-bdfb-35db59c6c206.mp4

## 👹 QA
Give it a try, and share some feedback


README.md
```diff
@@ -42,53 +42,37 @@ optic diff openapi.yaml --base main --check
 
 [Read breaking change documentation](https://useoptic.com/docs/breaking-changes)
 
-## Easily adopt OpenAPI
+## Easily adopt OpenAPI and keep your specs up-to-date
```
acunniffe, line 46: updated the readme in the repo with usage

projects/openapi-cli/src/captures/capture-storage.ts
```diff
@@ -0,0 +1,29 @@
...
+    .createHash('md5')
+    .update(resolvedFilepath)
+    .digest('hex');
+
+  const trafficDirectory = Path.join(
+    tmpDirectory,
+    'optic',
+    'captures',
+    specPathHash
+  );
```
acunniffe, line 24: Traffic is stored in `/tmp` by default to make it easier to use the CLI. You don't need to output hars, remember the name, then delete them

projects/openapi-cli/src/captures/system-proxy.ts
```diff
@@ -0,0 +1,90 @@
...
+  constructor(
+    private proxyUrl: string,
+    private feedback: Awaited<ReturnType<typeof createCommandFeedback>>
+  ) {}
+  private stopCommand: (() => Promise<void>) | undefined = undefined;
+
+  async start(device: string | undefined) {
+    const { port } = url.parse(this.proxyUrl);
+
+    if (platform === 'mac') {
```
acunniffe, line 22: right now the magic only works on Mac. There are clear paths to turn it on w/ windows, but that's something we should validate demand for first via mixpanel

projects/openapi-cli/src/operations/patches/index.ts
```diff
@@ -48,7 +50,7 @@ export class OperationPatch {
     try {
       const result = JsonPatch.applyPatch(
         operation,
-        [...OperationPatch.operations(patch)],
+        JsonPatch.deepClone([...OperationPatch.operations(patch)]),
```
acunniffe, line 53: This fixed all the shape diff issues we were having. Patches are mutated just like the spec is mutated. So simulating any patches that added new objects to the heap, was changing the patches that were actually applied on save. Nasty bug you hit if you have any non-trivial # of patches

projects/openapi-cli/src/operations/streams/undocumented.ts
```diff
@@ -74,27 +93,25 @@ export class UndocumentedOperations {
     spec: OpenAPIV3.Document,
     specUpdates?: AsyncIterable<OpenAPIV3.Document>
   ): UndocumentedOperations {
+    const hostBaseMap: { [key: string]: string } = {};
 
-    const basePaths =
-      spec.servers?.map((server) => {
-        // add absolute in case url is relative (valid in OpenAPI, ignored when absolute)
-        const parsed = new Url.URL(server.url, 'https://example.org');
+    spec.servers?.forEach((server) => {
```
acunniffe, line 98: Robust fix for the basePath logic. The main issues we've been having opened on the repo relate to this. The previous logic was not using the host (mistake) and brittle with trailing slashes.

projects/optic/src/init.ts
```diff
@@ -71,5 +76,12 @@ export const initCli = async () => {
   registerRulesetUpload(rulesetSubcommands, cliConfig);
   registerRulesetInit(rulesetSubcommands, cliConfig);
 
+  // commands for tracking changes with openapi
+  cli.addCommand(await captureCommand());
+  cli.addCommand(await newCommand());
+  cli.addCommand(await captureCertCommand());
+  cli.addCommand(await clearCommand());
+  cli.addCommand(await verifyCommand());
+
```
acunniffe, line 85: Added these to `optic`

projects/openapi-cli/src/captures/capture-storage.ts
Alexandre-Herve: I noticed this watching your demo video, this is really nice! I expect we'll need to let them override this default at some point.

projects/openapi-cli/src/captures/streams/sources/har.ts
```diff
@@ -84,8 +85,15 @@ export class HarEntries {
 
       let requestBodyBuffer = interaction.request.body.buffer;
       let requestContentType = interaction.request.headers['content-type'];
-      let requestBodyEncoded =
-        requestContentType && requestBodyBuffer.toString('base64');
+
+      const requestBodyEncoded =
+        requestContentType &&
+        (await toBase64(
```
Alexandre-Herve, line 91: `content-encoding` could be an [array of strings](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding).

Asserting the type here lets `toBase64` crash implicitly, which doesn't feel right. It would be cleaner to let the `toBase64` function accept a list of strings as input and explicitly throw in the event of a list being received, with an appropriate error message.

Then later we can update `toBase64` to support a list when we need it.

projects/openapi-cli/src/captures/streams/sources/har.ts
```diff
@@ -84,8 +85,15 @@ export class HarEntries {
...
+          requestBodyBuffer,
+          (interaction.response.headers['content-encoding'] as
+            | string
+            | undefined) || undefined
```
Alexandre-Herve, line 95: It's recommended to use `?? undefined` instead of `|| undefined` to avoid accidentally turning falsy but meaningful values such as `false` or `0` to `undefined` (doesn't make a difference here, this is more a general comment)

projects/openapi-cli/src/captures/system-proxy.ts
```diff
@@ -0,0 +1,90 @@
...
+      this.stopCommand = async () => {
+        await Promise.all([
+          runCommand(`networksetup -setwebproxystate "${name}" off`),
+          runCommand(`networksetup -setsecurewebproxystate "${name}" off`),
+        ]);
+        this.feedback.notable(`Mac System Proxy settings cleared`);
+      };
+    } else {
+      console.log(
+        `automatic proxy configuration is not supported on ${platform}`
```
Alexandre-Herve, line 48: Maybe give them a way to request it? A ProductBoard link?

projects/openapi-cli/src/commands/verify.ts
```diff
@@ -0,0 +1,301 @@
...
+            spec,
+            sourcemap
+          );
+          analytics.push({
+            event: 'openapi.verify.patch',
+            properties: patchStats,
+          });
+        }
+
+        console.log(chalk.gray('-'.repeat(process.stdout.columns) + '\n'));
```
Alexandre-Herve, line 125: What's the difference between `feedback` and directly logging to the console? I would have the feeling that sending everything to `feedback` to determine the level of logging etc would make sense, but I'm not sure I have the big picture in mind.

Alexandre-Herve (approved):
Looks really cool, feels like it makes a lot of sense

projects/openapi-cli/src/captures/capture-storage.ts
acunniffe: You can output to a file by passing `-o file.har` to `capture` 
And verify can take a file in (to override the tmp dir) the same way `--har file.har` 

Do you think we need more? Or just to document those options well?

projects/openapi-cli/src/captures/streams/sources/har.ts
acunniffe: Oh interesting. I thought that was just the header type -- didn't realize it was part of HTTP to apply the encoding in the order of the array. 

I'll look into this more tonight. 

Separately there's the discussion of whether we want encoding in our HARs. 
Right now the capture flow decodes the message before writing it to disk. There are tradeoffs (some information is lost) and it makes marginally slower?? but verify faster. 

We could also just save the bodies as-is and do the decoding on the read side (possibly being unable to decode some messages)

projects/openapi-cli/src/commands/verify.ts
acunniffe: There shouldn't be a difference. I'll pass the feedback around and centralize these calls

projects/openapi-cli/src/commands/capture.ts
```diff
@@ -29,27 +30,46 @@ export async function captureCommand(): Promise<Command> {
   let certCommand = await captureCertCommand();
```
niclim, line 30: This looks like it's registered twice (here, and in the `optic init.ts` file)

projects/openapi-cli/src/commands/reporters/feedback.ts
```diff
@@ -1,14 +1,13 @@
 import { Command } from 'commander';
 import { Writable } from 'stream';
 import { trackEvent, flushEvents } from '../../segment';
-
+import chalk from 'chalk';
 export async function createCommandFeedback(
```
niclim, line 5: This doesn't need to be async - which I think is causing all the other commands to be "async"

projects/optic/src/init.ts
```diff
@@ -71,5 +76,12 @@ export const initCli = async () => {
   registerRulesetUpload(rulesetSubcommands, cliConfig);
   registerRulesetInit(rulesetSubcommands, cliConfig);
 
+  // commands for tracking changes with openapi
+  cli.addCommand(await captureCommand());
+  cli.addCommand(await newCommand());
+  cli.addCommand(await captureCertCommand());
+  cli.addCommand(await clearCommand());
+  cli.addCommand(await verifyCommand());
```
niclim, line 84: Do we want these to be under some sort of subcommand? Do you think it'll clash with any of the existing commands / commands we're planning to build?

projects/openapi-cli/src/captures/streams/sources/har.ts
```diff
@@ -192,6 +203,43 @@ export class HarEntries {
...
+ */
+async function toBase64(
+  buffer: Buffer,
+  encoding: string | string[] | undefined
+): Promise<string> {
+  try {
+    const firstEncoding = Array.isArray(encoding) ? encoding[0] : encoding;
+
+    if (Array.isArray(encoding) && encoding.length > 1)
+      throw new Error('multiple content-encodinings are not supported');
```
niclim, line 220: `encodinings`

projects/openapi-cli/src/commands/capture-cert.ts
```diff
@@ -24,9 +35,31 @@ export async function captureCertCommand(): Promise<Command> {
...
         certStore.delete();
-        return await feedback.notable(
-          'CA certificate deleted for use by proxy'
-        );
+        return feedback.notable('CA certificate deleted for use by proxy');
+      }
+
+      console.log(`We take privacy seriously. You should understand how Optic's TLS capturing works:
+
+THe Optic proxy is a man-in-the-middle proxy that only runs on and logs data to your local machine.
```
niclim, line 43: `The`

projects/openapi-cli/src/commands/capture-cert.ts
```diff
@@ -24,9 +35,31 @@ export async function captureCertCommand(): Promise<Command> {
...
+3. The Optic proxy is also a man-in-the-middle. It tries to read traffic to your API hostnames: i.e. localhost:3005 or https://api.example.com to learn and verify API behaviors
+
+By default, there is no way for Optic to read any TLS traffic. If you want to use Optic to read TLS traffic, it will need a trusted CA Certification. This setup wizard will:
+
+1. Generate a CA Certificate on your local machine (it will not phone it home)
+2. Add that certificate to your trust chain 
+3. Use that Cert to terminate TLS and log traffic from the target hostnames. 
+
+If you do this, you are man-in-the-middling yourself. Optic will theoretically be able to see any TLS traffic when the 'capture' command is activated.   
+You are welcome to read the source code github.com/opticdev/optic
```
niclim, line 56: `https://` i think is needed so it registers as a link in a shell (i think)

niclim (approved):

projects/openapi-cli/src/commands/capture.ts
acunniffe: good catch. removed and tested. I'm not sure why it was done that way before

projects/openapi-cli/src/commands/reporters/feedback.ts
acunniffe: good catch. We changed Chalk to the version that's compatible with our code so we don't need to dynamically import it anymore. I've updated all the usages / downstreams

projects/optic/src/init.ts
acunniffe: I actually think `optic oas [...]` isn't bad. I'll make that update

PR #1446, opened 2023-01-04
add in lint-staged + husky

niclim:
## 🍗 Description

Part of moving the open source stuff out of monorail meant we lost the pre-commit hooks. adding the code back. to enable this yourself, run the following commands
```
$ npx husky-init && yarn

# then update your .husky/pre-commit file to look like
$ cat .husky/pre-commit

#!/bin/sh
. "$(dirname "$0")/_/husky.sh"

yarn run lint-staged

```

## 📚 References

## 👹 QA


.gitignore
```diff
@@ -15,3 +15,4 @@ build
 .nyc_output
 
 .env
+.husky
```
niclim, line 18: This is what we have in monorail. Alternatively we could commit the `.husky/pre-commit` file so all the lint-staged will be the same. The reason we didn't have this in monorail is because there's non-JS code in the monorepo and dont want a hard dep on npm / JS for making commits (which init husky does)

nalanj (approved):

PR #1453, opened 2023-01-04
Update diff to handle uploads

niclim:
## 🍗 Description

PR includes changes to the `optic diff` command so that we can upload from diff. This includes the following changes:
- Adds authentication and more git context into the CLI config initialization
- `--ruleset` as an arg. The way rulesets are selected are by: looking at a the `--ruleset` arg, then looking at the `x-optic-ruleset` key on the spec, then looking at the `optic.dev.yml` config (which defaults to breaking changes if no config found)
- If a user is logged in / token is passed in:
  - [future pr] A user can pass in `spec-1-id` and `spec-2-id` which can diff specs in cloud (TBD what format)
  - If a spec has `x-optic-url` and there's no changes from the git (i.e. git state is clean)
    - [future pr] then we upload the spec with `git:` tags AND upload results + tag run

I think we're safe to merge this, since the extra features added is a new `--ruleset` keyword, and the authenticated flow is only triggered by optic login (not implemented) or by passing `optic token` as an env variable

I'll implement the following in subsequent PRs:
- Connecting up the APIs
- Supporting cloud spec ids

## 📚 References

## 👹 QA


projects/optic/src/config.ts
```diff
@@ -24,45 +28,28 @@ export type RawYmlConfig = {
...
   ruleset: [{ name: 'breaking-changes', config: {} }],
+  isAuthenticated: false,
+  client: createOpticClient('no_token'),
 };
 
 const ajv = new Ajv();
 const configSchema = {
   type: 'object',
   properties: {
-    files: {
```
niclim, line 51: Removed because we aren't using this anymore

projects/optic/src/config.ts
```diff
@@ -177,3 +163,31 @@ export const initializeRules = async (config: RawYmlConfig) => {
 
   config.ruleset = [...rulesetMap.values()];
 };
+
+export async function initializeConfig(): Promise<OpticCliConfig> {
```
niclim, line 167: Moved the code from init into here, to be able to add authentication + client initalization here. CLIConfig might not be the correct word for this anymore though - could rename to something else

projects/optic/src/commands/diff/diff.ts
```diff
@@ -134,10 +139,15 @@ const openBrowserToPage = async (url: string) => {
 
 const getBaseAndHeadFromFiles = async (
   file1: string,
-  file2: string
+  file2: string,
+  config: OpticCliConfig
 ): Promise<[ParseResult, ParseResult]> => {
   try {
-    return Promise.all([getFileFromFsOrGit(file1), getFileFromFsOrGit(file2)]);
+    // TODO update function to try download from spec-id cloud
```
niclim, line 146: I'll do this in a follow up PR

projects/optic/src/utils/spec-loaders.ts
```diff
@@ -2,26 +2,146 @@ import fs from 'node:fs/promises';
...
+  ParseOpenAPIResult,
+  parseOpenAPIWithSourcemap,
+} from '@useoptic/openapi-io';
+import { OpticCliConfig, VCS } from '../config';
+import { resolveGitRef } from './git-utils';
 
 const exec = promisify(callbackExec);
 
-export type { ParseResult };
+export type ParseResult = ParseOpenAPIResult & {
```
niclim, line 19: I moved this code in from `optic-ci`, we're diverging the implementations since we need to also track which git sha these files came from (for tagging / uploading)

projects/optic/src/utils/spec-loaders.ts
```diff
@@ -2,26 +2,146 @@ import fs from 'node:fs/promises';
...
+      return {
+        ...(await parseOpenAPIFromRepoWithSourcemap(
+          input.name,
+          config.root,
+          input.branch
+        )),
+        isEmptySpec: false,
+        context: {
+          vcs: 'git',
+          sha: await resolveGitRef(input.branch),
```
niclim, line 116: We need to resolve the git ref since we can pass in branch names here. i.e. branch names will be converted into the shas they represent

projects/optic/src/commands/diff/diff.ts
```diff
@@ -258,34 +286,56 @@ const getDiffAction =
...
+      }
+
+      const shouldUploadResults =
+        (shouldUploadBaseSpec || baseParseResult.isEmptySpec) &&
+        (shouldUploadHeadSpec || headParseResult.isEmptySpec);
+
+      if (shouldUploadResults) {
+        // TODO upload results
+      }
+    }
```
niclim, line 337: I'm going to implement this in a follow up PR

projects/optic/src/utils/spec-loaders.ts
```diff
@@ -2,26 +2,146 @@ import fs from 'node:fs/promises';
...
+      };
+    }
+    case SpecVersionFrom.file:
+      return {
+        ...(await parseOpenAPIWithSourcemap(
+          path.resolve(workingDir, input.filePath)
+        )),
+        isEmptySpec: false,
+        context:
+          config.vcs?.type === VCS.Git && config.vcs.status === 'clean'
```
niclim, line 127: We only want to associate a local file if the git status is clean (i.e no uncommitted changes). We could change the definition of clean later on to mean only files that have `.yml` or `.json` changes - right now it's _any_ change.

projects/optic/src/commands/diff/diff.ts
```diff
@@ -258,34 +286,54 @@ const getDiffAction =
...
```
niclim, line 335: I'm going to implement this in a follow up PR but I want to make sure this logic makes sense.

We only upload if there's an API Id we can detect (this will be pulled from the optic-url, first from the head spec and then from the base spec if head spec does not have it (i.e. when spec is empty)).

We then only upload the specs if it is unchanged in git. We only upload the results if both the base + head are unchanged in git OR they are empty.

projects/optic/src/__tests__/integration/diff.test.ts
```diff
@@ -12,7 +12,10 @@ jest.setTimeout(30000);
 
 describe('diff', () => {
   test('two files, no repo or config', async () => {
-    const workspace = await setupWorkspace('diff/files-no-repo');
+    const workspace = await setupWorkspace('diff/files-no-repo', {
+      repo: true,
```
nalanj, line 16: Seems like maybe the test name here should be updated if we're including a repo on the test.

projects/optic/src/commands/diff/diff.ts
nalanj: > We then only upload the specs if it is unchanged in git. We only upload the results if both the base + head are unchanged in git OR they are empty.

I'm not sure I understand this. Why wouldn't we upload them if they're changed?

projects/optic/src/commands/diff/generate-rule-runner.ts
```diff
@@ -1,31 +1,59 @@
...
 type InputPayload = Parameters<typeof prepareRulesets>[0];
 
+const getStandardToUse = async (
+  options: {
+    specRuleset?: string;
+    rulesetArg?: string;
+    config: OpticCliConfig;
+  }
+): Promise<OpticCliConfig['ruleset']> => {
+  // We always take the --ruleset arg as priority, then the ruleset on the API spec (from the head), then fallback to the optic.dev.yml config
```
nalanj, line 23: > then fallback to the optic.dev.yml config

Aidan mentioned wanting to get rid of optic.dev.yml. Is that going forward?

projects/optic/src/config.ts
nalanj: Seems fine to me.

projects/optic/src/utils/spec-loaders.ts
nalanj: I think the correct way on it would be to look at all of the files that compose the spec in the working directory and determine if any of them have changed. I think just checking the entire directory is fine for now, though.

nalanj (approved):
Approving with a few questions and comments.

projects/optic/src/commands/diff/diff.ts
niclim: Ah what I mean here is that the working directory is clean

projects/optic/src/__tests__/integration/diff.test.ts
niclim: Ah this should actually be `repo: false` - updated test

projects/optic/src/commands/diff/generate-rule-runner.ts
niclim: I don't see a reason we need to - I figure we can keep around this and use it as a fallback for an entire repo config (which also has a default fallback of breaking-changes on)

What's the reason we'd want to get rid of it (cc @acunniffe)

PR #1474, opened 2023-01-06
Implement `optic api add`

niclim:
## 🍗 Description

This PR implements `optic api add`. This can be called in two ways:
- `optic api add` - adds all detected openapi files in a repo
- `optic api add <file_name> --history-depth n` - adds a single api file with history up to a depth of n (0 is all history)

If token is in multiple orgs, we prompt for which org to add to. If a user does not select a ruleset, we prompt to see if they would like to create a standard or use an existing standard.

Screen Shot 2023-01-06 at 2 04 04 PM

Also changed `ruleset` -> `standard`

TODO:
- [x] add notes for ruleset
- [x] write x-optic-url + x-optic-ruleset
- [x] add tests

TODO in follow up PR
- [ ] connect up APIs (fetch org from APIs, upload api, upload spec)

## 📚 References

## 👹 QA


projects/optic/src/commands/api/add.ts
```diff
@@ -0,0 +1,215 @@
...
+      logger.error('Must be logged in to add APIs. Log in with `optic login`');
+      process.exitCode = 1;
+      return;
+    } else if (!config.vcs || config.vcs?.type !== VCS.Git) {
+      logger.error('Must have git in path and be in a git repo to add apis');
+      process.exitCode = 1;
+      return;
+    }
+
+    if ('todo fetch orgs related to token') {
```
niclim, line 88: @nalanj I think we'll need an endpoint that returns orgs a token has access to here

projects/optic/src/commands/api/add.ts
```diff
@@ -0,0 +1,215 @@
...
+        continue;
+      }
+
+      const opticUrl = parseResult.jsonLike[OPTIC_URL_KEY];
+      const apiId = opticUrl
+        ? 'todo get optic id from url'
+        : 'todo make API call';
+
+      logger.info(`Found OpenAPI Spec at ${path}`);
+      for await (const sha of shas) {
```
niclim, line 130: Here we traverse the history until we hit:
- an invalid API
- an empty spec

I'm not convinced this is correct, and maybe instead we should just ignore the empty / invalid specs and continue, instead of stopping. Some valid cases:
- Someone uploads an invalid API, and then fixes it in a follow up PR
- Someone deletes and API and then reverts the changes

projects/optic/src/__tests__/integration/api-add.test.ts
```diff
@@ -0,0 +1,90 @@
...
+    oldEnv = { ...process.env };
+    process.env.LOG_LEVEL = 'info';
+  });
+
+  afterEach(() => {
+    jest.resetAllMocks();
+    process.env = { ...oldEnv };
+  });
+
+  describe('no interaction', () => {
```
niclim, line 38: I'm not really sure how we can get the integration test to run with interaction / what needs to be changed - I just forced this test to not prompt for anything

projects/optic/src/commands/api/add.ts
nalanj: I just dug through detecting renames and I don't think it's worth the effort. So yeah, I'd say just stop when you get to the first copy of the spec, and hopefully teams don't rename them too terribly often. We could write some kind of plumbing spec upload code for teams that want to walk renames.

nalanj (approved):
Handful of comments, but overall looks really good and exactly what I had imagined.

projects/optic/src/__tests__/integration/api-add.test.ts
```diff
@@ -0,0 +1,90 @@
...
+
+describe('optic api add', () => {
+  let oldEnv: any;
+  beforeEach(() => {
+    oldEnv = { ...process.env };
+    process.env.LOG_LEVEL = 'info';
+  });
+
+  afterEach(() => {
+    jest.resetAllMocks();
```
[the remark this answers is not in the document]
niclim: Ah I did set it, we don't need this here

projects/optic/src/commands/api/add.ts
```diff
@@ -0,0 +1,282 @@
...
+      logger.error(
+        'Invalid argument combination: Cannot set a history-depth !== 1 when no spec path is provided'
+      );
+      process.exitCode = 1;
+      return;
+    } else if (!config.isAuthenticated) {
+      logger.error('Must be logged in to add APIs. Log in with `optic login`');
+      process.exitCode = 1;
+      return;
+    } else if (!config.vcs || config.vcs?.type !== VCS.Git) {
```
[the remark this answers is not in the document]
niclim: Yeah, I just modified the code to handle this case

PR #1475, opened 2023-01-06
Spectral plus ruleset

acunniffe:
## 🍗 Description
Run your spectral configs within Optic. Some rulesets can run `added`, some `always`. 

## 📚 References
Config:
```yaml
spectral: 
   added:
     - ./latest-pagination.yml 
   always:
     - ./metadata.yml
     - ./operations.yml
```

## 👹 QA


projects/standard-rulesets/src/spectral-oas-v6/index.ts
```diff
@@ -41,7 +41,7 @@ export class SpectralOasV6Ruleset extends SpectralRule {
             error.keyword === 'enum'
               ? `${error.message} ${error.params.allowedValues}`
               : error.message;
-          return `- ruleset/naming${error.instancePath} ${message}`;
+          return `- ruleset/spectral-oas-v6${error.instancePath} ${message}`;
```
acunniffe, line 44: I think this was a typo from a previous release

projects/standard-rulesets/src/spectral/index.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  async runRules(inputs: {
+    context: any;
+    nextFacts: IFact[];
+    currentFacts: IFact[];
+    changelog: IChange[];
+    nextJsonLike: OpenAPIV3.Document<{}>;
+    currentJsonLike: OpenAPIV3.Document<{}>;
+  }): Promise<Result[]> {
+    const rulesets = await this.preparedRulesets();
```
acunniffe, line 48: some choices to make here...

right now if even 1 of the rulesets you try to load fails you'll get an error as you run the rules. If we push that error to the fromOpticConfig it'll fail before trying to run, but those have currently been sync and this would make them async

We could swallow errors and warn if any of these failed to load or run. I think it's fine to fail them all right now, but was curious what @niclim imagined for rulessets that fail to stand up or run

projects/standard-rulesets/src/spectral/index.ts
```diff
@@ -0,0 +1,105 @@
...
+
+    const allResults = await Promise.all(
+      rulesets.map((ruleset) => ruleset.runRules(inputs))
+    );
+
+    return allResults.flat(1);
```
acunniffe, line 54: this should retain order between runs

projects/standard-rulesets/src/spectral/index.ts
```diff
@@ -0,0 +1,105 @@
...
+    }
+  }
+}
+
+async function uriToSpectral(uri: string) {
+  const spectral = new Spectral();
+  const ruleset: Ruleset = await bundleAndLoadRuleset(uri, { fs, fetch });
+  // setting explicitly because of a poor choice in Spectral core. The instanceof is fragile when you load code / types from different modules and combine them (like we have to with bundle and core)
+  // https://github.com/stoplightio/spectral/blob/a1bd6d29b473aff257dbf66264ebdf471fae07cc/packages/core/src/spectral.ts#L87
+  spectral.ruleset = ruleset;
```
acunniffe, line 103: this is annoying, but I don't think we have a good way around it. Spectral is using instanceOf and the types are different since they load in from different modules (bundle and core)

projects/standard-rulesets/src/spectral/index.ts
```diff
@@ -0,0 +1,105 @@
...
+      items: {
+        type: 'string',
+        description: 'URI of spectral ruleset file (file or URL)',
+      },
+    },
+  },
+};
+const validateConfigSchema = ajv.compile(configSchema);
+
+export class SpectralRulesets {
```
niclim, line 29: I think you might need to extend `ExternalRuleBase` here (similar to spectral rule) to get the correct `isInstance` check in the rule runner

niclim (approved):
looks good

projects/standard-rulesets/src/index.ts
```diff
@@ -6,10 +6,12 @@ import { BreakingChangesRuleset } from './breaking-changes';
 import { NamingChangesRuleset } from './naming-changes';
 import { ExamplesRuleset } from './examples';
 import { SpectralOasV6Ruleset } from './spectral-oas-v6';
+import { SpectralRulesets } from './spectral';
 
 export const StandardRulesets = {
   'breaking-changes': BreakingChangesRuleset,
   naming: NamingChangesRuleset,
   'spectral-oas-v6': SpectralOasV6Ruleset,
```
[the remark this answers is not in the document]
acunniffe: Agreed -- we did not do much with it.

projects/standard-rulesets/src/spectral/index.ts
acunniffe: done

projects/standard-rulesets/src/index.ts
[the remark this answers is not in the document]
niclim: I think there's some mentions of it in the README (in standard rulesets / rulesets-base) - would also need to clean that up as well

PR #1477, opened 2023-01-09
implement `optic diff-all`

niclim:
## 🍗 Description

PR implements:
- optic diff-all, compare-to defaults to the working dir, and compare-from defaults to HEAD~1
- can set a `--ruleset` (same logic as diff)
- can open --check
- uploads specs at the end, and opens in web view
- implements a --json key (same as diff)

TODO:
- [x] finish implementing JSON changelog
- [x] write test for json changelog

In a future PR
- connect this command to APIs

## 📚 References

## 👹 QA


projects/optic/src/utils/spec-loaders.ts
```diff
@@ -136,15 +136,18 @@ async function specFromInputToResults(
...
+  strict: boolean
 ): Promise<ParseResult> => {
   const file = await specFromInputToResults(
     parseSpecVersion(filePathOrRef, defaultEmptySpec),
     config,
     process.cwd()
   ).then((results) => {
-    validateOpenApiV3Document(results.jsonLike, results.sourcemap);
+    validateOpenApiV3Document(results.jsonLike, results.sourcemap, {
+      strictOpenAPI: strict,
```
niclim, line 149: Fix to strictly check OpenAPI on `to` files, and not to strictly check on `from` files.

The default is strict check, we want to _not_ strictly check API files on from files, since you can never edit a `from` file.

projects/optic/src/utils/git-utils.ts
```diff
@@ -49,3 +49,25 @@ export const resolveGitRef = async (ref: string): Promise<string> =>
...
+    const cb = (err: unknown, stdout: string, stderr: string) => {
+      if (err || stderr || !stdout) reject(err || stderr);
+      resolve(
+        stdout
+          .trim()
+          .split('\n')
+          .filter((path) => !!path)
+      );
+    };
+    const command = `toplevel=$(git rev-parse --show-toplevel) && \
```
niclim, line 66: Extracted this command out from `api add` to be able to use it for diff-all (modifying to grep from a git tree optionally)

projects/optic/src/commands/diff/diff-all.ts
```diff
@@ -0,0 +1,365 @@
...
+        )
+      );
+      logger.info(
+        'Run the `optic api add` command to add these specs to optic'
+      );
+      logger.info(warnings.missingOpticUrl.map((f) => f.path).join('\n'));
+      logger.info('');
+    }
+
+    if (warnings.unparseableFromSpec.length > 0) {
```
niclim, line 312: I'm not sure if this should fail with `exitCode=1` - i'm leaning towards no, because if you have an unparsable from spec, you shouldn't fail since there's nothing you can do here, but we cannot run the spec

projects/optic/src/commands/diff/diff-all.ts
```diff
@@ -0,0 +1,365 @@
...
+      }
+    );
+
+    const { warnings, results } = await computeAll(
+      candidatesMap,
+      config,
+      options
+    );
+
+    if (warnings.missingOpticUrl.length > 0) {
```
niclim, line 299: I think we landed process.exitCode !== 1 in this case, but happy to change either way

projects/optic/src/commands/diff/diff-all.ts
```diff
@@ -0,0 +1,365 @@
...
+      (typeof fromParseResults.jsonLike[OPTIC_URL_KEY] === 'string' &&
+        toParseResults.isEmptySpec)
+    ) {
+      const { specResults, checks } = await compute(
+        [fromParseResults, toParseResults],
+        config,
+        options
+      );
+
+      if (config.isAuthenticated) {
```
niclim, line 164: This feels a little weird to be here (this will mean we run the diff and upload and open in web, before moving on to the next command).

The reason it's here is because `from`, `to` and `results` can take up a lot of memory, and sequentially processing it means we don't need to load all results into memory. 

That being said, there's a couple of things:
- bulk-compare (from optic-ci) loads all comparisons into memory, and then dumps out results / uploads and works for our current usage
- It might be less of an issue when we optimize the changelog / results output

I think it's probably _fine_ here, but it can lead to _some_ partial runs if there's exceptions thrown

projects/optic/src/commands/diff/diff-all.ts
```diff
@@ -0,0 +1,365 @@
...
+          meta
+        );
+        const anonymousId = await getAnonId();
+        trackEvent('optic.diff_all.view_web', anonymousId, {
+          compressedDataLength: compressedData.length,
+        });
+        await open(`${config.client.getWebBase()}/cli/diff#${compressedData}`, {
+          wait: false,
+        });
+      }
```
niclim, line 212: Same comment as above (it feels weird to open as part of the compute flow)

projects/optic/src/commands/diff/compressResults.ts
```diff
@@ -0,0 +1,59 @@
+import zlib from 'node:zlib';
```
niclim, line 1: Moved from `diff`

projects/optic/src/commands/diff/compute.ts
```diff
@@ -0,0 +1,42 @@
+import { generateSpecResults } from '@useoptic/openapi-utilities';
```
niclim, line 1: also moved from `diff`

projects/optic/src/commands/diff/diff-all.ts
```diff
@@ -0,0 +1,452 @@
...
+        `Error: optic diff-all must be called from a git repository.`
+      );
+      process.exitCode = 1;
+      return;
+    }
+
+    if (options.json) {
+      // For json output we only want to render json
+      logger.setLevel('silent');
+    }
```
niclim, line 333: If there's a better way to do this we can - i also thoguht maybe we could just dump a json output file as well

projects/optic/src/commands/diff/diff-all.ts
```diff
@@ -0,0 +1,452 @@
...
+    results,
+  };
+}
+
+type Result = Awaited<ReturnType<typeof compute>> & {
+  fromParseResults: ParseResult;
+  toParseResults: ParseResult;
+  from?: string;
+  to?: string;
+};
```
niclim, line 237: I'm not a fan of this results format, it's pretty massive given it includes the entire specs + results, which can mean we are using a ton of memory. I originally had this inside the `compute` function, but moved it out for code simplicity (to handle JSON / terminal output diffs computed from results)

I think this is _ok_ for now, since this is how `bulk-compare` works, but I'm pretty sure we'll run into issues with a large number of diff comparisons

This will be made better when we have the new results format which will drastically reduce the memory usage, however keeping all files in memory will keep the memory usage high

Thoughts?

projects/optic/src/commands/diff/diff-all.ts
```diff
@@ -0,0 +1,452 @@
...
+  compareFrom: string;
+  ruleset?: string;
+  check: boolean;
+  web: boolean;
+  json: boolean;
+};
+
+// Match up the to and from candidates
+// This will return the comparisons we can try to run
+function matchCandidates(
```
niclim, line 82: Here we pull both the sets of `from sha` and `to sha` and match up the results.

We later disregard certain files once we read the contents (below)

projects/optic/src/commands/diff/diff-all.ts
```diff
@@ -0,0 +1,447 @@
...
+    missingOpticUrl: [],
+    unparseableFromSpec: [],
+    unparseableToSpec: [],
+  };
+
+  const results: Result[] = [];
+
+  for await (const [_, candidate] of candidatesMap) {
+    // try load both from + to spec
+    let fromParseResults: ParseResult;
```
niclim, line 130: Here we'll only compare if:
- `to` spec is parseable (with strict validation)
- `from` spec is parseable (with not strict validation)
- `to` spec has the x-optic-url (we've opted in) OR from spec has `x-optic-url` AND to spec is empty (i.e. we've deleted the `to` spec)

projects/optic/src/commands/diff/diff-all.ts
nalanj: I think it's fine until it's not

projects/optic/src/commands/diff/diff-all.ts
nalanj: I think this is fine for now. We'll def. hear about it if it's an issue.

nalanj (approved):
Looks good. I def. think we'll hit perf issues over time, but I also think it'd be pre optimizing for us to worry too much about it right now.

PR #1482, opened 2023-01-10
spectral plus using spectral CLI

acunniffe:
## 🍗 Description
Instead of eating all the responsibility for running yaml spectral rules, downloading `.js` rules running with the spectral runtime, etc we're just piping to the spectral CLI for spectral plus now. 

Case when spectral fails to load ruleset
Screenshot 2023-01-10 at 2 37 29 PM
n

Case when spectral can load ruleset and delivers results 
Screenshot 2023-01-10 at 2 40 33 PM

Case where user does not have spectral CLI installed 
Screenshot 2023-01-10 at 2 41 22 PM

## 📚 References

## 👹 QA


projects/rulesets-base/src/extended-rules/spectral-rule.ts
```diff
@@ -145,3 +161,50 @@ export class SpectralRule extends ExternalRuleBase {
...
+  spectralRuleset: string,
+  afterSpecAbsolutePath: string
+): Promise<string> {
+  const resultsOutput = path.join(
+    os.tmpdir(),
+    `spectral-output-${Math.floor(Math.random() * 100000)}.json`
+  );
+
+  return new Promise((resolve, reject) => {
+    const spectralRunning = spawn(
```
acunniffe, line 175: Spectral has a bug that truncates stdout, they suggest using `--o` and a results file

projects/rulesets-base/src/extended-rules/spectral-rule.ts
```diff
@@ -145,3 +161,50 @@ export class SpectralRule extends ExternalRuleBase {
...
   }
 }
+
+async function runSpectral(
+  spectralRuleset: string,
+  afterSpecAbsolutePath: string
+): Promise<string> {
+  const resultsOutput = path.join(
+    os.tmpdir(),
+    `spectral-output-${Math.floor(Math.random() * 100000)}.json`
```
niclim, line 171: a `nanoid` or `uuid` might be more appropriate here, but this should also be fine (cant remember which one we have in rulesets base but we have at least one of them)

projects/standard-rulesets/package.json
```diff
@@ -20,13 +20,11 @@
   },
   "dependencies": {
     "@stoplight/spectral-core": "^1.8.1",
-    "@stoplight/spectral-ruleset-bundler": "^1.5.0",
     "@stoplight/spectral-rulesets": "^1.14.1",
```
niclim, line 23: Correct me if I'm wrong, but we can remove spectral-core and rulesets from standard rulesets? We just need to keep it around in the `rulesets-base` package for now

niclim (approved):

PR #1483, opened 2023-01-10
Implement `optic spec push` command

niclim:
## 🍗 Description

Implements the optic spec push command. 

Validation:
- Spec must exist in current working dir and must be parsable / pass validation
- Spec must be added (i.e. have a x-optic-url key)
- At least 1 tag must be added (either via git, or manually)

Tags:
- are automatically added when in git, and working dir is clean
- must be alphanumeric or `: - _` characters

TODO
- [x] Write tests

## 📚 References

## 👹 QA


projects/optic/src/__tests__/integration/spec-push.test.ts
```diff
@@ -0,0 +1,87 @@
...
+      workspace,
+      'spec push ./spec.yml --tag env:production,the-favorite-api'
+    );
+    expect(code).toBe(0);
+    expect(
+      normalizeWorkspace(workspace, sanitizeOutput(combined))
+    ).toMatchSnapshot();
+  });
+
+  test('requires at least one tag to push', async () => {
```
nalanj, line 53: Should we relax this for general-purpose-ness of the command? I could see wanting to push a spec up and tag it later or even some situation where I'd want to push a spec from one place, push from another place, and then do a run against it.

In some ways I feel like this command is plumbing for building up all sorts of custom workflows, so I think we should keep it as flexible as possible.

projects/optic/src/commands/spec/push.ts
```diff
@@ -0,0 +1,139 @@
...
+    .command('push', { hidden: true })
+    .configureHelp({
+      commandUsage: usage,
+    })
+    .addHelpText('after', helpText)
+    .description('Push a spec version')
+    .argument('[spec_path]', 'path to file to push')
+    .option(
+      '--tag <tag>',
+      'Adds additional tags to the spec version. If the spec is unchanged from git, the git tag will automatically be included. Tags must be alphanumeric or the - _ : characters'
```
nalanj, line 37:
```suggestion
      'Adds additional tags to the spec version. In git repositories with a clean working directory, a git tag will automatically be included. Tags must be alphanumeric or the - _ : characters'
```

nalanj (approved):

projects/optic/src/__tests__/integration/spec-push.test.ts
niclim: I don't have strong opinions here, I just wasnt sure how we were intending to display this and if specs would just get lost if we didn't have a tag - updating and relaxing the requirement

PR #1490, opened 2023-01-11
Connect optic diff to apis

niclim:
## 🍗 Description

PR connects up the `optic diff` function to the backend endpoints.

## 📚 References

## 👹 QA


projects/optic/src/utils/cloud-specs.ts
```diff
@@ -0,0 +1,78 @@
...
+    api_id: apiId,
+    checksum,
+  });
+
+  await uploadFileToS3(
+    result.check_results_url,
+    Buffer.from(JSON.stringify(opts.specResults))
+  );
+
+  await opts.client.createRun({
```
niclim, line 72: @nalanj Could we also return the web url in `createRun` - alternatively, I could construct the URL from the run id, but I'd need to fetch the org id somewhere and in the diff / diff-all cases, it's sometime ambiguous where the org id comes from

projects/optic/src/utils/cloud-specs.ts
```diff
@@ -0,0 +1,78 @@
...
+}
+
+export async function uploadRun(
+  apiId: string,
+  opts: {
+    fromSpecId: string;
+    toSpecId: string;
+    client: OpticBackendClient;
+    specResults: {
+      // TODO change this type
```
niclim, line 54: Will be updated in https://github.com/opticdev/optic/pull/1506

projects/optic/src/__tests__/integration/__snapshots__/diff.test.ts.snap
```diff
@@ -1002,3 +1002,29 @@ Configure check rulesets in optic cloud or your local optic.dev.yml file.
...
+      at [4m$workspace$/spec.json:24:565[24m
+
+
+
+1 operation changed
+[32m[1m0 checks passed[22m[39m
+[31m[1m1 checks failed[22m[39m
+
+Configure check rulesets in optic cloud or your local optic.dev.yml file.
+Uploaded results of diff to http://localhost:3000/organizations/org-id/runs/api-id
```
niclim, line 1027: errr this is wrong, but fixed in https://github.com/opticdev/optic/pull/1516

projects/optic/src/commands/diff/diff.ts
```diff
@@ -232,22 +245,57 @@ const getDiffAction =
...
+      } else if (baseParseResult.isEmptySpec) {
+        baseSpecId = EMPTY_SPEC_ID;
       }
-      if (shouldUploadHeadSpec) {
-        // TODO upload spec
+      if (headParseResult.context && specDetails) {
+        let tags: string[] = [];
+        if (headParseResult.context.vcs === VCS.Git) {
+          const currentBranch = await Git.getCurrentBranchName();
+          tags = [`git:${headParseResult.context.sha}`, `git:${currentBranch}`];
```
nalanj, line 273: I would change the branch tag prefix to be something like `branch` or `gitbranch`

projects/optic/src/utils/cloud-specs.ts
```diff
@@ -0,0 +1,78 @@
...
+  const result = await opts.client.prepareSpecUpload({
+    api_id: apiId,
+    spec_checksum,
+    sourcemap_checksum,
+  });
+  if ('upload_id' in result) {
+    await Promise.all([
+      uploadFileToS3(
+        result.spec_url,
+        Buffer.from(JSON.stringify(opts.spec.jsonLike))
```
nalanj, line 28: You'll need to stable stringify these too, or S3 will reject the upload as not matching the checksum.

projects/optic/src/utils/checksum.ts
```diff
@@ -0,0 +1,10 @@
+import { createHash } from 'crypto';
+import stableStringify from 'json-stable-stringify';
+
+export function computeChecksum(file: any): string {
+  const hash = createHash('sha256');
+
+  hash.update(stableStringify(file));
```
nalanj, line 7: It may make more sense to just stable stringify before we checksum and then use it to create the checksum and to upload, since s3 will reject a blob that doesn't match the requested checksum.

nalanj (approved):
Left a few comments - specifically the checksum ones are fairly important to note.

PR #1491, opened 2023-01-11
Create new compare specs functions

niclim:
## 🍗 Description

The goal of this PR is to create a compare spec function that can run in parallel with the current diff flow and outputs the new output format. I ended up deciding to keep around the old facts to changelog work for the rule runner migration, and instead just handle the outputs there. This was because we want to mimic that behavior, and it's going to take a pretty big rule runner migration to update therule runner to use the new diff objects.

I figure we can migrate the rule runner when we need to handle that case. It's more important for us to upload a new format, and have this function take a performance hit to get this out faster

Note that this PR does not change anything how things currently run - it's just the plumbing for the code here.

## 📚 References

## 👹 QA


projects/openapi-utilities/src/openapi3/traverser.ts
```diff
@@ -0,0 +1,505 @@
+import { jsonPointerHelpers as jsonPointer } from '@useoptic/json-pointer-helpers';
```
niclim, line 1: This is straight copy pasted from the other traverser, but only emits JSON paths + where it comes from.

We can extra enough data from here from the JSON path + type to get:
- location
- actual values

projects/standard-rulesets/src/spectral/index.ts
```diff
@@ -87,6 +94,55 @@ export class SpectralRulesets extends ExternalRuleBase {
 
     return allResults.flat(1);
   }
+
+  async runRulesV2(inputs: {
```
niclim, line 98: Again, also copy pasted from runRules and slightly modified for different inputs / outputs

projects/rulesets-base/src/rule-runner/rule-runner.ts
```diff
@@ -189,4 +209,99 @@ export class RuleRunner {
...
+    // TODO reimplement the rule runner so we don't need to generate legacy fact types here
+    const beforeFacts = traverseSpec(inputs.fromSpec);
+    const afterFacts = traverseSpec(inputs.toSpec);
+    const changelog = factsToChangelog(beforeFacts, afterFacts);
+
+    const openApiFactNodes = groupFacts({
+      beforeFacts,
+      afterFacts,
+      changes: changelog,
+    });
```
niclim, line 242: Ideally, we'd be using the new traverser function here that I have and generate a fact tree, which would then be able to be used to determine which rules to run.

Since there's a lot of differences in this, I'd probably have to end up reimplementing the entire rule running, which if that was the case we should just refactor this.

If we want to:
- invest in performance
- invest in rule writing ergonomics

We shoud refactor this function

projects/openapi-utilities/src/compare-specs/compare-specs.ts
```diff
@@ -0,0 +1,59 @@
...
+  const toSpec = to.isEmptySpec ? {} : to.jsonLike;
+
+  const diffs = diff(fromSpec, toSpec);
+
+  const results = await ruleRunner.runRules({
+    context: {}, // TODO
+    diffs,
+    fromSpec: from.jsonLike,
+    toSpec: to.jsonLike,
+  });
```
niclim, line 52: The idea here is that sourcemaps are uploaded, so we don't need to include them in the results (and can interpret them in subsequent usage)

niclim:
Some interpreter helpers I've added which might give a better understanding of how i plan to use the output here https://github.com/opticdev/optic/pull/1495

projects/openapi-utilities/src/openapi3/traverser.ts
niclim: Also, see comment here https://github.com/opticdev/optic/pull/1495#discussion_r1068574852 - had to add a `path` fact type due to how the diff works and how we need to interpret the diff

acunniffe (approved):

PR #1492, opened 2023-01-11
Make infer paths more accurate

acunniffe:
## 🍗 Description
Two big improvements within: 

1.  When documenting an API you don't know, sometimes you can't tell if the inference is right. Now Optic prints an example path right underneath the inferred pattern. Hopefully if you see it and it looks screwy you know before you realize before you add it to the openapi. Once you add the weird cases by hand the inference gets really good
2. We were getting a lot of weird learnings where certain obvious keywords were being learned as resource names. Like `api` `v2` etc. The logic is going to end up being a ton of heuristics no matter what we do -- inference is never clean. Now there are checks that certain keywords or patterns do not become resource names in the tree 

## 📚 References

## 👹 QA

### DIFF
--- a/projects/openapi-cli/src/specs/patches/generators/missing-path.ts
+++ b/projects/openapi-cli/src/specs/patches/generators/missing-path.ts
@@ -41,6 +41,9 @@ export function* missingPathPatches(
               in: 'path',
               name: parameterName,
               required: true,
+              schema: {
+                type: 'string',
+              },
             };
           }
         ),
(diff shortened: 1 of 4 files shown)


projects/openapi-cli/src/specs/patches/generators/missing-path.ts
```diff
@@ -41,6 +41,9 @@ export function* missingPathPatches(
               in: 'path',
               name: parameterName,
               required: true,
+              schema: {
```
acunniffe, line 44: fix reported by @philsturgeon who noticed that these were being added without schemas and failing spectral's oas checks.

PR #1495, opened 2023-01-12
Create diff helpers

niclim:
## 🍗 Description

This PR targets https://github.com/opticdev/optic/pull/1491

This PR introduces a bunch of interpreter functions for openapi 3 for the diff case. In a follow up PR, I'll use these functions to replace usage of JSONChangelog/TerminalChangelog to use the new diff format.

TODO
- [ ] Write tests

## 📚 References

## 👹 QA

### DIFF
--- a/projects/openapi-utilities/src/openapi3/traverser.ts
+++ b/projects/openapi-utilities/src/openapi3/traverser.ts
@@ -1,7 +1,4 @@
-import {
-  jsonPointerHelpers as jsonPointer,
-  jsonPointerHelpers,
-} from '@useoptic/json-pointer-helpers';
+import { jsonPointerHelpers as jsonPointer } from '@useoptic/json-pointer-helpers';
 import { OpenAPIV3, OpenAPIV3_1 } from 'openapi-types';
 // This is the V2 traverser version which is intended to be used for
 // more efficient storage + better compatibility between versions
@@ -77,7 +74,7 @@ export function getRaw<T extends OpenApiV3TraverserFact['type']>(
   spec: OpenAPIV3.Document,
   fact: Extract<OpenApiV3TraverserFact, { type: T }>
 ): RawItem<T> {
-  return jsonPointerHelpers.get(spec, fact.location.jsonPath);
+  return jsonPointer.get(spec, fact.location.jsonPath);
 }
 
 export class OpenApiV3Traverser implements Traverse<OpenAPIV3.Document> {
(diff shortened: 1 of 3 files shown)


projects/openapi-utilities/src/openapi3/types.ts
```diff
@@ -0,0 +1,110 @@
...
+  | 'component-schema-example';
+
+type _OpenApiV3TraverserFact<T extends V3FactType> = {
+  location: {
+    jsonPath: string;
+  };
+  type: T;
+};
+
+export type OpenApiV3TraverserFact<T extends V3FactType> = T extends T
```
niclim, line 25: I needed this indirection / TS magic to get `if (fact.type === '...') -> narrowed fact`

projects/openapi-utilities/src/openapi3/traverser.ts
```diff
@@ -41,6 +27,97 @@ const isObject = (value: any) => {
...
 };
 
+export function getRaw<T extends V3FactType>(
+  spec: OpenAPIV3.Document,
+  fact: OpenApiV3TraverserFact<T>
+): FactRawItem<T> {
+  return jsonPointer.get(spec, fact.location.jsonPath);
+}
+
+export function getLocation<T extends V3FactType>(
```
niclim, line 37: Because we are responsible for generating the JSON paths, we can rely on stable JSON path generation here. For this, we can encapsulate version differences between V2/V3 / other OAS versions. We can infer basically everything from JSON paths

projects/openapi-utilities/src/openapi3/traverser.ts
```diff
@@ -63,10 +145,16 @@ export class OpenApiV3Traverser implements Traverse<OpenAPIV3.Document> {
 
     for (let [pathPattern, paths] of Object.entries(this.input.paths || {})) {
       const traverser = this;
+      yield {
+        location: {
+          jsonPath: jsonPointer.append('', 'paths', pathPattern),
+        },
+        type: 'path',
+      };
```
niclim, line 153: @acunniffe I needed to add this because
- the diff will dig down to the lowest level where there's a diff between two objects
- The case where we add a path `/paths/~1api~1users` is not a diff at an operation level, it's the level above. The way we group diffs into relevant nodes is by looking at the nearest parent fact, which in this case is a specification. This wasn't a problem with previous diffs because we were comparing facts to facts, instead of where the diffs occured

projects/openapi-utilities/src/openapi3/types.ts
```diff
@@ -0,0 +1,115 @@
+import { OpenAPI, OpenAPIV3 } from 'openapi-types';
+
+export type V3FactType =
```
niclim, line 3: @Alexandre-Herve This was the best i could get the types to - this (and the usage in `traverser + `group-diffs` is what i want to achieve with this. Any suggestions on how I can get away without some of the type casing in `getLocation` etc?

Alexandre-Herve: No I don't know why it's not inferring that's odd 🤔 I'll try again later see if I can find something.

acunniffe (approved):

PR #1501, opened 2023-01-13
Update JSON changelog to handle new results format

niclim:
## 🍗 Description

Note - this PR's build will fail until all changelog components have been moved over. Splitting out into multiple PRs to be reviewed separately, and will merge in all together (which will switch over usage)

I initially attempted to keep the results shape the same, but I don't think it's actually correct. Some notable differences with how this works is:
- If a node is `added` or `removed`, we won't actually traverse down and log extra attributes. Instead, the node that was added will be "attribute added"
  - e.g. if we add a new operation, in the old changelog, `requests` would have been "added", in this changelog, it's only on the operation level. This is (I think) more correct for a changelog, but for rule runners we'd need to make sure all child nodes are triggered.
- We only recognize if a node is changed if that node's fact has changed, and not any of it's children (this is old behavior, but I think this was inconsistently implemented before?)

I think it's also ok for us to change the output of this, since this is only used in `optic diff` and we've kept the same format

## 📚 References

## 👹 QA

### DIFF
--- a/projects/openapi-utilities/src/utilities/json-changelog.ts
+++ b/projects/openapi-utilities/src/utilities/json-changelog.ts
@@ -1,135 +1,69 @@
-import {
-  groupChangesAndRules,
-  OpenApiEndpointChange,
-  RequestChange,
-  ResponseChange,
-  BodyChange,
-} from './group-changes';
-import { ChangeVariant, OpenApiKind, OpenApiFact } from '../openapi3/sdk/types';
-import isEqual from 'lodash.isequal';
-import omit from 'lodash.omit';
+import { OpenAPIV3 } from 'openapi-types';
+import { jsonPointerHelpers } from '@useoptic/json-pointer-helpers';
 
-const getChange = (
-  change: ChangeVariant<any> | null
-): 'added' | 'removed' | 'changed' =>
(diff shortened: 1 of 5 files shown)


projects/openapi-utilities/src/utilities/json-changelog.ts
```diff
@@ -144,62 +78,110 @@ type JsonChangelog = {
...
+  specs: { from: OpenAPIV3.Document; to: OpenAPIV3.Document },
+  groupedChanges: GroupedDiffs
 ): JsonChangelog {
   const results: JsonChangelog = { operations: [] };
 
-  const { changesByEndpoint, specification } = groupedChanges;
-  for (const [_, endpointChange] of changesByEndpoint) {
-    results.operations.push(getEndpointLogs(endpointChange));
+  const { paths } = groupedChanges;
+  for (const [path, pathObject] of Object.entries(paths)) {
```
niclim, line 87: This felt a little insane to do - i ended up moving the path logic into the groupDiffs function so we don't need to inteprete what "adding a path" means https://github.com/opticdev/optic/pull/1505/commits/6995a8d9939ea7a146dd887629d97717e2a00d59

acunniffe (approved):

PR #1517, opened 2023-01-16
Connect `optic spec push` to apis

niclim:
## 🍗 Description

## 📚 References

## 👹 QA

### DIFF
--- a/projects/optic/src/__tests__/integration/__snapshots__/spec-push.test.ts.snap
+++ b/projects/optic/src/__tests__/integration/__snapshots__/spec-push.test.ts.snap
@@ -3,16 +3,16 @@
 exports[`optic spec push can push a spec to a repo 1`] = `
 "Automatically adding the git sha git:COMMIT-HASH and branch git:master as tags 
 
-Uploading spec for api todo get optic id from url with tags env:production, the-favorite-api, git:COMMIT-HASH, git:master
-Succesfully uploaded spec to Optic. View the spec here TODODODODOO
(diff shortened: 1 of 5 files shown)


projects/optic/src/__tests__/integration/__snapshots__/spec-push.test.ts.snap
```diff
@@ -3,16 +3,16 @@
 exports[`optic spec push can push a spec to a repo 1`] = `
 "Automatically adding the git sha git:COMMIT-HASH and branch git:master as tags 
 
-Uploading spec for api todo get optic id from url with tags env:production, the-favorite-api, git:COMMIT-HASH, git:master
-Succesfully uploaded spec to Optic. View the spec here TODODODODOO
+Uploading spec for api at https://app.useoptic.com/organizations/org-id/apis/api-id with tags env:production, the-favorite-api, git:COMMIT-HASH, git:master
+Succesfully uploaded spec to Optic. View the spec here http://localhost:3000/organizations/org-id/specs/spec-id
```
niclim, line 7: I'm not doing any logic to look between environments, which is why this snapshot looks odd (we set a prod url, but the env look up points to local).

This shouldn't be a problem in regular use - I could also just update the spec to match the test env

nalanj: I do stuff like that in tests all the time.

nalanj (approved):

PR #1526, opened 2023-01-17
Analytics for ruleset usage

niclim:
## 🍗 Description

Notes:
- Will trigger one event per diff (`diff.rulesets`)
- This means diff_all will trigger `n` events for `n` diffs

## 📚 References

## 👹 QA


acunniffe (approved):

acunniffe:
This works with the default rulesets too right? `--check` without an optic yml is breaking changes?

niclim:
Yeah it will be with every ruleset - and our default is to run breaking changes checks if there's no config

PR #1527, opened 2023-01-17
Implement skeleton for `optic ci comment`

niclim:
## 🍗 Description

This PR has a lot of the meat for optic ci comment (taken from monorail + adapted to our new format)

The idea I have here is:
- If `CI===true`, optic diff and optic diff-all will dump a CI results file (see `./utils/ci-data.ts`)
- Calling `optic ci comment` after this will read that data (and error if not found)
- Will comment / update based on the same rules we had for monorail

In follow up PRs I'm going to:
- Implement the github / gitlab connectors for API calls
- update diff to dump a CIRun file if `CI===true`
- update diff-all to dump a CIRun file if `CI===true`

## 📚 References

## 👹 QA

### DIFF
--- a/projects/openapi-utilities/src/index.ts
+++ b/projects/openapi-utilities/src/index.ts
@@ -93,6 +93,7 @@ export {
   getOperationsModifsLabel,
   countOperationsModifications,
   getLabel,
+  getOperationsChangedLabel,
 } from './utilities/count-changed-operations';
 
 export { compareSpecs } from './compare-specs/compare-specs';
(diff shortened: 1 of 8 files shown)


nalanj (approved):
Looks great. I'd assume we could also add something that checks if $GITHUB_STEP_SUMMARY is set and if so writes some output to it so we'd end up with a job summary for the GitHub action as well.

niclim:
> I'd assume we could also add something that checks if $GITHUB_STEP_SUMMARY is set and if so writes some output to it so we'd end up with a job summary for the GitHub action as well

Yeah I think so, I haven't looked into it but I presume it's there

PR #1542, opened 2023-01-24
make it easier to set TLS for mac

acunniffe:
## 🍗 Description
Improved the DX for Mac

1. To accept the cert, you are directed to run `setup-tls` with sudo and it does it for you
2. If you get a cert error while capturing (indicating you did not do step 1), you get instructions to do step 1. This is a lot better than what we did before...yell "make sure you did this..." and hope people did. 
Screenshot 2023-01-23 at 7 04 30 PM

I think we can mimic this pattern on linux and then windows. Users on discord have been giving advice on which native commands to run

## 📚 References

## 👹 QA

### DIFF
--- a/projects/openapi-cli/src/cli.ts
+++ b/projects/openapi-cli/src/cli.ts
@@ -10,7 +10,7 @@ import { verifyCommand } from './commands/verify';
 import { CliConfig, readConfig } from './config';
 import { trackEvent } from './segment';
 import { clearCommand } from './commands/clear';
-import { captureCertCommand } from './commands/setup-tls';
+import { setupTlsCommand } from './commands/setup-tls';
 
 export async function makeCli(config: CliConfig) {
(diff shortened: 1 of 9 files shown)


projects/openapi-cli/src/cli.ts
```diff
@@ -41,10 +41,10 @@ export async function runCli(packageManifest?: {
...
   }).notify();
 
-  console.log('oas commands are now part of the optic cli');
-  console.log(
-    'Updated documented is at: http://useoptic.com/docs/document-existing-api\n\n'
-  );
+  // console.log('oas commands are now part of the optic cli');
+  // console.log(
+  //   'Updated documented is at: http://useoptic.com/docs/document-existing-api\n\n'
+  // );
```
niclim, line 47: Lets just delete this here

niclim (approved):

PR #1552, opened 2023-01-25
Put Optic Cloud extensions under `openapi` tag

acunniffe:
## 🍗 Description
Users said they were hard to find if they had to scroll to the bottom

## 📚 References

## 👹 QA

### DIFF
--- a/projects/optic/src/commands/api/add.ts
+++ b/projects/optic/src/commands/api/add.ts
@@ -121,14 +121,6 @@ async function promptForStandard(
 ): Promise<string | undefined> {
   const existingStandards = await client.getOrgStandards(orgId);
 
-  // if empty, create default without interrupting first time users
-  if (existingStandards.length === 0) {
-    const standard = await client.createOrgStandard(orgId, [
-      { name: 'breaking-changes', config: {} },
-    ]);
-    return standard.slug;
-  }
-
   const rulesetResponse = await prompts([
     {
       type: 'confirm',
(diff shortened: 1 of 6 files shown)


projects/optic/src/commands/api/write-to-file.ts
```diff
@@ -34,23 +39,69 @@ export async function writeYml(
...
+
+      return (
+        cleanedYaml.substring(0, indexOfOpenApiLineNewLine) +
+        (extensionLines.length ? '\n' + extensionLines.join(`\n`) : '') +
+        cleanedYaml.substring(indexOfOpenApiLineNewLine)
+      );
+    } else {
+      throw new Error('openapi: 3.x.x line not found');
+    }
+  } catch (e) {
```
acunniffe, line 96: wrote this with a pretty robust fallback mechanism...if we get no-ops re: `openapi` is spelled wrong or other weirdness resulting in an invalid patch, it just write it to yaml naivelly.

projects/optic/src/commands/api/add.ts
```diff
@@ -121,6 +121,14 @@ async function promptForStandard(
...
   const existingStandards = await client.getOrgStandards(orgId);
 
+  // if empty, create default without interrupting first time users
+  if (existingStandards.length === 0) {
+    const standard = await client.createOrgStandard(orgId, [
+      { name: 'breaking-changes', config: {} },
+    ]);
+    return standard.slug;
+  }
+
```
niclim, line 131: Is this intended to be part of this PR?

If yes, should we add any logging here to show that a standard has been added?

niclim (approved):
Do we also need to handle JSON as well?

projects/optic/src/commands/api/add.ts
acunniffe: nope removing this

acunniffe:
> Do we also need to handle JSON as well?

I suppose we should . I'll update

PR #1553, opened 2023-01-25
Better add command styling

acunniffe:
## 🍗 Description
Now featuring spinners, links that look like links, and more descriptive error messages

## 📚 References
Screenshot 2023-01-25 at 12 00 29 PM

## 👹 QA

### DIFF
--- a/projects/optic/src/commands/api/add.ts
+++ b/projects/optic/src/commands/api/add.ts
@@ -3,7 +3,7 @@ import prompts from 'prompts';
 import open from 'open';
 import path from 'path';
 import { wrapActionHandlerWithSentry } from '@useoptic/openapi-utilities/build/utilities/sentry';
-
+import ora from 'ora';
 import { OpticCliConfig, VCS } from '../../config';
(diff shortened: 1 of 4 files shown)


projects/optic/src/commands/api/add.ts
```diff
@@ -121,6 +121,14 @@ async function promptForStandard(
 ): Promise<string | undefined> {
   const existingStandards = await client.getOrgStandards(orgId);
 
+  // if empty, create default without interrupting first time users
+  if (existingStandards.length === 0) {
+    const standard = await client.createOrgStandard(orgId, [
+      { name: 'breaking-changes', config: {} },
+    ]);
+    return standard.slug;
+  }
```
niclim, line 130: Also - intended to be here? If yes then we might want to add some logging?

niclim (approved):

projects/optic/src/commands/api/add.ts
acunniffe: No. I don't think we want people to have to think about standards on first run.

PR #1560, opened 2023-01-26
make add --all a flag

acunniffe:
## 🍗 Description
I think this makes more sense to a user than telling them to run it naked. Also matches git. 

## 📚 References

## 👹 QA


niclim (approved):
Maybe in the future we could do paths to specs instead of an `--all` flag, but this is fine for now

e.g. `optic add api .` or `optic add api folder_name`

PR #1566, opened 2023-01-27
create a denormalize function

niclim:
## 🍗 Description

Creates a denormalize function that copies shared path parameters -> each method

TODO
- [x] Write tests
- [x] Figure out source map issue with nested refs

In a future PR
- [ ] connect this to diff, diff-all, api add and spec push

## 📚 References

## 👹 QA

### DIFF
--- a/projects/openapi-io/src/denormalizers/__tests__/__snapshots__/denormalize.test.ts.snap
+++ b/projects/openapi-io/src/denormalizers/__tests__/__snapshots__/denormalize.test.ts.snap
@@ -1,0 +1,204 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`denormalize denormalizes shared path parameters 1`] = `
+{
+  "jsonLike": {
+    "info": {
+      "title": "some thing",
+      "version": "v0",
+    },
+    "openapi": "3.0.1",
+    "paths": {
+      "/example": {
+        "get": {
+          "parameters": [
+            {
+              "description": "something here",
+              "in": "query",
+              "name": "something-else",
+              "required": true,
+              "schema": {
+                "example": "d5b640e5-d88c-4c17-9bf0-93597b7a1ce2",
+                "format": "uuid",
+                "type": "string",
+              },
+            },
+            {
+              "description": "should-not-override",
+              "in": "query",
+              "name": "should-not-override",
+              "required": true,
+              "schema": {
+                "example": "d5b640e5-d88c-4c17-9bf0-93597b7a1ce2",
+                "format": "uuid",
+                "type": "string",
+              },
+            },
+            {
+              "description": "something here",
+              "in": "query",
+              "name": "something",
+              "required": true,
+              "schema": {
+                "example": "d5b640e5-d88c-4c17-9bf0-93597b7a1ce2",
+                "format": "uuid",
+                "type": "string",
+              },
+            },
+          ],
+          "responses": {
+            "200": {
+              "content": {
+                "application/json": {
+                  "schema": {
+                    "type": "string",
+                  },
+                },
+              },
+              "description": "some thing",
+            },
+          },
+        },
+        "post": {
+          "parameters": [
+            {
+              "description": "should not be overriden",
+              "in": "query",
+              "name": "should-not-override",
+              "required": true,
+              "schema": {
+                "example": "d5b640e5-d88c-4c17-9bf0-93597b7a1ce2",
+                "format": "uuid",
+                "type": "string",
+              },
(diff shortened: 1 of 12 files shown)


projects/openapi-io/src/denormalizers/denormalize.ts
```diff
@@ -0,0 +1,57 @@
...
+              const newPointer = jsonPointerHelpers.compile([
+                'paths',
+                pathKey,
+                method,
+                'parameters',
+                String(operation.parameters.length),
+              ]);
+
+              parse.sourcemap.logPointerToRootFile(newPointer, oldPointer);
+              operation.parameters.push(parameter);
```
acunniffe, line 45: can't `path.parameters` be undefined here?

projects/openapi-io/src/parser/sourcemap.ts
```diff
@@ -52,6 +50,18 @@ export class JsonSchemaSourcemap {
     }
   }
 
+  logPointerInFile(
+    filePath: string,
+    sourcePointer: string,
+    targetPointer: string
+  ) {
```
niclim, line 57: Rather than trying to reconstruct json pointers that included the path, I updated the sourcemap so we can pass in a file path.

projects/openapi-io/src/parser/sourcemap.ts
```diff
@@ -72,31 +82,3 @@ export class JsonSchemaSourcemap {
     }
   }
 }
-export function resolveJsonPointerInYamlAst(
```
niclim, line 75: This is unused, it was split out previously into `sourcemap` and `sourcemapReader` (which this belongs to the latter). It's duplicated in sourcemapReader where it's actually used

projects/openapi-io/src/denormalizers/__tests__/__snapshots__/denormalize.test.ts.snap
```diff
@@ -0,0 +1,204 @@
...
+    type: string
+    format: uuid
+    example: d5b640e5-d88c-4c17-9bf0-93597b7a1ce2
+",
+        "index": 1,
+        "path": "/Users/nicholaslim/Desktop/repos/optic/projects/openapi-io/src/denormalizers/__tests__/specs/parameters.yml",
+        "sha256": "85d23de44c126a054df04ca9720a69b0842935868c6d0a771baa37f0bee832ba",
+      },
+    ],
+    "refMappings": {
```
niclim, line 187: Here's the key part for the sourcemap - this points to the paramters where they originate from and where they end up in the resultant document

projects/openapi-io/src/denormalizers/__tests__/__snapshots__/denormalize.test.ts.snap
```diff
@@ -0,0 +1,204 @@
...
+
+exports[`denormalize denormalizes shared path parameters 1`] = `
+{
+  "jsonLike": {
+    "info": {
+      "title": "some thing",
+      "version": "v0",
+    },
+    "openapi": "3.0.1",
+    "paths": {
```
niclim, line 11: Here we've removed the shared parameter paths, and placed them onto each method. Notice how `post` does not get overriden even if we've defined something twice. 

We'll pick the _most_ specific parameter (i.e. the one on the method function)

projects/openapi-io/src/parser/sourcemap.ts
acunniffe: I think this will get weird if you the shared parameter was a ref. 

```
paths: 
  /xyz:
    parameters:
      - $ref: ./shared#parameters/auth
```

acunniffe: I don't see any downside to doing it correctly (by reading from the sourcemap)

niclim: It is reading the sourcemap https://github.com/opticdev/optic/pull/1566/files#diff-3d28014487c4e359675a91ff0ed1099359b36761c608fdfb91bfd7605b746a26R54-R65

and i've added it in as a test case here https://github.com/opticdev/optic/pull/1566/files#diff-c10d156f1637218487b466d310f1c89d779fd91f5f52accdd80b5b1422ec883eR24

niclim: This is just to avoid having to do:
`filePath + jsonpointer` and then splitting that same file path again

acunniffe: Sorry missed that. I see it now and this makes sense.

PR #1572, opened 2023-01-30
Add CLI nudges to try other commands

acunniffe:
## 🍗 Description

## 📚 References

## 👹 QA


projects/openapi-cli/src/commands/verify.ts
```diff
@@ -185,6 +187,12 @@ export async function verifyCommand(): Promise<Command> {
 
       await flushEvents();
 
+      if (Boolean(options.document) && !hasOpticUrl) {
```
acunniffe, line 190: every time you document a new endpoint & don't have the optic url

projects/optic/src/commands/diff/diff.ts
```diff
@@ -177,11 +181,14 @@ const runDiff = async (
...
-      logger.info(
-        `Configure check standards in optic cloud or your local optic.dev.yml file.`
-      );
+    logger.info('')
+
+    if (!hasOpticUrl) {
+      logger.info(chalk.blue.bold(`See the full history of this API by running "optic add ${path.parse(baseFile.sourcemap.rootFilePath).base}"`))
+    }
+
+    if (!config.isInCi && specResults.results.length) {
```
acunniffe, line 190: this is a little annoying -- you may see it after setting up CI, but hard to know if you have the GH action set up from here w/o running the API. I think we can figure out what to do as we hear from users.

projects/optic/src/commands/diff/diff.ts
```diff
@@ -177,11 +181,14 @@ const runDiff = async (
...
 
-    if (!config.isInCi) {
-      logger.info('');
-      logger.info(
-        `Configure check standards in optic cloud or your local optic.dev.yml file.`
-      );
+    logger.info('')
+
+    if (!hasOpticUrl) {
+      logger.info(chalk.blue.bold(`See the full history of this API by running "optic add ${path.parse(baseFile.sourcemap.rootFilePath).base}"`))
```
niclim, line 187: You won't actually add the full API history unless you add `optic add spec --history-depth 0`

projects/optic/src/commands/diff/diff.ts
niclim: This isn't going to do what you expect - this will always show up if your running this locally, regardless whether you have set something up in CI or not. (`config.isInCi` refers to wheter the current run is in CI (i.e. `process.env.CI === true`)

I suggest something different - how about:
- at the end of a diff run ask if you want to set up optic in CI and gives you three options
  - yes (runs optic ci setup) - e.g. running something like https://github.com/opticdev/optic/blob/main/projects/optic/src/commands/ci/setup.ts#L83-L87
  - maybe later (asks again in 1 week)
  - no (stops asking)

We can store this in `~/.config/optic/config.json` https://github.com/opticdev/optic/blob/main/projects/optic/src/config.ts#L190-L213

acunniffe: That feels a bit too in your face -- going right into a prompt at the end of a diff. Let's think a bit more on this today

niclim: Alternatively, we can log comments and do the same idea:
- will keep asking you every 1 week until you either:
  - run `optic ci setup`
  - run `optic mute ...` or something like that

projects/optic/src/__tests__/integration/__snapshots__/spec-push.test.ts.snap
```diff
@@ -1,9 +1,9 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`optic spec push can push a spec to a repo 1`] = `
-"Automatically adding the git sha git:COMMIT-HASH and branch gitbranch:master as tags 
+"Automatically adding the git sha git:COMMIT-HASH and branch gitbranch:main as tags
```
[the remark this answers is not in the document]
acunniffe: sure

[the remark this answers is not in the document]
niclim: fix in [33bb707](https://github.com/opticdev/optic/pull/1572/commits/33bb70745e27b34c2057b7175da6fca9ee447418)

niclim (approved):

PR #1575, opened 2023-01-31
Update error handler to look for forbidden

niclim:
## 🍗 Description

## 📚 References

## 👹 QA

### DIFF
--- a/projects/optic/src/error-handler.ts
+++ b/projects/optic/src/error-handler.ts
@@ -14,8 +14,10 @@ export const errorHandler = <Args extends any[], Return extends any>(
       if (UserError.isInstance(e)) {
         console.error(e.message);
       } else if (
-        (e instanceof BadRequestError || e instanceof ForbiddenError) &&
-        /Invalid optic token/i.test(e.message)
+        (e instanceof BadRequestError &&
+          e.source === 'optic' &&
+          /Invalid token/i.test(e.message)) ||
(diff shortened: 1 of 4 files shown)


projects/optic/src/error-handler.ts
```diff
@@ -14,8 +14,8 @@ export const errorHandler = <Args extends any[], Return extends any>(
       if (UserError.isInstance(e)) {
         console.error(e.message);
       } else if (
-        e instanceof BadRequestError &&
-        /Invalid token/i.test(e.message)
+        (e instanceof BadRequestError || e instanceof ForbiddenError) &&
```
nalanj, line 17: I think we'd want to check if it's ever forbidden without a care about the message or bad request with the matching message. I.e.:

```
else if (e instanceof ForbiddenError || (e instanceof BadRequestError && /Invalid optic token/i.test(e.message))) {
```

niclim: The problem is if we make Forbidden requests to other APIs, since this is a global handler - we'd need to look for Forbidden from just from optic. 

We could create our own optic error stuff, or identify it someway though. Let me put up something to handle this

niclim: Ok I updated this to be able to trace errors from the client generated - so we're now looking for forbidden and bad request

Not a huge fan of how I implemented this, but it works

nalanj (approved):

PR #1583, opened 2023-02-01
fix date yaml

acunniffe:
## 🍗 Description
This popped up in testing today. Because JSON Ref Resolver uses JSON loader for YAML instead of core it tries to instantiate date keys. we don't want custom JS types for any of our parsing so I've disabled this

## 📚 References

## 👹 QA


projects/openapi-io/src/parser/insourced-yaml.js
```diff
@@ -0,0 +1,61 @@
...
+  async parse(file) {
+    // eslint-disable-line require-await
+    let data = file.data;
+    if (Buffer.isBuffer(data)) {
+      data = data.toString();
+    }
+
+    if (typeof data === 'string') {
+      try {
+        return yaml.load(data, { schema: core });
```
acunniffe, line 52: insourced for this change

PR #1586, opened 2023-02-02
remove discovery from setup and standards from add

acunniffe:
## 🍗 Description

## 📚 References

## 👹 QA

### DIFF
--- a/projects/optic/src/__tests__/integration/api-add.test.ts
+++ b/projects/optic/src/__tests__/integration/api-add.test.ts
@@ -85,7 +85,7 @@ describe('optic api add', () => {
       }
       const { combined, code } = await runOptic(
         workspace,
-        'api add ./spec.yml --standard blah'
+        'api add ./spec.yml'
       );
 
       expect(code).toBe(0);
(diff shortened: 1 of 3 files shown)


projects/optic/src/__tests__/integration/api-add.test.ts
```diff
@@ -85,7 +85,7 @@ describe('optic api add', () => {
       }
       const { combined, code } = await runOptic(
         workspace,
-        'api add ./spec.yml --standard blah'
```
acunniffe, line 88: I removed this entirely because the chances someone knows the id of their standard was basically 0 anyway. Web and or optic yaml makes more sense

niclim (approved):
Where / how do we let users know where how they can add the `x-optic-ruleset` key?

nalanj (approved):

acunniffe:
> Where / how do we let users know where how they can add the `x-optic-ruleset` key?

Maybe on standards. WE can give them a codebox with the standard URl set and tell them to copy / paste

Issue #1600, opened 2023-02-05
Overly strict OpenAPI Spec 3.1 JsonSchema enforcement


SeanKeenan-at:
There are three properties I'm seeing fail: 
 - `const` (similar to `enum`)
 - `propertyNames` on objects
 - [unevaluatedProperties](https://json-schema.org/understanding-json-schema/reference/object.html#unevaluated-properties) on objects (Available in `2019-09`)

I believe while these were invalid in OpenAPI 3.0, 3.1 has brought closer alignment to the JSONSchema spec (and up to metaschema 2020-12) and I think the above should be valid. All of these also pass spectral lint.

I believe the changes need to be made to this file: `projects/openapi-io/src/validation/validation-schemas.ts`

acunniffe:
Thanks @SeanKeenan-at -- we'll loosen up the validation around 3.1 keywords. 

It may be a while before anyone has time to write breaking change rules that take some of these keywords into account. They're probably easy to write code for, but difficult to think through the requirements for. If you have any interest in helping to think through that let us know. 

I'll update this issue later in the week once we have a release out

SeanKeenan-at:
Wow, that was a fast response - thank you! And totally understand.

I think in-so-far as breaking changes:
 - I believe `const` can be treated _exactly_ as an `enum`, as long as you first convert the single element to an array. (As in, converting between `const: 'example'` and an `enum: ['example']` doesn't even need to be picked up as a change IMO).
 - `propertyNames` + `unevaluatedProperties` are a bit trickier to think through.

I threw up a PR for the `const` treatment if that's helpful! #1601

PR #1601, opened 2023-02-05
Add breaking changes const support

SeanKeenan-at:
## 🍗 Description
Adds support for `const` in breaking-changes detection.

Works by treating `const` the same as an `enum` with exactly one value. This means that converting between:
`const: 'test'` and `enum: ['test']` is not considered a breaking change.

## 📚 References
https://github.com/opticdev/optic/issues/1600

## 👹 QA
New automated test coverage


SeanKeenan-at:
cc @acunniffe

projects/standard-rulesets/src/breaking-changes/__tests__/parameter-enum-breaking-changes.test.ts
```diff
@@ -108,6 +108,117 @@ describe('breaking changes ruleset - parameter enum change', () => {
...
+            get: {
+              parameters: [
+                {
+                  name: 'test',
+                  in: location,
+                  schema: {
+                    type: 'string',
+                    const: 'a',
+                    // Casting because const does not exist in OpenAPIV3.NonArraySchemaObject
+                  } as OpenAPIV3.NonArraySchemaObject,
```
SeanKeenan-at, line 131: I don't love this, but it seems to be an external-to-optic library. I tried to at least scope the casts to the minimum size I could (vs casting the whole `OpenAPIV3.Document`

acunniffe: Yeah it looks like we have a few typing issues to think through before we can fix this. Since Optic currently won't even read the `const` schemas we'll get that work up and ping you when you can use the correct typing. Probably going to take a day or two -- we're launching a few other features today first.

acunniffe:
Schema changes merged and are in this branch now. I can take this over and get it over the finish line tomorrow.

acunniffe (approved):

PR #1604, opened 2023-02-07
Verify token

nalanj:
## 🍗 Description

Verify token on login

## 📚 References

## 👹 QA

### DIFF
--- a/projects/optic/src/commands/login/login.ts
+++ b/projects/optic/src/commands/login/login.ts
@@ -71,7 +71,7 @@ Create an account and generate a personal access token at ${chalk.underline.blue
     const result = await newClient.verifyToken();
 
     if (result.user) {
-      identify(result.user.userId, result.user.email);
+      identify(result.user.email);
       trackEvent('cli.login');
       await flushEvents();
     }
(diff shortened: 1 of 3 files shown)


projects/optic/src/commands/login/login.ts
```diff
@@ -56,24 +61,43 @@ Create an account and generate a personal access token at ${chalk.underline.blue
...
+  try {
+    const result = await newClient.verifyToken();
 
-    logger.info(
-      chalk.green(
-        `Successfully saved your personal access token to ${USER_CONFIG_PATH}`
-      )
-    );
+    if (result.user) {
+      identify(result.user.userId, result.user.email);
```
niclim, line 74: We'll want to send the https://github.com/opticdev/optic/blob/main/projects/openapi-utilities/src/utilities/segment.ts#L8 as the user id instead which I think will help associate events with the email.

Alternatively I think there's an `alias` command that might be appropriate here

niclim (approved):

projects/optic/src/commands/login/login.ts
niclim: Yeah I think that should work - we'll just want to see the result of that

PR #1609, opened 2023-02-09
Make comment command also write to stdout

acunniffe:
## 🍗 Description
A lot of developers click into our CI action expecting to see changelog results + links. This makes sense since that's where they usually find data like this. 

I think we should log a basic comment-like message out at the very end of our comment process. This is also a nice fallback if you don't have a token...you still get something

## 📚 References

## 👹 QA

### DIFF
--- a/projects/optic/src/commands/ci/comment/comment.ts
+++ b/projects/optic/src/commands/ci/comment/comment.ts
@@ -9,6 +9,7 @@ import {
 import { logger } from '../../../logger';
 import { CommentApi, GithubCommenter, GitlabCommenter } from './comment-api';
 import { errorHandler } from '../../../error-handler';
+import chalk from 'chalk';
 
 const usage = () => `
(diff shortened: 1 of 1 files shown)


projects/optic/src/commands/ci/comment/comment.ts
```diff
@@ -153,4 +154,27 @@ const getCiCommentAction =
         await commenter.createComment(body);
       }
     }
+
+    logger.log(body);
+    if (data.failed.length || data.noop.length)
+      logger.log(
+        chalk.red(`${data.failed.length + data.noop.length} failures`)
```
niclim, line 161: A noop isn't a failure - it's when specs didn't change.

niclim (approved):

niclim (changes requested):
Actually now that I think about this, this is in the wrong place. The `optic diff-all` command should be rendering this out instead of comment, comment only runs if there's a comment token

I think optic diff-all already renders out terminal changelog + results data and the web url - if anything we should be updating the diff logging output

acunniffe:
I mean that's fine - we just need to give the links. That's the key part. closing this for now

PR #1619, opened 2023-02-10
Run strict validation on diff variant with 2 files

niclim:
## 🍗 Description

In the case we run `optic diff file1 file2` we should run strict validation. The reason we usually don't is because `optic diff file --base ...` means we cannot fix the before spec, but in this diff instance, we should be able to.

## 📚 References

## 👹 QA


acunniffe (approved):

acunniffe:
Never thought of that but makes total sense

PR #1628, opened 2023-02-13
pass org id to patch ruleset

niclim:
## 🍗 Description

Forgot to do this for the patch endpoint, whcih is also hit as part of the upload flow

TODO
- [ ] bump patch version

## 📚 References

## 👹 QA


niclim:
Ok tested with this pre-release, looks like it's now working (with the backend endpoint updates as well)

nalanj (approved):

PR #1632, opened 2023-02-14
fix example validator pre-processing

acunniffe:
## 🍗 Description

## 📚 References

## 👹 QA

### DIFF
--- a/projects/standard-rulesets/src/examples/requireValidExamples.ts
+++ b/projects/standard-rulesets/src/examples/requireValidExamples.ts
@@ -59,21 +59,19 @@ export function validateSchema(
 }
 
 function strictAdditionalProperties(schema: any) {
-  if (typeof schema === 'object') {
-    if (Array.isArray(schema)) {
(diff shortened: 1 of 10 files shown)


projects/standard-rulesets/src/examples/requireValidExamples.ts
```diff
@@ -59,21 +59,19 @@ export function validateSchema(
...
-        schema.type === 'object' &&
-        !schema.hasOwnProperty('additionalProperties')
-      ) {
-        schema.additionalProperties = false;
-      }
-
-      Object.values(schema).forEach(strictAdditionalProperties);
+  if (Array.isArray(schema)) {
+    schema.forEach((item) => strictAdditionalProperties(item));
+  } else if (Object.prototype.toString.call(schema) === '[object Object]') {
```
niclim, line 64: This is a really odd check - can we instead do a `typeof schema === 'object' && schema !== null`?

niclim (approved):

PR #1655, opened 2023-02-17
Fix reading files in git with spaces

niclim:
## 🍗 Description

Reading a file `openapi with spaces.yml` does not work in our git ref resolver. This is because the git resolver receives the file as a url encoded string (`openapi%20file%20with%20space.yml`) which is then passed to `git show` and subsequently fails.

## 📚 References

https://github.com/opticdev/monorail/issues/2923

## 👹 QA


acunniffe (approved):
NIce 👍

PR #1660, opened 2023-02-21
fix: stableVersion breaks `prerelease` keyword versioning

notnmeyer:
## 🍗 Description

when `stableVersion` is set, bumping by `prerelease` breaks. ive replicated 2 different behaviors, one where a prerelease bump doesn't change the version at all, and another where it increments the _patch_ version by 1, and then subsequent bumps do nothing.

for example,
```
➜ jq .version package.json
"0.39.5-0"

➜ yarn run version prerelease
...

➜ jq .version package.json
"0.39.6-0"

➜ yarn run version prerelease
...

➜ jq .version package.json
"0.39.6-0"
```

this looks like https://github.com/yarnpkg/berry/issues/3868. i implemented the same workaround mentioned there with some additions for workspaces support, and it seems to work,

```
➜ jq .version package.json
"0.39.5-0"

➜ yarn run version prerelease
...

➜ jq .version package.json
"0.39.5-1"

➜ yarn run version prerelease
...

➜ jq .version package.json
"0.39.5-2"
```

if we'd rather not add this crud just to preserve bumping with the `prerelease` keyword, we can update our release docs instead to mention that prereleases need to be explicitly versioned, e.g., `yarn run version 0.0.0-0`

## 📚 References

## 👹 QA

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,8 +15,9 @@
   ],
   "scripts": {
     "release": "gh release create --target=$(git branch --show-current) v$(node -e \"process.stdout.write(require('./package.json').version)\")",
-    "version": "yarn workspaces foreach -v version",
-    "prepare": "husky install"
+    "version": "yarn run rmStableVersion && yarn workspaces foreach -v version",
+    "prepare": "husky install",
+    "rmStableVersion": "./scripts/remove-stableVersion"
   },
   "devDependencies": {
     "@types/prettier": "^2",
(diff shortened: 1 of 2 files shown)


notnmeyer:
hrm, i appear to have forgotten about this and i think the fact that it hasn't come up at all is a pretty good signal this isnt terribly important.

PR #1668, opened 2023-02-23
Fix regression where openapi paths would not be grouped together if variable in path changed

acunniffe:
## 🍗 Description
Openapi diff will now line up comparisons for operations / children again even when variables change in paths

## 📚 References

## 👹 QA


projects/openapi-utilities/src/diff/__tests__/__snapshots__/diff.test.ts.snap
```diff
@@ -180,6 +180,15 @@ exports[`diff openapi diff with array values diffs for primitive values 1`] = `
 ]
 `;
 
+exports[`diff openapi diff with array values diffs operations even if variable names change  1`] = `
+[
+  {
+    "after": "/paths/~1user~1{user_Id}/get/operationId",
+    "before": "/paths/~1user~1{userId}/get/operationId",
```
acunniffe, line 187: It can now see the operationId changing between these endpoints

projects/openapi-utilities/src/diff/diff.ts
```diff
@@ -60,12 +62,16 @@ export function typeofDiff(
 }
 
 // Diffs two objects, generating the leaf nodes that have changes
-export function diff(before: any, after: any): ObjectDiff[] {
+export function diff(
+  before: any,
+  after: any,
+  initialPath: string = ''
```
acunniffe, line 68: I added this because it is the correct way to do `arrayIdFn` and `objectIdFn`

projects/openapi-utilities/src/diff/diff.ts
```diff
@@ -84,13 +90,13 @@ export function diff(before: any, after: any): ObjectDiff[] {
     // for objects, we just use the key
     if (Array.isArray(before.value) && Array.isArray(after.value)) {
       const allValues = [...before.value, ...after.value];
-      const arrayIdFn: (v: any, i: number) => string = allValues.every((v) =>
-        isParameterObject(v)
```
acunniffe, line 88: this is not really correct. It works most of the time, but it's not specific to the parameter arrays we are actually trying to catch.

projects/openapi-utilities/src/diff/diff.ts
```diff
@@ -122,17 +129,40 @@ export function diff(before: any, after: any): ObjectDiff[] {
           afterPath,
         });
       }
-    } else if (!Array.isArray(before.value) && !Array.isArray(after.value)) {
+    } else if (
+      !Array.isArray(before.value) &&
+      !Array.isArray(after.value) &&
+      Object.prototype.toString.call(before.value) &&
```
acunniffe, line 135: I think this was running before on primitives, but `Object.keys(...)` was returning `[]`

projects/openapi-utilities/src/diff/openapi-matchers.ts
```diff
@@ -0,0 +1,18 @@
+import { jsonPointerHelpers } from '@useoptic/json-pointer-helpers';
+import { getParameterIdentity } from './array-identifiers';
+
+export const isPathParameterArray = (pointer: string): boolean => {
```
acunniffe, line 4: more precise way of checking if we are in a path parameter array

projects/openapi-utilities/src/diff/openapi-matchers.ts
```diff
@@ -0,0 +1,18 @@
...
+    jsonPointerHelpers.matches(pointer, [
+      'paths',
+      '**',
+      '!parameters',
+      'parameters',
+    ])
+  );
+};
+
+export const isPathsMap = (pointer: string): boolean => {
```
acunniffe, line 16: easy to check for path map

projects/json-pointer-helpers/src/json-pointers/json-pointer-helpers.ts
```diff
@@ -55,6 +57,37 @@ function relative(pointer: string, from: string) {
...
+function matches(pointer: string, pattern: string[]): boolean {
+  const components = parse(pointer);
+
+  if (components.length !== pattern.length) {
+    return false;
+  }
+
+  return components.every((comp, index) => minimatch(comp, pattern[index]));
+}
+function endsWith(pointer: string, pattern: string[]): boolean {
```
acunniffe, line 80: I think this will be helpful to have things like `enum`, `allOf` and `type`.

projects/openapi-utilities/src/diff/diff.ts
```diff
@@ -84,13 +90,13 @@ export function diff(before: any, after: any): ObjectDiff[] {
     // for objects, we just use the key
     if (Array.isArray(before.value) && Array.isArray(after.value)) {
       const allValues = [...before.value, ...after.value];
-      const arrayIdFn: (v: any, i: number) => string = allValues.every((v) =>
-        isParameterObject(v)
-      )
-        ? getParameterIdentity
-        : allValues.every((v) => typeof v !== 'object')
```
acunniffe, line 91: not really sure what case this catches so left it as-is.

what is it for?

projects/openapi-utilities/src/diff/__tests__/diff.test.ts
```diff
@@ -128,6 +132,37 @@ describe('diff openapi', () => {
       expect(diffWithChanges.length).toBe(2);
       expect(diffWithChanges).toMatchSnapshot();
     });
+
+    test('diffs operations even if variable names change ', () => {
```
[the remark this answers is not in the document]
acunniffe: I added a test @niclim https://github.com/opticdev/optic/pull/1668/files#diff-bca1031775f57baa9a46397850ec3e585e9dc1137eeffb45f011e214c9a83bd4R192

Does this snapshot look correct to you?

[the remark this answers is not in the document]
niclim: Yeah thats what I expect for the snapshot  (i.e. we track the same identify between before and after)

niclim (approved):
looks great

PR #1672, opened 2023-02-27
Add no-missing-urls to diff-all options

niclim:
## 🍗 Description

Add an option to return `exitCode = 1` if diff-all detects a spec that does not have an `x-optic-url`

A user can also scope this down by specifying a `match` or `ignore` glob to ignore false positives

TODO
- [ ] Also add this to the github action options

## 📚 References

## 👹 QA

### DIFF
--- a/projects/optic/src/commands/diff/diff-all.ts
+++ b/projects/optic/src/commands/diff/diff-all.ts
@@ -79,6 +79,11 @@ comma separated values (e.g. "**/*.yml,**/*.json")'
     .option('--upload', 'upload specs', false)
     .option('--web', 'view the diff in the optic changelog web view', false)
     .option('--json', 'output as json', false)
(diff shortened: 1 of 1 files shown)


projects/optic/src/commands/diff/diff-all.ts
```diff
@@ -79,6 +79,11 @@ comma separated values (e.g. "**/*.yml,**/*.json")'
     .option('--upload', 'upload specs', false)
     .option('--web', 'view the diff in the optic changelog web view', false)
     .option('--json', 'output as json', false)
+    .option(
+      '--no-missing-urls',
```
acunniffe, line 83: I don't like this name. `fail-on-untracked-openapi`

acunniffe (approved):

PR #1674, opened 2023-02-28
allOf diffing

acunniffe:
## 🍗 Description
Specifically targets objects + allOf objects (the primary use case) and does not try to work with anything that isn't an object. 
- `object` -> `allOf` 
- `allOf` -> `object`
- `allOf(n) <> allOf(n)` where n can be different

When we see an allOf of objects being compared to another allOf or an object: 
- manually create the correct comparisons for properties and enqueue them
- manually diff the required 
- enqueue a diff for the other attributes of objects

Future
- we may want to add validation rules for impossible allofs -- no other tools do this, but really easy to make something that can't exist in the real world like 
```
allOf:
 -  type: object
    properties: []...
 - type: object
   additionalProperties: false
```

## 📚 References

## 👹 QA

### DIFF
--- a/projects/openapi-utilities/src/diff/__tests__/diff.test.ts
+++ b/projects/openapi-utilities/src/diff/__tests__/diff.test.ts
@@ -1,6 +1,7 @@
 import { describe, test, expect } from '@jest/globals';
 import { diff, ObjectDiff, reconcileDiff } from '../diff';
 import { before, after } from './mock-data';
+import { jsonPointerHelpers } from '@useoptic/json-pointer-helpers';
 
 describe('diff openapi', () => {
   test('can diff an openapi spec', () => {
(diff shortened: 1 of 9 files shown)


projects/openapi-utilities/src/diff/__tests__/diff.test.ts
```diff
@@ -128,11 +133,562 @@ describe('diff openapi', () => {
...
+          '200',
+          'content',
+          'application/json',
+          'schema',
+        ])
+      );
+
+      expect(diffResult.length).toBe(2);
+    });
+    test('removals of properties', () => {
```
niclim, line 348: This is a case where we won't be able to show anything in changelog (if you look at the trail generated `"before": "/paths/~1users/get/responses/200/content/application~1json/schema/required/0",` you can see that it will point to something that does not exist in the after case.

acunniffe: I may be missing something but how is this any different than removing a normal property? We have to anchor that somewhere (its parent)? 

We also have this issue in the normal changelog for the required if you go from `required: ['a']` -> `_` instead of `required: []`. 

The before paths don't need to resolve in the after spec, that's why there are separate paths right?

PR #1677, opened 2023-02-28
export lib and rulesets in optic lib

niclim:
## 🍗 Description

Export functions to allow users to bundle rulesets / and customize their own CLI 

```javascript
import { initCli, setRulesets }  from 'lib';

(async () => {
  const cli = await initCli();
  setRulesets([your own rulesets])

  cli.parse(process.argv);
})()
```

## 📚 References

## 👹 QA

### DIFF
--- a/projects/optic/src/commands/diff/generate-rule-runner.ts
+++ b/projects/optic/src/commands/diff/generate-rule-runner.ts
@@ -1,5 +1,5 @@
 import path from 'path';
-import { OpticCliConfig } from '../../config';
+import { ConfigRuleset, OpticCliConfig } from '../../config';
 import { StandardRulesets } from '@useoptic/standard-rulesets';
 import {
   RuleRunner,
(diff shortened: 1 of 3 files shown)


projects/optic/src/commands/diff/generate-rule-runner.ts
```diff
@@ -11,6 +11,12 @@ import { loadCliConfig } from '../../config';
 import { logger } from '../../logger';
 import chalk from 'chalk';
 
+const baseRulesets: (Ruleset | ExternalRule)[] = [];
```
niclim, line 14: This will act as an extension and you can still define `--ruleset` or other things (this will be additive)

projects/optic/src/commands/diff/generate-rule-runner.ts
```diff
@@ -64,7 +70,15 @@ export const generateRuleRunner = async (
   let warnings: string[] = [];
 
   if (checksEnabled) {
-    const standard = await getStandardToUse(options);
+    let defaultStandard: ConfigRuleset[] = [];
+    if (injectedRulesets.length > 0) {
+      logger.debug('Using injected rulesets from CLI bundle');
+      rulesets.push(...injectedRulesets);
+    } else {
+      defaultStandard = [{ name: 'breaking-changes', config: {} }];
```
niclim, line 78: We only set the default standard if there are no injected rulesets

acunniffe:
Looks good. After our calls tomorrow let's get a migration guide up

PR #1682, opened 2023-03-05
--document with $ref

acunniffe:
## 🍗 Description
Generated specs can be thousands of lines long, and difficult for a team to pick up and maintain.  This PR takes the first steps towards documenting with `$refs` 

- If a `component.schema` exists that is close[1] to a new operation's request or response body, Optic will use that schema instead of inlining it. 
- If a `component.schema` exists that matches the `items` in a new operation's request or response array, it will be used instead of inlining that schema. 
- If you are documenting multiple operations at once (ie `--document all), Optic will use previous schemas to inform what it generates next. 

[1] close via schema closeness algorithm (also in this PR)

Future: 
- Make it easy to rename the generated schemas 
- Patch with $ref (not just document) 

Screenshot 2023-03-05 at 11 29 15 AM

## 📚 References

## 👹 QA


projects/openapi-cli/src/commands/diffing/document.ts
```diff
@@ -513,7 +513,28 @@ export function addOperations(
         })(documentedBodies)
       );
 
+      const addedPaths = new Set<string>();
```
acunniffe, line 516: only try to refactor added paths, don't touch stuff that was already there.

projects/openapi-cli/src/shapes/closeness/closeness.ts
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  Object.entries(other).forEach(([key, value]) => {
+    tuples.push([jsonPointerHelpers.append(trail, key), value]);
+  });
+
+  return tuples;
+}
+
+export function computeClosenessCached(
```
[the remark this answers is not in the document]
niclim: I think if we wanted to cache this, we could cache `setB` of all the component schemas

projects/openapi-cli/src/shapes/closeness/schema-inventory.ts
```diff
@@ -0,0 +1,273 @@
...
+  async *refsForAdditions(
+    addedPaths: Set<string>,
+    spec: OpenAPIV3.Document
+  ): SpecPatches {
+    if (addedPaths.size === 0) return [];
+    const sorted = Array.from(addedPaths).sort();
+
+    for await (let added of sorted) {
+      const addedSchema = jsonPointerHelpers.get(spec, added);
+      const arrayItems = arrayItemPaths(addedSchema, added);
```
niclim, line 61: I think a completely fair optimization here is to only go 1 level down

projects/openapi-cli/src/shapes/closeness/closeness.ts
```diff
@@ -0,0 +1,124 @@
...
+      matchingIntersectSize =
+        matchingIntersectSize + intersection.size / union.size;
+    } else {
+      matchingIntersectSize = matchingIntersectSize + 1;
+    }
+  });
+
+  return matchingIntersectSize / keyUnionSize;
+}
+export function computeCloseness(
```
niclim, line 119: How does this work with nested refs? Does oas cli unwrap / deref references / references in component schemas?

projects/openapi-cli/src/shapes/closeness/schema-inventory.ts
```diff
@@ -0,0 +1,273 @@
...
+                },
+              ],
+            };
+            yield patch;
+          }
+        }
+      }
+      //create a root schema
+      if (!matchedSub && !matchedRoot && isARootSchema) {
+        const refName = refNameGenerator(added);
```
niclim, line 147: Is this always correct to generate ref for a root schema? Perhaps we should only scope this to root schemas that are an object?

niclim:
I am a little concerned about the performance, since we're computing the matrix of closeness between component schemas and every added schema (and all of it's children).

The performance improvements we have are:
- cache the component schema keys as a `set` (avoids walking component schemas multiple times)
- only look 1 level deep for schema matches in `added` schemas

projects/openapi-cli/src/shapes/closeness/closeness.ts
acunniffe: yeah on each cycle. I have a test case for that.

niclim (approved):

PR #1683, opened 2023-03-06
Clean up raw loading to make it possible for us to error when a file …

nalanj:
…that looks like a spec fails

## 🍗 Description

## 📚 References

## 👹 QA

### DIFF
--- a/projects/optic/src/commands/diff/diff-all.ts
+++ b/projects/optic/src/commands/diff/diff-all.ts
@@ -158,16 +158,32 @@ async function computeAll(
     // Cases we run the comparison:
     // - if to spec has x-optic-url
     // - if from spec has x-optic-url AND to spec is empty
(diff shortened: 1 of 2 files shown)


projects/optic/src/commands/diff/diff-all.ts
```diff
@@ -158,16 +158,32 @@ async function computeAll(
...
-      if (!specPathToLoad) {
-        logger.debug(
-          `Skipping comparison from ${candidate.from} to ${candidate.to} both are undefined`
-        );
-        continue;
+      rawSpec = await loadRaw(specPathToLoad);
+    } catch (e) {
+      if (e instanceof Error && e['probablySpec']) {
+        console.error(`Error parsing ${specPathToLoad}:`);
+        console.error(e.message);
```
niclim, line 175: Do we want to use the `logger` here?

niclim (approved):

PR #1689, opened 2023-03-08
Connect api coverage to spec patches

niclim:
## 🍗 Description

Connects up coverage to diffs, to specify which body nodes (or operations) have diffs

TODO
- [x] Test out coverage

## 📚 References

## 👹 QA

### DIFF
--- a/projects/optic/src/commands/oas/coverage/api-coverage.ts
+++ b/projects/optic/src/commands/oas/coverage/api-coverage.ts
@@ -124,6 +124,7 @@ export class ApiCoverageCounter {
         patch.diff?.kind === 'AdditionalProperty' ||
         patch.diff?.kind === 'MissingRequiredProperty'
       ) {
+        operation.diffs = true;
         const isResponse = parts[3] === 'responses';
         if (isResponse) {
           const [, _pathPattern, _method, , statusCode] = parts;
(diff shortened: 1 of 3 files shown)


projects/optic/src/commands/oas/coverage/api-coverage.ts
```diff
@@ -112,6 +115,38 @@ export class ApiCoverageCounter {
...
+    const [_, pathPattern, method] = parts;
+    const operation = this.coverage.paths[pathPattern]?.[method];
+    if (operation) {
+      if (
+        patch.diff?.kind === 'UnmatchedType' ||
+        patch.diff?.kind === 'AdditionalProperty' ||
+        patch.diff?.kind === 'MissingRequiredProperty'
+      ) {
+        operation.diffs = true;
+        const isResponse = parts[3] === 'responses';
```
acunniffe, line 129: please use https://github.com/opticdev/optic/blob/3fd14c632bc27f65fd474ca63f4a9fce731ecaa1/projects/json-pointer-helpers/src/json-pointers/json-pointer-helpers.ts#L60

acunniffe (approved):

acunniffe:
@niclim -- I'm not sure how this will work if you added a new response status code / body (that's a patch too)

niclim:
I think it should work as expected - document + patch happens first, and then we run another pass + reload the spec (if I understand the code correctly).

We reload the spec, and we use that as the base

PR #1691, opened 2023-03-08
fix generation for gitlab to checkout branch

niclim:
## 🍗 Description

Gitlab checks out using a detached head (https://gitlab.com/gitlab-org/gitlab/-/issues/15409) - we can force it to pull the branch name by adding a `git checkout command`. This is important, otherwise we cannot add a `gitbranch` tag to push events

## 📚 References

## 👹 QA


nalanj:
I don't think this is what we'd want because by default gitlab is checking out the speculative merge commit. We need to arbitrarily attach the branch name from within the CI code instead, I'm pretty sure. The same issue is true on GitHub right now.

PR #1692, opened 2023-03-08
Add warning if HEAD

niclim:
## 🍗 Description

## 📚 References

## 👹 QA


projects/optic/src/commands/diff/upload-diff.ts
```diff
@@ -48,6 +48,13 @@ export async function uploadDiff(
       const currentBranch = await Git.getCurrentBranchName();
       if (currentBranch !== 'HEAD') {
         tags.push(sanitizeGitTag(`gitbranch:${currentBranch}`));
+      } else {
+        logger.warn(
```
nalanj, line 52: Would be good to only show this warning if there's not a manually set tag from the CLI.

nalanj (approved):

PR #1693, opened 2023-03-09
add `--tag` on diff and diff-all

niclim:
## 🍗 Description

Will also need to update all generated code to start using `--tag` in diff-all snippets

## 📚 References

## 👹 QA

### DIFF
--- a/projects/optic/src/commands/diff/diff-all.ts
+++ b/projects/optic/src/commands/diff/diff-all.ts
@@ -62,6 +62,10 @@ export const registerDiffAll = (cli: Command, config: OpticCliConfig) => {
       'HEAD~1'
     )
     .option(
+      '--tag <tag>',
+      'Adds additional tags to the HEAD spec. Should be used in conjunction with `--upload`'
+    )
+    .option(
       '--match <match>',
       'a glob to match specs (e.g. "**/*.yml" or "**/specifications/*.json"). Also takes \
       comma separated values (e.g. "**/*.yml,**/*.json")'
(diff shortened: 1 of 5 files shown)


projects/optic/src/commands/diff/upload-diff.ts
```diff
@@ -43,13 +47,16 @@ export async function uploadDiff(
 
   if (specs.to.context && specDetails) {
     let tags: string[] = [];
+    tags.push(...getTagsFromOptions(options.tag));
     if (specs.to.context.vcs === VCS.Git) {
       tags.push(`git:${specs.to.context.sha}`);
       const currentBranch = await Git.getCurrentBranchName();
```
nalanj, line 53: Should we only do this if a gitbranch tag isn't set from the command line?

projects/optic/src/commands/diff/diff.ts
```diff
@@ -67,6 +67,10 @@ export const registerDiff = (cli: Command, config: OpticCliConfig) => {
       '--standard <standard>',
       'run comparison with a locally defined standard, if not set, looks for the standard on the [x-optic-standard] key on the spec, and then the optic.dev.yml file.'
     )
+    .option(
+      '--tag <tag>',
```
nalanj, line 71: Would `--head-tag` be clearer here?

projects/optic/src/commands/diff/diff-all.ts
```diff
@@ -61,6 +61,10 @@ export const registerDiffAll = (cli: Command, config: OpticCliConfig) => {
       'the base ref to compare against. Defaults to HEAD~1',
       'HEAD~1'
     )
+    .option(
+      '--tag <tag>',
```
nalanj, line 65: Would `--head-tag` be clearer here?

nalanj (approved):

PR #1694, opened 2023-03-09
Connect upload to verify

niclim:
## 🍗 Description

Connects the verify command to the backend to upload verification data.

The following combinations aren't usable with verify
- `--upload` + `--document` or `--patch`
- `--upload` not in a git repo OR git repo working dir is not clean
- spec doesn't have an `x-optic-url`

## 📚 References

## 👹 QA

### DIFF
--- a/projects/optic/src/commands/oas/verify.ts
+++ b/projects/optic/src/commands/oas/verify.ts
@@ -4,7 +4,7 @@ import path from 'path';
 import * as fs from 'fs-extra';
 
 import { createCommandFeedback, InputErrors } from './reporters/feedback';
-import { flushEvents, trackCompletion, trackEvent } from './lib/segment';
+import { flushEvents, trackEvent } from './lib/segment';
 import { trackWarning } from './lib/sentry';
 import * as AT from './lib/async-tools';
 import { OpenAPIV3, readDeferencedSpec } from './specs';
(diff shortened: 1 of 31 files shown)


projects/optic/src/commands/oas/verify.ts
```diff
@@ -188,6 +216,38 @@ export async function verifyCommand(): Promise<Command> {
...
+          process.exitCode = 1;
+          return;
+        }
+
+        const { orgId, apiId } = opticUrlDetails;
+        const tags: string[] = [];
+        if (config.vcs?.type === VCS.Git) {
+          tags.push(`git:${config.vcs.sha}`);
+          const currentBranch = await Git.getCurrentBranchName();
+          tags.push(sanitizeGitTag(`gitbranch:${currentBranch}`));
```
niclim, line 236: Do you think it's worth adding `--tag` here? I don't think we'll run into the detached head state in verify if it's used locally. 

And for now, we can only really run this locally (i.e. there isn't a good way to run this in CI b/c of captures)

projects/optic/src/commands/oas/verify.ts
```diff
@@ -38,8 +45,33 @@ export async function verifyCommand(): Promise<Command> {
...
+      if (options.upload) {
+        if (options.document || options.patch) {
+          console.error(
+            'Cannot run upload with --document or --patch options.'
+          );
+          process.exitCode = 1;
+          return;
+        } else if (
+          config.vcs?.type !== VCS.Git ||
+          config.vcs.status === 'dirty'
```
niclim, line 66: I'm not sure this is too strict, but it's not very useful in our current system to run verify --upload outside of a git repo / with a "dirty" working dir.

The spec wouldn't have any tags, and wouldn't really show up anywhere

projects/optic/src/commands/oas/verify.ts
```diff
@@ -188,6 +221,62 @@ export async function verifyCommand(): Promise<Command> {
...
+                Object.values(OpenAPIV3.HttpMethods).includes(method as any)
+              ) {
+                const endpointChecksum = computeEndpointChecksum(
+                  path,
+                  method,
+                  methodObj as OpenAPIV3.OperationObject
+                );
+                const pathHashes = endpointHashes[path] ?? {};
+                pathHashes[method] = endpointChecksum;
+                endpointHashes[path] = pathHashes;
```
niclim, line 266: I'm computing the endpoint hashes separately, outside of the coverage datastructure. Keying this with the non-denormalized path

PR #1701, opened 2023-03-15
bundle commmand

acunniffe:
## 🍗 Description
Proper bundle command for specs. Will leave everything dereferenced except Schemas, Examples and Parameters. Those $refs will be insourced. 

limitations: 
This will always output a valid OpenAPI specification, but to do that consistently, means it will not always use pretty names for components with conflicts or imported from other files. 

## 📚 References

## 👹 QA


acunniffe:
Built this late last night. Giving to users today for feedback

acunniffe:
@niclim when reviewing this the best way to think about what the code is doing is reading the sourcemap, and selectively "undoing" the dereference step. 

We're allowing `components` . `schemas|examples|parameters` -- so any `$ref` usage in one those slows will first be copied to `components[type]`, and then all of the usages of that `$ref` will be replaced with `{$ref: <new-path>}`. 

This is done recursively tail-first, which is the weirdest requirement because depth !== level of nesting 1:1 for annoying reasons ie openapi has reponses and request bodies that start in different places, parameters can be in operations, or path, etc.

projects/optic/src/commands/bundle/bundle.ts
```diff
@@ -0,0 +1,507 @@
...
+    .argument('[file_path]', 'openapi file to bundle')
+    .option('-o [output]', 'output file name')
+    .option(
+      '--filter-x-extensions [extensions]',
+      'extensions to filter when truthy value set'
+    )
+    .option(
+      '--include-x-extensions [extensions]',
+      'extensions to filter when truthy value set'
+    )
```
niclim, line 47: I'm not convinced this is the right place for filtering. This feels like it should be another command that's run before and piped into this function, otherwise we're setting ourselves up to support `--filter-x-extensions` / `--include-x-extensions` on _every_ optic command

projects/optic/src/commands/bundle/bundle.ts
```diff
@@ -0,0 +1,507 @@
...
+    let parsedFile: ParseResult;
+    if (filePath) {
+      parsedFile = await getSpec(filePath, config);
+
+      const updatedSpec = bundle(parsedFile.jsonLike, parsedFile.sourcemap);
+
+      if (includeExtensions.length) {
+        Object.entries(updatedSpec.paths).forEach(([path, operations]) => {
+          Object.entries(operations!).forEach(([key, operation]) => {
+            if (key === 'parameters') return;
```
niclim, line 87: Should use the `OpenAPIV3.HttpMethods` check

projects/optic/src/commands/bundle/bundle.ts
```diff
@@ -0,0 +1,507 @@
...
+              }
+            }
+          });
+        });
+      }
+
+      if (filterExtensions.length) {
+        Object.entries(updatedSpec.paths).forEach(([path, operations]) => {
+          Object.entries(operations!).forEach(([key, operation]) => {
+            if (key === 'parameters') return;
```
niclim, line 111: Should use the `OpenAPIV3.HttpMethods` check

projects/optic/src/commands/bundle/bundle.ts
```diff
@@ -0,0 +1,507 @@
...
+        });
+
+      if (o) {
+        // write to file
+        const outputPath = path.resolve(o);
+        await fs.writeFile(
+          outputPath,
+          isYaml(o) ? yamlOut() : JSON.stringify(updatedSpec, null, 2)
+        );
+        console.log('wrote bundled spec to ' + path.resolve(o));
```
niclim, line 145: should use `../logger.ts` instead of console.logs so we can control the log levels

projects/optic/src/commands/bundle/bundle.ts
```diff
@@ -0,0 +1,507 @@
...
+        }
+      }
+    } else {
+      console.error('No specification found');
+      process.exitCode = 1;
+      return;
+    }
+  };
+
+const methods = `{get,post,put,delete,patch,head,options}`;
```
niclim, line 161: could do `{${OpenAPIV3.HttpMethods.values().join(',')}}`

niclim (commented):
can we also add tests in the `integrations` folder

projects/optic/src/commands/bundle/bundle.ts
acunniffe: Do you know how to hide options in commander. Talked to folks about hiding these and then moving them off

niclim (approved):

PR #1704, opened 2023-03-15
fix `optic api add` to push spec when not in git repo

niclim:
## 🍗 Description

Two fixes
- if a user is outside of a git repo, push the current spec
- if a user is in a git repo, but the repo has uncommitted changes, exit optic api add and log an error
  - I thought about being a little more lax here, and telling them to run `optic spec push` instead, but I think this is easier to explain and be stricter

## 📚 References

## 👹 QA

### DIFF
--- a/projects/optic/src/commands/api/add.ts
+++ b/projects/optic/src/commands/api/add.ts
@@ -261,6 +261,18 @@ export const getApiAddAction =
       );
       process.exitCode = 1;
       return;
+    } else if (config.vcs?.type === VCS.Git && config.vcs.status === 'dirty') {
+      logger.error(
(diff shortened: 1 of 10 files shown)


projects/optic/src/commands/api/add.ts
```diff
@@ -253,6 +261,18 @@ export const getApiAddAction =
       );
       process.exitCode = 1;
       return;
+    } else if (config.vcs?.type === VCS.Git && config.vcs.status === 'dirty') {
```
Alexandre-Herve, line 264: I don't understand this case, why not let them add an untracked API when in a git repo?

PR #1750, opened 2023-03-22
fix line width oas

acunniffe:
## 🍗 Description
Generated specs with long Ref names can end up with multiple line and the weird yaml `\` syntax

## 📚 References

## 👹 QA


projects/openapi-io/src/write/index.ts
```diff
@@ -1,7 +1,10 @@
 import yaml from 'yaml';
 
 export function writeYaml(document: any, indent: 2 | 4 = 2) {
```
niclim, line 3: do we use the `indent` variable anywhere? If not can we remove it?

niclim (approved):

PR #1751, opened 2023-03-22
oas document command

acunniffe:
## 🍗 Description
Now you document your APIs by running 
`optic oas document --all` or `optic oas document "get /xyz" "post /abc"`

Captures are more explicit 

```
» Verifying API behavior with traffic from last 10 captures. Reset captures with "optic oas clear todo-api.yaml"` 
```

Verify doesn't have a document flag anymore. 

## 📚 References

## 👹 QA

### DIFF
--- a/projects/optic/src/commands/oas/capture.ts
+++ b/projects/optic/src/commands/oas/capture.ts
@@ -127,7 +127,6 @@ export async function captureCommand(config: OpticCliConfig): Promise<Command> {
       }
 
       let destination: Writable = fs.createWriteStream(inProgressName);
-      let enterPressed = false;
 
       const commandRunner = new RunCommand(proxyUrl, feedback);
 
@@ -148,7 +147,6 @@ export async function captureCommand(config: OpticCliConfig): Promise<Command> {
             lines.close();
           };
           if (runningCommand) await commandRunner.kill();
-          enterPressed = true;
 
           sourcesController.signal.addEventListener('abort', onAbort);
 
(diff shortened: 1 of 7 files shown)


projects/optic/src/commands/oas/captures/capture-storage.ts
```diff
@@ -6,7 +6,7 @@ import crypto from 'crypto';
 
 export async function captureStorage(
   filePath: string
-): Promise<[boolean, string]> {
+): Promise<[boolean, string, number]> {
```
niclim, line 9: Can we change this into an object with keys? Easier to understand if it's keyed rather than a tuple for larger tuple sizes

projects/optic/src/commands/oas/diffing/document.ts
```diff
@@ -73,18 +75,19 @@ export async function addIfUndocumented(
 }
 
 async function computeOperationsToAdd(
-  input: string,
+  operationsToAdd: ParsedOperation[],
+  isAddAll: boolean,
```
niclim, line 79: Instead of having isAddAll, why not just pass the entire list operations if `isAddAll` and the smaller list of operations.

This avoids having to pass down 2 laters of fn `addIfUndocumented` and `computeOperationsToAdd`

projects/optic/src/commands/oas/verify.ts
```diff
@@ -65,6 +66,16 @@ export function verifyCommand(config: OpticCliConfig): Command {
     )
     .action(async (specPath) => {
       const options = command.opts();
+
+      if (options.document) {
+        return console.log(
+          `--document flag has been removed. You are looking for "optic oas document ${path.relative(
```
niclim, line 72: This sounds a little awkward - `Run: optic oas document <spec> {document args}`

niclim (commented):
mostly nits, but I think that document and patch should be the same command (maybe named differently

niclim (changes requested):

projects/optic/src/commands/oas/diffing/document.ts
acunniffe: We don't know the operations at the earlier stage

projects/optic/src/commands/oas/verify.ts
```diff
@@ -52,8 +52,6 @@ export function verifyCommand(config: OpticCliConfig): Command {
     )
     .option('--har <har-file>', 'path to HttpArchive file (v1.2, v1.3)')
     .option('--exit0', 'always exit 0')
-    .option('--document <operations>', 'HTTP method and path pair(s) to add')
-    .option('--patch', 'Patch existing operations to resolve diffs')
```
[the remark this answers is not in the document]
acunniffe: If we're moving these to the root I don't think it makes sense to anymore. We also tell them the follow up commands to run throughout

niclim (approved):

projects/optic/src/commands/oas/capture.ts
```diff
@@ -24,10 +24,44 @@ import { platform } from './lib/shell-utils';
...
 import { runVerify } from './verify';
 import { OpticCliConfig } from '../../config';
+import Path from 'path';
+import os from 'os';
+const tmpDirectory = os.tmpdir();
 
 export async function captureCommand(config: OpticCliConfig): Promise<Command> {
   const command = new Command('capture');
 
+  const clear = new Command('clear');
```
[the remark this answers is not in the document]
acunniffe: done

PR #1772, opened 2023-03-28
only open changelog web when has results

niclim:
## 🍗 Description

## 📚 References

## 👹 QA

### DIFF
--- a/projects/optic/src/commands/diff/diff.ts
+++ b/projects/optic/src/commands/diff/diff.ts
@@ -285,35 +285,36 @@ const getDiffAction =
         (!options.check || diffResult.specResults.results.length === 0)
       ) {
         logger.info('Empty changelog: not opening web view');
-      }
-      const analyticsData: Record<string, any> = {
-        isInCi: config.isInCi,
-      };
-
-      if (!maybeUrl) {
-        const meta = {
-          createdAt: new Date(),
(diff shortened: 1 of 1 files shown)


projects/optic/src/commands/diff/diff.ts
```diff
@@ -285,35 +285,36 @@ const getDiffAction =
...
-          baseParseResult,
-          headParseResult,
-          diffResult.specResults,
-          meta
-        );
-        analyticsData.compressedDataLength = compressedData.length;
-        logger.info('Opening up diff in web view');
-        maybeUrl = `${config.client.getWebBase()}/cli/diff#${compressedData}`;
-        await flushEvents();
+        if (!maybeUrl) {
```
nalanj, line 293: I must not be understanding, but why would we open the maybeUrl if it's falsy?

niclim: Hmm so the git diff is a little hard to read - https://github.com/opticdev/optic/blob/04af3bd5a719776f3342757c06057878cb7224b1/projects/optic/src/commands/diff/diff.ts#L289-L316

maybeUrl refers to the uploaded url, if there isn't a url we set it as the compressed cli/diff route

nalanj (approved):

PR #1806, opened 2023-04-04
Fix bundle order

acunniffe:
## 🍗 Description
Fixes a bug bundling nested refs

## 📚 References

## 👹 QA

### DIFF
--- a/projects/json-pointer-helpers/src/json-pointers/json-pointer-helpers.ts
+++ b/projects/json-pointer-helpers/src/json-pointers/json-pointer-helpers.ts
@@ -60,7 +60,7 @@ function relative(pointer: string, from: string) {
 function startsWith(
   pointer: string,
   pattern: string[],
-  exact: boolean = false
+  options: { exact: boolean } = { exact: false }
 ): boolean {
   const components = parse(pointer);
   const sliced = components.slice(0, pattern.length);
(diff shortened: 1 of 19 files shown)


projects/json-pointer-helpers/src/json-pointers/json-pointer-helpers.ts
```diff
@@ -57,15 +57,21 @@ function relative(pointer: string, from: string) {
   return compile(targetDecoded.slice(fromDecoded.length));
 }
 
-function startsWith(pointer: string, pattern: string[]): boolean {
+function startsWith(
+  pointer: string,
+  pattern: string[],
+  exact: boolean = false
```
niclim, line 63: Can we change this to be an `options` arg with key exact so it's easier to read when it's used downstream

niclim (commented):
Could you explain what this fixes?

acunniffe:
> Could you explain what this fixes?

When we bundle nested references, we need to rewrite JSON Trails ie `/path/method/responses/200/..../schema/.../$ref` becomes `/components/NewName/$ref`. 

The code that decided which references to rewrite was relying on using a string `startsWith` instead of using the JSON Path Matchers. This meant that you might match two similar, but different refs:

```
/...the-same-path/.../customer
/...the-same-path/.../customerContact
```
Would match even though they are siblings and different. The fix uses our existing code to compare the actual path, not their string representation. I added the exact flag as a precaution in case someone had contentType `**/**` which minimatch would respect.

projects/json-pointer-helpers/src/json-pointers/json-pointer-helpers.ts
acunniffe: Done

niclim (approved):
Could be good to include a test here as well

PR #1810, opened 2023-04-06
Add api url

nalanj:
## 🍗 Description

Adds a new `add-api-url` command to Optic that will simply take in an API URL and add it to the given spec file.

## 📚 References

## 👹 QA


niclim (approved):
Do you think this should go on `api` instead of `spec`? I'm not sure where this would fit with our current distinctions

nalanj:
@niclim I just figured it goes under spec because it touches on an individual spec file rather than touching a whole api. I'm not sure what the distinction between spec or api for us is.

nalanj:
@niclim merge or should we talk to @acunniffe about where it goes?

acunniffe:
I feel like this is better as a "add --no-verify" flag or something like that. Is the intent that this doesn't try to upload, but will make an API in optic cloud?

nalanj:
@acunniffe No. It's just for marking up the spec with the x-optic-url. It comes up if you're doing things like updating a public spec in a job or are generating a spec and need to add the url to the file without it being added to the cloud since it's already there.

niclim:
I think it's fine to just throw in spec for now, I don't think we have a strong distinction between API or spec right now, and since we think it's not a common use case we can throw it in, and then just alias it later

PR #1818, opened 2023-04-10
Allow --validation strict or loose in diff and diff-all

niclim:
## 🍗 Description

Allows a user to specify `--validation strict / loose` which defaults to strict (the default values)

Also fixes a bug with loose validation where we weren't failing on below
```yml
responses:
  "202": null
```

## 📚 References

## 👹 QA

### DIFF
--- a/projects/optic/src/commands/diff/diff-all.ts
+++ b/projects/optic/src/commands/diff/diff-all.ts
@@ -1,4 +1,4 @@
-import { Command } from 'commander';
+import { Command, Option } from 'commander';
 import pm from 'picomatch';
 import { OpticCliConfig, VCS } from '../../config';
 import { findOpenApiSpecsCandidates } from '../../utils/git-utils';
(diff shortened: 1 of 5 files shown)


projects/optic/src/commands/diff/diff-all.ts
```diff
@@ -79,6 +79,14 @@ comma separated values (e.g. "**/*.yml,**/*.json")'
       '--standard <standard>',
       'run comparison with a locally defined standard, if not set, looks for the standard on the [x-optic-standard] key on the spec, and then the optic.dev.yml file.'
     )
+    .addOption(
+      new Option(
+        '--validation <validation>',
```
niclim, line 84: It's kind of more correct to call it `--head-validation`, however this would be inconsistent with diff since if you run `diff file1 file2` that --validation applies to both

nalanj (approved):

PR #1825, opened 2023-04-11
fix

acunniffe:
## 🍗 Description
Fixes bundle command. 

We're using a tree walking algorithm to handle nested refs, but sometimes a nested ref containing another nested ref will break the assumption that we can patch these in reverse depth order. 

The correct long term fix is to change how this works, but for now I think we can apply operations sequentially and test their targets. 

## 📚 References

## 👹 QA

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "openapi-workspaces",
   "license": "MIT",
   "private": true,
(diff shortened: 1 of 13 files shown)


niclim (approved):
Can we also include an integration test as well to run on a nested ref?

PR #1830, opened 2023-04-12
Update spectral rule to read severity from spectral result

niclim:
## 🍗 Description

Updates spectral rule to read from spectral severity

Based off https://github.com/opticdev/optic/pull/1829

## 📚 References

## 👹 QA


acunniffe (approved):

projects/rulesets-base/src/extended-rules/spectral-rule.ts
```diff
@@ -54,6 +64,15 @@ function toOpticResult(
   };
 }
 
+// This will treat info + hint as a info in Optic
+function severityToOpticSeverity(spectralSev: DiagnosticSeverity): Severity {
+  return spectralSev === DiagnosticSeverity.Error
```
acunniffe, line 69: looks good. If I'm reading this right it looks like we only support `error` `warn` and `info`? seems reasonable.

niclim: Yes thats correct

niclim:
There's a bug here that I need to fix, spectral rules don't appear to be running correctly on added in certain cases - need to spend some time to figure out and fix this

projects/rulesets-base/src/extended-rules/spectral-rule.ts
```diff
@@ -186,8 +205,21 @@ export class SpectralRule extends ExternalRuleBase {
...
           changesByJsonPath[fact.location.jsonPath];
+
+        if (
+          !maybeChange &&
+          jsonPointerHelpers.matches(fact.location.jsonPath, [
+            'paths',
+            '**',
+            `{${Object.values(OpenAPIV3.HttpMethods).join(',')}}`,
+          ])
+        ) {
```
niclim, line 218: @acunniffe just an FYI, this is kind of weird but it's the same thing as the groupDiffs w/ path.

There's a few solutions to fixing this, one of which is some sort of internal spec format, the other being a better abstraction around the fact trees / jsonPath -> Fact type

PR #1839, opened 2023-04-14
forward effective_at to tags for api history add

Alexandre-Herve:
## 🍗 Description

Trying to do the following:

- when adding historical specs form `api add`, use `spec.effective_at` as `spec_tag.effective_at`
- otherwise just let bwts use `new Date()` as `spec.effective_at`

- [x] Need to adapt the `/spec/id/tag` route first.

@niclim am I doing it correctly?

## 📚 References

## 👹 QA

### DIFF
--- a/projects/optic/src/client/optic-backend.ts
+++ b/projects/optic/src/client/optic-backend.ts
@@ -137,13 +137,15 @@ export class OpticBackendClient extends JsonHttpClient {
     git_name?: string;
     git_email?: string;
     commit_message?: string;
+    forward_effective_at_to_tags?: boolean;
   }): Promise<{ id: string }> {
     return this.postJson(`/api/specs`, spec);
   }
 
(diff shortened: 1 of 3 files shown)


niclim:
I think you'll also need to update the spec push command which calls `uploadSpec` 

- On new spec https://github.com/opticdev/optic/blob/main/projects/optic/src/utils/cloud-specs.ts#L70
- On existing spec https://github.com/opticdev/optic/blob/main/projects/optic/src/utils/cloud-specs.ts#L86

Alexandre-Herve:
After putting some more thoughts in this morning I think we're good re: `api add` and `spec push`:
- the default is that we don't forward spec effective_at
- in the `api add` command we do request to forward effective_at when creating the spec and when tagging it

We also upload specs in `verify` and `diff` commands but I think we don't want to handle history there either.

niclim (approved):

PR #1843, opened 2023-04-16
require url to get example postman

acunniffe:
## 🍗 Description

## 📚 References

## 👹 QA


acunniffe:
I think it's ok to have a request w/o a response, but not the other way around. The spec would still be populated with useful paths, and parameters and as you save more example responses they'd show up as expected. 

It's the response w/o a concrete, actual, path that was causing the issues.

niclim (approved):

PR #1853, opened 2023-04-19
Add `optic api create` command

niclim:
## 🍗 Description

Adding an optic api create command, this would be similar to the github create repo case where you dont have a spec but want a place holder.

this fits into the spec generation case, where you would want to specify a URL and add to it when generating, rather than creating a spec and then adding the url

## 📚 References

## 👹 QA

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "openapi-workspaces",
   "license": "MIT",
   "private": true,
-  "version": "0.42.10",
+  "version": "0.42.11",
   "workspaces": [
     "projects/json-pointer-helpers",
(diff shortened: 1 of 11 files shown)


acunniffe:
Is this better on cloud?

niclim:
> Is this better on cloud?

Hmm, maybe? I thought it was better here since we try pull `default_branch`, `web_url` and get you in the CLI, but I think we could have both

I'll put up a PR to add this into cloud as well, and if think that having both is confusing, I can close this PR

nalanj (approved):

Issue #1857, opened 2023-04-20
Reverse proxy mode


Manouchehri:
**Describe the enhancement**
In some situations, like serverless functions, setting a proxy isn't trivial; setting a different server URL would be trivial though!

**Additional information**
https://docs.mitmproxy.org/stable/concepts-modes/#reverse-proxy

Example usage:
```
dave@mbp corellium-openapi % optic capture openapi.json https://moda.enterprise.corellium.com/api --reverse-proxy
» Reverse proxy running on https://localhost:8000.
 help  Press [ Enter ] to finish capturing requests
⠋ 2 requests captured
```

And the usage would look like:
```
curl -kv "https://localhost:8000/api/example"
```

niclim:
Hi, thanks for using Optic and the issue. We just released `0.42.12` which includes the `--reverse-proxy` command for capture which should work as you described.

_(this was actually under `--no-system-proxy` but we renamed it to `--reverse-proxy` as its easier to understand / more discoverable)_

Manouchehri:
Hmm, mind testing this for me? I can't get it to work.

```sh
optic capture --reverse-proxy openapi.json https://moda.enterprise.corellium.com
```

```
dave@mbp corellium-openapi % optic capture --reverse-proxy openapi.json https://moda.enterprise.corellium.com
» Proxy running on https://localhost:8000. System proxy updated
 help  Press [ Enter ] to finish capturing requests

Not seeing any traffic captured? Make sure your HTTP Client is using the proxy: https://www.useoptic.com/docs/oas-reference/client-guides
⠼ 2 requests captured
```

```
dave@mbp ~ % curl -kv "https://localhost:8000/api/v1/ready"
*   Trying 127.0.0.1:8000...
* Connected to localhost (127.0.0.1) port 8000 (#0)
* ALPN: offers h2
* ALPN: offers http/1.1
* [CONN-0-0][CF-SSL] (304) (OUT), TLS handshake, Client hello (1):
* [CONN-0-0][CF-SSL] (304) (IN), TLS handshake, Server hello (2):
* [CONN-0-0][CF-SSL] (304) (IN), TLS handshake, Unknown (8):
* [CONN-0-0][CF-SSL] (304) (IN), TLS handshake, Certificate (11):
* [CONN-0-0][CF-SSL] (304) (IN), TLS handshake, CERT verify (15):
* [CONN-0-0][CF-SSL] (304) (IN), TLS handshake, Finished (20):
[...]

PR #1866, opened 2023-04-21
fix: reverse-proxy against remote urls

niclim:
## 🍗 Description

running `optic capture spec https://api.useoptic.com --reverse-proxy` would strip out the protocol, which meant `https` servers would sometimes return a 301 (because we tried to hit the http:// endpoint)

additionally, we were not updating the host header, leading to certain servers rejecting requests. manually updating the host header fixes this (and it appears redirection also works now with the correct host header)

## 📚 References
https://github.com/opticdev/optic/issues/1857

## 👹 QA

### DIFF
--- a/projects/optic/src/commands/oas/captures/streams/sources/proxy.ts
+++ b/projects/optic/src/commands/oas/captures/streams/sources/proxy.ts
@@ -95,6 +95,7 @@ export class ProxyInteractions {
         beforeRequest: onTargetedRequest,
         forwarding: {
           targetHost: targetOrigin,
+          // bug: updateHostHeader isn't rewriting the request header - we're manually rewriting this in `beforeRequest`
           updateHostHeader: true,
         },
         trustAdditionalCAs: options.targetCA || [],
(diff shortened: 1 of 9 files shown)


projects/optic/src/commands/oas/captures/streams/sources/proxy.ts
```diff
@@ -39,7 +39,7 @@ export class ProxyInteractions {
       targetCA?: Array<{ cert: Buffer | string }>;
     } = {}
   ): Promise<[ProxyInteractions, string, string]> {
-    let { host, protocol } = new URL(targetHost);
+    let { host, protocol, origin: targetOrigin } = new URL(targetHost);
     if (targetHost.includes('/')) {
       // accept urls to be passed in rather than pure hosts
       targetHost = host;
```
niclim, line 45: This function is a little overloaded, but we're generating the transparent proxy and the mockttp server here. I'm fixing specifically the mockttp server case, and leaving the transparent proxy host logic intact

acunniffe (approved):

PR #1867, opened 2023-04-21
support nullable properly across versions

acunniffe:
## 🍗 Description
Optic was generating invalid schemas sometimes because `3.1` requires you supports a null type while `3.0` only supports `nullable`
## 📚 References

## 👹 QA

### DIFF
--- a/projects/optic/src/commands/oas/shapes/patches/generators/type.ts
+++ b/projects/optic/src/commands/oas/shapes/patches/generators/type.ts
@@ -8,11 +8,13 @@ import { SchemaObject } from '../../schema';
 import { jsonPointerHelpers } from '@useoptic/json-pointer-helpers';
 import { Schema } from '../../schema';
 import { ShapeLocation } from '../..';
+import { SupportedOpenAPIVersions } from '@useoptic/openapi-io';
 
 export function* typePatches(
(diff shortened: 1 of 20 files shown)


projects/optic/src/commands/oas/shapes/patches/generators/type.ts
```diff
@@ -76,32 +78,108 @@ export function* typePatches(
...
-      PatchImpact.Addition,
-      !shapeContext.location
-        ? PatchImpact.BackwardsCompatibilityUnknown
-        : 'inRequest' in shapeContext.location
-        ? PatchImpact.BackwardsCompatible
-        : PatchImpact.BackwardsIncompatible,
-    ],
-    groupedOperations: makeOneOfOperations(),
-  };
+  if (openAPIVersion === '3.0.x') {
```
niclim, line 87: Can we write tests for these 4 different code paths?

niclim (approved):

projects/optic/src/commands/oas/shapes/patches/generators/type.ts
acunniffe: So there are two existing tests which have been updated for when a field becomes nullable. And another two when a null field is first seen

PR #1874, opened 2023-04-24
Update to run against URLs

niclim:
## 🍗 Description

Updates to allow diff / other things against a URL (not a optic hosted spec - will be in a different PR).

Example usage - `optic diff null: https://raw.githubusercontent.com/opticdev/optic/main/projects/optic/specs/test-spec.yml --web`

## 📚 References

## 👹 QA

### DIFF
--- a/projects/optic/src/commands/diff/diff.ts
+++ b/projects/optic/src/commands/diff/diff.ts
@@ -207,13 +207,14 @@ const runDiff = async (
     }
 
     logger.info('');
-
-    if (!hasOpticUrl) {
+    if ((!hasOpticUrl && headFile.from === 'file') || headFile.from === 'git') {
+      const relativePath = path.relative(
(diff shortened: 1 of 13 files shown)


projects/optic/src/commands/diff/diff.ts
```diff
@@ -208,13 +207,14 @@ const runDiff = async (
     }
 
     logger.info('');
-
-    if (!hasOpticUrl) {
+    if ((!hasOpticUrl && headFile.from === 'file') || headFile.from === 'git') {
```
niclim, line 210: This was incorrect before:
- It logged when it was an empty spec
- It added a path that wasn't copy paste-able (i.e. it only showed the base path)
- It logged out the base file, not the head file path (i.e. could have been a `git:...`

PR #1883, opened 2023-04-26
Support --include-uncommitted-changes and upload with cloud spec

niclim:
## 🍗 Description

Two fixes:
- add an `--include-uncommitted-changes` flag to diff and diff-all
- the upload flow to handle cloud specs (i.e. we don't upload them, but instead we'll point at the already uploaded spec)

TODO
- [ ] write a test for this upload flow after https://github.com/opticdev/optic/pull/1882 lands

## 📚 References

## 👹 QA

### DIFF
--- a/projects/optic/src/commands/diff/upload-diff.ts
+++ b/projects/optic/src/commands/diff/upload-diff.ts
@@ -39,25 +39,26 @@ export async function uploadDiff(
   let baseSpecId: string | null = null;
   let headSpecId: string | null = null;
   if (specs.from.context && specDetails) {
-    const tags =
-      specs.from.context.vcs === VCS.Git
-        ? [`git:${specs.from.context.sha}`]
-        : [];
-    baseSpecId = await uploadSpec(specDetails.apiId, {
-      spec: specs.from,
-      client: config.client,
-      tags,
-      orgId: specDetails.orgId,
-    });
+    if (specs.from.context.vcs === VCS.Git) {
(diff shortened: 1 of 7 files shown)


projects/optic/src/commands/diff/upload-diff.ts
```diff
@@ -30,34 +29,30 @@ export async function uploadDiff(
...
-    });
+    if (specs.from.context.vcs === VCS.Git) {
+      const tags = [`git:${specs.from.context.sha}`];
+      baseSpecId = await uploadSpec(specDetails.apiId, {
+        spec: specs.from,
+        client: config.client,
+        tags,
+        orgId: specDetails.orgId,
+      });
+    } else if (specs.from.context.vcs === VCS.Cloud) {
```
niclim, line 44: Fixes the upload flow when using `--compare-from cloud:tag` - since we'll set the spec id to whatever we downloaded from cloud

projects/optic/src/utils/organization.ts
```diff
@@ -3,7 +3,7 @@ import { OpticBackendClient } from '../client';
 
 export async function getOrganizationFromToken(
   client: OpticBackendClient,
-  message: string
+  message: string | false
```
niclim, line 6: Updates so that we can pass in false to just select the first org in certain cases (i.e. generated or CI)

nalanj (approved):

Issue #1608, opened 2023-02-09
Failed to set ECDH curve


isamauny:
Hi there,

I am trying to use Optic to capture traffic to a TLS endpoint. I did setup the TLS cli certificate, but still hit the $subject issue. 

My target endpoint uses the following ciphers:

# TLS 1.3 (suites in server-preferred order)
TLS_AES_256_GCM_SHA384 (0x1302)   ECDH x25519 (eq. 3072 bits RSA)   FS	256
TLS_CHACHA20_POLY1305_SHA256 (0x1303)   ECDH x25519 (eq. 3072 bits RSA)   FS	256
TLS_AES_128_GCM_SHA256 (0x1301)   ECDH x25519 (eq. 3072 bits RSA)   FS	128
# TLS 1.2 (suites in server-preferred order)
TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256 (0xc02f)   ECDH x25519 (eq. 3072 bits RSA)   FS	128
TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384 (0xc030)   ECDH x25519 (eq. 3072 bits RSA)   FS	256
TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256 (0xcca8)   ECDH x25519 (eq. 3072 bits RSA)   FS	256

Should this be working or have I hit a limitation of the embedded proxy ? 
BTW, I have tried proxying to same API with a 1.3 MITM proxy and it works.

Thanks (and hi @acunniffe !)

acunniffe:
Oh hi @isamauny ! We'll look into this more today. 

In the meantime if you did manage to get a HAR out of MITM proxy, did you know you can pipe that directly into Optic? 

```
optic oas capture [openapi] [target-host] --har /path/to/file.har
```

That should unblock you as we look into this

acunniffe:
@isamauny 

Can you share the openssl version on your local machine? And also the API server you are targeting? 

If you have time, could you also try targetting another API host to see if it's service specific. For instance `https://api.github.com`?

Manouchehri:
> Oh hi @isamauny ! We'll look into this more today.
> 
> In the meantime if you did manage to get a HAR out of MITM proxy, did you know you can pipe that directly into Optic?
> 
> ```
> optic oas capture [openapi] [target-host] --har /path/to/file.har
> ```
> 
> That should unblock you as we look into this

Is it still okay to use `oas`? I see it's marked as deprecated.

[...]

PR #1891, opened 2023-04-28
Fix/ECDH curve error 

niclim:
## 🍗 Description

The ECDH curve originated from a downstream dependencies (mockttp issue: https://github.com/httptoolkit/mockttp/issues/139)

Also includes the updateHostHeader fix so I'm removing the workaround I added here.

## 📚 References
https://github.com/opticdev/optic/issues/1608
## 👹 QA

### DIFF
--- a/projects/optic/src/commands/oas/captures/streams/sources/proxy.ts
+++ b/projects/optic/src/commands/oas/captures/streams/sources/proxy.ts
@@ -37,13 +37,16 @@ export class ProxyInteractions {
     options: {
       ca?: ProxyCertAuthority;
       targetCA?: Array<{ cert: Buffer | string }>;
-    } = {}
+      mode: 'reverse-proxy' | 'system-proxy';
+    }
   ): Promise<[ProxyInteractions, string, string]> {
-    let { host, protocol } = new URL(targetHost);
+    let { host, protocol, origin } = new URL(targetHost);
     if (targetHost.includes('/')) {
       // accept urls to be passed in rather than pure hosts
       targetHost = host;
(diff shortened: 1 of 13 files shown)


acunniffe (approved):

niclim:
looks liek there's a different bug with the mockttp lib, needs an upstream fix before we can merge this (it's adding an extra slash to forwarded requests)

projects/optic/src/commands/oas/captures/streams/sources/proxy.ts
```diff
@@ -37,14 +37,17 @@ export class ProxyInteractions {
...
+    }
   ): Promise<[ProxyInteractions, string, string]> {
-    let { host, protocol } = new URL(targetHost);
+    let { host, protocol, origin } = new URL(targetHost);
     if (targetHost.includes('/')) {
       // accept urls to be passed in rather than pure hosts
       targetHost = host;
     }
 
+    const forwardHost = options.mode === 'reverse-proxy' ? origin : targetHost;
```
niclim, line 49: Something changed downstream in mockttp that we need different hosts to forward to in 

Takes into account:
- reverse proxy needs the protocol as the forward host 
- system proxy needs the host (https://github.com/opticdev/optic/pull/1888 - fixed regression here)

I'm not entirely sure how this all fits together but I was able to hit github's api via this

acunniffe: Interesting. Might be good to document this in the code too -- I wonder if it may change again

PR #1897, opened 2023-05-02
feat: package optic as a single executable app

notnmeyer:
## 🍗 Description

⚠️ needs a lot of testing

## 📚 References

## 👹 QA

### DIFF
--- a/projects/optic/package.json
+++ b/projects/optic/package.json
@@ -112,5 +112,19 @@
     "update-notifier": "^5.1.0",
     "url-join": "^4.0.1",
     "whatwg-mimetype": "^3.0.0"
+  },
+  "pkg": {
+    "scripts": [
+      "build/commands/ruleset/upload.js"
+    ],
+    "assets": [
+      "../../node_modules/open/xdg-open",
+      "../../node_modules/vm2/lib/setup-sandbox.js"
+    ],
+    "targets": [
+      "node18-macos-arm64",
+      "node18-macos-x64"
+    ],
+    "outputPath": "../../dist"
   }
 }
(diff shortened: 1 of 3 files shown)


notnmeyer:
https://github.com/vercel/pkg#targets

> macOS: possible to build x64 on arm64 with Rosetta 2 but not opposite.

this will be somewhat sticky since GHA doesn't offer apple silicon macs yet as far as i can tell. we can build x64 which will run under rosetta2, but if we want proper builds for folks to install locally we'll need somewhere to build them and get set up to sign them, etc. 

i'll create a project to track all the ancillary parts around this, but i'm thinking after some light manual testing, we're OK to merge this (since its not connected to any release flows yet). we'll get some smoke testing in place, connect this to our existing release process, etc.

projects/optic/package.json
```diff
@@ -112,5 +112,21 @@
...
     "url-join": "^4.0.1",
     "whatwg-mimetype": "^3.0.0"
+  },
+  "pkg": {
+    "scripts": [
+      "build/commands/ruleset/upload.js"
+    ],
+    "assets": [
+      "../../node_modules/open/xdg-open",
+      "../../node_modules/vm2/lib/setup-sandbox.js"
```
notnmeyer, line 122: this file doesn't get a warning during the build, but if it's excluded you get a runtime error,

```
➜ dist/optic-macos-arm64 --help
pkg/prelude/bootstrap.js:1872
      throw error;
      ^

Error: File '/**/optic/node_modules/vm2/lib/setup-sandbox.js' was not included into executable at compilation stage. Please recompile adding it as asset or script.
```

niclim (approved):

PR #1910, opened 2023-05-08
feat: attach linux sea to releases

notnmeyer:
## 🍗 Description

attaches linux optic binaries to releases. the builds happen on linux, which is fine for now, but we'll need something a little more complicated when we add mac builds to the mix.

signed and notarized macOS builds are coming, we're working on getting registered with the apple developer program.

## 📚 References
- closes https://github.com/opticdev/monorail/issues/4273

## 👹 QA


notnmeyer:
ok, linux, alpine (musl), and windows assets are being uploaded to the releases page. https://github.com/opticdev/optic/releases/tag/v0.43.1-3

niclim (approved):

PR #1924, opened 2023-05-17
Bundle fix/nested components circular & mapping support

acunniffe:
## 🍗 Description
Fixes support for nested component bundling and rewriting [mapping](https://swagger.io/docs/specification/data-models/inheritance-and-polymorphism/) -- these aren't $refs, they are pointers to $refs so you have to rewrite them. This might make sense to do in the dereference stage at some point, but for now I think this is the correct place to do it. We can move it closer to core as we get better support for discriminators  

## 📚 References

## 👹 QA

### DIFF
--- a/projects/optic/src/commands/bundle/bundle.ts
+++ b/projects/optic/src/commands/bundle/bundle.ts
@@ -542,10 +542,8 @@ function bundleMatchingRefsAsComponents<T>(
       );
     });
 
-    const copy = JSON.parse(JSON.stringify(ref.component));
-
     ref.component = jsonpatch.applyPatch(
-      copy,
+      ref.component,
       sortby(
         nestedRefUsageUpdates,
         (i) => -jsonPointerHelpers.decode(i.path).length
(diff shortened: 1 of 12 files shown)


projects/optic/src/commands/bundle/json-iterator.ts
```diff
@@ -0,0 +1,26 @@
+import { jsonPointerHelpers } from '@useoptic/json-pointer-helpers';
+
+interface IteratorResult {
+  pointer: string;
+  value: any;
+}
+
+export function* jsonIterator(
```
acunniffe, line 8: thank you chat gpt

projects/optic/src/commands/bundle/bundle.ts
```diff
@@ -599,6 +626,49 @@ function bundleMatchingRefsAsComponents<T>(
     }
   }, specCopy);
 
+  if (options.rewriteMapping) {
+    let $refs = await $RefParser.resolve(sourcemap.rootFilePath);
+
+    const refKeys = $refs.paths();
+
+    // handle mapping
+    const mappingPattern = [
```
acunniffe, line 635: 1. find mapping within schemas
2. associate the $ref with the resolved ref graph 
3. if it exists apply a patch. It should exist, but it's also valid to use a URL or other pointer type here.

niclim (approved):

PR #1931, opened 2023-05-18
chore: only build linuxstatic for linux, rename binaries

notnmeyer:
## 🍗 Description

- linux: rather than producing one statically and one dynamically linked binary, just build the former. it should run everywhere.
    ```
    ➜ docker run -it --rm -v ./dist/:/dist -t alpine /dist/optic-linux-arm64 --help &>/dev/null ; echo $status
    0

    ➜ docker run -it --rm -v ./dist/:/dist -t ubuntu /dist/optic-linux-arm64 --help &>/dev/null ; echo $status
    0
    ```
- renames binaries so that the arch and OS matches valid GOARCH/GOOS values. this might seem like an odd choice, but Go has well established and widely used defaults for these values. additionally, there's a good set of open-source shell functions that also uses those defaults for platform detection.

ultimately, both of these changes will make it simpler to create an install script that does platform detection.

## 📚 References

## 👹 QA

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -134,7 +134,7 @@ jobs:
           RELEASE_TAG: ${{ github.event.release.tag_name }}
         run: |
           task pkg:build -- \
-            -t node18-linux-arm64,node18-linux-x64,node18-alpine-arm64,node18-alpine-x64,node18-win-arm64,node18-win-x64 \
+            -t node18-linuxstatic-arm64,node18-linuxstatic-x64,node18-win-arm64,node18-win-x64 \
             --no-bytecode --public --public-packages "*"
           task pkg:archive pkg:upload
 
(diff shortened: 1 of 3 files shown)


notnmeyer:
the archives ultimately end up looking like,

```
➜ ls -l dist/archives/
total 450000
-rw-r--r--@ 1 nate  staff       554 May 18 08:04 checksums.txt
-rw-r--r--@ 1 nate  staff  41627592 May 18 08:04 optic-darwin-amd64.tar.gz
-rw-r--r--@ 1 nate  staff  39983276 May 18 08:04 optic-darwin-arm64.tar.gz
-rw-r--r--@ 1 nate  staff  38381643 May 18 08:04 optic-linux-amd64.tar.gz
-rw-r--r--@ 1 nate  staff  38008266 May 18 08:04 optic-linux-arm64.tar.gz
-rw-r--r--@ 1 nate  staff  37137038 May 18 08:04 optic-windows-amd64.exe.tar.gz
-rw-r--r--@ 1 nate  staff  31887312 May 18 08:04 optic-windows-arm64.exe.tar.gz
```

niclim (approved):

PR #1944, opened 2023-05-23
feat(ci): release macOS binaries

notnmeyer:
## 🍗 Description

while we aren't set up to sign and notarize our macOS builds yet, we're going to start attaching them to releases anyway. this pr adds a dedicated workflow for building/releasing the binaries. it runs on macOS runners, which can build all the platforms in question.

## 📚 References

## 👹 QA

hard to test this from a branch. after merging i'll ship a prerelease and fix up any issues.

### DIFF
--- a/.github/workflows/release-binaries.yml
+++ b/.github/workflows/release-binaries.yml
@@ -1,0 +1,96 @@
+name: release-binaries
+
+on:
+  workflow_dispatch: {}
+  release:
+    types:
+      - published
+
+env:
+  SKIP_ANNOUNCE: 'false'
+  SKIP_PUBLISH: 'false'
+
+jobs:
+  verify-prerelease-tag:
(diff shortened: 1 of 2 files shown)


.github/workflows/release-binaries.yml
```diff
@@ -0,0 +1,42 @@
...
+    inputs:
+      release_tag_name:
+        required: true
+        type: string
+        description: >
+          The tag that a release has been previously created from. Binaries will be attached to this release.
+
+jobs:
+  release:
+    runs-on: macos-latest-xl
```
notnmeyer, line 21: 12-core macOS runner 🚀 https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

PR #1959, opened 2023-05-26
Exit with error code 1 when no diff-all specs

niclim:
## 🍗 Description

exits with error code 1 when no specs to diff-all

## 📚 References

## 👹 QA


projects/optic/src/utils/git-utils.ts
```diff
@@ -164,7 +164,7 @@ export const findOpenApiSpecsCandidates = async (
 ): Promise<string[]> =>
   new Promise((resolve, reject) => {
     const cb = (err: unknown, stdout: string, stderr: string) => {
-      if (err || stderr || !stdout) reject(err || stderr);
```
niclim, line 167: Not sure if this was working before, but if no yml/json files are foud that have openapi in the repo, this fails early

nalanj (approved):

PR #1973, opened 2023-05-31
add a server + server ready check to capture

niclim:
## 🍗 Description

Adds a server + server ready check command to optic capture - useful when you want this to run in CI and run this code against a server that pulls traffic

Usage:
`optic capture oas.yml http://localhost:3030 --command "npm run test" --server "npm run start --server-ready-check "curl http://localhost:3030/healthcheck`

TODO
- [ ] Update documentation
- [ ] Update website tutorials

## 📚 References

## 👹 QA

### DIFF
--- a/projects/optic/src/commands/oas/capture.ts
+++ b/projects/optic/src/commands/oas/capture.ts
@@ -29,6 +29,9 @@ import { OpticCliConfig } from '../../config';
 import { clearCommand } from './capture-clear';
 import { createNewSpecFile } from '../../utils/specs';
 import { logger } from '../../logger';
+import url from 'url';
+import ora from 'ora';
+import { UserError } from '@useoptic/openapi-utilities';
 
 export async function captureCommand(config: OpticCliConfig): Promise<Command> {
   const command = new Command('capture');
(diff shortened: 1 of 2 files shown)


projects/optic/src/commands/oas/capture.ts
```diff
@@ -203,62 +245,65 @@ export async function captureCommand(config: OpticCliConfig): Promise<Command> {
...
-                    }
-                  );
-                } catch (e) {
-                  console.log(e);
-                }
+      await completing;
+      if (serverRunner) serverRunner.kill();
+      if (!interactiveCapture) {
+        await complete();
+        process.exit(exitCode ?? 0);
```
niclim, line 294: This was kind of weird and I'm not entirely sure what's happening here - we're not properly tearing down something in our spawn process (`--command` and `--server` are both affected)

Using the hammer approach to just exit in non-interactive mode, but we should look into what's causing this

projects/optic/src/commands/oas/capture.ts
```diff
@@ -203,62 +245,65 @@ export async function captureCommand(config: OpticCliConfig): Promise<Command> {
...
-                    feedback,
-                    {
-                      printCoverage: false,
-                    }
-                  );
-                } catch (e) {
-                  console.log(e);
-                }
+      await completing;
+      if (serverRunner) serverRunner.kill();
```
nalanj, line 291: You probably want to cleanly shut down the server here. Usually a process manager is going to have some sort of shutdown timeout and send a kill only after that expires. We could probably mimic docker (although it's technically not quite correct) and send a sigterm, wait up to 10 seconds, and then send a sigkill.

nalanj (approved):
This seems reasonable considering how this command is already written, but this code overall has a ton of problems with not handling signals to catch when processes under it die and clean everything up correctly.

PR #1975, opened 2023-06-01
add fastify-capture project

Alexandre-Herve:
## 🍗 Description

Add a `fastify-capture` project meant to be published as a standalone NPM package.

## 📚 References

## 👹 QA


projects/fastify-capture/package.json
```diff
@@ -0,0 +1,34 @@
...
+  "types": "build/index.d.ts",
+  "scripts": {
+    "build": "yarn tsc --build --verbose && yarn babel ./src --out-dir ./build --extensions \".ts\",\".tsx\"",
+    "clean": "rm -rf build",
+    "test": "ENVIRONMENT=test jest --colors --passWithNoTests",
+    "local:run": "yarn ts-node src/index.ts"
+  },
+  "dependencies": {
+    "cookie": "^0.5.0",
+    "fastify": "^4.17.0",
```
niclim, line 14: should this be a dev dep since it's only used for typing? Or should it be a peer dependency?

projects/fastify-capture/src/index.ts
```diff
@@ -0,0 +1,195 @@
...
+import http from 'http';
+
+type FastifyCaptureOptions = {
+  harOutputDir?: string;
+};
+
+export const fastifyCapture = (options: FastifyCaptureOptions = {}) => {
+  const entries: Entry[] = [];
+
+  const harOutputDir = options.harOutputDir
```
niclim, line 16: Should we default to the optic capture tmp dir? I think we could call this something like "optic-har-capture"+project-path for the capture spec hash

niclim (approved):

projects/fastify-capture/package.json
Alexandre-Herve: Yeah looks like it's correctly stripped out of the build, let's put it as dev dependency

projects/fastify-capture/src/index.ts
Alexandre-Herve: [This](https://github.com/opticdev/optic/blob/12827ccacf54e497b0bf5512a5be4ab5ed562fc3/projects/optic/src/commands/oas/captures/capture-storage.ts#L4)? 

Looks a bit fancy for the job I don't know. With a local folder people can find the output easily I think.

projects/fastify-capture/package.json
Alexandre-Herve: Yeah looks like it's correctly stripped out of the build, let's do this

PR #1990, opened 2023-06-08
allow additional properties for openapi 3_1

niclim:
## 🍗 Description

additional properties is false on 3_1 but true on 3 - this breaks cases like `webhooks`

## 📚 References

## 👹 QA


projects/openapi-io/src/validation/validation-schemas.ts
```diff
@@ -161,7 +160,7 @@ const openapi3_1_schema_object = {
   'x-custom-validator': 'validateSchema',
   description:
     'The Schema Object allows the definition of input and output data types. These types can be objects, but also primitives and arrays. This object is an extended subset of the JSON Schema Specification Wright Draft 00.  For more information about the properties, see JSON Schema Core and JSON Schema Validation. Unless stated otherwise, the property definitions follow the JSON Schema.',
-  additionalProperties: false,
+  additionalProperties: true,
```
acunniffe, line 163: this works, we don't do anything with webhooks yet re: validating them (lint) or diffing them (diff). We should just note that somewhere.

acunniffe (approved):

projects/openapi-io/src/validation/validation-schemas.ts
niclim: Agreed, we're not treating webhooks as first class yet - when we do (i.e. render + handle webhooks to run rules, etc) we should update the schema validation here

PR #2002, opened 2023-06-13
Nested unused $ref removal 

acunniffe:
## 🍗 Description

## 📚 References

## 👹 QA

### DIFF
--- a/projects/optic/src/commands/bundle/bundle.ts
+++ b/projects/optic/src/commands/bundle/bundle.ts
@@ -146,20 +146,7 @@ const bundleAction =
         });
       }
 
-      let noUnused = false;
-      while (!noUnused) {
-        //write tmp file
-        const tmpOutput = path.join(
-          path.dirname(filePath),
-          `tmp-openapi-${Date.now()}.json`
-        );
-        await fs.writeFile(tmpOutput, JSON.stringify(updatedSpec, null, 2));
-        const spec = await getSpec(tmpOutput, config);
-        const result = removeUnusedComponents(updatedSpec, spec.sourcemap);
-        updatedSpec = result.spec;
-        noUnused = result.removals === 0;
-        await fs.unlink(tmpOutput);
-      }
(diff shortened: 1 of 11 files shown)


projects/optic/src/commands/bundle/bundle.ts
```diff
@@ -146,15 +146,20 @@ const bundleAction =
...
+        const tmpOutput = path.join(
+          path.dirname(filePath),
+          `tmp-openapi-${Date.now()}.json`
+        );
+        await fs.writeFile(tmpOutput, JSON.stringify(updatedSpec, null, 2));
+        const spec = await getSpec(tmpOutput, config);
+        const result = removeUnusedComponents(updatedSpec, spec.sourcemap);
+        updatedSpec = result.spec;
+        noUnused = result.removals === 0;
+        await fs.unlink(tmpOutput);
```
niclim, line 161: - can we instead mutate the spec in place and detect usage once we've bundled everything?
- can we avoid looping and redoing this multiple times?

I know its probably the fastest way to implement this, but it opens us to a whole host of other issues which are completely avoidable

niclim (changes requested):

projects/optic/src/commands/bundle/bundle.ts
acunniffe: We have to hit the file system at least once. That's the only way our sourcemaps can track refs. They need that abstraction to know which refs point at the same files/pointers. 

We can probably make it only hit the file system once though and do some fancy logic in memory to mutate the sourcemap as we remove stuff. Or change the sourcemap parser to be able to parse bundled specs (without any dependencies)

niclim (approved):
we can do this in 1 pass by:
- collect refs by traversing all keys not in components
- when we hit any ref, we traverse the component (e.g. if we hit `#components/schemas/Something` we then recursively traverse those keys and add it to a set)

i suppose this is fine for now though

PR #2010, opened 2023-06-15
Update diff + diff-all output

niclim:
## 🍗 Description

Updates the diff + diff-all output such that:
- diff + diff-all have headers for each spec they're diffing at the top
  - there's preview documentation here as well as operation change + rule summaries
  - the name of the spec and the path
- Rules are nested to the relevant change (e.g. query param rules will go under the relevant operation -> query param)
- Reworked how rules are rendered + different colors for different warn levels

Used https://www.notion.so/useoptic/CI-Output-2b43865e15bb4fcfb9bbcb5558c06f25 to capture output - but there's some differences here:
- didn't use the ruleset name because we don't ahve that in the rule; we can update to use it but right now i'm using the rule name instead

Some notes:
- Because of how we're rendering fields, we render the grouping of added, changed and removed fields, and then all the rules under that. In the future we can render these out in individual lines and nest rules under them. Right now it's just all grouped under body
- This is a pretty big change that kind of needed to go together (lots of impacted parts) - released a pre-release `0.45.5-1` which we should test before releasing
- In a future PR i'm going to fix the rule pointers to use git links when in CI

Probably best to install `0.45.5-1` and try it out from that and see any rough edges here

## 📚 References

## 👹 QA


projects/openapi-utilities/src/openapi3/group-diff.ts
```diff
@@ -231,44 +293,99 @@ function normalizeRequiredDiff(
 
 export function groupDiffsByEndpoint(
```
niclim, line 294: had to update this to group rules together with the same node - this impacts:
- the rule result renderer (still used by `lint`)
- the json changelog (i.e. needed to update the json changelog to check the diff count)
- the terminal changelog (which was completely rewritten)

acunniffe:
Can we try to pad the left side of all the errors and break at the right point so they never wrap?

Alexandre-Herve (approved):

acunniffe:
Screenshot 2023-06-16 at 6 45 08 AM

^ this looks off as well? What's the thinking behind the commas?

niclim:
> Can we try to pad the left side of all the errors and break at the right point so they never wrap?

Could you explain why this is important? I don't think it's going to be that trivial to implement and I'm not sure I understand that avoiding text wrapping is going to be much different that word breaks.

> ^ this looks off as well? What's the thinking behind the commas?

Could you elaborate? This is how it was implemented before to be more information dense. I can split this out into individual rows but it'll take me some more time to do this (and group the rules)

acunniffe:
> Could you explain why this is important?
It just looks disorganized. I know the terminal breaking is hard. What if we did separate lines for the trail and the file? Two arbitrarily long things back to back is hard. Also I can't remember the terminal linking convention but maybe we can do the file relative to the working directory to make it shorter. 

> Could you elaborate? This is how it was implemented before to be more information dense.

So how does this look when any of these propeties in the comma separated list are `changed`? Do they get their own rule then? Is it all grouped underneath the in the same area? I'd default to giving each node it's own row. That's just how people scan things. 

I really like what you did with the colors. 

`[what it is] (white) [change status] (color)` 
     `error|warning|info (with color)` 

I think this format encourages quick scanning and makes you focus on the things that really matter (failed a rule)

niclim:
Let me see what I can do to make this output more succinct + i'll group fields together

I'll work on:
-  making the json path + terminal linking more succinct
- group field + rule changes and give them their own nodes

niclim:
Ok updated to look something like this 

Main changes:
- Uses a relative path from the cwd for sourcemap links
- Does not show json path of rule failure - they should be nested to where they affect so there's no need to render the json path anymore
- renders each property as a different line and nests rules under them

```
x Swagger Petstore Updated /Users/nicholaslim/Desktop/repos/optic/projects/optic/src/__tests__/integration/workspaces/diff/petstore/petstore-base.json
Operations: 5 operations added, 2 changed, 1 removed
x  Checks: 61/96 passed 

specification details:
- /servers/2 added
- /info/title changed

✔ GET /pet/findByStatus: added
  

✔ GET /pet/findByTags: added
  

✔ GET /pet/{petId}: added
  

✔ POST /pet/{petId}: added
  

✔ DELETE /pet/{petId}: added
  

x PUT /user/{username}: 
  - request: 
    - body */*: 
      - property /schema/properties/phone: removed
      - property /schema/properties/bio: added
      - property /schema/properties/userStatus: changed
      
        x [prevent request property type changes] expected request body property 'userStatus' to not change type. This is a breaking change.
        at src/__tests__/integration/workspaces/diff/petstore/petstore-updated.json:1347:48590
        
      - property /schema/properties/lastName: changed
      - property /schema/properties/username: changed
      
        x [prevent changing request property to required] cannot make a request property required. This is a breaking change.
        at src/__tests__/integration/workspaces/diff/petstore/petstore-updated.json:1342:48332
        

x GET /user/{username}: 
  - response 200: 
    - body application/json: 
      - property /schema/properties/phone: removed
      
        x [prevent removing response property] cannot remove response property 'phone'. This is a breaking change.
        at src/__tests__/integration/workspaces/diff/petstore/petstore-base.json:848:29942
        
      - property /schema/properties/bio: added
      - property /schema/properties/userStatus: changed
      
        x [prevent response property type changes] expected response body property 'userStatus' to not change type. This is a breaking change.
        at src/__tests__/integration/workspaces/diff/petstore/petstore-updated.json:1289:46579
        
      - property /schema/properties/lastName: changed
      
        x [prevent making response property optional] cannot make required response property 'lastName' optional. This is a breaking change.
        at src/__tests__/integration/workspaces/diff/petstore/petstore-updated.json:1286:46420
        
      - property /schema/properties/username: changed
    - body application/xml: 
      - property /schema/properties/phone: removed
      
        x [prevent removing response property] cannot remove response property 'phone'. This is a breaking change.
        at src/__tests__/integration/workspaces/diff/petstore/petstore-base.json:827:29057
        
      - property /schema/properties/bio: added
      - property /schema/properties/userStatus: changed
      
        x [prevent response property type changes] expected response body property 'userStatus' to not change type. This is a breaking change.
        at src/__tests__/integration/workspaces/diff/petstore/petstore-updated.json:1268:45660
        
      - property /schema/properties/lastName: changed
      
        x [prevent making response property optional] cannot make required response property 'lastName' optional. This is a breaking change.
        at src/__tests__/integration/workspaces/diff/petstore/petstore-updated.json:1265:45501
        
      - property /schema/properties/username: changed
  - response 404: 
    - body application/json: added
      - property : added

```

Screenshot 2023-06-16 at 2 23 46 PM

acunniffe:
Looking good

acunniffe (approved):

PR #2028, opened 2023-06-23
History follow renames

Alexandre-Herve:
## 🍗 Description

Follow file renames when logging history. Later we can apply this to API add as well.

## 📚 References

## 👹 QA


niclim (approved):
nice - it's something we don't handle in `api add` as well if we want to fix that there as well later

PR #2034, opened 2023-06-26
track install method

niclim:
## 🍗 Description

## 📚 References

## 👹 QA


notnmeyer (approved):
we'll need to sort out how to include the container image as an installation method as well (but not in this pr). it'll be a little tricky since the image installs the binary.

i think to detect the optic docker container we look for `INSTALLATION_METHOD == "binary" && /etc/machine-id == "optic-docker"`

Issue #2039, opened 2023-06-27
Cannot view diff report


davidkeaveny:
When I run the `optic diff` against my OpenAPI specification, the operation completes successfully and gives me a brief summary; if I then run it again with the `--web` setting so I get the full report, I receive an `ENAMETOOLONG` error. I'm assuming that the error here is occurring when it tries to open a browser to display the report?

I am running the following against Optic 0.46.0 on Windows 11 (22H2) and Node 18.16.0:

```
optic diff open-api.yaml --base main --web
```

which ends with the following:

```
GET /employees/{employeeId}/roster-shifts:
  - query parameter PageSize: removed
  - query parameter PageNumber: removed
  - query parameter pageNumber: added

  - query parameter pageSize: added

  - response 200:
    - body application/octet-stream: removed
      - property : removed
    - body application/json: added
      - property : added

...and 94 other endpoints

See the full history of this API by running "optic api add open-api.yaml --history-depth 0"
Opening up diff in web view
Error: spawn ENAMETOOLONG
    at ChildProcess.spawn (node:internal/child_process:413:11)
    at Object.spawn (node:child_process:757:9)
    at baseOpen (C:\Users\david.keaveny\AppData\Roaming\npm\node_modules\@useoptic\optic\node_modules\open\index.js:224:34) {
  errno: -4064,
  code: 'ENAMETOOLONG',
  syscall: 'spawn'
}
spawn ENAMETOOLONG
```

Is the web report written somewhere in the temporary file system, so that I could have a look at it? Or is there any chance of having an explicit write-to-file option in the diff?

niclim:
Hi @davidkeaveny - I just released a new version `0.47.4` that tries works around this issue on windows - could you try it out and let me know if that works for you?

davidkeaveny:
Hi @niclim - many thanks for the update, I've upgraded to 0.47.5, and now when I run the same command as before, it doesn't give the error as above, but writes the diff report to the screen. That's a big improvement for me, so many thanks again!

PR #2041, opened 2023-06-27
add fallback for open with large url

niclim:
## 🍗 Description

Windows terminals have trouble opening the web url (long url encoded via a hash) - trying to work around this by writing a html file that redirects to the original url and writing that to a tmp dir.

## 📚 References

https://github.com/opticdev/optic/issues/2039

## 👹 QA

`0.47.4-0` prerelease version can be used for testing

### DIFF
--- a/.github/workflows/release-container-image.yml
+++ b/.github/workflows/release-container-image.yml
@@ -50,7 +50,7 @@ jobs:
 
       - name: Login to Amazon ECR Public
         id: login-ecr-public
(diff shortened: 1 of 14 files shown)


acunniffe:
Got my VM working again. Node and optic are installing. will report back first thing tomorrow.

acunniffe:
Hmm so it didn't work the first time I ran it -- hung terminal, but then eventually it did. Not sure if there's a permissions or file write thing. Couldn't reproduce. But yeah -- works now! nice


Screenshot 2023-06-27 at 9 41 53 PM

acunniffe (approved):

PR #2042, opened 2023-06-27
More error silencing

niclim:
## 🍗 Description

Silences errors in sentry for:
- invalid url for optic capture
- resolver errors

## 📚 References

## 👹 QA

### DIFF
--- a/projects/optic/src/commands/oas/captures/streams/sources/proxy.ts
+++ b/projects/optic/src/commands/oas/captures/streams/sources/proxy.ts
@@ -23,6 +23,8 @@ import portfinder from 'portfinder';
 import globalLog from 'log';
 import invariant from 'ts-invariant';
 import chalk from 'chalk';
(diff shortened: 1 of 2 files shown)


projects/optic/src/commands/oas/captures/streams/sources/proxy.ts
```diff
@@ -41,7 +43,19 @@ export class ProxyInteractions {
       proxyPort?: number;
     }
   ): Promise<[ProxyInteractions, string, string]> {
-    let { host, protocol, origin } = new URL(targetHost);
+    let host: string;
+    let protocol: string;
+    let origin: string;
+    try {
+      ({ host, protocol, origin } = new URL(targetHost));
+    } catch (e) {
```
acunniffe, line 51: should we make it an error to have a path? 

Since it's confusing to users and doesn't actually do anything?

acunniffe (approved):

PR #2043, opened 2023-06-27
Fix `optic update` interactions + add tests

niclim:
## 🍗 Description

Writes integration tests for `optic update` - Some bugs from the snapshots I need to investigate
- [x] Request bodies aren't documented from the snapshots (they're set in the har)
- [x] When passing a server prefix, we're getting unexpected behaviors (all documented paths but with empty bodies)

## 📚 References

## 👹 QA

### DIFF
--- a/projects/optic/src/commands/oas/captures/getInteractions.ts
+++ b/projects/optic/src/commands/oas/captures/getInteractions.ts
@@ -8,12 +8,61 @@ import { HarEntries } from './streams/sources/har';
 import { trackWarning } from '../lib/sentry';
 import * as AT from '../lib/async-tools';
 import { PostmanCollectionEntries } from './streams/sources/postman';
+import { OpenAPIV3 } from '@useoptic/openapi-utilities';
+
+async function* handleServerPathPrefix(
+  interactions: CapturedInteractions,
+  spec: OpenAPIV3.Document
+): CapturedInteractions {
+  const hostBaseMap: { [key: string]: string } = {};
+
+  spec.servers?.forEach((server) => {
+    try {
+      // add absolute in case url is relative (valid in OpenAPI, ignored when absolute)
+      const parsed = new URL(server.url);
+
+      const pathName = parsed.pathname;
+      // remove trailing slash
+      if (pathName.endsWith('/') && pathName.length > 1) {
+        hostBaseMap[parsed.host] = pathName.substring(0, pathName.length - 1);
+      } else {
+        hostBaseMap[parsed.host] = pathName;
+      }
+    } catch (e) {}
+  });
+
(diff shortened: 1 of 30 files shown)


projects/optic/src/commands/oas/shapes/schema.ts
```diff
@@ -114,12 +114,7 @@ export class Schema {
   }
 
   static isPolymorphic(schema: SchemaObject) {
```
niclim, line 116: uh this was wrong so this block never evaluated true https://github.com/opticdev/optic/blob/main/projects/optic/src/commands/oas/shapes/streams/patches.ts#L26 when an empty object like `schema: {}` was passed in

acunniffe (approved):

projects/optic/src/commands/oas/captures/getInteractions.ts
```diff
@@ -8,12 +8,61 @@ import { HarEntries } from './streams/sources/har';
 import { trackWarning } from '../lib/sentry';
 import * as AT from '../lib/async-tools';
 import { PostmanCollectionEntries } from './streams/sources/postman';
+import { OpenAPIV3 } from '@useoptic/openapi-utilities';
+
+async function* handleServerPathPrefix(
```
niclim, line 13: Moved this function up to the top level interactions handler - we now adjust the path based on the OpenAPI server key and do not emit interactions when not relevant. This was in multiple places in the code base and was inconsistently implemented (implemented in some places and not other)

projects/optic/src/commands/oas/operations/infer-path-structure.ts
```diff
@@ -395,7 +395,7 @@ function reducePathPattern(component: PathComponentCandidate): string {
 
 function fragmentize(path: string): string[] {
   if (path.length === 0 || !path.startsWith('/')) {
-    throw new Error(`Malformed path '${path}'`);
```
niclim, line 398: I'm not sure why this was throwing an error - a base path is a valid path, and should be treated as an empty array (similarly with other parts of the code)

acunniffe: Yes this looks correct. I'm not 100% sure how it'd get there, but I don't think this code handles the case where you get something like "api/v4" -- where "api" doesn't have the required leading "/".

niclim: I see - I updated this to have a check like:
```
if (!path.startsWith('/')) {
  path = `/${path};`
}
```

this will return `[]` for `''`, `'/'` and also correctly handle `api/v4` -> fragmentizing it correctly (which i'm not sure how it'd be set that way since all of this is coming from hars)

PR #2048, opened 2023-06-28
add finish setting up cloud to diff + diff-all output

acunniffe:
## 🍗 Description
Screenshot 2023-06-28 at 2 07 31 PM

## 📚 References

## 👹 QA


projects/optic/src/commands/diff/diff.ts
```diff
@@ -457,41 +459,42 @@ const getDiffAction =
...
-        };
-
-        if (!maybeChangelogUrl) {
-          const meta = {
-            createdAt: new Date(),
-            command: ['optic', ...process.argv.slice(2)].join(' '),
-            file1,
-            file2,
-            base: options.base,
+    if (config.isInCi && !options.upload)
```
niclim, line 462: This will break the `--web` option - this will only allow `.web` to trigger in ci?

niclim (changes requested):

niclim (approved):

PR #2064, opened 2023-07-05
feat: capture v2 initial work

notnmeyer:
## 🍗 Description

adds initial support for a streamlined capture flow, via a `capture` block in `optic.yml`. the configuration specifies the details of an API: how to start it, which endpoints to make requests to, etc, and then captures the requests using the reverse-proxy. currently, you you would then run `optic update...` to update your openapi spec.

the config looks like,

```
capture:
  openapi.yml:
    server:
      command: go run main.go
      url: http://localhost:8080
      ready_endpoint: /
      ready_interval: 5000
    requests:
      - path: /
      - path: /users
      - path: /users/create
        verb: POST
        data:
          name: Hank
```

the new capture flow is used when both are true:

- there is a `capture` block in the optic.yml
- a target url is not specified at the cli (so just `optic capture openapi.yml`)

not intended for real use yet, additional features, smoothing, and polishing are coming shortly.

## 📚 References
- https://github.com/orgs/opticdev/projects/71

## 👹 QA


niclim:
@notnmeyer i pushed in a top level capture command and moved your code under it

For now, I just put the new command behind `optic beta capture` - it still will call the `optic capture` v1 flow, but lets us move stuff around easier to have this separate command since i want to remove / change the wording / deprecate some of the old commands + help text

niclim (approved):

PR #2069, opened 2023-07-06
feat(cli): throttle capture requests

notnmeyer:
## 🍗 Description

adds configurable concurrency for the new capture flow requests. via a `capture[].config` block in optic.yml.

```
capture:
  openapi.yml:
    config:
      request_concurrency: 100
    server: ...
    requests: [...]
```

## 📚 References
- closes https://github.com/opticdev/monorail/issues/4593

## 👹 QA


projects/optic/src/config.ts
```diff
@@ -35,6 +35,11 @@ const DefaultOpticCliConfig: OpticCliConfig = {
 };
 
 const CaptureConfigData = Type.Object({
+  config: Type.Optional(
+    Type.Object({
+      request_concurrency: Type.Number(),
```
niclim, line 40: Think request_concurrency should be optional as well (if we add more config) - 
```suggestion
      request_concurrency: Type.Optional(Type.Number()),
```

projects/optic/src/commands/capture/capture.ts
```diff
@@ -182,7 +183,12 @@ const getCaptureAction =
 
     // make requests
     console.log(trafficDirectory);
-    const requests = makeRequests(captureConfig.requests, proxyUrl);
+    const concurrency = captureConfig.config?.request_concurrency || 25;
```
niclim, line 186: Should the default here be something lower like 5 or something? Requests should resolve fairly quickly so higher concurrency might not mean faster due to contention

niclim (approved):

PR #2071, opened 2023-07-06
feat(capture): skip running local server

notnmeyer:
## 🍗 Description

allows skipping the local server (executed via `capture[].server.command`) the following ways,
 
- [x] adds an option to the capture command (`--server-override <url>`) that skips executing `capture[].server.command` and sends proxy traffic to the override URL instead. useful for testing against a remote server or different environment. 
    - `optic beta capture -r openapi.yml --server-override https://example.com`

- [x] `capture[].server.command` is now optional. omitting it assumes the server is started externally (or already running). proxy traffic is forwarded to `capture[].server.url` or the value of `--serverOverride`, if set. a minimal capture configuration now looks like,
    
    ```
    capture:
      openapi.yml:
        server:
          url: https://example.com
        requests: [...]
    ```

## 📚 References
- closes https://github.com/opticdev/monorail/issues/4581

## 👹 QA

demo of capture running when `server.command` is not specified. it will run the server ready same as if it were starting the server and then send requests through, 
demo


projects/optic/src/commands/capture/capture.ts
```diff
@@ -127,58 +134,56 @@ const getCaptureAction =
...
+        console.log(data.toString());
+      });
+    }
 
     // wait until server is ready
-    await new Promise((r) =>
-      setTimeout(r, captureConfig.server.ready_interval)
-    );
-
-    if (captureConfig.server.ready_endpoint) {
```
notnmeyer, line 155: i struggle to think that theres a situation where we don't want to wait until the server is ready

projects/optic/src/commands/capture/capture.ts
```diff
@@ -127,58 +134,56 @@ const getCaptureAction =
...
+    const readyEndpoint = captureConfig.server.ready_endpoint || '/';
+    const readyInterval = captureConfig.server.ready_interval || 1000;
+    const readyUrl = urljoin(serverUrl, readyEndpoint);
 
-      const timeout = 10 * 60 * 1_000; // 10 minutes
-      const now = Date.now();
-      const spinner = ora('Waiting for server to come online...');
-      spinner.start();
-      spinner.color = 'blue';
+    const timeout = 10 * 60 * 1_000; // 10 minutes
```
notnmeyer, line 166: this would be a good thing to move to `capture.config`

niclim (approved):

projects/optic/src/commands/capture/capture.ts
notnmeyer: the point about the root endpoint not being available is valid.

as far as not needing to check for an external server—consider my demo. in that example our "external" server is not ready (because i hadnt started it before starting capture). the ready check gets us the pleasant behavior of not needing to have the proxy and the server be ready in a strict order.

that said, i can certainly imagine situations where failing faster is preferable. so yeah, i suppose you're right here. we'll only wait if a ready_endpoint is supplied.

notnmeyer: so here's the thing: we _have_ to have some sort of wait to handle the local server case. using my go test server, which will start about as fast as anything we can expect, capture sends the requests before the server is ready. we weren't seeing that before because we had that one superfluous seeming timeout right before dropping into the loop.  so i think we need to either:

1. keep the arbitrary sleep of one internal before we start server checking
2. or, require ready_endpoint and always ready check

notnmeyer: going with #1

projects/optic/src/commands/capture/capture.ts
```diff
@@ -197,13 +202,20 @@ const getCaptureAction =
 
     Promise.all(requests)
       .then(() => {
-        process.kill(-app.pid!);
-        sourcesController.abort();
+        // stop the app server
+        if (!options.serverOverride && captureConfig.server.command) {
+          process.kill(-app.pid!);
+        }
```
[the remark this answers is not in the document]
notnmeyer: fixed this

PR #2083, opened 2023-07-10
feat(capture): configurable timeout

notnmeyer:
## 🍗 Description

- makes the server ready timeout configurable
- lowers the default timeout to 3 minutes, which still seems very generous to me

## 📚 References

## 👹 QA


projects/optic/src/config.ts
```diff
@@ -46,6 +46,7 @@ const CaptureConfigData = Type.Object({
     url: Type.String(),
     ready_endpoint: Type.Optional(Type.String()),
     ready_interval: Type.Optional(Type.Number()),
+    ready_timeout: Type.Optional(Type.Number()),
```
niclim, line 49: Not sure where but we should document that this is in milliseconds

niclim (approved):

PR #2085, opened 2023-07-10
feat(capture): run custom commands to generate capture traffic

notnmeyer:
## 🍗 Description

- adds a `capture.requests_command.(command|proxy_variable)` fields to optic.yaml allowing arbitrary commands to be run and send traffic to the proxy. can be used in conjunction with `capture.requests`.

- `capture.requests_command.proxy_variable` is the name of the environment variable that will be injected into process.env when running `capture.requests_command.command`. the value will be the URL for the running reverse proxy. the default values are `OPTIC_PROXY`, `http://localhost:8000`. 

an example using [hurl](https://hurl.dev),

```
➜ cat optic.yml
capture:
  openapi.yml:
    server:
      command: go run main.go
      url: http://localhost:8080
    requests_command:
      command: hurl hurl/*.hurl
      # variables must be prefixed with HURL_ to be available in hurl
      proxy_variable: HURL_OPTIC_PROXY

➜ cat hurl/requests.hurl
GET {{OPTIC_PROXY}}/
GET {{OPTIC_PROXY}}/users
POST {{OPTIC_PROXY}}/users/create
{
    "name": "Hank"
}

➜ optic beta capture -r openapi.yml                                                                               ⠋✔ Server check passed
3 requests captured. Run `optic update --all` to document updates.
```

## 📚 References
- closes https://github.com/opticdev/monorail/issues/4628

## 👹 QA

### DIFF
--- a/projects/optic/src/commands/capture/capture.ts
+++ b/projects/optic/src/commands/capture/capture.ts
@@ -156,7 +156,7 @@ const getCaptureAction =
     });
 
     // start the app
-    let app: ChildProcessWithoutNullStreams;
+    let app: ChildProcessWithoutNullStreams | undefined = undefined;
     if (!options.serverOverride && captureConfig.server.command) {
       const cmd = captureConfig.server.command.split(' ')[0];
       const args = captureConfig.server.command.split(' ').slice(1);
@@ -164,7 +164,7 @@ const getCaptureAction =
 
       // log error output from the app
       app.stderr.on('data', (data) => {
-        console.log(data.toString());
+        logger.error(data.toString());
       });
     }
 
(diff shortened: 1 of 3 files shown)


notnmeyer:
side note: the yaml schema is already feeling a little bolted-on as we're figuring this out. created https://github.com/orgs/opticdev/projects/71/views/1?pane=issue&itemId=33025164 to discuss tightening it up.

projects/optic/src/commands/capture/capture.ts
```diff
@@ -197,43 +202,71 @@ const getCaptureAction =
...
+          });
+        });
+        reqCmd.stderr.on('data', (data) => {
+          logger.error(data.toString());
+        });
+      }
+    } catch (error) {
+      logger.error(error);
+    } finally {
+      await Promise.all([requestsPromise, reqCmdPromise]);
```
niclim, line 248: I think this should go right above the catch block (i.e. we should wait for it as part of the `try`. `await`ing a Promise will throw an error, which means we want any potential error to be thrown in the `try` part

projects/optic/src/commands/capture/capture.ts
```diff
@@ -197,43 +202,71 @@ const getCaptureAction =
...
+              reject();
+            }
+          });
+        });
+        reqCmd.stderr.on('data', (data) => {
+          logger.error(data.toString());
+        });
+      }
+    } catch (error) {
+      logger.error(error);
```
niclim, line 246: I'm not too sure on this, but maybe we should throw the error here so that we bail out of the rest of the command.

Alternatively, we can use `await Promise.allSettled` which will wait for all of our requests to finish and we can disregard any requests that failed. What do you think?

projects/optic/src/commands/capture/capture.ts
```diff
@@ -197,43 +202,71 @@ const getCaptureAction =
...
+    let requestsPromise: Promise<any> = Promise.resolve();
+    let reqCmdPromise: Promise<void> = Promise.resolve();
+    try {
+      if (captureConfig.requests) {
+        const requests = makeRequests(
+          captureConfig.requests,
+          proxyUrl,
+          captureConfig.config?.request_concurrency || 5
+        );
+        requestsPromise = Promise.all(requests);
```
niclim, line 219: I think we actually want to use `Promise.allSettled` here - Promise.all will exit early if any request fails, where we may want to wait until all requests finish before deciding what to do with the errors

projects/optic/src/commands/capture/capture.ts
```diff
@@ -197,43 +202,71 @@ const getCaptureAction =
...
+    let count = 0;
     for await (const har of harEntries) {
+      count++;
       captures.addHar(har);
+      logger.debug(`Captured ${har.request.url}`);
     }
     await captures.writeHarFiles();
 
+    logger.info(
+      `${count} requests captured. Run \`optic update --all\` to document updates.`
```
niclim, line 265:
```suggestion
      `${count} requests captured`
```

We're removing the need to run `optic update` after this. I'm also not sure that rendering the request count captured is super important but we can figure that out later. We might want to instead show the endpoints we've gotten requests on (and # of unmatched requests)

projects/optic/src/commands/capture/capture.ts
niclim: Thinking about it, we might want to:
- Collect the errors that are thrown from here
- log them out as a warning (e.g. `some requests failed, etc`) - but we should still be able to continue

Failed requests can be inconsistent (and there might be cases you want to fail on them) but for the most part I think we can continue

projects/optic/src/commands/capture/capture.ts
notnmeyer: > I'm also not sure that rendering the request count captured is super important but we can figure that out later

yeah, im cool to remove it later. it's mostly useful for me now to make sure something happens. otherwise i `LOG_LEVEL=debug` and its a bit verbose

projects/optic/src/commands/capture/capture.ts
notnmeyer: yeah, i do think its correct to continue. since the command can be anything, like a script that runs multiple commands, i think its possible that we may be able to salvage some of the requests. or at least, that was my thinking at the time.

notnmeyer: @niclim were you thinking something like this, https://github.com/opticdev/optic/pull/2085/commits/4e54cb0d47a3d84a1fc98228818ec57e21bbc9b3?

niclim: Yep that looks good

niclim (approved):

PR #2086, opened 2023-07-10
Store capture hars by endpoints

niclim:
## 🍗 Description

This PR updates captures to store them by endpoint id rather than all in a single har - this grouping will make it easier for us to update hars endpoint by endpoint

## 📚 References

## 👹 QA


projects/optic/src/commands/capture/capture.ts
```diff
@@ -224,11 +223,13 @@ const getCaptureAction =
...
       });
 
-    for await (const observation of observations) {
-      if (observation.kind === CaptureObservationKind.InteractionCaptured) {
-        console.log(`Captured ${observation.path}`);
-      }
+    for await (const har of harEntries) {
+      captures.addHar(har);
     }
+    await captures.writeHarFiles();
```
niclim, line 229: This isn't technically necessary right now, but will let us reused cached requests + may be helpful for us to debug + get users to send us debug information

notnmeyer (approved):

projects/optic/src/commands/capture/storage.ts
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    const unmatched = path.join(trafficDir, `unmatched.har`);
+    this.paths.set(UNMATCHED_PATH, {
+      path: unmatched,
+      hars: [],
+    });
+  }
+
+  addHar(har: HttpArchive.Entry) {
```
niclim, line 70: Collects this to be synchronous - likely we don't need streams here as it's likely we're going to get requests in the range of 100-1000 (maybe)

PR #2097, opened 2023-07-12
chore(capture): mild refactor

notnmeyer:
## 🍗 Description

moves bits of logic from getCaptureAction() into smaller function. the intention was to avoid changing the logic much and just split it out into smaller units. whenever we deprecate and remove the original capture bits, there's an opportunity to move some of these changes into the lower-level abstractions. very allergic to breaking or otherwise trying to modifying the v1 capture flow.

## 📚 References

## 👹 QA

### DIFF
--- a/projects/optic/src/commands/capture/capture.ts
+++ b/projects/optic/src/commands/capture/capture.ts
@@ -247,8 +247,18 @@ const getCaptureAction =
       if (app) {
         process.kill(-app.pid!);
       }
+
+      if (errors.length > 0) {
+        logger.error('finished with errors:');
+        errors.forEach((error, index) => {
+          logger.error(`${index}:\n${error}`);
+        });
+      }
     }
 
+    //
+    // process proxy interactions into hars
+    //
     const harEntries = HarEntries.fromProxyInteractions(proxy.interactions);
     const captures = new GroupedCaptures(
       trafficDirectory,
@@ -262,13 +272,6 @@ const getCaptureAction =
         `Captured ${har.request.method.toUpperCase()} ${har.request.url}`
       );
     }
-    if (errors.length > 0) {
-      logger.error('finished with errors:');
-      errors.forEach((error, index) => {
-        logger.error(`${index}:\n${error}`);
-      });
-    }
-
     await captures.writeHarFiles();
     let hasAnyEndpointDiffs = false;
 
@@ -324,6 +327,7 @@ const getCaptureAction =
       );
     }
 
+    // learn path patterns
     if (options.update && options.interactive) {
       logger.info('');
       logger.info(
--- a/projects/optic/src/commands/capture/capture.ts
+++ b/projects/optic/src/commands/capture/capture.ts
@@ -242,41 +242,31 @@ const getCaptureAction =
 
     let errors: any[] = [];
     try {
-      let requestsPromise: Promise<any> = Promise.resolve();
+      // send requests
+      let sendRequestsPromise: Promise<any> = Promise.resolve();
       if (captureConfig.requests) {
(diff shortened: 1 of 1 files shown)


projects/optic/src/commands/capture/capture.ts
```diff
@@ -128,187 +155,117 @@ const getCaptureAction =
...
-      }
+    // start proxy
+    const proxy = new ProxyInstance(
+      options.serverOverride || captureConfig.server.url
+    );
+    await proxy.start(
+      options.proxyPort ? Number(options.proxyPort) : undefined
     );
+
+    // parse optic.yml
```
niclim, line 201: This isn't correct - this loads the spec

projects/optic/src/commands/capture/capture.ts
```diff
@@ -128,187 +155,117 @@ const getCaptureAction =
...
     // make requests
-    const captures = new GroupedCaptures(
-      trafficDirectory,
-      specToOperations(spec.jsonLike).flatMap((o) =>
-        o.methods.map((m) => ({ method: m, path: o.pathPattern }))
-      )
-    );
-
     let errors: any[] = [];
     try {
```
niclim, line 229: `app = await startApp` should be included in this try catch block. Otherwise any errors thrown while starting the app will not be caught and will not tear down the proxy server

projects/optic/src/commands/capture/capture.ts
```diff
@@ -492,3 +450,145 @@ function makeRequests(
...
+}
+
+async function specExists(filePath: string): Promise<boolean> {
+  try {
+    await fs.stat(filePath);
+  } catch {
+    return false;
+  }
+  return true;
+}
```
niclim, line 551: nit: I think this is an unnecessary layer of indirection - we can just put it in the `setup` function since it's only called once, and the setup fn is fairly straightforward

niclim (approved):
just small comments but overall looks good

projects/optic/src/commands/capture/capture.ts
notnmeyer: wrapped it in its own try-catch, https://github.com/opticdev/optic/pull/2097/commits/ac9fe617551983793c3c9f05d85044b4e0bb334a—otherwise there's bigger changes to make since we don't always run the server.

also now throwing an error when the server exits with non-zero code. doesn't quite handle all early termination situations (like an early exit with 0) but we're expecting the server to behave reasonably to a certain extent.

projects/optic/src/commands/capture/capture.ts
```diff
@@ -581,6 +581,13 @@ async function startApp(
   app.stderr.on('data', (data) => {
     logger.error(data.toString());
   });
+
+  app.on('exit', (code) => {
+    if (code! > 0) {
+      throw `Unexpected exit from the server with exit code: ${code}`;
+    }
+  });
```
niclim, line 589: So I'm not sure this will work as we expect - it's throwing inside of the fn here which may error after the server starts and will throw in a separate function stack

Instead I think this should be set _outside_ of this function and we should be tracking whether we expect the code to fail (i.e. with a variable `expectAppRunning`)

niclim: Ok so this still exits, but it results in an uncaught exception - I'm not sure what the best way to handle this is right now - will need to try some things to fix this - we can probably merge it and add a note here to handle this

Also we should throw ```new Error(`Unexpected exit from the server with exit code: ${code}`)``` instead

notnmeyer: commented it for now 👍

PR #2100, opened 2023-07-13
fix: install xdg-utils

notnmeyer:
## 🍗 Description

- vercel/pkg can't bundle xdg-open with the binary it produces. the CLI uses this to try and open web browsers. the absence of xdg-open entirely leads to errors when doing something that uses it,

```
➜ docker run --rm docker.io/useoptic/optic:local login
Generate a token below
Create an account and generate a personal access token at https://app.useoptic.com/user-settings/personal-access-token/new
? Enter your token here: › Error: spawn xdg-open ENOENT
    at ChildProcess._handle.onexit (node:internal/child_process:283:19)
    at onErrorNT (node:internal/child_process:476:16)
    at process.processTicksAndRejections (node:intern
```

installing xdg-utils will at least smooth out the experience somewhat,

```
➜ docker run --rm docker.io/useoptic/optic:local login
Generate a token below
Create an account and generate a personal access token at https://app.useoptic.com/user-settings/personal-access-token/new
? Enter your token here: › dfdsfd
```

this won't be useful for `diff --web`, but for a certain size of URL, it wouldn't matter anyway. `xdg-open` exits non-0 when there is no browser to open a link in, so optic would be able to look at that to determine when we couldn't launch a browser.

## 📚 References

## 👹 QA

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@ FROM alpine:latest
 
 ARG OPTIC_CLI_VERSION=latest
 
-RUN apk --no-cache add git curl
+RUN apk --no-cache add git curl xdg-utils
 RUN echo "optic-docker" > /etc/machine-id
 RUN set -e; sh -c "$(curl -s --location https://install.useoptic.com/install.sh)" -- $OPTIC_CLI_VERSION /usr/local/bin
 


niclim:
I'd need to spend some time to look into this but I don't know if we get the exit code from `open` https://github.com/opticdev/optic/blob/main/projects/optic/src/commands/login/login.ts#L54

If no error is thrown we don't really have a way to know if this was opened or not (which means we actually want xdg not to be installed in this case)

PR #2102, opened 2023-07-13
Fix/server running

niclim:
## 🍗 Description

Couple of things in this PR:
- used promise race to bail early if the server unexpected quit
- had to move a couple of things around to get this to work correctly

## 📚 References

## 👹 QA


projects/optic/src/commands/capture/capture.ts
```diff
@@ -204,47 +204,54 @@ const getCaptureAction =
...
-        logger.error(err);
-        process.exitCode = 1;
-        return;
-      }
-    }
 
-    // make requests
     let errors: any[] = [];
     try {
+      let bailout: Bailout = { didBailout: false, promise: Promise.resolve() };
```
niclim, line 218: Had to put this in an object so that:
- We can pass it around to different functions
- When we pass this into functions and mutate the value inside (i.e. didBailout) we need the other functions to see the change (objects are copied by reference, primitives are copied by value)

projects/optic/src/commands/oas/captures/streams/sources/proxy.ts
```diff
@@ -196,7 +196,8 @@ export class ProxyInteractions {
     });
 
     function onAbort(e) {
-      capturingProxy.reset();
+      capturingProxy.stop();
+      transparentProxy.stop();
```
niclim, line 200: This was being called but because the `interactions` stream wasn't being consumed, the process never exited

projects/optic/src/commands/capture/capture.ts
```diff
@@ -567,41 +543,78 @@ async function setup(filePath: string): Promise<string> {
...
+      .then((res) => String(res.status).startsWith('2'))
+      .catch((e) => {
+        logger.debug(e);
+        return false;
+      });
+
+  const serverReadyPromise = new Promise(async (resolve) => {
+    let done = false;
+    // We need to bail out if the server shut down, otherwise we never conclude this promise chain
+    while (!done && !bailout.didBailout) {
```
niclim, line 603: We need to bail out here otherwise this basically infinite loops (and node tried to wait for this to exit but it never does)

projects/optic/src/commands/capture/capture.ts
```diff
@@ -567,41 +543,78 @@ async function setup(filePath: string): Promise<string> {
...
+        throw new UserError('Server check timed out.');
+      }
+      await wait(readyInterval);
+    }
+    if (bailout.didBailout) spinner.fail('Server unexpectedly exited');
+    else spinner.succeed('Server check passed');
+    resolve(null);
+  });
+
+  await Promise.race([serverReadyPromise, bailout.promise]);
```
niclim, line 619: Here's the key part with bailout - we're waiting for either:
- the server to finish
- the bailout promise to reject

projects/optic/src/commands/capture/capture.ts
```diff
@@ -567,41 +543,78 @@ async function setup(filePath: string): Promise<string> {
...
   });
 
-  // TODO: this doesn't get caught by the calller but doesn't seem to leave the proxy hanging.
-  // lets find a better way.
-  app.on('exit', (code) => {
-    if (code! > 0) {
-      throw Error(`Unexpected exit from the server with exit code: ${code}`);
-    }
-  });
+  const bailout: Bailout = {
```
niclim, line 558: We wrap the `app.on` with a promise and use that wherever we would usually wait for something (i.e. server ready OR requests in flight) 

This can then be `Promise.race` which will resolve or reject with the first resolve OR rejection

projects/optic/src/commands/capture/capture.ts
```diff
@@ -204,47 +204,54 @@ const getCaptureAction =
...
-    }
 
-    // make requests
     let errors: any[] = [];
     try {
+      let bailout: Bailout = { didBailout: false, promise: Promise.resolve() };
+      if (!options.serverOverride && captureConfig.server.command) {
+        [app, bailout] = startApp(captureConfig.server.command, serverDir);
+        // If we don't bail out (i.e. the server is still running), we need the promise to be passed down to the next request
+        if (captureConfig.server.ready_endpoint) {
```
niclim, line 222: I split this up because before if you raise an error between starting the app, and the ready endpoint being set, the app var is never assigned and which means the `finally` block does not clean up the app

projects/optic/src/commands/capture/capture.ts
```diff
@@ -204,47 +204,54 @@ const getCaptureAction =
...
+          await waitForServer(
+            bailout,
+            captureConfig.server.ready_endpoint,
+            readyInterval,
+            timeout,
+            proxy.targetUrl
+          );
+        }
+      }
+      // TODO handle error handling properly
```
niclim, line 232: So this doesn't actually catch errors raised by requests in this block - didn't handle this in this PR but we can handle it separately (not sure off the top of my head for this)

notnmeyer (approved):

PR #2106, opened 2023-07-14
handle lint and diff file doesn't exist

niclim:
## 🍗 Description

With how we've changed spec loading, we don't throw errors on empty specs (because it's valid to compare against a gitref empty) - this means we have to throw on empty spec at some other point

## 📚 References

## 👹 QA


niclim:
Actually this is wrong - doesnt take into account cloud specs - there's a different fix here

PR #2118, opened 2023-07-17
feat(capture): combine requests* into requests.send and requests.run

notnmeyer:
## 🍗 Description

combines `requests` and `requests_command` configuration into `requests.(send|run)`. like before, you can use either or both. the config will wind up looking like this,

```
capture:
  openapi.yml:
    server:
      command: go run main.go
      url: http://localhost:8080
    requests:
      run:
        command: hurl hurl/*.hurl
        proxy_variable: HURL_OPTIC_PROXY
      send:
        - path: /
        - path: /users
        - path: /users/create
          method: POST
          data:
            name: Hank⏎  
```

if `requests` is missing entirely or an empty object,
```
➜ optic beta capture openapi.yml
"requests.send" or "requests.run" must be specified in optic.yml
```

## 📚 References
- closes https://github.com/opticdev/monorail/issues/4640

## 👹 QA

### DIFF
--- a/projects/optic/src/config.ts
+++ b/projects/optic/src/config.ts
@@ -34,7 +34,7 @@ const DefaultOpticCliConfig: OpticCliConfig = {
   isInCi: process.env.CI === 'true',
 };
 
-const Request = Type.Object({
+const RequestSend = Type.Object({
   path: Type.String(),
   method: Type.Optional(
     Type.String({
@@ -43,6 +43,13 @@ const Request = Type.Object({
   ),
   data: Type.Optional(Type.Object({})),
 });
+
+const RequestRun = Type.Optional(
+  Type.Object({
+    command: Type.String(),
+    proxy_variable: Type.Optional(Type.String()),
+  })
+);
 
 const CaptureConfigData = Type.Object({
   config: Type.Optional(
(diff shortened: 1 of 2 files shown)


projects/optic/src/config.ts
```diff
@@ -58,18 +65,19 @@ const CaptureConfigData = Type.Object({
...
   }),
-  requests: Type.Optional(Type.Array(Request)),
-  requests_command: Type.Optional(
+  requests: Type.Optional(
     Type.Object({
-      command: Type.String(),
-      proxy_variable: Type.Optional(Type.String()),
+      // one of these is technically required, but that's enforced at runtime
+      run: Type.Optional(RequestRun),
+      send: Type.Optional(Type.Array(RequestSend)),
```
notnmeyer, line 72: @niclim is there a good way to express that at least one of these needs to exist at the type level? would avoid needing to make a runtime check.

niclim: I think you can do a `OneOf` union of 
```
Type.OneOf([ // not sure the exact signature of this
Type.Object({  run: RequestRun, send: Type.Optional(Type.Array(RequestSend)), }),
Type.Object({  run: Type.Optional(RequestRun), send: Type.Array(RequestSend), })
])
```

I'm not really sure if that's clearer - what you have is fine though and will work

notnmeyer: ill take a quick look. i think in the long run it is clearer to do it here, otherwise we need to remember that there's random validation logic in the actual capture flow. i like the idea of knowing before we get there whether or not the config is 100% legit.

notnmeyer: after looking i don't think typebox does oneOf.

PR #2121, opened 2023-07-17
Support `postman` and `har` inputs for capture 2.0

niclim:
## 🍗 Description

Updates to support `postman` and `har` inputs. I moved out the capture / server / request generation into a separate file since it's pretty big and self contained (i.e. you make requests and run a server to get interactions out)

## 📚 References

## 👹 QA

### DIFF
--- a/projects/optic/src/commands/capture/capture.ts
+++ b/projects/optic/src/commands/capture/capture.ts
@@ -120,38 +120,9 @@ const getCaptureAction =
     targetUrl: string | undefined,
     options: CaptureActionOptions
   ) => {
-    const pathFromRoot = path.relative(config.root, path.resolve(filePath));
-    const captureConfig = config.capture?.[pathFromRoot];
-
     // capture v1
     if (targetUrl !== undefined) {
       await captureV1(filePath, targetUrl, config, command);
-      return;
-    }
-
-    // verify capture v2 config is present
-    if (targetUrl !== undefined || captureConfig === undefined) {
-      logger.error(`no capture config for ${filePath} was found`);
-      // TODO log error and run capture init or something - tbd what the first use is
-      process.exitCode = 1;
-      return;
-    }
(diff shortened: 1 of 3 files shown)


projects/optic/src/commands/capture/capture.ts
```diff
@@ -66,6 +58,17 @@ export function registerCaptureCommand(cli: Command, config: OpticCliConfig) {
       'add new endpoints in interactive mode. must be run with --update',
       false
     )
+    .option(
+      '--postman  <postman-collection-file>',
+      'path to postman collection'
+    )
+    .option('--har <har-file>', 'path to har file (v1.2, v1.3)')
+    .addOption(
```
niclim, line 66: moved this up above the deprecated commands

projects/optic/src/commands/capture/capture.ts
```diff
@@ -185,96 +157,63 @@ const getCaptureAction =
...
         ...p,
         path: p.pathPattern,
       }))
     );
-    for await (const har of harEntries) {
-      captures.addHar(har);
-      logger.debug(
-        `Captured ${har.request.method.toUpperCase()} ${har.request.url}`
+
+    if (options.har) {
```
niclim, line 172: if we specify har / postman we can skip the server running entirely

projects/optic/src/commands/capture/storage.ts
```diff
@@ -34,15 +34,23 @@ export async function getCaptureStorage(filePath: string): Promise<string> {
 }
 
 export class GroupedCaptures {
```
niclim, line 36: So I'm storing both interactions and hars here because we only have fns to convert postman -> interactions.

In the future we'd probably want everything to be directly stored as interactions but there's some other work to do to support this

notnmeyer (approved):

PR #2123, opened 2023-07-18
parse body while reading interactions

acunniffe:
## 🍗 Description
We had readable streams for the bodies in each interaction. This made it hard to debug without providing any performance benefit. The interactions are streamed and all the content in the body was already in memory. Having another buffer here doesn't gain any performance. 

## 📚 References

## 👹 QA


projects/optic/src/commands/capture/sources/body.ts
```diff
@@ -1,107 +1,48 @@
...
   // TODO: add support for stream parsing JSON (see stream-json package)
   // static async jsonStream() {}
 
   static async text(body: CapturedBody) {
-    const buffer = await consumeBody(body);
-    return new TextDecoder().decode(buffer);
-  }
-}
-
-async function consumeBody(body: CapturedBody): Promise<Buffer> {
```
acunniffe, line 67: not needed.

projects/optic/src/commands/capture/sources/captured-interactions.ts
```diff
@@ -57,8 +57,11 @@ export class CapturedInteraction {
         requestPostData.text,
         requestPostData.encoding as BufferEncoding | undefined
       );
+
+      const asText = new TextDecoder().decode(buffer);
+
       requestBody = CapturedBody.from(
-        buffer,
+        asText,
```
acunniffe, line 64: This is the big change -- instead of creating a buffer JUST to stream it later back into text, we use the buffer's built in encoding to parse the string data in the HAR base64 or gzip into a normal string.

projects/optic/src/commands/oas/operations/diffs/visitors/request-body.test.ts
```diff
@@ -147,9 +147,9 @@ describe('visitRequestBody', () => {
     ]).toHaveLength(0);
     expect([
       ...visitRequestBody(subtypeMisMatch, requestBodySpec),
-    ]).toHaveLength(1);
+    ]).toHaveLength(0);
```
acunniffe, line 150: Before this would allow you to still learn a body if the content type didn't match. Not sure we actually want this so I changed the test. It just causes weird things to be learned later on.

projects/optic/src/commands/capture/sources/body.ts
```diff
@@ -1,107 +1,48 @@
...
-  stream: Readable | null;
+  body: string | null;
 }
 
 export class CapturedBody {
   static from(
-    source: Readable | Buffer | ArrayBuffer | string | null = null,
+    stringValue: string | null,
     contentType: string | null = null,
     size = 0
```
niclim, line 13: I'm not sure we actually need size anywhere since we're no longer streaming the value

niclim (approved):

PR #2133, opened 2023-07-18
Fix ref refactor addition

niclim:
## 🍗 Description

When i ported over coverage i didn't realize that added paths actually mean json paths in the spec rather than the endpoint path 🙃 

## 📚 References

## 👹 QA


notnmeyer (approved):

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "openapi-workspaces",
   "license": "MIT",
   "private": true,
-  "version": "0.47.13",
+  "version": "0.47.14",
```
niclim, line 5: going to publish again because our older version seems to have a json pointer helpers version already published https://github.com/opticdev/optic/actions/runs/5591403588

PR #2136, opened 2023-07-19
Flip the order of inputs to the diff command

gabsong:
When running the diff command, the actual order of inputs is `optic diff old_file new_file` or
`optic diff --base branch new_file`

## 🍗 Description
This makes an improvement the manual for `optic diff` to make it more clear.

## 📚 References

## 👹 QA
I was confused by the current help manual. I don't know if others feel the same. It was not clear which one is the old version spec and the new version spec. I did not change variable names, but it may help to indicate what is the old one and what is the new one.


niclim (approved):
Thanks for the contribution!

PR #2137, opened 2023-07-19
fix could not learn json schema

niclim:
## 🍗 Description

fixes this fn block - looks like if you pass `undefined` you'll throw an error here 

https://opticdev.sentry.io/issues/4324398092/?alert_rule_id=9659341&alert_type=issue&project=6089645&referrer=slack

## 📚 References

## 👹 QA

### DIFF
--- a/projects/optic/src/commands/oas/shapes/schema.ts
+++ b/projects/optic/src/commands/oas/shapes/schema.ts
@@ -136,7 +136,7 @@ function initialSchema(
   rootInput: any,
   openAPIVersion: SupportedOpenAPIVersions
 ): OpenAPIV3.SchemaObject {
-  if (rootInput === null) {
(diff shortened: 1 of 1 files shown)


projects/optic/src/commands/oas/shapes/schema.ts
```diff
@@ -136,7 +136,7 @@ function initialSchema(
   rootInput: any,
   openAPIVersion: SupportedOpenAPIVersions
 ): OpenAPIV3.SchemaObject {
-  if (rootInput === null) {
+  if (rootInput === null || rootInput === undefined) {
```
niclim, line 139: I think it's correct here to treat a keyed input but it returning `undefined` to be a `nullable` schema

PR #2138, opened 2023-07-19
Update transpilation target to be node16

niclim:
## 🍗 Description

Updates our transpilation target to be node16+ - before this was generating ES5 which the sourcemaps that were generated are harder to debug. node16+ looks like it's ES2021+ compatible from this table https://node.green/

TODO
- [x] create prerelease and test
- [x] test ruleset imports

## 📚 References

## 👹 QA
prerelease with version v0.47.15-1

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "openapi-workspaces",
   "license": "MIT",
   "private": true,
-  "version": "0.47.14",
+  "version": "0.47.15-1",
   "workspaces": [
     "projects/json-pointer-helpers",
     "projects/openapi-io",
@@ -28,5 +28,6 @@
     "**/*.+(js|jsx|ts|tsx|json|css)": [
       "yarn prettier --write"
     ]
-  }
+  },
+  "stableVersion": "0.47.14"
 }
--- a/projects/fastify-capture/tsconfig.json
+++ b/projects/fastify-capture/tsconfig.json
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "ES2020",
+    "target": "ES2021",
     "rootDir": "src",
     "declaration": true,
     "declarationMap": true,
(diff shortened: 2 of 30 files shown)


projects/rulesets-base/src/custom-rulesets/load-ruleset.ts
```diff
@@ -0,0 +1,5 @@
+// Handle ESM and CJS modules
+export function loadRuleset(path: string) {
```
niclim, line 2: So we were using dynamic imports before which effectively got compiled down to this. 

This require + esm interop allowed us to handle both CJS + ESM ruleset files, and dynamic imports _only_ supported ESM.

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "openapi-workspaces",
   "license": "MIT",
   "private": true,
-  "version": "0.47.14",
+  "version": "0.47.15-1",
```
Alexandre-Herve, line 5: 👀

projects/fastify-capture/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es5",
+    "target": "ES2020",
```
Alexandre-Herve, line 3: is there a specific reason why not `ES2021`?

projects/standard-rulesets/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es5",
+    "target": "ES2020",
```
Alexandre-Herve, line 3: same here wondering why not 2021

Alexandre-Herve (approved):

package.json
niclim: not planning on releasing a new release version yet - will keep it at the prerelease for any future prerelease bumps

projects/fastify-capture/tsconfig.json
niclim: Oops - fixing

PR #2146, opened 2023-07-20
feat: `optic capture init`

notnmeyer:
## 🍗 Description

adds support for writing a sample `capture` block to an existing optic.yml file, or printing a sample block to stdout.

- write to optic.yml: `optic beta capture init openapi.yml`
- print to stdout: `optic beta capture init openapi.yml --stdout`

a couple cases for consideration:
- if no optic.yml exists, one will be created with the generated capture block.
- if a capture block exists and has content,
    ```
    ➜ optic beta capture init openapi.yml
    optic.yml already contains a capture config for the file openapi.yml. This command would overwrite the existing configuration. Make changes manually, or view a sample capture configuration with: `optic capture init openapi.yml --stdout`
    ```
## 📚 References

## 👹 QA


projects/optic/src/commands/capture/capture.ts
```diff
@@ -131,6 +139,31 @@ const getCaptureAction =
       return;
     }
 
+    if (options.init) {
+      // no optic.yml is present but the command would attempt to write
+      if (!config.configPath && !options.stdout) {
```
niclim, line 144: If this is the case we should write to `config.root` - this is either the git root, or the cwd

projects/optic/src/commands/capture/capture.ts
```diff
@@ -131,6 +139,31 @@ const getCaptureAction =
...
+        Object.keys(config.capture).length !== 0 &&
+        !options.stdout
+      ) {
+        logger.info(
+          'Your optic.yml already contains a `capture` block with contents and this command would modify it. Make modifications to your capture config by editing optic.yml. If you want to see a complete capture block example, run `optic beta capture openapi.yml --init --stdout` instead.'
+        );
+        return;
+      }
+
+      initCaptureConfig(filePath, options.stdout!, config.configPath!);
```
niclim, line 163: So what I think we should do here instead is:
- If there's no file, we write a file to `config.root`
- If there is a no capture key - we write a `capture:` key
- If there's an existing key that matches the openapi file name
  - we log a warnning
- If not, we can inject the file here

Not entirely sure what functions we can use to inject into a yml doc and preserve comments but we can pair on this

projects/optic/src/utils/write-optic-config.ts
```diff
@@ -0,0 +1,21 @@
...
+import { logger } from '../logger';
+
+// merges a yaml document into the existing optic.yml and writes it to disk. if newDocument's top-level
+// key already exists in the Optic config, it is overwritten.
+export function updateOpticConfig(
+  newDocument: yaml.Document.Parsed,
+  opticConfigPath: string
+) {
+  try {
+    let opticYml = yaml.parseDocument(fs.readFileSync(opticConfigPath, 'utf8'));
```
niclim, line 12: Instead of `readFileSync`, we should use `await fs.readFile` which is exported from `import fs from 'node:fs/promises'`

projects/optic/src/utils/write-optic-config.ts
```diff
@@ -0,0 +1,21 @@
...
+
+    const topLevelKey = Object.keys(newDocument.toJSON())[0];
+    opticYml.set(topLevelKey, newDocument.get(topLevelKey));
+
+    fs.writeFileSync(opticConfigPath, opticYml.toString());
```
niclim, line 17: same here, use `await fs.writeFile`

projects/optic/src/utils/write-optic-config.ts
```diff
@@ -0,0 +1,21 @@
...
+  } catch (err) {
+    logger.error(err);
```
niclim, line 19: Instead of logging this error here, we might want to let the error throw up to the top level and handle it in the `capture.ts` file

projects/optic/src/commands/capture/init.ts
```diff
@@ -0,0 +1,71 @@
+import yaml from 'yaml';
+import { updateOpticConfig } from '../../utils/write-optic-config';
+import { logger } from '../../logger';
+
+export function initCaptureConfig(
+  oasFile: string,
```
niclim, line 6: This file name need to be relative to the git root https://github.com/opticdev/optic/blob/main/projects/optic/src/commands/capture/capture.ts#L176

projects/optic/src/commands/capture/capture.ts
notnmeyer: > If there's an existing key that matches the openapi file name
> we log a warning and log out instead
> If not, we can inject the capture config with the file name

i don't know that i agree that its important for this to support inserting more than 1 capture[oas_file] section. once you have the first one, you're bootstrapped—you've seen the complete example already. adding more capture configuration manually is trivial at that point, right?

notnmeyer: > If there's no file, we write a file to config.root
> If there is a no capture key - we write a capture: key

these are handled

niclim: You're right that we're likely going to use this for the the first use for the majority of the time, but I don't think it unreasonable for someone to run capture init when they've already added a capture config.

It's probably not that important to have for the first launch of this but I think it does add a level of polish here

niclim (approved):
Last thought - should we move this into a separate command like `optic capture init`? The `optic capture` command is starting to have a lot of different options and it might be clearer to split out into a separate command?

notnmeyer:
sure, that'd be fine

projects/optic/src/commands/capture/init.ts
```diff
@@ -0,0 +1,75 @@
...
+            # path: required
+            # method: optional, default: GET
+            - path: /
+              method: GET
+            - path: /users/create
+              method: POST
+              # optional, if omitted on a POST, default: {}
+              data:
+                name: Hank
+  `;
```
notnmeyer, line 74: any issues with the default value here? this is more or less the example ive been using in `opticdev/openapi-gen-go-gin`

projects/optic/src/commands/capture/capture.ts
```diff
@@ -111,6 +113,8 @@ export function registerCaptureCommand(cli: Command, config: OpticCliConfig) {
   cli.addCommand(command);
 }
 type CaptureActionOptions = {
+  init?: boolean;
+  stdout?: boolean;
```
niclim, line 117:
```suggestion

```

projects/optic/src/commands/capture/init.ts
```diff
@@ -0,0 +1,75 @@
...
+  return `
+    capture:
+      ${oasFile}:
+        config:
+          # the number of parallel requests to make when using 'requests.send'
+          # optional, default: 4
+          request_concurrency: 4
+        server:
+          # a command to run your server
+          # optional, if ommitted Optic assumes the server is running or started elsewhere
```
niclim, line 37:
```suggestion
          # optional, if omitted Optic assumes the server is running or started elsewhere
```

projects/optic/src/commands/capture/init.ts
```diff
@@ -0,0 +1,75 @@
...
+          command: your-server-command
+          # the url where your server can be reached once running
+          # required, can be overridden with '--server-override'
+          url: http://localhost:8080
+          # a readiness endpoint for Optic to validate before sending requests
+          # optional, default: /
```
niclim, line 43: I don't think we're defaulting this value to `/` - if you don't pass us anything we don't hit the endpoint

niclim (approved):

projects/optic/src/commands/capture/init.ts
notnmeyer: you're right — i'll update that comment

PR #2151, opened 2023-07-21
Prepare capture release

niclim:
## 🍗 Description

Moves capture command out of `optic beta capture` -> `optic capture` and deprecates the old commands

Any old capture command called with `optic capture <file> <url>` will call the old capture methods

## 📚 References

## 👹 QA


notnmeyer:
should this be more than a patch release?

notnmeyer (approved):

PR #2165, opened 2023-07-26
chore(capture): add top level comment to capture config

notnmeyer:
## 🍗 Description

- de-nates the capture comments—capitalize things, use punctuation

## 📚 References

## 👹 QA

there is a placeholder for a capture documentation link that we don't have published yet. don't merge this until there are docs and we've included a link.

### DIFF
--- a/projects/optic/src/commands/capture/init.ts
+++ b/projects/optic/src/commands/capture/init.ts
@@ -31,7 +31,7 @@ export function captureConfigExample(oasFile: string) {
       ${oasFile}:
         config:
           # The number of parallel requests to make when using 'requests.send'.
-          # Optional: default 4
+          # Optional: default: 4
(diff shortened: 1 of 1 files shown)


Alexandre-Herve (approved):

projects/optic/src/commands/capture/init.ts
```diff
@@ -26,50 +26,52 @@ export async function initCaptureConfig(
 // returns a complete Capture block example
 export function captureConfigExample(oasFile: string) {
   return `
+    # Configures the 'optic capture' flow. View the documentation for details about how capture works: https://www.useoptic.com/docs/capturing-traffic
```
notnmeyer, line 29: url doesn't work yet, but will as soon as we publish the corresponding doc updates which should happen shortly.

PR #2168, opened 2023-07-27
reject instead of throw error

niclim:
## 🍗 Description

by throwing we raise an uncaught exception, where we instead should be rejecting the promise so it's handled in our try catch

## 📚 References

## 👹 QA


notnmeyer (approved):
kicked the tires, it works

PR #2171, opened 2023-07-27
feat(capture): add headers to requests.send

notnmeyer:
## 🍗 Description

allows setting headers on `requests.send`... requests. to keep the verbosity and repetition down you can use anchors and aliases.

```yaml
send:
  - path: /
    headers: &headers
      x-some-header: awesome
      content-type: application/json;charset=utf-8
  - path: /users/create
    method: POST
    data:
      name: Hank
    headers:
      <<: *headers
      x-additional-header: also-awesome
```

```
===BEGIN HEADERS===
Content-Type: [application/json;charset=utf-8]
Accept: [*/*]
User-Agent: [node-fetch/1.0 (+https://github.com/bitinn/node-fetch)]
Accept-Encoding: [gzip,deflate]
Connection: [close]
X-Some-Header: [awesome]
===END HEADERS===
[GIN] 2023/07/27 - 13:02:37 | 200 |      54.375µs |       127.0.0.1 | GET      "/"

===BEGIN HEADERS===
X-Some-Header: [awesome]
Content-Length: [15]
Accept-Encoding: [gzip,deflate]
Connection: [close]
Content-Type: [application/json;charset=utf-8]
X-Additional-Header: [also-awesome]
Accept: [*/*]
User-Agent: [node-fetch/1.0 (+https://github.com/bitinn/node-fetch)]
===END HEADERS===
[GIN] 2023/07/27 - 13:02:37 | 201 |     241.833µs |       127.0.0.1 | POST     "/users/create"
```

todo: 
- [x] add a special case for content-type: if its supplied in the yaml, use that value, otherwise set it to `application/json;charset=utf-8`
- [x] update website docs
- [x] update CLI docs

## 📚 References

## 👹 QA

### DIFF
--- a/projects/optic/src/commands/capture/actions/captureRequests.ts
+++ b/projects/optic/src/commands/capture/actions/captureRequests.ts
@@ -143,7 +143,24 @@ function sendRequests(
 
     r.data ? (opts['body'] = JSON.stringify(r.data)) : '{}';
 
-    if (r.headers) opts['headers'] = r.headers;
+    // add headers to the request
+    if (r.headers) {
+      // convert all map keys to lowercase for ease of use
+      const headers = Object.keys(r.headers).reduce(
+        (acc, key) => {
+          acc[key.toLowerCase()] = r.headers![key];
+          return acc;
+        },
+        {} as { [key: string]: string }
+      );
+
+      opts['headers'] = headers;
+    }
+
+    // if a content-type header is not set, add it
+    if (opts['headers'] && !opts['headers'].hasOwnProperty('content-type')) {
+      opts['headers']['content-type'] = 'application/json;charset=utf-8';
+    }
 
     return limiter.schedule(() =>
       fetch(`${proxyUrl}${r.path}`, opts)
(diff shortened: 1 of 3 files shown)


projects/optic/src/config.ts
```diff
@@ -43,6 +43,7 @@ const RequestSend = Type.Object({
     })
   ),
   data: Type.Optional(Type.Object({})),
+  headers: Type.Optional(Type.Array(Type.Record(Type.String(), Type.String()))),
```
niclim, line 46: Should this instead be just a record? I don't think we need to send multiple arrays of headers

e.g. `{ header-1: asd, header-2: 123 }` if you want to specify multiple

niclim (approved):

projects/optic/src/commands/capture/actions/captureRequests.ts
```diff
@@ -139,13 +139,33 @@ function sendRequests(
...
     let verb = r.method || 'GET';
-    let opts = { method: verb };
+    let opts = {
+      method: verb,
+      headers: {},
+    };
+
+    if (r.data) opts['body'] = JSON.stringify(r.data);
+
+    try {
```
niclim, line 149: I think you can remove this try block - I think this will only throw when `r.headers` isn't the shape we expect, and since we're running it through validation this shouldn't happen

niclim (approved):

PR #2173, opened 2023-07-27
Write capture test for overlapping paths

niclim:
## 🍗 Description

Was debugging something and wrote this test, but looks like it's working as expected. Since I already wrote it I think it's worth merging to cover this case (requests to `/books` `/books/status` and `/books/{bookId}`) should have the correct specificity

## 📚 References

## 👹 QA

### DIFF
--- a/projects/optic/src/__tests__/integration/capture.test.ts
+++ b/projects/optic/src/__tests__/integration/capture.test.ts
@@ -49,6 +49,18 @@ describe('capture with requests', () => {
       expect(normalizeWorkspace(workspace, combined)).toMatchSnapshot();
       expect(code).toBe(1);
     });
+
+    test.only('verifies spec with endpoint specificity', async () => {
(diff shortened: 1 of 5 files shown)


projects/optic/src/__tests__/integration/capture.test.ts
```diff
@@ -49,6 +49,18 @@ describe('capture with requests', () => {
       expect(normalizeWorkspace(workspace, combined)).toMatchSnapshot();
       expect(code).toBe(1);
     });
+
+    test.only('verifies spec with endpoint specificity', async () => {
```
Alexandre-Herve, line 53: forgotten `only` here

Alexandre-Herve (approved):

projects/optic/src/__tests__/integration/capture.test.ts
niclim: ah thanks

PR #2176, opened 2023-07-31
Sentry error fixes

niclim:
## 🍗 Description

Fixes a couple of sentry errors
- Add a guard when `requestPostData.text` is undefined
- always try to unwrap captured body text

## 📚 References

## 👹 QA

### DIFF
--- a/projects/optic/src/commands/capture/sources/body.ts
+++ b/projects/optic/src/commands/capture/sources/body.ts
@@ -25,13 +25,8 @@ export class CapturedBody {
 
   static async json(body) {
     const text = await CapturedBody.text(body);
-    if (text) {
-      return JSON.parse(text);
-    }
+    return JSON.parse(text!);
   }
-
-  // TODO: add support for stream parsing JSON (see stream-json package)
-  // static async jsonStream() {}
 
   static async text(body: CapturedBody) {
     return body.body;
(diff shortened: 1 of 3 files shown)


projects/optic/src/commands/capture/sources/body.ts
```diff
@@ -25,14 +25,9 @@ export class CapturedBody {
 
   static async json(body) {
     const text = await CapturedBody.text(body);
-    if (text) {
-      return JSON.parse(text);
-    }
+    return JSON.parse(text!);
```
niclim, line 28: Always try to parse this so that we raise an error on undefined or something else (handled in the above case)

projects/optic/src/commands/oas/shapes/streams/documented-bodies.ts
```diff
@@ -264,7 +263,7 @@ async function decodeCapturedBody(
     try {
       value = await CapturedBody.json(capturedBody);
     } catch (err) {
-      SentryClient.captureException(err);
+      logger.error(err);
```
niclim, line 266: I think this should be in the responsibility of the user to fix this here - changing this to log error instead of sentry error.

This could be a malformed har file, or postman file that causes this error. Captured data / interactions here should be correctly formed from our code (if there's a mismatch, this is something an integration test should be catching this)

Alexandre-Herve (approved):

PR #2182, opened 2023-08-01
Fix/test flakiness with integration tests

niclim:
## 🍗 Description

Tests seem to be flaky because of the spinner + competing with other std out in certain cases and in tests. I'm updating this to not run the spinner when in tests as we've run into a lot of flakiness

## 📚 References

## 👹 QA


notnmeyer (approved):
hahaha, amazing

PR #2183, opened 2023-08-02
Surface command output when running request.command

niclim:
## 🍗 Description

## 📚 References

## 👹 QA

### DIFF
--- a/projects/optic/src/commands/capture/actions/captureRequests.ts
+++ b/projects/optic/src/commands/capture/actions/captureRequests.ts
@@ -195,10 +195,6 @@ async function runRequestsCommand(
     reqCmd.on('exit', (code) => {
       if (code === 0) {
         resolve();
-      } else if (code) {
-        process.exitCode = code;
-        logger.error(`command ${command} failed with exit code ${code}`);
-        resolve();
       } else {
         reject();
       }
--- a/projects/optic/src/commands/capture/actions/captureRequests.ts
+++ b/projects/optic/src/commands/capture/actions/captureRequests.ts
@@ -19,6 +19,25 @@ import { getSpinner } from '../../../utils/spinner';
 
 const defaultServerReadyTimeout = 10_000; // 10s
 
+// Clears logging so that the spinner is always flushed to the bottom until it is completed
+const loggerWhileSpinning = {
+  debug: (spinner?: ora.Ora, ...msg: any[]) => {
+    spinner?.clear();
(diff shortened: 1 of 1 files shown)


projects/optic/src/commands/capture/actions/captureRequests.ts
```diff
@@ -195,14 +195,18 @@ async function runRequestsCommand(
...
+        resolve();
       } else {
         reject();
       }
     });
   });
 
   reqCmd.stdout.on('data', (data) => {
-    logger.debug(data.toString());
+    logger.info(data.toString());
```
niclim, line 209: I think it's correct here to long data as it comes out in `info` - where server only logs on `debug`. Reasoning here is that the command _may_ also be tests that you want to see the test output in. 

We might also need to do some formatting / handle interactions with the spinner here, but I think that can come later

notnmeyer:
example output,

```
➜ npx ts-node ../optic/projects/optic/src/index.ts capture openapi.yml
⠋ Sending requests to server{"message":"Hello, world!"}
[{"id":1,"name":"John"},{"id":2,"name":"Bob"},{"id":3,"name":"Alice"}]
{"id":4,"name":"Hank"}
✔ Successfully captured requests

3 unmatched interactions
New endpoints are only added in interactive mode.
Run with `--update interactive` to add new endpoints
Hint: optic capture openapi.yml --update interactive
```

i kind of ignored how ugly the output can be when the output was behind debug, but if we're displaying this by default then we should add a newline after "Sending requests to server"

niclim:
Hmmm i agree, it's a little more complicated with how the spinner works (the spinner assumes it's the last thing to print to stdout), let me see what i can do here, i saw some workarounds from the ora github issues page

notnmeyer:
depending on how annoying that workaround is, we could probably just insert an `echo;` before the contents of `requests.run.command`

niclim:
@notnmeyer I updated this so that the spinner is always flushed to the bottom until it is completed (i.e. everything inside of the create requests should be better spaced so we're logging in the right areas)

```
➜  my-openapi-specs git:(main) ✗ npx ts-node ../optic/projects/optic/src/index.ts capture abe.yml 
asdasdasdasda
 a
 a 
 a 
 a 
 a 
 a 


sd

✔ Successfully captured requests
✔ GET /books
  ✓ 200 response

3 unmatched interactions
```

notnmeyer (approved):
yep, works for me

PR #2192, opened 2023-08-04
Ci setup refactor

Alexandre-Herve:
## 🍗 Description

- removed safeties around uncommitted changes
- always try to identify specs by path unless x-optic-url is set
- simplified the ci setup command
- deprecated the `--generated` command
- various logging fixes and improvements

I'm going to do a prerelease that we can test thoroughly before releasing a major version.

## 📚 References

## 👹 QA


projects/optic/ci/configs/github.yml
```diff
@@ -23,8 +27,12 @@ jobs:
...
           # If false, standard check failures will show in PR comments and
           # in Optic Cloud but will not cause the action to fail
-          standards_fail: %standards_fail
+          standards_fail: true
 
           # If you have more than one spec, separate matches with commas
-          # (openapi.yml,other.yml)
-          additional_args: --match %match
+          # Examples of match: "openapi.yml" or "openapi.yml,other.yml" or "*.spec.yml"
+          additional_args: --match ""
```
niclim, line 34: maybe the default here could be `openapi.yml`?  Either way, what happens if the user doesn't change this and runs diff-all in ci? Do we silently fail or pass? I can't remember how this works off the top of my head but I think this should fail so the user knows there's something to fix here

projects/optic/src/commands/diff/diff-all.ts
```diff
@@ -609,6 +609,9 @@ function applyGlobFilter(
 
 const getDiffAllAction =
   (config: OpticCliConfig) => async (options: DiffAllActionOptions) => {
+    if (options.generated) {
+      logger.warn(chalk.yellow.bold(`the --generated option is deprecated`));
```
niclim, line 613:
```suggestion
      logger.warn(chalk.yellow.bold(`the --generated option is deprecated, diff works without the --generated option`));
```

niclim (approved):
I think this looks good!

projects/optic/ci/configs/github.yml
Alexandre-Herve: This will match any spec by default, which I think is a great default. When no api is found, diff-all throws

PR #2194, opened 2023-08-04
Fix/non root server urls

niclim:
## 🍗 Description

Fixes the case where the server is not at the host root - e.g. `http://example.com/prefix`

Sending requests via the requests.send, requests.cmd and the ready check should all be adjusted. e.g.
a request to `/books` should be routed to `/prefix/books` if server. This request should be modeled in the OpenAPI file as `/books` (not `/prefix/books`).

## 📚 References

## 👹 QA

### DIFF
--- a/projects/optic/src/__tests__/integration/capture.test.ts
+++ b/projects/optic/src/__tests__/integration/capture.test.ts
@@ -113,6 +113,25 @@ describe('capture with requests', () => {
       ).toMatchSnapshot();
     });
 
+    test('handle servers not at root of host', async () => {
+      process.env.SERVER_PREFIX = '/api';
+      const workspace = await setupWorkspace('capture/with-server');
+      await setPortInFile(workspace, 'optic.yml');
+
+      const { combined, code } = await runOptic(
+        workspace,
+        `capture openapi-prefixed-url.yml --update automatic`
+      );
+      expect(normalizeWorkspace(workspace, combined)).toMatchSnapshot();
+      expect(code).toBe(0);
+      expect(
+        await fs.readFile(
+          path.join(workspace, 'openapi-prefixed-url.yml'),
+          'utf-8'
+        )
+      ).toMatchSnapshot();
+    });
+
     test('handle server path prefixes in spec', async () => {
       process.env.SERVER_PREFIX = '/api';
       const workspace = await setupWorkspace('capture/with-server');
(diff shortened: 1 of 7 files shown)


projects/optic/src/commands/capture/sources/proxy.ts
```diff
@@ -56,19 +63,10 @@ export class ProxyInteractions {
       );
       throw new UserError();
     }
-    if (targetHost.includes('/')) {
-      // accept urls to be passed in rather than pure hosts
-      targetHost = host;
-    }
```
niclim, line 62: Uhm I'm confused what this is trying to do, but this would always return true, since new URL needs a `protocol` (`http://` or `https://`)

projects/optic/src/commands/capture/sources/proxy.ts
```diff
@@ -56,19 +63,10 @@ export class ProxyInteractions {
...
+    targetHost = host;
 
     const forwardHost = options.mode === 'reverse-proxy' ? origin : targetHost;
 
-    if (protocol)
-      invariant(
-        targetHost.match(/^([a-zA-Z0-9\-]+\.)*[a-zA-Z0-9\-]+(:\d+)?$/),
-        'targetHost must be a valid host (hostname:port)'
-      ); // port optional
-
```
niclim, line 71: again, not sure what this is trying to do since I don't think this would ever latch since `new URL` throws an error on malformed urls

projects/optic/src/commands/capture/sources/proxy.ts
```diff
@@ -152,6 +155,9 @@ export class ProxyInteractions {
         timingEvents: timingEvents as TimingEvents,
       };
       requestsById.set(request.id, request);
+      return {
+        url: prefixedUrl,
```
niclim, line 159: This adjusts the url that we forward to

projects/optic/src/__tests__/integration/capture.test.ts
```diff
@@ -113,6 +113,25 @@ describe('capture with requests', () => {
       ).toMatchSnapshot();
     });
 
+    test('handle servers not at root of host', async () => {
+      process.env.SERVER_PREFIX = '/api';
```
notnmeyer, line 117: just to make sure im not missing something. is the env var just for test convenience, or is SERVER_PREFIX a valid way to set the prefix when running optic?

niclim: This is for test convenience - so that the server that I've build will only respond with prefixes

notnmeyer: ok, cool, that's what it looked like 👍

notnmeyer (approved):

PR #2204, opened 2023-08-08
Remove more unused code

niclim:
## 🍗 Description

Removes more unused code referring to old optic-ci workflows

## 📚 References

## 👹 QA


projects/openapi-utilities/src/legacy-ci-types.ts
```diff
@@ -0,0 +1,10 @@
+import { ResultWithSourcemap } from './types';
+import { IChange } from './openapi3/sdk/types';
+
+export type CompareFileJson = {
+  results: ResultWithSourcemap[];
+  changes: IChange[];
+  // TODO remove this on 2022 Aug
```
Alexandre-Herve, line 7: 2022?

Alexandre-Herve (approved):

projects/openapi-utilities/src/legacy-ci-types.ts
niclim: I think we're actually using this in OCFE, which I need to go and check when we can remove - this will be gone when ocfe is gone

PR #2207, opened 2023-08-09
fix: denormalize operation before computing checksum

Alexandre-Herve:
## 🍗 Description

Denormalize operation before computing checksum to match other checksums.

There is a type mismatch though:
```
Type '(ReferenceObject | ParameterObject)[]' is not assignable to type  'ParameterObject[]'. 
```

Not sure how to handle it

## 📚 References

## 👹 QA


niclim (commented):
Hmm I think we might need to actually denormalize the entire spec - since this is used in the api coverage calculation, maybe it's best if we denormalize the entire spec here https://github.com/opticdev/optic/blob/main/projects/optic/src/commands/capture/coverage/api-coverage.ts#L21

We might need to change the denormalization function to optionally take in a sourcemap as well for this to work

niclim (approved):

PR #2218, opened 2023-08-10
handle generated specs in ci setup

Alexandre-Herve:
## 🍗 Description

Minimalistic fix for the generated vs non generated in cloud issue. This is roughly equivalent to what was there before in this regard and I think it's enough to release Optic and have something "working", but I think we need to design something better in the long run.

## 📚 References

## 👹 QA


niclim (approved):
Should we also log that you should look at the file and change where there's the 🔧 icon?

PR #2224, opened 2023-08-11
fix capture not forwarding the correct headers

niclim:
## 🍗 Description

Sometimes, when forwarding requests sentry would add additional headers for tracing (it may also be messing up the existing headers as well). This would be attached whenever we forwarded requests using `capture`. Certain gateways and servers don't like it when you add in extra headers. Disabling this will no longer add sentry headers to http requests

## 📚 References

## 👹 QA

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "openapi-workspaces",
   "license": "MIT",
   "private": true,
-  "version": "0.49.0",
+  "version": "0.49.1-1",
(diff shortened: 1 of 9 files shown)


notnmeyer:
can we get a prerelease built from this branch for testing?

niclim:
`0.49.1-1` is building

notnmeyer (approved):
yep, this is working for me

PR #2227, opened 2023-08-14
Fix adding a host header to a capture request rerouting to different place

niclim:
## 🍗 Description

A curl request like: `curl -H "host: useoptic.com" localhost:8000/pets` when running something like `optic capture --server-override example.com` results in our proxy server forwarding a request to the set host header, instead of the proxy target

Note that I've _only_ fixed the capture 2.0 flow, since the code that was handling this I think was related to the system proxy code. I've forked the proxy code so capture 2.0 should fix this behavior.

Running something like this with the capture config should now work
```
# shell
optic capture pets.yml --server-override https://vr0yklzhdl.execute-api.us-east-1.amazonaws.com/dev
# optic.yml
capture:
  pets.yml:
    server:
      command: your-server-command
      url: http://localhost:8080
    requests:
      run:
        command: "sh -c \"curl $OPTIC_PROXY/pets --header 'Authorization: redacted' -H 'host: useoptic.com' --header 'content-type: application/json'\""
    config:
      request_concurrency: 4

```

publishing `0.49.1-3` which we can use to test this

## 📚 References

## 👹 QA

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "openapi-workspaces",
   "license": "MIT",
   "private": true,
-  "version": "0.49.1-1",
+  "version": "0.49.1-2",
   "workspaces": [
     "projects/json-pointer-helpers",
     "projects/openapi-io",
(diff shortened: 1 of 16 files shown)


notnmeyer (approved):
👍 working for me

PR #2229, opened 2023-08-15
Handle non json responses and bodies

niclim:
## 🍗 Description

This build will be broken until we fix _another_ bug that this uncovered where if you add 2 schemas in the same patch it explodes. I'm going to create another PR that targets this PR that will fix this issue

This PR fixes:
- `invalid json response body` when returned a non-json response
- Documents non-json responses as `schema: type: string` 

## 📚 References

## 👹 QA

### DIFF
--- a/projects/optic/src/commands/capture/actions/captureRequests.ts
+++ b/projects/optic/src/commands/capture/actions/captureRequests.ts
@@ -147,7 +147,7 @@ function sendRequests(
   proxyUrl: string,
   concurrency: number,
   spinner?: ora.Ora
-): Promise<void>[] {
+): Promise<any>[] {
   const limiter = new Bottleneck({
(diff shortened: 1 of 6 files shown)


projects/optic/src/commands/capture/actions/captureRequests.ts
```diff
@@ -180,11 +180,9 @@ function sendRequests(
     }
 
     return limiter.schedule(() =>
-      fetch(urljoin(proxyUrl, r.path), opts)
-        .then((response) => response.json())
```
niclim, line 184: This is what was throwing the error, we were assuming everything was json and this threw an error when it wasn't json - the rest of the code sort of handled this

notnmeyer (approved):

PR #2242, opened 2023-08-21
extend CI check

niclim:
## 🍗 Description

I threw in more cases like `ci=1` or `ci=yes` or `ci=y` but im not sure what else we might want to cover here

## 📚 References

## 👹 QA

### DIFF
--- a/projects/optic/src/config.ts
+++ b/projects/optic/src/config.ts
@@ -32,7 +32,8 @@ const checkIsInCi = (): boolean => {
     lowerCase === 'true' ||
     lowerCase === '1' ||
     lowerCase === 'yes' ||
-    lowerCase === 'y'
+    lowerCase === 'y' ||
+    lowerCase === 'on'
   );
 };
 


projects/optic/src/config.ts
```diff
@@ -32,7 +32,11 @@ const DefaultOpticCliConfig: OpticCliConfig = {
   ruleset: undefined,
   isAuthenticated: false,
   client: createOpticClient('no_token'),
-  isInCi: process.env.CI === 'true',
+  isInCi:
+    process.env.CI === 'true' ||
+    process.env.CI === '1' ||
+    process.env.CI === 'yes' ||
+    process.env.CI === 'y',
```
notnmeyer, line 39: are these case sensitive?

notnmeyer:
here are yaml's boolean values: https://yaml.org/type/bool.html

i think we should use 'em

notnmeyer:
oh, and we should _never_ suggest in documentation that anyone use any value for the `CI` env var other than "true". since `CI` is set automatically by most (all?) CI providers deviating from the default is here-be-dragons territory.

niclim:
Yeah totally, this is just to catch any case where it's set somewhere else inadvertently

PR #2243, opened 2023-08-21
Improve init output instructions

niclim:
## 🍗 Description

Having launched capture, we've seen a couple of instances where users haven't gotten from `capture init` -> completing a capture sucessfully.

This PR:
- Removes some of the configuration from the capture example that isn't necessary for your first use. I added a link to the website where the full capture config reference can be found and people can use it
- Updates some of the wording in the generated config to be more actionable + added 🔧 (similar to ci-setup)
- Added explicit next steps when running

Output
```
optic capture init asd.yml
✔ Wrote capture config to optic.yml

Next:
- Edit the configuration to match your setup (🔧)
    A full reference for the configuration can be found at https://www.useoptic.com/docs/capturing-traffic#configuration-reference
- Then run: optic capture asd.yml --update interaction
```

Also fixes:
- `optic capture init` overwriting existing config in optic.yml
- moves around some code so we can rely on TS type inferences

## 📚 References

## 👹 QA

### DIFF
--- a/projects/optic/src/commands/capture/init.ts
+++ b/projects/optic/src/commands/capture/init.ts
@@ -10,7 +10,7 @@ export async function initCaptureConfig(
   skipConfigUpdate: boolean,
   config: OpticCliConfig
 ): Promise<string | undefined> {
-  const captureExample = captureConfigExample(oasFile);
+  const captureExample = captureConfigExample(oasFile, skipConfigUpdate);
   const parsedExample = yaml.parseDocument(captureExample);
 
   if (skipConfigUpdate) {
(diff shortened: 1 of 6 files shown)


notnmeyer (changes requested):
this produces invalid yaml when writing to a file

projects/optic/src/commands/capture/init.ts
```diff
@@ -26,55 +34,35 @@ export async function initCaptureConfig(
...
-          # Optional: default: 4
-          request_concurrency: 4
+      # 🔧 Specify either 'requests.run' or 'requests.send' to generate requests to hit your server
+      requests:
+        # ℹ️ Requests should be sent to the Optic proxy, the address of which is injected into 'run.command's env as OPTIC_PROXY (or the value of 'run.proxy_variable').
+        run:
+          # 🔧 Specify a command that will generate traffic
+          command: your-test-command
+          # 🔧 Ensure your command uses the OPTIC_PROXY environment variable. This ensures test traffic is hit
+          proxy_variable: OPTIC_PROXY
```
notnmeyer, line 56: this should be said differently. i read it a couple times and still find it confusing.

> OPTIC_PROXY is added to your command's env and contains the URL of an Optic's local reverse proxy. Your command should send its requests to this URL.

acunniffe:
Should we give people a link here instead? https://www.useoptic.com/docs/generate-openapi-from-tests#option-1-local-api--test-suite

I think specific examples are more helpful at first than references. I want to find a case similar to my own, and then adapt the provided config to my use case. If I need to do more than that I can to the reference. This is in a component called capturing-traffic-shared if we want to share that with them.

niclim:
> this produces invalid yaml when writing to a file

Weird, I am not sure why it was doing that - i think it was producing JSON (technically valid) - I just moved a couple things around so we didn't have to get + set values from a yaml document. Going to also add some tests here

niclim:
> Should we give people a link here instead? https://www.useoptic.com/docs/generate-openapi-from-tests#option-1-local-api--test-suite
> 
> I think specific examples are more helpful at first than references. I want to find a case similar to my own, and then adapt the provided config to my use case. If I need to do more than that I can to the reference. This is in a component called capturing-traffic-shared if we want to put it on a reference page, and have it included in the generate and verify docs

This seems confusing - we're telling them on that page to run `optic capture init`, which immediately would send the back to that page without giving them anything extra? 

The reason we're linking the reference docs is because we're omitting certain bits of config. You may need it for certain set ups, but for most people they won't need to reach for that

notnmeyer:
the other thing is that the --stdout no longer produces something that can be directly piped to your optic.yml (because the top level capture key is gone). previously, you could add a functional capture block to a new file or an existing optic.yml with, `optic capture init openapi.yml --stdout >> optic.yml`.

now, doing that or copying and even and pasting the example into your optic.yml with produce errors.

making `capture init openapi.yml` and `capture init openapi.yml --stdout` produce different output is seems questionable.

niclim:
@notnmeyer 

- Updated --stdout to include the `capture: + filename` indents
- Fixed output of YAML when freshly initialized

notnmeyer (approved):

Issue #2259, opened 2023-08-24
Requests run command is run in a new window on windows.


juliangrube1988:
**Describe the bug**
The specified command from requests/run/command in optic.yml is run in a new Window. If the command fails users cant see the output.

**To Reproduce**
Steps to reproduce the behavior on Windows:
1. add nx run foo to optic.yml requests/run/command
2. run optic capture
3. new window opens

**Expected behavior**
Everything is printed to the same terminal just like on a linux machine. The opening of a new window might be due to: 
NodeJs > process > spawn > windowsHide (is not set).

**Details (please complete the following information):**
 - Windows 11
 - Optic: 0.49.3
- NodeJS version: 18.14.0

acunniffe:
Hey @juliangrube1988 I've opened #2260 that makes the change you suggested. It seems to have worked on my Windows 10 VM, but I haven't tested in on Windows 11 yet. 

Can you give it a try? It's in a prerelase of Optic `0.49.4-0`

`npm install -g @useoptic/optic@0.49.4-0`

Legiew:
Hey, we tested the new version, but the changes in the last version didn't change the reported behaviour.
I had a detailed look and tried out to set in runRequestsCommand the detached property of the spawn call to false. That solved all the issues in combination with windowsHide: true.
Why is the detached property set to true? Optic waits for the command to finish, so there is no way, that the command runs longer than optic itself right? Or did I miss something?

juliangrube1988:
We used powershell. Cunniffe ***@***.***>:﻿
Some of our tests aren't passing when we have detached:false -- can you share more information about your shell? Is it WSL? Powershell? So we can do a better reproduction

notnmeyer:
should be fixed in https://github.com/opticdev/optic/releases/tag/v0.50.1. builds will be up in a few minutes.

PR #2260, opened 2023-08-24
Hide Window shell windows when capturing

acunniffe:
## 🍗 Description
During capture we spawn 1-2 processes (proxy + test command if it's set). On windows machines that defaults to opening new shell windows. We need to explicitly set it to `false` to prevent this. 

Fixes https://github.com/opticdev/optic/issues/2259

## 📚 References
https://nodejs.org/api/child_process.html

## 👹 QA
Manual test on my windows 10 VM

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "openapi-workspaces",
   "license": "MIT",
   "private": true,
-  "version": "0.49.3",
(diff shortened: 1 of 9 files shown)


notnmeyer:
fixed in https://github.com/opticdev/optic/pull/2308

PR #2263, opened 2023-08-28
have patches regenerate after becoming nullable

acunniffe:
## 🍗 Description
Null patches on 3.1 were leading to duplicate entries in the `type` array which is invalid JSON schema. We were not regenerating the diffs so every `null` example was adding a new `"null"` entry to type arrays.
Fixes https://github.com/opticdev/monorail/issues/4931
image

## 📚 References

## 👹 QA

### DIFF
--- a/projects/optic/src/commands/capture/patches/patchers/shapes/handlers/type.ts
+++ b/projects/optic/src/commands/capture/patches/patchers/shapes/handlers/type.ts
@@ -198,7 +198,7 @@ export function* typePatches(
             value: schemaType,
           }),
         ],
-        shouldRegeneratePatches: false,
+        shouldRegeneratePatches: true,
         interaction,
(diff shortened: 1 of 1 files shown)


projects/optic/src/commands/capture/patches/__tests__/__snapshots__/patches.test.ts.snap
```diff
@@ -9234,7 +9180,6 @@ exports[`generateEndpointSpecPatches OAS version 3.1.0 schema with arrays with p
                                 "type": [
                                   "string",
                                   "null",
-                                  "null",
```
acunniffe, line 9237: what we're fixing. was wrong in our tests

notnmeyer (approved):

PR #2265, opened 2023-08-30
Make capture init example runnable

acunniffe:
## 🍗 Description
The capture init example we had before could not be run and the instructions branched in a lot of places. 

These changes add two different configs to the file:
- The first (uncommented) one is an example using `requests.send`, and public API that can be run immediately. Our docs can tell them to run it right away to see how Optic works
- The second is a real config for running tests with the basic options specified and a link to the full reference. 

I'm planning to update the docs to tell people to run capture right after init to see an example. OR change the `server.url` and try with that. 

## 📚 References

## 👹 QA

### DIFF
--- a/projects/optic/src/commands/capture/init.ts
+++ b/projects/optic/src/commands/capture/init.ts
@@ -27,34 +27,39 @@ export function captureConfigExample(
 ) {
   return `
 ${skipConfigUpdate ? `capture:\n  ${oasFile}:` : ''}
+    # 🔧 Runnable example with simple get requests. 
+    # Run with "optic capture ${oasFile} --update interactive" 
+    # You can change the server and the 'requests' section to experiment
     server:
-      # 🔧 Update this to the command to run your server.
-      # Optional: If omitted, Optic assumes the server is running or started elsewhere.
-      command: your-server-command
-      # 🔧 Update this url to where your server can be reached.
-      # Required: Can be overridden with '--server-override'.
-      url: http://localhost:8080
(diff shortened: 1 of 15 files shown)


Alexandre-Herve (approved):

projects/optic/src/commands/capture/init.ts
```diff
@@ -27,34 +27,45 @@ export function captureConfigExample(
...
+      - path: /orgs/opticdev/repos
+        method: GET   
+      - path: /orgs/facebook/repos
+        method: GET   
+      - path: /orgs/opticdev/repos
+        method: GET
+      - path: /orgs/facebook/repos
+        method: GET
+      - path: /orgs/github/repos
+        method: GET
```
notnmeyer, line 54: GET is the default, i think we should remove all the unnecessary "method" lines save for the first one where we can add a comment that it is the default method.

notnmeyer:
this is fine, but one of the original intentions of `capture init` was for the sample output to be a mostly self-documenting representation of capture configuration. these changes push the reference documentation to a demo and removes much of the possible configuration options.

acunniffe:
> this is fine, but one of the original intention of capture init was for the sample output to be a mostly self-documenting representation of capture configuration

Yeah I think this is not as useful as giving them something that works so they can feel what Optic is going to do for them. We do link to the full config reference and give a second "real" example of how they should set up Optic. I think this is worth trying for a week and seeing what happens

Issue #2257, opened 2023-08-24
rulesets-base uses ajv but it is not marked as a dependency


jgresty:
The rulesets-base package uses `ajv` here: https://github.com/opticdev/optic/blob/a83f34f84ae513527328af866e42b1ef18c1764a/projects/rulesets-base/src/custom-rulesets/resolve-ruleset.ts#L1

However it is not listed as a dependency in its package.json, meaning that any consumer of this package will need to pull in ajv separately.

niclim:
Ah, thanks for reporting - I've put up a PR https://github.com/opticdev/optic/pull/2275 which will be included in the next release

PR #2275, opened 2023-09-06
add ajv dep to ruleset-base

niclim:
## 🍗 Description

fixes https://github.com/opticdev/optic/issues/2257

## 📚 References

## 👹 QA


acunniffe:
I think this is fine, but I know we've run into issues with multiple AJV versions across our package.json in the past.

niclim:
I'm using the same ajv version as `optic` - this might actually fix that issue if this was always underspecified

Alexandre-Herve (approved):

PR #2308, opened 2023-09-18
fix: display `requests.run.command` errors in the same powershell window

notnmeyer:
## 🍗 Description

fixes the issue mentioned in https://github.com/opticdev/optic/issues/2259, where `requests.run.command` error output is hidden in a new short-lived powershell window.

`server.command` has the same issue, but is more complicated to fix. see [aidan's pr](https://github.com/opticdev/optic/pull/2260) for details there. considering that windows powershell users are a small percentage of our users, i think its fine to fix the reported issue without considering everywhere we use `spawn()`.

## 📚 References
- closes https://github.com/opticdev/optic/issues/2259
- obviates https://github.com/opticdev/optic/pull/2260

## 👹 QA

### DIFF
--- a/projects/optic/src/commands/capture/actions/captureRequests.ts
+++ b/projects/optic/src/commands/capture/actions/captureRequests.ts
@@ -202,8 +202,11 @@ async function runRequestsCommand(
         ...process.env,
         [proxyVar]: proxyUrl,
       },
-      detached: true,
+      detached: false,
       shell: true,
+      // hide child process powershell windows on Windows systems. this only works with "detached: false".
+      // https://nodejs.org/api/child_process.html#child_processexeccommand-options-callback
+      windowsHide: true,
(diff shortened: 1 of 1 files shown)


projects/optic/src/commands/capture/actions/captureRequests.ts
```diff
@@ -202,8 +202,11 @@ async function runRequestsCommand(
         ...process.env,
         [proxyVar]: proxyUrl,
       },
-      detached: true,
+      detached: false,
```
notnmeyer, line 205: originally, we added this to address issues where optic would hang and processes wouldn't shut down cleanly, or was it being able to identify the PID to kill?. in haven't been able to concoct a scenario like that on mac or windows with this change.

niclim (approved):

PR #2410, opened 2023-10-18
Update comment with capture

Alexandre-Herve:
## 🍗 Description

## 📚 References

## 👹 QA

### DIFF
--- a/projects/optic/src/commands/ci/comment/common.ts
+++ b/projects/optic/src/commands/ci/comment/common.ts
@@ -25,7 +25,7 @@ const getChecksLabel = (
     exemptedFailingChecks > 0 ? `, ${exemptedFailingChecks} exempted` : '';
 
   return failingChecks > 0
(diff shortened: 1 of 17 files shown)


projects/optic/src/commands/ci/comment/common.ts
```diff
@@ -80,9 +101,11 @@ ${
 <thead>
 <tr>
 <th>API</th>
-<th>Operation Changes</th>
-<th>Checks</th>
+<th>Changes</th>
+<th>Rules</th>
 ${anyCompletedHasWarning ? '<th>Warnings</th>' : ''}
+<th>Tests</th>
```
niclim, line 107: Should we maybe only render this if there's at least 1 capture run? Or maybe add instructions on how to get started with capture here

If we do the second, we'll need to add some information about how to convert from the `diff-all` or `GHA` or `gitlab` pipeline to `run`

niclim (approved):

Issue #2387, opened 2023-10-11
Support OpenAPI schema refactoring in breaking changes detection


niclim:
**Describe the bug**

In OpenAPI 3 / 3.1 / JSONSchema a schema can be defined by a type string (`object`, `string`), type array (`[string, number]`), oneOf / anyOf / allOf / not.

Right now, the breaking changes ruleset does not allow transitions between types, and oneOf / anyOf and does not compute the effective difference between oneOf additions / removals + updates (i.e. removing fields in child A, but child B being a superset of child A)

In general, we should allow:
- expansion of types in parameters and request bodies
- narrowing of types in response bodies

Some cases we'll need to consider
- adding / removing an array item to a `oneOf` or `anyOf`
- refactoring a type array (i.e. `[string, number]`) to a `oneOf`
- transitions from `oneOf` to `anyOf`
- refactoring from an object to a `oneOf`

Related: https://github.com/opticdev/optic/issues/2372

niclim:
This is fixed - the breaking change ruleset will now let you refactor between `type: 'object'`, `type: [string, object]`, `oneOf` and `anyOf` and correctly detect when the effective set of changes is a breaking change (i.e. narrowed in response, expanded in request)

This will be released in `0.50.11`

PR #2417, opened 2023-10-19
Breaking change polymorphic refactoring

niclim:
## 🍗 Description

PR to allow refactors in breaking changes when types are equivalent. In general, request bodies are allowed to be narrowed (i.e. you can remove things from request bodies because it's not breaking) and response bodies are allowed to be expanded (i.e. you can add things to responses).

This PR handles things like refactoring an `object` -> `oneOf` or `typeArrays` (i.e. `[string, number]`) to `oneOf` and will figure out the if all of the before oneOf items have at least 1 valid transition

From a high level, this PR:
- updates the traditional breaking change checks to not run if the field is inside of a union property (`oneOf` or `anyOf`)
- runs a separate rule that handles `oneOf` or `anyOf`
  - this rule creates a keymap that is path to values, and 
  - if there is a union type (i.e. type array, oneOf, etc), we create multiple keymaps
    - for a transition not to be narrowed, every before item must have at least 1 corresponding after item that is not narrowed
    - for a transition not to be expanded, every after item must have at leat 1 corresponding before item that is not expanded

Notes about polymorphism
- in this PR, we aren't considering the difference between `oneOf` and `anyOf` - there's some transitions that were breaking here but this requires more thought on how to implement.
- `allOf` is handled at the denormalization step (i.e. merged into an object)
- `not` is not handled here

There's a few things I want to implement in follow up PRs:
- Update to surface reasons why a union was not valid (i.e. JSON path x/y/z did not have a valid transition)
- refactoring to a single handler for breaking changes

## 📚 References
https://github.com/opticdev/optic/issues/2387
## 👹 QA

### DIFF
--- a/projects/optic/src/__tests__/integration/__snapshots__/diff.test.ts.snap
+++ b/projects/optic/src/__tests__/integration/__snapshots__/diff.test.ts.snap
@@ -9,7 +9,7 @@ exports[`diff file doesn't exist 1`] = `
 exports[`diff petstore diff 1`] = `
 "[31mx [39m[1mSwagger Petstore Updated[22m [90mpetstore-base.json[39m
 [1mOperations: [22m5 operations added, 16 changed, 1 removed
-[31mx [39m [1mChecks:[22m 160/211 passed 
+[31mx [39m [1mChecks:[22m 192/243 passed 
 
 specification details:
 - /servers/2 [32madded[39m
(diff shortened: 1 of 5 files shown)


Alexandre-Herve (approved):

yonilerner:
Incredible, thank you

PR #2477, opened 2023-10-31
chore(cli): update diff help output

notnmeyer:
## 🍗 Description

- update diff's help output
- adds short command forms for some of the options. made an educated guess about the common ones folks might want to abbreviate.

## 📚 References

## 👹 QA

### DIFF
--- a/projects/optic/src/commands/diff/diff.ts
+++ b/projects/optic/src/commands/diff/diff.ts
@@ -114,8 +114,8 @@ export const registerDiff = (cli: Command, config: OpticCliConfig) => {
     .option('-c, --check', 'Enable checks', false)
     .option('-u, --upload', 'Upload run to cloud', false)
     .option('-w, --web', 'View the diff in the optic changelog web view', false)
-    .option('-j, --json', 'Output as json', false)
-    .option('-l, --last-change', 'Find the last change for this spec', false)
+    .option('--json', 'Output as json', false)
+    .option('--last-change', 'Find the last change for this spec', false)
     .option(
       '--generated',
       '[deprecated] Optic no longer differentiates generated and non-generated specifications'
--- a/projects/optic/src/commands/diff/diff.ts
+++ b/projects/optic/src/commands/diff/diff.ts
@@ -52,8 +52,8 @@ const description = 'Run a diff between two API specs';
 
 const usage = () => `
   optic diff <file_path> --base <base>
-  optic diff <file_to_compare_against> <file_path>
-  optic diff <file_to_compare_against> <file_path> --check`;
+  optic diff [file_to_compare_against] <file_path>
+  optic diff [file_to_compare_against] <file_path> --check`;
 
 const helpText = `
 Example usage:
--- a/projects/optic/src/commands/diff/diff.ts
+++ b/projects/optic/src/commands/diff/diff.ts
@@ -51,9 +51,9 @@ type DiffActionOptions = {
 const description = 'Run a diff between two API specs';
 
 const usage = () => `
-  optic diff <file_path> --base <base>
-  optic diff [file_to_compare_against] <file_path>
-  optic diff [file_to_compare_against] <file_path> --check`;
+  optic diff [file_path] --base <base>
+  optic diff [file_to_compare_against] [file_path]
+  optic diff [file_to_compare_against] [file_path] --check`;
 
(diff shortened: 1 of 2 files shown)


niclim (approved):

projects/optic/src/commands/diff/diff.ts
```diff
@@ -78,47 +78,48 @@ export const registerDiff = (cli: Command, config: OpticCliConfig) => {
...
     .addHelpText('after', helpText)
     .description(description)
-    .argument('[file_to_compare_against]', 'path to file to compare with')
-    .argument('[file_path]', 'path to file to compare')
-    .option(
-      '--base <base>',
-      'the base ref to compare against. Defaults to HEAD. Also supports optic cloud tags (cloud:tag_name)',
-      'HEAD'
+    .argument('[file_to_compare_against]', 'Path to file to compare with')
+    .argument('[file_path]', 'Path to file to compare')
```
acunniffe, line 82: should we rename these to "before" and "after"? feels like that's what people are trying to get to anyway as they parse this.

projects/optic/src/commands/diff/diff.ts
```diff
@@ -78,47 +78,48 @@ export const registerDiff = (cli: Command, config: OpticCliConfig) => {
...
         .choices(['strict', 'loose'])
         .default('strict')
     )
-    .option('--check', 'enable checks', false)
-    .option('--upload', 'upload run to cloud', false)
-    .option('--web', 'view the diff in the optic changelog web view', false)
-    .option('--json', 'output as json', false)
+    .option('-c, --check', 'Enable checks', false)
+    .option('-u, --upload', 'Upload run to cloud', false)
+    .option('-w, --web', 'View the diff in the optic changelog web view', false)
```
acunniffe, line 116: Optic should be uppercase here

acunniffe (approved):

projects/optic/src/commands/diff/diff.ts
notnmeyer: i can't decide how i feel about it, 

```
➜ npx ts-node $HOME/code/optic/projects/optic/src/index.ts diff --help | grep -A2 Arguments
Arguments:
  before                         Path to file to compare with
  after                          Path to file to compare

➜ optic diff --help | grep -A2 Arguments
Arguments:
  file_to_compare_against    path to file to compare with
  file_path                  path to file to compare
```

yeah, i cant decide what i like. `diff --help` uses "file1" and "file2", which is about as distilled and simple as you could make it.

projects/optic/src/commands/diff/diff.ts
```diff
@@ -48,15 +48,15 @@ type DiffActionOptions = {
...
 
-const description = `run a diff between two API specs`;
+const description = 'Run a diff between two API specs';
 
 const usage = () => `
   optic diff <file_path> --base <base>
-  optic diff <file_to_compare_against> <file_path>
-  optic diff <file_to_compare_against> <file_path> --check`;
+  optic diff [file_to_compare_against] <file_path>
+  optic diff [file_to_compare_against] <file_path> --check`;
```
notnmeyer, line 56: this is informational, but im following commander's syntax here and differentiating optional vs required arguments.

projects/optic/src/init.ts
```diff
@@ -54,6 +54,7 @@ export const initCli = async (
   } = {}
 ): Promise<Command> => {
   cli.name('optic');
+  cli.helpOption('-h, --help', 'Display help for the command');
```
notnmeyer, line 57: overrides the default option thats appended to everything with the same text, but capitalized

projects/optic/src/commands/diff/diff.ts
notnmeyer: ok @acunniffe i tried the file1/file2 thing. what do you think?

```
➜ npx ts-node $HOME/code/optic/projects/optic/src/index.ts diff openapi.yml diff --help
Usage:
  optic diff <file> --base <base>
  optic diff [file1] <file2>
  optic diff [file1] <file2> --check

Run a diff between two API specs

Arguments:
  file1                          Path to file to compare with
  file2                          Path to file to compare
```

notnmeyer: went with file1/2

PR #2479, opened 2023-11-01
chore(cli): update capture help

notnmeyer:
## 🍗 Description

- set of --help consistency changes

## 📚 References

## 👹 QA

### DIFF
--- a/projects/optic/src/commands/capture/capture.ts
+++ b/projects/optic/src/commands/capture/capture.ts
@@ -39,79 +39,80 @@ import { flushEvents, trackEvent } from '../../segment';
 import { getOpticUrlDetails } from '../../utils/cloud-urls';
 import sortBy from 'lodash.sortby';
 import * as Git from '../../utils/git-utils';
+import { getEndpointDiffs } from '@useoptic/openapi-utilities/build/openapi3/group-diff';
 
 const indent = (n: number) => '  '.repeat(n);
 
 export function registerCaptureCommand(cli: Command, config: OpticCliConfig) {
   const command = new Command('capture');
+  // not super sure why we have to override the help here
+  command.helpOption('-h, --help', 'Display help for the command');
 
   command.addCommand(clearCommand());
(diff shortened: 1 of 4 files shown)


projects/optic/src/commands/capture/capture.ts
```diff
@@ -39,79 +39,80 @@ import { flushEvents, trackEvent } from '../../segment';
...
 import sortBy from 'lodash.sortby';
 import * as Git from '../../utils/git-utils';
+import { getEndpointDiffs } from '@useoptic/openapi-utilities/build/openapi3/group-diff';
 
 const indent = (n: number) => '  '.repeat(n);
 
 export function registerCaptureCommand(cli: Command, config: OpticCliConfig) {
   const command = new Command('capture');
+  // not super sure why we have to override the help here
+  command.helpOption('-h, --help', 'Display help for the command');
```
notnmeyer, line 49: @niclim any thoughts on why overriding the help text is necessary here? its set in what i thought was a global place, https://github.com/opticdev/optic/blob/main/projects/optic/src/init.ts#L57

i didnt need to explicitly set it for the `diff` subcommand, and its already displaying correctly for `login` which i haven't touched. im sure there some difference in how the `capture` command is wired up, but im not seeing it.

niclim: I haven't confirmed this but it looks like capture creates it's own `const command ...` and `cli.addCommand(command)` where as `diff` registers something like `cli.command('diff')`

niclim (approved):

projects/optic/src/commands/capture/capture.ts
notnmeyer: ah, you're right—i see it now 👍

notnmeyer: made some mild changes to the main capture command so this works without overriding the help options. its still necessary for `capture clear|init`, looks like a little more of a refactor than its worth.

PR #2526, opened 2023-11-09
fix sourcemap when using custom headers

niclim:
## 🍗 Description

Fixed sourcemap generation when using custom headers - missed this part in the last PR

## 📚 References

## 👹 QA

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "openapi-workspaces",
   "license": "MIT",
   "private": true,
-  "version": "0.51.0",
+  "version": "0.51.1-0",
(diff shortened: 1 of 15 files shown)


notnmeyer (approved):

niclim:
Ok tested this out using prerelease 

Adding an Optic config like this:
```
external_refs:
  resolve_headers:
    - headers:
        Authorization: 'Bearer {{ TOKEN }}'
```

And then hitting a private repo's github contents like
```
TOKEN=gh_pat optic diff null: <url_to_content>
```

works successfully

PR #2545, opened 2023-11-16
Feature/lint gpt

Alexandre-Herve:
## 🍗 Description

## 📚 References

## 👹 QA


projects/standard-rulesets/src/lintgpt/index.ts
```diff
@@ -0,0 +1,307 @@
...
+                responses.before?.location.jsonPath!
+              )
+            : undefined,
+        });
+      });
+    });
+
+    ///// run the rules
+
+    const operationsRules = this.requirementRules.filter(
```
acunniffe, line 202: @Alexandre-Herve -- does this support parameters / headers? Is that coming next. Just want to make sure I write down the limitations accurately in the docs. I know it's just busy work to iterate those so we can release it after v1 if we are ready to get the bulk of this out.

projects/standard-rulesets/src/lintgpt/openai.ts
```diff
@@ -0,0 +1,9 @@
+import { OpenAI } from 'openai';
+
+export const openai: OpenAI = (() => {
```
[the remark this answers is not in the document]
Alexandre-Herve: Forgot to remove it from there, good catch!

projects/standard-rulesets/src/lintgpt/index.ts
Alexandre-Herve: I just copied what you had done there, but this can be added later yes

Issue #2557, opened 2023-11-27
Replace deprecated analytics-node in @useoptic/optic


jgresty:
The package analytics-node has been deprecated, [see github repo for details](https://github.com/segmentio/analytics-node), and should be replaced with @segment/analytics-node.

This is raising flags in automated scanners since it transitively depends on a vulnerable version of axios.

acunniffe:
Will take a look at this today. Looks like a drop-in replacement

acunniffe:
This should be out in `0.53.5`

PR #2558, opened 2023-11-28
Fix/upgrade segment package

acunniffe:
## 🍗 Description
Addresses https://github.com/opticdev/optic/issues/2557. Upgrades to the latest Segment Node package under their vendor namespace on npm `@segment`

## 📚 References

## 👹 QA
I'm shipping a prerelease to make sure it works in builds


acunniffe:
Prerelease works. Ran it locally and got events in Segment

PR #2561, opened 2023-11-28
parent in property rule context

acunniffe:
## 🍗 Description
Put the parent schema in the property schema rules. 
## 📚 References

## 👹 QA

### DIFF
--- a/projects/rulesets-base/src/rule-runner/data-constructors.ts
+++ b/projects/rulesets-base/src/rule-runner/data-constructors.ts
@@ -46,6 +46,35 @@ const createFactsWithRaw = <T extends OpenApiKind>(
 
   return factsWithRaw;
 };
+const createPropertyFactsWithRaw = <T extends OpenApiKind>(
+  nodeDetailMap: Map<string, NodeDetail<T>>,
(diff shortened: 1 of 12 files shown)


projects/rulesets-base/src/rule-runner/data-constructors.ts
```diff
@@ -46,6 +46,35 @@ const createFactsWithRaw = <T extends OpenApiKind>(
...
+  for (const [name, nodeDetail] of nodeDetailMap.entries()) {
+    const nodeFact = nodeDetail[key];
+    if (nodeFact) {
+      const parentPath = jsonPointerHelpers.pop(
+        jsonPointerHelpers.pop(nodeFact.location.jsonPath)
+      );
+      factsWithRaw.set(name, {
+        ...nodeFact,
+        raw: jsonPointerHelpers.get(openApiSpec, nodeFact.location.jsonPath),
+        parent: jsonPointerHelpers.get(openApiSpec, parentPath),
```
acunniffe, line 71: This is just a pointer so we're not dup'ing this in memory. It looks very verbose in the snapshots though

PR #2594, opened 2023-12-11
remove method and path from property location context

niclim:
## 🍗 Description

## 📚 References

## 👹 QA

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "openapi-workspaces",
   "license": "MIT",
   "private": true,
-  "version": "0.53.9",
+  "version": "0.53.10",
   "workspaces": [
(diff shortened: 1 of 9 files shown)


acunniffe:
This is too broad. We need it for operations since they don't have the path and method in the `OpenAPI.OperationObject` type. 

We don't want it for the properties, and probably the request / response bodies. The properties are the big one though

niclim:
> This is too broad. We need it for operations since they don't have the path and method in the `OpenAPI.OperationObject` type.
> 
> We don't want it for the properties, and probably the request / response bodies. The properties are the big one though

Could you elaborate? My understanding is that these changes only method and path from the properties location context

acunniffe (approved):

PR #2636, opened 2024-01-02
fix polymoprhic with nullable patching

niclim:
## 🍗 Description

fixes the case where nullable is next to a polymorphic keyword (allOf, oneOf, anyOf)

## 📚 References

## 👹 QA

### DIFF
--- a/projects/optic/src/commands/capture/patches/__tests__/patches.test.ts
+++ b/projects/optic/src/commands/capture/patches/__tests__/patches.test.ts
@@ -980,6 +980,65 @@ describe('generateEndpointSpecPatches', () => {
         expect(patches).toMatchSnapshot();
         expect(specHolder.spec).toMatchSnapshot();
       });
+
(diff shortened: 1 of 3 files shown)


projects/optic/src/commands/capture/patches/__tests__/patches.test.ts
```diff
@@ -980,6 +980,65 @@ describe('generateEndpointSpecPatches', () => {
         expect(patches).toMatchSnapshot();
         expect(specHolder.spec).toMatchSnapshot();
       });
+
+      test('with nullable', async () => {
+        // nullable keyword only valid in 3.0.1
+        if (version !== '3.0.1') return;
```
acunniffe, line 986: I believe it's valid in `3.0.0` and `3.0.1`

acunniffe (approved):
Looks good -- let's get this out and back to them

PR #2639, opened 2024-01-02
downgrade typebox

niclim:
## 🍗 Description

The binary build was broken because `0.32.0` typebox uses the `exports` field in package.json to build support for ESM, but `pkg` does not respect it https://github.com/vercel/pkg/issues/1873

## 📚 References

## 👹 QA

### DIFF
--- a/projects/optic/package.json
+++ b/projects/optic/package.json
@@ -61,7 +61,7 @@
     "@octokit/rest": "^19.0.0",
     "@segment/analytics-node": "^1.1.3",
     "@sentry/node": "^7.74.0",
-    "@sinclair/typebox": "^0.32.0",
+    "@sinclair/typebox": "0.31.28",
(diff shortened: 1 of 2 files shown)


notnmeyer (approved):

notnmeyer:
we've seen enough "X broke the binaries" issues that im feeling extra motivated to get some sort of testing around them. annoyingly (and my fault), we dont have an atomic release process. the binaries and container image wait for the npm packages to release first. its a remnant of how things were added on over time.

what would be great is building all the artifacts first, doing whatever verification, and then publishing everything.

@niclim was this failing during `optic run`? we could do something like https://github.com/opticdev/bookstore-example/blob/main/.github/workflows/optic.yml between building the binaries and publishing them. that won't do anything for atomicity, but could at least prevent us from pushing broken binary builds.

Issue #2535, opened 2023-11-13
Optic throws an error when receiving multipart/form-data requests


mithodin:
**Describe the bug**
When uploading a file to an endpoint that takes a multipart/form-data request body, optic throws the error
```
[request body] '' does not match type object.
```

**To Reproduce**
1. Clone the minimal example at https://github.com/mithodin/optic-multipart-form-data-repro
2. Run `npm install`, then `npm run optic:test`
3. See error

**Expected behavior**
optic should pass successfully and report a compliant request body.

niclim:
Thanks for the repro case - it looks like we aren't parsing form data / multipart form data and is currently being interpreted as a string (which is why it's saying `'' does not match type object`.

I believe this will need to be added here https://github.com/opticdev/optic/blob/main/projects/optic/src/commands/capture/patches/patchers/shapes/documented-bodies.ts#L224, and we'll need to pull in something to parse form / multipart form data here

niclim:
Hi, version `0.53.21` now supports form data and multipart-form data - thanks for reporting and let us know if there's anything else you run into!

PR #2655, opened 2024-01-11
Add form support to `capture`

niclim:
## 🍗 Description

Adds support for `application/x-www-form-urlencoded` and `multipart/form-data` content types.

- x-www-form-encoded is a simple query string we parse
- multipart form data is in a different format - there are parsers available, but they're mainly designed as streaming consumers to be used in web frameworks. Wrote a more specific one for our usecase (i.e. pull out field name and value)

Note that these fields are similar to query parameters (i.e. you can send multiple keys and some parsers will handle as arrays, etc) but we haven't handled it here since there are many ways to define arrays in forms and or JSON objects.

## 📚 References
https://github.com/opticdev/optic/issues/2535
## 👹 QA

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "openapi-workspaces",
   "license": "MIT",
   "private": true,
-  "version": "0.53.20",
+  "version": "0.53.21",
   "workspaces": [
     "projects/json-pointer-helpers",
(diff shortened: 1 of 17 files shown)


acunniffe (approved):
Looks good Nic

PR #2669, opened 2024-01-18
Fix/tests for node 20

niclim:
## 🍗 Description

## 📚 References

## 👹 QA

### DIFF
--- a/projects/optic/src/commands/oas/captures/proxy.test.ts
+++ b/projects/optic/src/commands/oas/captures/proxy.test.ts
@@ -1,13 +1,13 @@
 import { it, describe, expect, beforeAll, afterAll } from '@jest/globals';
 
 import * as mockttp from 'mockttp';
-import bent from 'bent';
 import { collect } from '../lib/async-tools';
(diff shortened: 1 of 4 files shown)


notnmeyer (approved):

projects/optic/src/commands/oas/captures/proxy.test.ts
```diff
@@ -1,13 +1,13 @@
...
-import bent from 'bent';
 import { collect } from '../lib/async-tools';
 import fetch from 'node-fetch';
 import https from 'https';
 import UrlJoin from 'url-join';
 import { httpsOverHttp } from 'tunnel';
 import { ProxyInteractions, ProxyCertAuthority } from './proxy';
+import portfinder from 'portfinder';
 
 describe('ProxyInteractions', () => {
```
niclim, line 12: Removed the dep `bent` (not sure why it was here) and swithced it out to use node-fetch. Also had to manually specify the portfinder range so the tests didn't conflict and didn't open on the same port (caused by jest tests running in parallel)

PR #2677, opened 2024-01-22
Order capture patch summaries

niclim:
## 🍗 Description

I noticed that node20 tests were being flaky, particularly around `optic capture` outputs. Not sure why node 20 specifically had differnet outputs, but I realized that patches were being summarized in the order the interactions happened, meaning we could have out of order patch summaries... E.g.

```
POST /form
  [31m× [39mRequest Body
  # Here 200 response body comes between the request bodies
  [31m[200 response body] body is not documented[39m
  [31m[request body] 'name' is not documented (/properties)[39m
  [31m[request body] 'surname' is not documented (/properties)[39m
```

This PR orders the patch summaries so that we list them by request, then response (sorted by status code ascending)

## 📚 References

## 👹 QA


notnmeyer (approved):
ah, sweet — i was seeing some weird stuff on a PR last week too

acunniffe (approved):

PR #2729, opened 2024-02-12
Merge nested allof

niclim:
## 🍗 Description

This PR merges nested allOf keys (e.g. `allOf: [ { allOf: [...] } ]`)

This PR also changes how we handle invalid allOf definitions. Instead of ignoring the `allOf` if there's a `type: string` and not flattening allOf, we ignore the `type: string` variant and flatten the rest (assuming there is an object). I think this is likely a better user experience, especially downstream in the rest of our app (e.g. allOf is assumed to not be there, and if it's invalid we'll get things like "every key in the allOf is removed" rule errors).

TODO
- [x] Surface reason / errors / warnings as part of denormalizing

## 📚 References

## 👹 QA


acunniffe (approved):
Yeah this makes sense. And if we did anything else, I think we'd do validation about teams defining conflicting types

PR #2743, opened 2024-02-20
update case functionality

niclim:
## 🍗 Description

## 📚 References

## 👹 QA


acunniffe (approved):
This is the fix for the case insensitive header thing right? Let's ship it to $Customer

PR #2751, opened 2024-02-26
fix(optic): don't loop forever when running from `/`

notnmeyer:
## 🍗 Description

i noticed what i thought was an issue with our docker image, where any invocation of optic would hang without output and require a `kill -9` to stop. after some stracing and validating where it first sprung up i narrowed it down to the seemingly innocuous way we walk the file tree to find a config.

when `$PWD` is `/` the walk enters an infinite loop. this most noticeable in the image because we leave its `WORKDIR` set at the root, `/`. you can however replicate this on macOS as well with,

```shell
cd / && optic --version
```

just be ready to grab the pid so you can kill it :)

## 📚 References

## 👹 QA

### DIFF
--- a/projects/optic/src/config.ts
+++ b/projects/optic/src/config.ts
@@ -157,7 +157,7 @@ export async function detectCliConfig(
 ): Promise<string | undefined> {
   let currentDir = process.cwd();
   // search up from pwd to specified dir
-  while (topLevelDir.length < currentDir.length) {
+  while (topLevelDir.length <= currentDir.length) {
     const expectedDevYmlPath = path.join(currentDir, OPTIC_DEV_YML_NAME);
     const expectedYmlPath = path.join(currentDir, OPTIC_YML_NAME);
     try {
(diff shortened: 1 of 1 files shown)


projects/optic/src/config.ts
```diff
@@ -157,7 +157,7 @@ export async function detectCliConfig(
 ): Promise<string | undefined> {
   let currentDir = process.cwd();
   // search up from pwd to specified dir
-  while (topLevelDir.length <= currentDir.length) {
+  while (topLevelDir.length < currentDir.length) {
```
niclim, line 160: I think we want to keep this conditional so we can check if the top level dir (e.g. `/user/some-project`) for the optic.yml

Perhaps instead we can do something like this on L173
```javascript
const next = path.dirname(currentDir)
if (next === currentDir) break
else currentDir = next
```

notnmeyer: cool, ill give it a try — looks like what im doing doesn't pass the tests anyway. thanks 👍

niclim (approved):

PR #2772, opened 2024-03-04
implement swagger 2 validation

niclim:
## 🍗 Description

## 📚 References

## 👹 QA

### DIFF
--- a/projects/openapi-io/src/validation/validation-schemas.ts
+++ b/projects/openapi-io/src/validation/validation-schemas.ts
@@ -1,3 +1,1574 @@
+export const basic_swagger2_object = {
+  type: 'object',
+  required: ['swagger', 'info', 'paths'],
+  properties: {
+    swagger: {
+      type: 'string',
+      enum: ['2.0'],
+    },
+    info: {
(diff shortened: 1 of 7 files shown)


projects/openapi-io/src/validation/validation-schemas.ts
```diff
@@ -1,3 +1,1574 @@
...
+            patch: {
+              $ref: '#/definitions/operation',
+            },
+          },
+        },
+      },
+    },
+  },
+  definitions: {
+    operation: {
```
acunniffe, line 45: shouldn't this also have operationId produces consumes, parameters tags and summary?

acunniffe (approved):
Looks good. I got a little confused by the difference between basic and the full schema, but figured out you were using it when strict validation was turned off

niclim:
> Looks good. I got a little confused by the difference between basic and the full schema, but figured out you were using it when strict validation was turned off

We have the same idea with the openapi 3 validator, so this is to match the functionality (I think we initially did this to try match the minimum typings) but I could see us moving away to not validate anything here

